PR #9, opened 2022-12-01
README.md: Add a decription of relationship with rpm-ostree

cgwalters:
Closes: https://github.com/containers/bootc/issues/8


jlebon (approved):
Thanks, that helps a bit. :)

PR #15, opened 2022-12-02
Also use DeploymentStatus for upgrade

cgwalters:
Add an API to gather our extra metadata from a single deployment, such as the supported status.  Use it in the status path and the upgrade path, instead of gathering the supported state separately.


jmarrero (approved):
lgtm

PR #16, opened 2022-12-05
Use stage_tree_with_options

cgwalters:
This is a bug in the ostree Rust bindings, it can't express the difference between a zero-length array and an unset array.

And if we provide a zero-length array, ostree will happily remove all other kargs.


jmarrero (approved):
lgtm

PR #17, opened 2022-12-05
switch: Always create a new origin

cgwalters:
This way we won't leak state like rpm-ostree packages from a previous one.


jmarrero (approved):
lgtm

PR #18, opened 2022-12-05
Clean up staging flow

cgwalters:
Add a helper `stage()` which takes all the data we need, and use it in both the `upgrade` and `switch` paths.  This drops out an "internal error" message and avoids the "utils" code growth.


jmarrero (approved):
lgtm

PR #23, opened 2022-12-07
switch: Fix changing from an ostree remote

cgwalters:
When we added logic to prune previous containers, we broke switching from an ostree remote + ref.  We need to parse the ref to handle the remote.

Closes: https://github.com/containers/bootc/issues/19


jmarrero (approved):
lgtm

PR #25, opened 2022-12-07
Enable `cargo vendor-filterer`

cgwalters:
Add configuration and a GH action post-push which automatically generates a vendor tar.


jmarrero (approved):
lgtm

PR #28, opened 2022-12-16
ci/run-kola: Support pull secret for running in Prow

cgwalters:
This is needed to support fetching the image from the registry.


cgwalters:
OK this needs https://github.com/coreos/coreos-assembler/pull/3288

jmarrero (approved):
lgtm

PR #30, opened 2022-12-21
Add an `install` command

cgwalters:
Usage example with `/dev/vda` being an extra mounted disk:

```
$ podman run --privileged --pid=host --net=none --security-opt label=type:unconfined_t quay.io/cgwalters/c9s-oscore:latest bootc install /dev/vda
```

# Code and high level design

This merges a lot of code and logic from coreos-installer, coreos-assembler and some bits of fedora-coreos-config.

Unlike ostree, `bootc install` understands and manipulates block devices and bootloaders.

 - coreos-installer: ignition and blockdev.rs code
 - coreos-assember: sgdisk invocations and ostree repo setup translated to Rust
 - fedora-coreos-config: Things like transposefs, boot/root uuid handling are now done on the "install boot" here

Note this obviates much of the need for [generating disk images](https://github.com/coreos/fedora-coreos-tracker/issues/1151).  That said, it may also make sense to have `bootc install generate-iso`.

TODO items:

- Add support for e.g. `--copy-network` type flows as well as https://github.com/containers/bootc/issues/22
- Implement ppc64le and s390x (needs code from both coreos-assembler and coreos-installer)
- support installing from SELinux-disabled hosts (needs https://github.com/ostreedev/ostree/issues/2804 )
- Support *overwriting* disk we used to boot (`systemctl switch-root` to container image?)
- Support basic partition configs (e.g. `/var`)
- Related, investigate https://www.freedesktop.org/software/systemd/man/systemd-repart.html
- Also support running Ignition from the "install boot"; this would make debugging Ignition configs *dramatically* nicer
- Ensure we lock devices when partitioning https://systemd.io/BLOCK_DEVICE_LOCKING/
- Support config files (e.g. distro default filesystem specified in `/usr/lib/bootc/install.cfg`)
- Support (or default) to only installing EFI if booted that way, etc.
- Support systemd-boot
- Clean up/refactor install code to be more distribution-independent in general
- Consider merging bootupd in here as e.g. `bootc bootloader update` or so?
- Documentation
- Tests
- More explicitly test host/target kernel compatibility (e.g. filesystems)

Closes: https://github.com/containers/bootc/issues/24
Closes: https://github.com/containers/bootc/issues/1


cgwalters:
>  mount: /sys/fs/selinux: mount point does not exist.

OK yeah.  Let's see if we can support the exciting case of also installing from systems with SELinux disabled...

cgwalters:
OK yeah, this path of testing `bootc install` from the ubuntu system needs https://github.com/ostreedev/ostree/issues/2804

So, going to instead look at doing CI on Fedora for that.

cgwalters:
OK, we now have a CI test that proves this works.  I think it's about time to get this merged and do further followups (there'll be a lot) as distinct PRs on top of a proven working base.  Lifting draft!

cgwalters:
OK no real team on this repo for now so no peer review...

PR #32, opened 2023-01-09
Add an `internal-tests run-privileged-integration`

cgwalters:
status: Don't error out if not booted

For now, just print out nothing.  This helps a future integration
testing scenario.

---

Add an `internal-tests run-privileged-integration`

Add a feature to enable internal tests.  I originally started
by matching
https://github.com/ostreedev/ostree-rs-ext/commit/6543c855580da3a3f2a85bc8b4a294b4c92c2829
but since here the binary is really end user facing and not just
a demo, we shouldn't enable the feature by default.

If in the future we split off the tiny CLI bit out of the workspace,
then we could auto-enable but that seems not worth it.

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,7 @@ jobs:
         with:
           key: "build"
       - name: Build
-        run: cargo build --release
+        run: cargo build --release --features internal-testing-api
       - name: Upload binary
         uses: actions/upload-artifact@v2
         with:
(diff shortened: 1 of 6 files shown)


jmarrero (approved):
lgtm

cgwalters:
Hmm, need to do some tweaks to the Prow config here; I don't think we want to use approve/lgtm, just CI tests.  (But this is obviously subject to a larger debate for consistency inside the containers/ GH org)

PR #33, opened 2023-01-13
Small prep patches for `install`

cgwalters:
status: Use `if` instead of `.then(||)`

It's equally clear and avoids clippy lint conflicts between different
Rust versions.

---

cli: Add some error prefixing

On general principle.

---

Add a `reexec` helper

Prep for the `install` path where we will use this in
more ways.

---


jmarrero (approved):
lgtm

PR #35, opened 2023-01-17
Use version API from ostree-ext

cgwalters:
In preparation for https://github.com/ostreedev/ostree-rs-ext/pull/454


jmarrero (approved):
lgtm

PR #36, opened 2023-01-19
Add `cargo xtask` and packaging infrastructure

cgwalters:
First, this adds `cargo xtask` following
https://github.com/matklad/cargo-xtask/

We use this to write "external glue scripts" in Rust, not bash. Specifically we now have e.g. `cargo xtask vendor` which just wraps running `cargo vendor-filterer`.

Then build on that and add `cargo xtask package-srpm` which generates a `.src.rpm`.

And build on that by adding the requisite glue to have Fedora's COPR be able to understand it, so that we can get auto-built and shipped packages there.

This will make trying out bootc a bit easier.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[alias]
+xtask = "run --package xtask --"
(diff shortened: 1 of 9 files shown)


jmarrero (approved):
pretty cool!

PR #37, opened 2023-01-20
reexec: Generalize a bit

cgwalters:
The rexec logic was hardcoded to run `unshare -m`; move that bit into the caller so the core re-exec does literally just that.

Prep for other re-exec callers in the install code.


jmarrero (approved):
lgtm

PR #56, opened 2023-02-04
xtask: Inject timestamp into version

cgwalters:
I was hitting an issue with our COPR builds where we were getting *earlier* builds because the git hash was the first component of the version, and the leading numbers there don't increment, and RPM wants higher versions.

Force newer-is-better by injecting the date (up to the minute, not second because we're not going to build more than one a second).


jmarrero (approved):
lgtm

PR #60, opened 2023-02-07
install: Refactor selinux re-execution more

cgwalters:
We need to be able to use this in more places; rework things so that we have a handy function.


jmarrero (approved):
lgtm

PR #61, opened 2023-02-08
install: Clean up imageref handling

cgwalters:
Prep for takeover installs, but also just a general cleanup. Rather than having `ContainerPolicyAllowInsecure` in multiple places, have it just once with an explicit rationale.


cgwalters:
Decided to do a bigger PR here in https://github.com/containers/bootc/pull/62

PR #62, opened 2023-02-08
install: More cleanups

cgwalters:
install: Clean up imageref handling

Prep for takeover installs, but also just a general cleanup.
Rather than having `ContainerPolicyAllowInsecure` in multiple
places, have it just once with an explicit rationale.

---

install: Further clean up imageref handling

Prep for takeover installs.  Rather than carrying the
container env info around in `State`, carry just the
source image reference and digest separately.

This will make it easier to add code later for
takeover installs to use an image already in an ostree repo.

---

install: Split off baseline module

Since we added `install-to-filesystem`, the install code needs
a refactoring where we more cleanly split up our "default"
installer which in turn passes its logic to the filesystem
side.

---


jmarrero (approved):
lgtm

PR #65, opened 2023-02-17
bootloader: Switch to using bootupd to do BIOS

cgwalters:
install: use ghcr.io for oscore

See https://github.com/cgwalters/bootc-demo-base-images/commit/2917d146b86b339bae4b3bbf1f848301b48386aa

---

bootloader: Switch to using bootupd to do BIOS

This means we now get unified management.

---


cgwalters:
@HuijingHei This uses your changes to bootupd!

HuijingHei (approved):
LGTM

PR #66, opened 2023-02-20
install: Verify we're root

cgwalters:
This came up in chat. The error message right now is pretty obscure, we fail in `chcon` which might make you think something is wrong with SELinux.


jmarrero (approved):
lgtm

PR #67, opened 2023-02-25
install: Add systemd-style drop-in config files

cgwalters:
Add support for `/usr/lib/bootc/install/*.toml`
and `/etc/bootc/install/*.toml`
and `/run/bootc/install/*.toml` etc. which define config files for the install process.

Use this to configure the root filesystem type instead of hardcoding it.  We still default to xfs but it's now trivial for an OS/distribution to drop-in an override (or just edit the default toml).  Additionally derived builds can do so just as easily.

We then drop the `impl Default for Filesystem` so that the default isn't hardcoded in the binary anymore.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ all-test:
 
 install:
 	install -D -t $(DESTDIR)$(prefix)/bin target/release/bootc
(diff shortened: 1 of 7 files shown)


jmarrero (approved):
lgtm

PR #68, opened 2023-02-25
ci: Switch to `make test-bin-archive`

cgwalters:
build: Install bin-archive in target/

To avoid polluting the toplevel.

---

ci: Switch to `make test-bin-archive`

Right now we only install a single binary, but in the future
that will change (e.g. systemd units, config files).

---


jmarrero (approved):
lgtm

PR #70, opened 2023-02-26
xtask: Fix package builds

cgwalters:
No idea how it worked before; we don't want to include the diff output which can lead to us failing to parse an integer:

```
error: Packaging

Caused by:
    0: Finding git timestamp
    1: invalid digit found in string
```


jmarrero (approved):
lgtm

jmarrero:
> No idea how it worked before;

magic :magic_wand:

PR #75, opened 2023-03-12
 install: Add support for LUKS

cgwalters:
install: Drop `impl Default for Filesystem`

We require this to be configured externally now.

---

install: Add support for LUKS

This currently requires systemd 253 for
https://github.com/systemd/systemd/commit/03f36e9c0174f983e832c6ced80492ebefd34e46

---

### DIFF
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -576,6 +576,7 @@ fn skopeo_supports_containers_storage() -> Result<bool> {
 }
 
 pub(crate) struct RootSetup {
+    luks_device: Option<String>,
     device: Utf8PathBuf,
     rootfs: Utf8PathBuf,
     rootfs_fd: Dir,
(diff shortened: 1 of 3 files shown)


cgwalters:
One thing I'm a bit uncertain about here is that while systemd-cryptenroll is a much nicer replacement for the bash scripts in https://github.com/latchset/clevis it doesn't scope in NBDE/Tang like clevis either.  With my RHEL-families hat on it seems almost certain at least some images will need to support that.

Maybe one thing we end up doing with the install code is having things be more explicitly pluggable, so "tpm2-luks" can be backed by either systemd-cryptenroll or clevis.  It's probably not too hard to also add clevis support and then make this another install config option.

cgwalters:
This installs but fails to boot, looks like cryptsetup is refusing to mount because it wants a keyfile or pin.  Tried also injecting `luks.options=tpm2-device=auto` but no luck.  Going to need to debug the cryptsetup generator I guess.

cgwalters:
OK this needed https://github.com/cgwalters/bootc-demo-base-images/commit/ff46f764500e557270b5097a7bdf4b21c154403c

jmarrero (approved):
lgtm

PR #77, opened 2023-03-15
Add `usr-overlay`

cgwalters:
This also just forwards to `ostree admin unlock` the same as `rpm-ostree usroverlay` does.


jmarrero (approved):
lgtm

PR #79, opened 2023-03-17
takeover installs: prep0

cgwalters:
install: Make config module `pub(crate)` and configs serializable

Prep for takeover installs.

---

lsm: Add an API to set SELinux permissive

This is going to be needed for takeover installs.  There's
no point to trying to keep the LSM state running through
the whole thing because we're going to replace the OS anyways.

---

Bump requirement to latest ostree-ext

Prep for takeover installs.

---


jmarrero (approved):
lgtm

PR #84, opened 2023-03-21
install: Fall back to `setenforce 0` if needed

cgwalters:
I've opened https://github.com/containers/bootc/issues/83 to track a better fix.

### DIFF
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -178,6 +178,8 @@ pub(crate) struct State {
     pub(crate) source: SourceInfo,
     /// Force SELinux off in target system
     pub(crate) override_disable_selinux: bool,
+    #[allow(dead_code)]
+    pub(crate) setenforce_guard: Option<crate::lsm::SetEnforceGuard>,
     pub(crate) config_opts: InstallConfigOpts,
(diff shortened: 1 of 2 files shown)


cgwalters:
cc @rhatdan - Trying really hard here to not `setenforce 0` but failing.  For some reason I don't understand, the domain transition from a container image executing as `unconfined_t` running an `install_exec_t` file just doesn't transition to `install_t`.  Trying to do e.g. 
```
$ podman run --rm -ti --privileged --security-opt label=type:install_t quay.io/centos/centos:stream9 true
{"msg":"exec container process `/usr/bin/true`: Permission denied","level":"error","time":"2023-03-21T13:09:14.616124Z"}
```
fails, etc.

rhatdan:
This is a container-selinux issue.

https://github.com/containers/container-selinux/pull/222

rhatdan:
Fixed in https://github.com/containers/container-selinux/releases/tag/v2.206.0

cgwalters:
Nice, thanks!

The good thing at least here is that while this PR is a hack, it is at least designed such that we only fall back to `setenforce 0` if our "re-exec as install_t" trick doesn't work.

PR #89, opened 2023-03-23
cli: Add an alias `update` for `upgrade`

cgwalters:
yum/dnf does this, I think it's a friendly thing to do.  I actually typed `bootc update` and got reminded to do this.


jmarrero (approved):
lgtm

PR #96, opened 2023-04-25
install: Only invoke `chcon` if SELinux enabled in the source

cgwalters:
We shouldn't try to do any SELinux labeling if the source doesn't have it.

xref https://github.com/containers/bootc/issues/83


cgwalters:
cc @akdev1l

jmarrero (approved):
/lgtm

PR #100, opened 2023-05-02
WIP: Add support for configmaps

cgwalters:
This is part of https://github.com/containers/bootc/issues/22


cgwalters:
I was typing this up, but thinking about it...perhaps it should land in ostree-ext so it can be shared with rpm-ostree?

Or...maybe we publish a crate for bootc and pull *that* into rpm-ostree too...

cgwalters:
OK what I'm getting hung up on a bit here is exactly what the API is to upload objects outside of kubernetes.  It looks like [podman play kube](https://docs.podman.io/en/v4.1.1/markdown/podman-play-kube.1.html) basically punts on this and you point the command at an arbitrary directory...which is rather un-opinionated - and also has some annoying ergonomics because then you typically want to delete the files after they've been created.

The typical systemd-style pattern of supporting e.g. `/run/bootc/api` and `/etc/bootc/api` (but probably not `/usr/lib/bootc/api` since that gets weird around mutability) is perhaps a bit better than an arbitrary directory.

We'd want to have `bootc api get configmap` to dump all the configmap objects that bootc knows about.  (Or maybe just `bootc get`?)

PR #105, opened 2023-07-04
cli: Add a `--check` flag for update

### LINKED ISSUE #3: Add `upgrade --dry-run` or so
So that client-side automation tooling can poll for updates before downloading.  xref https://github.com/coreos/rpm-ostree/issues/4176 etc.

jbtrystram:
This simply pull the manifest of the iamge to see if any update and available. It could be named `--dry-run` but I wanted to be consistent with `rpm-ostree`.
Fixes #3

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -28,9 +28,13 @@ pub(crate) struct UpgradeOpts {
 
     #[clap(long)]
     pub(crate) touch_if_changed: Option<Utf8PathBuf>,
+
+    /// Check if an update is available without applying it
+    #[clap(long)]
+    pub(crate) check: bool,
 }
 
-/// Perform an upgrade operation
+/// Perform an switch operation
 #[derive(Debug, Parser)]
 pub(crate) struct SwitchOpts {
     /// Don't display progress
@@ -57,7 +61,7 @@ pub(crate) struct SwitchOpts {
     pub(crate) target: String,
 }
 
-/// Perform an upgrade operation
+/// Perform a status operation
 #[derive(Debug, Parser)]
 pub(crate) struct StatusOpts {
     /// Output in JSON format.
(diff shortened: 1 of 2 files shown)


lib/src/cli.rs
```diff
@@ -281,17 +285,39 @@ async fn upgrade(opts: UpgradeOpts) -> Result<()> {
...
+        match imp.prepare().await? {
+            PrepareResult::AlreadyPresent(c) => {
+                println!(
+                    "No changes available for {}. Latest digest: {}",
+                    imgref, c.manifest_digest
+                );
+            }
+            PrepareResult::Ready(p) => {
+                println!(
+                    "New manifest available for {}. Digest {}",
```
cgwalters, line 302: One less than ideal thing here is that it leaves open the problem of how programmatic access would work - if I'm scripting bootc, I'd have to parse this text designed for human readability.

We already have a `--touch-if-changed` CLI option which is designed for the non-check case, so let's honor it with `--check` too.

lib/src/cli.rs
```diff
@@ -281,17 +285,39 @@ async fn upgrade(opts: UpgradeOpts) -> Result<()> {
...
+                    imgref, p.manifest_digest
+                );
+            }
+        }
         return Ok(());
-    }
+    } else {
```
cgwalters, line 308: Since we have a `return` above I don't think we need to indent everything here, right?  (Though if you follow my above suggestion, the touch_if_changed handling would be outside the `else`)

cgwalters:
BTW, this will conflict with https://github.com/containers/bootc/pull/106 - I forgot you'd been looking at this issue.  How these two intersect is actually very interesting because semantically that's about adding a declarative flow...so to properly model things here we may need to split out a "discovered but not downloaded" state internally.

lib/src/cli.rs
jbtrystram: I updated the PR with a change in the return statement and moved the touch_if_changed handling outside of the else. Now it will carry on and honor the option. 
However that keeps the indentation.

jbtrystram:
> Two minor nits on the commit message
> 
>     * Typo "iamge"
> 
>     * Please drop the trailing "." in the subject line

I amended the commit message fixing that :)

jbtrystram:
> BTW, this will conflict with #106 - I forgot you'd been looking at this issue. How these two intersect is actually very interesting because semantically that's about adding a declarative flow...so to properly model things here we may need to split out a "discovered but not downloaded" state internally.

which PR do you want to merge first ? I'm fine waiting for your changes then rebasing this

cgwalters:
> which PR do you want to merge first ? I'm fine waiting for your changes then rebasing this

Mmmm...my inclination here is to merge your work first because I want to prioritize new contributors.  So let's do that.  Thanks for the PR!

PR #106, opened 2023-07-04
Rework API to use a Kubernetes CRD

cgwalters:
I was working on the configmap support slowly, but a big issue
it raised was how we expose the ability to mutate things.

Our current "API" is just an imperative CLI.  This would require
higher level tooling to manage Kubernetes style declarative system state.

Instead, let's use the really nice Rust `kube` crate to define a CRD;
this is what is output (in YAML form) via `bootc status` now.

We translate the imperative CLI verbs into changes to the `spec`
field.

However, things become more compelling when we offer a `bootc edit`
CLI verb that allows arbitrary changes to the spec.

I think this will become the *only* way to manage attached configmaps,
instead of having imperative CLI verbs like `bootc configmap add` etc.
At least to start.

---

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -19,6 +19,8 @@ hex = "^0.4"
 fn-error-context = "0.2.0"
 gvariant = "0.4.0"
 indicatif = "0.17.0"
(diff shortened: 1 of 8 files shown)


jmarrero (approved):
lgtm

PR #107, opened 2023-07-06
cli: Add an `edit` verb

cgwalters:
Depends https://github.com/containers/bootc/pull/106

---

cli: Add an `edit` verb

This interactively replaces the specified state, in a similar way
as `kubectl edit`.

As of right now, the only supported state to change is the desired
image.

But this will help unblock [configmap support].

[configmap support]: https://github.com/containers/bootc/issues/22

---

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -16,9 +16,11 @@ use ostree_ext::keyfileext::KeyFileExt;
 use ostree_ext::ostree;
 use ostree_ext::sysroot::SysrootLock;
 use std::ffi::OsString;
+use std::io::Seek;
 use std::os::unix::process::CommandExt;
 use std::process::Command;
 
+use crate::spec::Host;
 use crate::spec::HostSpec;
 use crate::spec::ImageReference;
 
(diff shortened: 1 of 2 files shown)


cgwalters:
One thing that probably needs to happen here is we need to implement more of an "api server" in having at least a `resourceVersion` equivalent to detect conflicts.

Also a random thought: we probably still do want to run as a systemd unit when changes are being made.  But we could just always serialize our state into `/run/bootc/status.json` perhaps, and let clients monitor it for changes via inotify?

jmarrero:
wdyt of adding an `apply` option for passing the crd yaml and use edit to do live edits?

cgwalters:
That's how it works right?  You can `bootc status > status.yaml` then edit or change as desired, then `bootc edit status.yaml` again.  To do things interactively you need to `bootc edit -` where `-` means stdin.

jmarrero:
yeah that is how it works, I was just wondering if it made sense to use the word 'apply' instead for that. As that is how kubectl and oc do it, which would be a more familiar interface I guess.

jmarrero (commented):
lgtm

cgwalters:
It's a good point!  I'm not sure... we could make `bootc apply` an alias?  Or we could always call it `apply`?

cgwalters:
To be clear I'm happy to change to `apply`, the more I think about the more I agree.

jmarrero:
Well I think we need both if we want the "familiar interface". IIRC apply is just for when we are passing the yaml file. https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#apply

Edit is more manual, expects the user to modify it with an editor:
https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#edit

But Idk, I see that edit also has a -f filename option which I am not sure if it just does the same thing as apply... If it does, then I guess we can ignore apply for now.

PR #114, opened 2023-08-15
 install: Compile (but error at runtime) on non-x86_64/aarch64 

### LINKED ISSUE #112: Build fails on s390x:
Build fails on s390x:
```
error[E0425]: cannot find value `EFIPN` in this scope
   --> /builddir/build/BUILD/rpm-ostree-2023.5.38.g0cbcf4ae/vendor/bootc-lib/src/install/baseline.rs:247:13
    |
247 |             EFIPN,
    |             ^^^^^ not found in this scope
error[E0425]: cannot find value `EFIPN_SIZE_MB` in this scope
   --> /builddir/build/BUILD/rpm-ostree-2023.5.38.g0cbcf4ae/vendor/bootc-lib/src/install/baseline.rs:248:26
    |
34  | pub(crate) const BOOTPN_SIZE_MB: u32 = 510;
    | ------------------------------------------- similarly named constant `BOOTPN_SIZE_MB` defined here
...
248 |             format!("0:+{EFIPN_SIZE_MB}M"),
    |                          ^^^^^^^^^^^^^ help: a constant with a similar name exists: `BOOTPN_SIZE_MB`
error[E0425]: cannot find value `EFIPN` in this scope
   --> /builddir/build/BUILD/rpm-ostree-2023.5.38.g0cbcf4ae/vendor/bootc-lib/src/install/baseline.rs:252:32
    |
252 |         Some(format!("{device}{EFIPN}"))
    |                                ^^^^^ not found in this scope
For more information about this error, try `rustc --explain E0425`.
error: could not compile `bootc-lib` due to 3 previous errors
Caused by:
  process didn't exit successfully: `rustc --crate-name bootc_lib --edition=2021 /builddir/build/BUILD/rpm-ostree-2023.5.38.g0cbcf4ae/
```
This blocks an rpm-ostree release as we are vendoring bootc in rpm-ostree.

jmarrero:
We already are checking for target_arch when using the constants. Just instantiating them when we are in a target arch makes the build fail when we are building on a arch that that hits the constraint we created for initiating the constant.

closes #112

### DIFF
--- a/lib/src/install/baseline.rs
+++ b/lib/src/install/baseline.rs
@@ -33,16 +33,8 @@ pub(crate) const BOOTPN: u32 = 3;
 // This ensures we end up under 512 to be small-sized.
 pub(crate) const BOOTPN_SIZE_MB: u32 = 510;
 pub(crate) const ROOTPN: u32 = 4;
-#[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
 pub(crate) const EFIPN: u32 = 2;
-#[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
 pub(crate) const EFIPN_SIZE_MB: u32 = 512;
-#[cfg(target_arch = "aarch64")]
-pub(crate) const RESERVEDPN: u32 = 1;
-#[cfg(target_arch = "ppc64")]
-pub(crate) const PREPPN: u32 = 1;
-#[cfg(target_arch = "ppc64")]
-pub(crate) const RESERVEDPN: u32 = 1;
 
 #[derive(clap::ValueEnum, Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "kebab-case")]


jmarrero:
draft while I test this locally. I just messed up my toolbox by upgrading to f39. So this is just a theory atm... I can't see anything else that could be wrong.

jmarrero:
It builds ok locally and I started up a s390x debug pod, ran:

```
cosa shell
sudo dnf install openssl-devel glib2-devel ostree-devel
cargo build
```
and got a good build:
```
   Compiling bootc v0.1.0 (/home/jenkins/agent/workspace/debug-pod/bootc/cli)
    Finished dev [optimized + debuginfo] target(s) in 2m 26s
[coreos-assembler]$ uname -a
Linux 1830b359aaec 6.3.12-200.fc38.s390x #1 SMP Thu Jul  6 03:39:39 UTC 2023 s390x GNU/Linux
```

cgwalters:
A procedural note, we don't usually include the file suffix ".rs" in the topic prefix.  This is about the install code (module).  And it's usually best in the subject message to explain a bit more what the goal is.  Something like:

> install: Compile (but error at runtime) on non-x86_64/aarch64

This all said, while I'm not opposed to this fix...thinking about this a bit more, what seems it'd be better (and about equally easy) is to change the build configuration in `Cargo.toml` to only enable the `install` feature on x86_64 and ppc64le.

Then we can drop many of the architecture conditionals in that code.  Conceptually we'd be going from `bootc install` on those arches saying "This isn't implemented" to "command not found" which I think is probably more right.

Hmm...I'm not totally sure Cargo has support for conditionally enabling a crate feature for a given arch (without resorting to a build script).  It does support this for dependencies https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies 

Well...in the end, let's just do some cleanup for this commit message and merge!

(Of course the real fix is to actually implement the install flow for those arches, but that's more involved)

cgwalters:
BTW just for reference this install logic is a Rust rewrite of https://github.com/coreos/coreos-assembler/blob/main/src/create_disk.sh

PR #116, opened 2023-08-16
lib: Dispatch to monomorphic function

cgwalters:
This mirrors https://github.com/ostreedev/ostree-rs-ext/pull/515

From looking at cargo bloat.


jmarrero (approved):
lgtm

PR #117, opened 2023-08-22
Cargo.toml: Update release profile, add releaselto

cgwalters:
- `release` should use `panic=abort` by default because we make heavy use of FFI and this is safest, and I don't think we need unwinding anyways
- The `releaselto` produces smallest binaries


jmarrero (changes requested):
lgtm but:

commit needs the `Signed-off-by:` line :pen:

jmarrero (approved):

PR #118, opened 2023-08-25
lib: Bump to ostree-ext 0.11.5

cgwalters:
And note the new `require_bootable()`, since we couldn't require that in the container image importer.


jmarrero (approved):
lgtm

PR #119, opened 2023-08-25
Two lint fixes

cgwalters:
tests: Drop unnecessary `drop`

Newer rustc warns about this, dropping a reference does
nothing.  I think this code used to handle unmount via drop.

---

build-sys: Opt in to resolver = 2

This silences a warning from newer Rust.

---


jmarrero (approved):

cgwalters:
man the signoff thing is annoying

PR #126, opened 2023-09-16
Add version and timestamp to image status fields

cgwalters:
These are very practical things to want; the switch to a k8s-native API lost these user-facing ergonomics.


jmarrero (approved):
lgtm

PR #129, opened 2023-09-22
Support for `install-to-filesystem --replace=alongside`

cgwalters:
install: Trim bind mount information

I'm seeing this appear when targeting from a container image.

---

WIP: Support for `install-to-filesystem --replace=alongside`

I was trying to be really ambitious in https://github.com/containers/bootc/pull/78
for the full "takeover" path.  This is a *much* *much* simpler variant
where we just:

- Blow away and reinitialize the `/boot` and `/boot/efi` partitions
- Write inside the existing filesystem, leaving the OS running

Then when we reboot, we'll just need to clean up the old OS
state (or optionally leave it).

---

### DIFF
--- a/lib/src/bootloader.rs
+++ b/lib/src/bootloader.rs
@@ -75,7 +75,7 @@ pub(crate) fn install_via_bootupd(
     let bootfs = &rootfs.join("boot");
     let bootfs = Dir::open_ambient_dir(bootfs, cap_std::ambient_authority())?;
 
-    {
+    if super::install::ARCH_USES_EFI {
         let efidir = bootfs.open_dir("efi")?;
         install_grub2_efi(&efidir, &grub2_uuid_contents)?;
     }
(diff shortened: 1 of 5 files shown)


cgwalters:
Demo:

```
$ podman run --privileged -v /:/target --pid=host --net=none --security-opt label=type:install_t quay.io/cgwalters/ostest bootc install-to-filesystem --replace=alongside /target
```

cgwalters:
Notes to myself, script for testing this:

```
#!/bin/bash
set -xeuo pipefail
install /run/hostsrv/src/github/containers/bootc/target/release/bootc /usr/bin
podman run --rm --privileged --env BOOTC_BOOTLOADER_DEBUG=1 --pid=host -v /usr/bin/bootc:/usr/bin/bootc -v /:/target --security-opt label=type:install_t registry.gitlab.com/centos/cloud/sagano/fedora-boot-tier-0-dev:38 bootc install-to-filesystem --target-no-signature-verification --karg=console=ttyS0,115200n8 --replace=alongside /target
```

Note that we effectively bind mount in `/usr/bin/bootc` from the host devenv all the way into the container for faster iteration.

PR #133, opened 2023-09-26
upgrade: Wire up --check to new API

cgwalters:
This changes bootc to use the new API we added to ostree-ext, which rpm-ostree was also adapted to use.

A key advantage of this is that the update is cached and not lost across reboot.


jmarrero (approved):
lgtm

PR #134, opened 2023-09-27
Deps: Bump to latest kube/k8s-openapi

cgwalters:
On general principle; they trimmed their supported version set which reduces dep size.  But it also drops out some old things like `base64@0.13.1` (how many times does a base64 library really need to break semver?).


jmarrero:
looks like both k8s-openapi and k8s-client are providing base64 now? Also signed-off stuff :cry:

jmarrero (approved):
/lgtm

PR #135, opened 2023-09-29
install: Do use zipl on s390x

cgwalters:
This pairs with
https://github.com/ostreedev/ostree/pull/3059/commits/e3d93a85713dbc4728d6271ecd23e05c88cf54f3

What we need to add is something more like
`ostree admin init-fs --edition=2023` that flips on modern defaults.


jmarrero (approved):
lgtm

PR #136, opened 2023-09-29
install: Give better error if not run in a podman container

cgwalters:
(We should probably detect docker and error out nicely there
 too, or of course support docker too)


jmarrero (approved):
lgtm

PR #137, opened 2023-10-02
Add support for `--replace-mode=alongside` for ostree target

cgwalters:
Ironically our support for `--replace-mode=alongside` breaks when we're targeting an already extant ostree host, because when we first blow away the `/boot` directory, this means the ostree stack loses its knowledge that we're in a booted deployment, and will attempt to GC it...

https://github.com/ostreedev/ostree-rs-ext/pull/550/commits/8fa019bfa821303cfb7a7f069ae2320f4c3800fa is a key part of the fix for that.

However, a notable improvement we can do here is to grow this whole thing into a real "factory reset" mode, and this will be a compelling answer to
https://github.com/coreos/fedora-coreos-tracker/issues/399

To implement this though we need to support configuring the stateroot and not just hardcode `default`.

### DIFF
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -20,8 +20,8 @@ use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
 
-use anyhow::Ok;
 use anyhow::{anyhow, Context, Result};
+use anyhow::{ensure, Ok};
 use bootc_utils::CommandRunExt;
 use camino::Utf8Path;
 use camino::Utf8PathBuf;
@@ -576,7 +576,7 @@ pub(crate) fn print_configuration() -> Result<()> {
 }
 
 #[context("Creating ostree deployment")]
-async fn initialize_ostree_root(state: &State, root_setup: &RootSetup) -> Result<Storage> {
+async fn initialize_ostree_root(state: &State, root_setup: &RootSetup) -> Result<(Storage, bool)> {
     let sepolicy = state.load_policy()?;
     let sepolicy = sepolicy.as_ref();
     // Load a fd for the mounted target physical root
@@ -584,17 +584,27 @@ async fn initialize_ostree_root(state: &State, root_setup: &RootSetup) -> Result
     let rootfs = root_setup.rootfs.as_path();
     let cancellable = gio::Cancellable::NONE;
 
+    let stateroot = state.stateroot();
+
+    let has_ostree = rootfs_dir.try_exists("ostree/repo")?;
+    if !has_ostree {
+        Task::new_and_run(
+            "Initializing ostree layout",
+            "ostree",
+            ["admin", "init-fs", "--modern", rootfs.as_str()],
+        )?;
+    } else {
+        println!("Reusing extant ostree layout");
+
+        let path = ".".into();
+        let _ = crate::utils::open_dir_remount_rw(rootfs_dir, path)
+            .context("remounting target as read-write")?;
+        crate::utils::remove_immutability(rootfs_dir, path)?;
+    }
+
     // Ensure that the physical root is labeled.
     // Another implementation: https://github.com/coreos/coreos-assembler/blob/3cd3307904593b3a131b81567b13a4d0b6fe7c90/src/create_disk.sh#L295
     crate::lsm::ensure_dir_labeled(rootfs_dir, "", Some("/".into()), 0o755.into(), sepolicy)?;
-
-    let stateroot = state.stateroot();
-
-    Task::new_and_run(
-        "Initializing ostree layout",
-        "ostree",
-        ["admin", "init-fs", "--modern", rootfs.as_str()],
-    )?;
 
     // And also label /boot AKA xbootldr, if it exists
     let bootdir = rootfs.join("boot");
@@ -619,6 +629,11 @@ async fn initialize_ostree_root(state: &State, root_setup: &RootSetup) -> Result
     let sysroot = ostree::Sysroot::new(Some(&gio::File::for_path(rootfs)));
     sysroot.load(cancellable)?;
 
+    let stateroot_exists = rootfs_dir.try_exists(format!("ostree/deploy/{stateroot}"))?;
+    ensure!(
+        !stateroot_exists,
+        "Cannot redeploy over extant stateroot {stateroot}"
+    );
     sysroot
         .init_osname(stateroot, cancellable)
         .context("initializing stateroot")?;
@@ -649,7 +664,7 @@ async fn initialize_ostree_root(state: &State, root_setup: &RootSetup) -> Result
     let sysroot = ostree::Sysroot::new(Some(&gio::File::for_path(rootfs)));
     sysroot.load(cancellable)?;
     let sysroot = SysrootLock::new_from_sysroot(&sysroot).await?;
-    Storage::new(sysroot, &temp_run)
+    Ok((Storage::new(sysroot, &temp_run)?, has_ostree))
 }
 
 #[context("Creating ostree deployment")]
@@ -657,6 +672,7 @@ async fn install_container(
     state: &State,
     root_setup: &RootSetup,
     sysroot: &ostree::Sysroot,
+    has_ostree: bool,
 ) -> Result<(ostree::Deployment, InstallAleph)> {
     let sepolicy = state.load_policy()?;
     let sepolicy = sepolicy.as_ref();
@@ -749,6 +765,7 @@ async fn install_container(
     options.kargs = Some(kargs.as_slice());
     options.target_imgref = Some(&state.target_imgref);
     options.proxy_cfg = proxy_cfg;
+    options.no_clean = has_ostree;
     let imgstate = crate::utils::async_task_with_spinner(
         "Deploying container image",
         ostree_container::deploy::deploy(&sysroot, stateroot, &src_imageref, Some(options)),
@@ -1295,10 +1312,11 @@ async fn install_with_sysroot(
     sysroot: &Storage,
     boot_uuid: &str,
     bound_images: &[crate::boundimage::ResolvedBoundImage],
+    has_ostree: bool,
 ) -> Result<()> {
     // And actually set up the container in that root, returning a deployment and
     // the aleph state (see below).
-    let (_deployment, aleph) = install_container(state, rootfs, &sysroot).await?;
+    let (_deployment, aleph) = install_container(state, rootfs, &sysroot, has_ostree).await?;
     // Write the aleph data that captures the system state at the time of provisioning for aid in future debugging.
     rootfs
         .rootfs_fd
@@ -1359,6 +1377,12 @@ async fn install_to_filesystem_impl(state: &State, rootfs: &mut RootSetup) -> Re
         .ok_or_else(|| anyhow!("No uuid for boot/root"))?;
     tracing::debug!("boot uuid={boot_uuid}");
 
+    // If we're doing an alongside install, then the /dev bootupd sees needs to be the host's.
+    ensure!(
+        crate::mount::is_same_as_host(Utf8Path::new("/dev"))?,
+        "Missing /dev mount to host /dev"
+    );
+
     let bound_images = if state.config_opts.skip_bound_images {
         Vec::new()
     } else {
@@ -1379,8 +1403,16 @@ async fn install_to_filesystem_impl(state: &State, rootfs: &mut RootSetup) -> Re
 
     // Initialize the ostree sysroot (repo, stateroot, etc.)
     {
-        let sysroot = initialize_ostree_root(state, rootfs).await?;
-        install_with_sysroot(state, rootfs, &sysroot, &boot_uuid, &bound_images).await?;
+        let (sysroot, has_ostree) = initialize_ostree_root(state, rootfs).await?;
+        install_with_sysroot(
+            state,
+            rootfs,
+            &sysroot,
+            &boot_uuid,
+            &bound_images,
+            has_ostree,
+        )
+        .await?;
         // We must drop the sysroot here in order to close any open file
         // descriptors.
     }
@@ -1521,7 +1553,8 @@ fn remove_all_in_dir_no_xdev(d: &Dir) -> Result<()> {
 
 #[context("Removing boot directory content")]
 fn clean_boot_directories(rootfs: &Dir) -> Result<()> {
-    let bootdir = rootfs.open_dir(BOOT).context("Opening /boot")?;
+    let bootdir =
+        crate::utils::open_dir_remount_rw(rootfs, BOOT.into()).context("Opening /boot")?;
     // This should not remove /boot/efi note.
     remove_all_in_dir_no_xdev(&bootdir)?;
     if ARCH_USES_EFI {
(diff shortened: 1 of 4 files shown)


omertuc:
Sorry @cgwalters , accidentally pushed to your fork instead of mine

cgwalters:
I think you can just take over this PR too if you want, or open a new PR from your fork - either way.

omertuc:
Rebased. Without any changes, I'm facing an issue where in an ostree system, the mounted `/` on the host system is an overlay (`-v` mounted into `/:/target` and so the `findmnt` `source` for it is `overlay` rather than a `/dev/...` and so it trips up [lsblk](https://github.com/containers/bootc/blob/e4f00af8c5485612d0f1f05d6e31ee6766ff254e/lib/src/install.rs#L1735) later on

I'll see how I can tweak it so that it finds the right device

omertuc:
With additional `-v /sysroot:/target -v /sysroot:/target/sysroot` mounts and `--stateroot foo`, this seems to work

omertuc:
@cgwalters thoughts on the above mounts? Do we want to require them for install on ostree targets, or should I figure out a way to make this work without them, using just the already-documented install mounts (i.e. `/:/target`)?

omertuc:
> > and so the findmnt source for it is overlay rather than a /dev/... and so it trips up lsblk later on
> 
> We should learn how to peel that. This is really the same thing as [bugzilla.redhat.com/show_bug.cgi?id=2308594](https://bugzilla.redhat.com/show_bug.cgi?id=2308594) and [ostreedev/ostree#3198](https://github.com/ostreedev/ostree/issues/3198) and [containers/composefs#280](https://github.com/containers/composefs/issues/280)
> 
> Short term the simplest is the same logic as the grub patch - detect overlayfs for `/` and check if `/sysroot` exists and is mounted, if so use that.

OK. Changed it so that when the target rootfs is an overlay, we'll implicitly try targetting `<original_target>/sysroot` instead. It wasn't working at first and was a bit of a headache for me to debug because apparently if you mount `/:/target` then inside the container `/target/sysroot` is read-only by default, and so `ensure_dir_labeled` was failing, as opposed to when you mount `/sysroot:/target` directly, in which case it's not read-only. Took me a while to track that down chasing red herrings, and I'm still not sure who's responsible for this behavior (kernel? podman?), but after I realized it I simply moved `ensure_dir_label` to run only after your added `let _ = crate::utils::open_dir_remount_rw...` and then the rest just worked.

Current code might need a bit of touch-ups, but do you think the direction is right? Should I clean it up and undraft?

cgwalters:
> It wasn't working at first and was a bit of a headache for me to debug because apparently if you mount /:/target then inside the container /target/sysroot is read-only by default, and so ensure_dir_labeled was failing, as opposed to when you mount /sysroot:/target directly, in which case it's not read-only. 

I think that's possibly because it's bootc that's special casing mounting `/sysroot` read-write - that's how we do it outside of a container at least.

lib/src/install.rs
```diff
@@ -1336,6 +1352,19 @@ async fn install_to_filesystem_impl(state: &State, rootfs: &mut RootSetup) -> Re
         .ok_or_else(|| anyhow!("No uuid for boot/root"))?;
     tracing::debug!("boot uuid={boot_uuid}");
 
+    // If we're doing an alongside install, then the /dev bootupd sees needs to be the host's.
```
cgwalters, line 1355: I think we can drop this code now and hard require `-v /dev:/dev`

lib/src/install.rs
```diff
@@ -1589,12 +1627,35 @@ pub(crate) async fn install_to_filesystem(
     if !st.is_dir() {
         anyhow::bail!("Not a directory: {root_path}");
     }
+
+    let inspect = crate::mount::inspect_filesystem(&fsopts.root_path)?;
+
+    let alternative_root = fsopts.root_path.join("sysroot");
+    let root_path = match inspect.source.as_str() {
```
cgwalters, line 1634: I recently changed this in https://github.com/containers/composefs/pull/303

I think we should detect the *type* of the filesystem and not its source string.

(Actually also today, ostree records whether it was booted via composefs; there's a file we can read, but eh)

lib/src/lsm.rs
```diff
@@ -333,7 +333,6 @@ pub(crate) fn ensure_dir_labeled_recurse(
     Ok(())
 }
 
-/// A wrapper for creating a directory, also optionally setting a SELinux label.
```
cgwalters, line 336: Why drop the docs?

lib/src/utils.rs
```diff
@@ -55,6 +58,44 @@ pub(crate) fn find_mount_option<'a>(
         .next()
 }
 
+/// Try to (heuristically) determine if the provided path is a mount root.
+pub(crate) fn is_mountpoint(root: &Dir, path: &Utf8Path) -> Result<Option<bool>> {
```
cgwalters, line 62: Since I wrote this code originally I decided it made sense generically so let's update to use https://github.com/coreos/cap-std-ext/pull/56

cgwalters (commented):
Thanks for picking this up!!

lib/src/lsm.rs
omertuc: probably misplaced vim `dd`

lib/src/utils.rs
```diff
@@ -55,6 +58,21 @@ pub(crate) fn find_mount_option<'a>(
         .next()
 }
 
+/// Given a target directory, if it's a read-only mount, then remount it writable
+#[context("Opening {target} with writable mount")]
+pub(crate) fn open_dir_remount_rw(root: &Dir, target: &Utf8Path) -> Result<Dir> {
```
cgwalters, line 63: This needs a `#[cfg(feature = install)]` looks like

omertuc:
The current experience is:

```bash
echo "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIOTVytyhnSfX20smAsNKYG5Zpz6vSzDZu22S8PCDJ2Iw omer" > authkeys

sudo podman run --rm --privileged -v $PWD/authkeys:/authkeys -v /dev:/dev -v /var/lib/containers:/var/lib/containers -v /:/target --pid=host --security-opt label=type:unconfined_t -e RUST_LOG=trace quay.io/otuchfel/bootc:latest bootc install to-existing-root --replace alongside --acknowledge-destructive --stateroot foobar ----root-ssh-authorized-keys /authkeys
```

Following that, when I reset, it immediately boots into the new `foobar` stateroot, there are no grub boot entries for the original one:

image

I assume this is not our desired experience, and I should look into having the original boot entry preserved?

cgwalters:
It's a great question. I think for `install --replace=alongside`, that is indeed the expectation by default.

Let's then target merging this as is?

I think then what we should look at is weaving in this functionality into https://github.com/containers/bootc/issues/404 right?

That said we may end up wanting to expose this as something like `bootc install --retain` or something? At least in the ostree case it's trivial to do, we just don't blow away `/boot` :smile: 

The *super* messy thing is interoperating with non-ostree-ready bootloader setups so `--retain` would probably have to just error out in that case for now.

IOW:

- Let's merge as is
- Look at `--retain` as a followup
?

omertuc:
sgtm

cgwalters:
OK so we should have CI coverage here...I think it's actually as easy as doing another install after an initial one *without* doing the manual "wipe ostree" stuff we have in one of the install tests right?

omertuc:
> OK so we should have CI coverage here...I think it's actually as easy as doing another install after an initial one _without_ doing the manual "wipe ostree" stuff we have in one of the install tests right?

Not sure I agree, ideally we should actually boot into the installed system first? But getting that into the current harness might be a bit tricky

cgwalters:
Yes, testing with reboots is going to require some more infrastructure here.

omertuc:
> Yes, testing with reboots is going to require some more infrastructure here.

I've added a test but as suspected, the test passes even on `main`, so it's not very helpful without a reboot

cgwalters:
We talked about this and realized that while the new test passes, it would pass already today because it's actually the "install --stateroot" that makes it work.

The main fix we need here is preserving existing deployments when we detect we're booted via ostree.

So actually a way we could test this is via our tmt tests instead.

But in the end again I'm good to merge as is. Since I wrote this PR I can't approve it, you (or someone else) needs to do so.

omertuc:
>So actually a way we could test this is via our tmt tests instead.

Branching this to a separate issue \<link will follow\>

omertuc:
Up until now I've tested the PR on a random RHEL AI (bootc) system I had lying around, and it worked, but now I'm trying it on Silverblue and it fails. This seems to be due to the fact that / (/target inside the bootc container) on silverblue, despite being a btrfs rw mount, cannot be written to, so our rw remount is futile

cgwalters:
>  silverblue, despite being a btrfs rw mount, cannot be written to

What error do we get? `-EROFS`? What does `findmnt` look like when our container is executed? I wonder if we're messing up a remount?

omertuc:
> > silverblue, despite being a btrfs rw mount, cannot be written to
> 
> What error do we get? `-EROFS`? What does `findmnt` look like when our container is executed? I wonder if we're messing up a remount?

```
Reusing extant ostree layout
DEBUG Loaded SELinux policy: b06a119416deb696a80dbb67248b66d3829780c9e72d376a5089097ace6de5f8
DEBUG Target . is a mountpoint, remounting rw
DEBUG Labeling .
TRACE Label for . is Some("system_u:object_r:root_t:s0")
ERROR Installing to filesystem: Creating ostree deployment: fchmod: Operation not permitted (os error 1)
```

```
bash-5.1# findmnt -J -v --output=SOURCE,TARGET,MAJ:MIN,FSTYPE,OPTIONS,UUID --mountpoint /target
{
   "filesystems": [
      {
         "source": "/dev/vda3",
         "target": "/target",
         "maj:min": "0:32",
         "fstype": "btrfs",
         "options": "rw,relatime,seclabel,compress=zstd:1,discard=async,space_cache=v2,subvolid=258,subvol=/root",
         "uuid": "41f3f4ae-1242-4ef1-bcd8-fd8706f51738"
      }
   ]
}
```

cgwalters:
> ERROR Installing to filesystem: Creating ostree deployment: fchmod: Operation not permitted (os error 1)

Is there maybe a selinux denial here? Failing on `fchmod` is really odd.

Also hmm we must be missing some `.context()` here...are we failing on `ensure_dir_labeled()`? I bet that's it...but it'd be good to know for sure

cgwalters:
@omertuc I think we need to get more of your PRs merged, you're doing good work! This one specifically I guess I'm uncertain if we should block on the silverblue-reinstall case; I suspect it's btrfs specific.

But I may look at it today...

omertuc:
> @omertuc I think we need to get more of your PRs merged, you're doing good work! This one specifically I guess I'm uncertain if we should block on the silverblue-reinstall case; I suspect it's btrfs specific.
> 
> But I may look at it today...

Yeah I was downloading Fedora CoreOS (as opposed to Silverblue) today to see if we hit similar issues, I'll report with what I find

cgwalters:
One big conflict was around how the install phase got split up, and threading through `has_ostree` between those parts would have been annoying.

I did https://github.com/containers/bootc/pull/872 instead which is just dead code to start, but moves the "we detected an extant repo" into the "big bag of state" we already had in `RootSetup`.

omertuc:
stop stop stop, you're working on your old stale branch, I've already solved all these conflicts

omertuc:
Forced pushed now to solve some tiny conflict on import lines, other than that there's no conflicts

omertuc:
Forced push again because of duplicate import

omertuc:
>I suspect it's btrfs specific.

>Yeah I was downloading Fedora CoreOS (as opposed to Silverblue) today to see if we hit similar issues, I'll report with what I find

I just tried Fedora CoreOS and I can confirm the same issue happens even with xfs, so it's not a btrfs issue

cgwalters:
> stop stop stop 😆 you're working on your old stale branch, I've already solved all these conflicts a while back

Oops! Sorry

omertuc:
Got around to taking a serious look at this read-only behavior on FCOS/Silverblue, it seems to simply be due to an immutable attribute on the ostree deployment, which gets preserved when that is mounted directly only `/`.

On bootc systems I assume this attribute gets lost due to usage of composefs/overlay?

cgwalters:
Oh duh of course. 

> On bootc systems I assume this attribute gets lost due to usage of composefs/overlay instead of a direct mount

Kind of, it's not lost so much as shadowed. But basically with composefs we don't *need* the immutable bit anymore, it was always a hack.

Although wait, there's two immutable bits potentially - one on the deployment root, and one on the *physical* `/`.

Ahh yes, see https://github.com/coreos/coreos-assembler/blob/58df48e0c237a638df5b57a475d3b80b1029baa5/src/osbuild-manifests/coreos.osbuild.x86_64.mpp.yaml#L560 

Basically let's change our code to run `chattr -i` before the chmod. We could try to preserve it but I don't think it's important.

omertuc:
> Basically let's change our code to run `chattr -i` before the chmod. We could try to preserve it but I don't think it's important.

Yep, force pushed with rebase and this change. However we're now hitting another problem:

```
TRACE exec: "ostree" "config" "--repo" "ostree/repo" "set" "sysroot.bootloader" "none"
ERROR Installing to filesystem: Creating ostree deployment: Subprocess failed: ExitStatus(unix_wait_status(256))
```

Looking into what's behind this one

cgwalters:
>  They seem to be tied from what I can see:

This is confusing for sure! In ostree there's the term "physical root" vs "booted root". When we're in the booted system, then `/` is the deployment/booted root, and `/sysroot` is the physical root.

But mounting the filesystem from outside, `/` is the physical root, and `/ostree/deploy/<stateroot>/<checksum>.<id>` is the deployment root.

I am pretty sure what we're hitting here at first is the physical root, not the deployment root.

> ERROR Installing to filesystem: Creating ostree deployment: Subprocess failed: ExitStatus(unix_wait_status(256))

Well that's busted, we should be getting stderr from that...is there nothing there?

omertuc:
`strace -y` 

```
[pid  4103] openat(3</target/sysroot/ostree/repo>, ".", O_WRONLY|O_CLOEXEC|O_TMPFILE, 0600) = -1 EROFS (Read-only file system)
```

Looks like inside the container, `/target` and `/target/sysroot` are separate mounts that have to be rw remounted separately... Another thing that didn't happen in existing bootc systems

Looking into this

omertuc:
When remounting both, installation seems to proceed smoothly, so this seems to be the final hurdle

cgwalters:
> Looks like inside the container, /target and /target/sysroot are separate mounts that have to be rw remounted separately...

Ahh I think I understand the root of confusion here. I think early on we need to detect "is this an ostree system" and if so, then we need to actually look at `/target/sysroot` and basically do most of our operations in terms of that.

But for non-ostree systems then `/` is already the physical root.

The install code may need some cleanups around this. In the non-ostree case we can assume that the physical root `/` is writable (otherwise nothing could happen), and in the ostree case we shouldn't need to remount `/` writable (it doesn't make sense to do so with composefs); we need to peel it and access the physical root at `/sysroot`.

omertuc:
> > Looks like inside the container, /target and /target/sysroot are separate mounts that have to be rw remounted separately...
> 
> Ahh I think I understand the root of confusion here. I think early on we need to detect "is this an ostree system" and if so, then we need to actually look at `/target/sysroot` and basically do most of our operations in terms of that.
> 
> But for non-ostree systems then `/` is already the physical root.
> 
> The install code may need some cleanups around this. In the non-ostree case we can assume that the physical root `/` is writable (otherwise nothing could happen), and in the ostree case we shouldn't need to remount `/` writable (it doesn't make sense to do so with composefs); we need to peel it and access the physical root at `/sysroot`.

So basically what you're suggesting is to change your original suggestion [here](https://github.com/containers/bootc/pull/137#issuecomment-2389403509) from "detect overlayfs" to "detect ostree system" ?

cgwalters:
> So basically what you're suggesting is to change your original suggestion https://github.com/containers/bootc/pull/137#issuecomment-2389403509 from "detect overlayfs" to "detect ostree system" ?

Either way is OK by me honestly. For grub, it does feel a bit cleaner not to do anything specific to ostree but to just check for `/sysroot`. In an ideal world perhaps we even "standardize" this `/sysroot` path for other theoretical non-ostree-but-filesystem/composefs based systems...

But for bootc we have other code specific to ostree anyways that aren't going to go away anytime soon so "detect ostree" is something we'll need to do regardless, whether it's early on or a bit later in the flow.

lib/src/install.rs
```diff
@@ -576,26 +576,40 @@ pub(crate) fn print_configuration() -> Result<()> {
...
+        println!("Reusing extant ostree layout");
+
+        let path = ".".into();
+        let _ = crate::utils::open_dir_remount_rw(rootfs_dir, path)
+            .context("remounting target as read-write")?;
+        crate::utils::remove_immutability(rootfs_dir, path)?;
+
+        let path = "sysroot".into();
+        let _ = crate::utils::open_dir_remount_rw(rootfs_dir, path)
+            .context("remounting target sysroot as read-write")?;
```
cgwalters, line 606: This one won't exist in the non-ostree case, that's why the install test GHA is failing (those tests install overtop the GHA runner).

This gets to what I was saying in that I think we'll need to detect this early on dispatch on it.

I think concretely I'd say we check for `ostree/repo` and there should be a `bool existing_ostree` in the `State` structure or so?

omertuc: Yeah I see I'll rework that tomorrow

omertuc: >This one won't exist in the non-ostree case

This code is already inside a code branch that's os-tree only. I'd say the test failures are due to a lacking cleanup operation in `fn reset_root` - it only deletes the ostree deployments direction, not the entire ostree repo

omertuc (approved):

PR #139, opened 2023-10-11
lib: Include license files when packaging

cgwalters:
While we're not shipping this on crates.io and probably won't soon, it's useful to be able to run a `cargo package` just in case.  The Fedora rust2rpm tool wants license files in the crate tarball.


jmarrero (approved):
/lgtm

PR #144, opened 2023-10-13
install: Relax restriction on separate /boot

cgwalters:
For example, the current RHEL 9.2 VM images (KVM/cloud-init guest, and at least GCP) don't use a separate `/boot` by default (probably because they don't support LUKS for `/` like CoreOS systems do).

### DIFF
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -561,7 +561,9 @@ async fn initialize_ostree_root_from_self(
             .context("Opening etc/fstab")
             .map(BufWriter::new)?
     };
-    writeln!(f, "{}", root_setup.boot.to_fstab())?;
+    if let Some(boot) = root_setup.boot.as_ref() {
+        writeln!(f, "{}", boot.to_fstab())?;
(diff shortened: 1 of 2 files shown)


jmarrero (approved):
lgtm

lib/src/install.rs
```diff
@@ -627,9 +629,10 @@ pub(crate) struct RootSetup {
     device: Utf8PathBuf,
     rootfs: Utf8PathBuf,
     rootfs_fd: Dir,
+    rootfs_uuid: Option<String>,
     /// If true, do not try to remount the root read-only and flush the journal, etc.
     skip_finalize: bool,
-    boot: MountSpec,
+    boot: Option<MountSpec>,
```
cgwalters, line 635: Tangentially related...it's in *precisely* this kind of code refactoring that Rust is great.  The compiler found all the places I needed to change.

jmarrero: That is awesome indeed!

PR #149, opened 2023-10-18
Drop k8s-openapi and kube crates

cgwalters:
These are massive overkill for what we need now, and in particular k8s-openapi contains huge amounts of generated code.

Reimplement the tiny subset of k8s metadata we need.


jmarrero (approved):
lgtm! :astonished:

PR #155, opened 2023-10-23
 install: Don't require EFI if booted via BIOS and doing alongside

cgwalters:
install: Track `is_alongside`

We're going to need to dispatch other behavior on this around
the bootloader, so make it a generic thing.

---

install: Don't require EFI if booted via BIOS and doing alongside

Today Anaconda defaults to setting up the bootloader with
source == target mode; on x86_64 that means if booted via BIOS
the system is installed just with BIOS, and EFI if booted via EFI.

When we're doing an alongside install we hence cannot require
an ESP.

---


cgwalters:
So with this (building on [previous change](https://github.com/containers/bootc/pull/129)) I've verified that I can take a stock Fedora Workstation install (i.e. dnf based etc.) and building a container image that has the latest bootc installed with this code like:

```
FROM quay.io/fedora/fedora-silverblue:39
RUN rpm-ostree install https://download.copr.fedorainfracloud.org/results/rhcontainerbot/bootc/fedora-39-x86_64/06558679-bootc/bootc-202310231752.gd87aff8c54-1.fc39.x86_64.rpm && ostree container commit
```

(or obviously in the future we should ship `dnf` here...)

Then on the target system:

```
# dnf -y install skopeo
# podman run --privileged -v /:/target --pid=host --net=none --security-opt label=type:unconfined_t quay.io/you/yourdesktop:latest bootc install-to-filesystem --replace=alongside --target-no-signature-verification /target 
# systemctl reboot
```

And when you reboot, you'll be able to access all your files in `/sysroot`; for example if you have VM images they'll be in `/sysroot/var/lib/libvirt`.  Note that Workstation uses a btrfs subvolume for `/home` by default which *won't* be mounted in this model (unless you explicitly enabled it in your container image!).  But you can find the old subvolume information in:
```
$ grep /home /sysroot/etc/fstab
UUID=2f59332e-bb92-429f-aad6-baa6d0dcd801 /home                   btrfs   subvol=home,compress=zstd:1 0 0
```

Then mount it like e.g.
`mount -t btrfs -o subvol=home,compress=zstd:1 UUID=2f59332e-bb92-429f-aad6-baa6d0dcd801 /mnt/oldhome`

And you'll be able to copy over files from your old home directory, etc.

PR #160, opened 2023-10-27
install: Add more image metadata to aleph

cgwalters:
Motivated by https://issues.redhat.com/browse/COS-2526

It's quite possible that the image sha will have been GCd, but the version number and timestamp inside the image (not just the file timestamp) are probably sufficient for general reference.


cgwalters:
Motivation here is same as https://github.com/coreos/fedora-coreos-tracker/blob/main/internals/README-internals.md#aleph-version

cgwalters:
also xref https://github.com/openshift/machine-config-operator/pull/2994

vrothberg (approved):
LGTM

PR #164, opened 2023-10-31
README.md: Add missing link to internal doc

pablomh:


vrothberg (commented):
Thanks for contributing, @pablomh !

CI fails because the commit needs the Signed-off-by signature (https://github.com/containers/bootc/pull/164/checks?check_run_id=18221847947).  You can do that via `git commit -s --amend` and repushing.

pablomh:
I noticed :/ I've added it and it should be in a better shape :)

PR #166, opened 2023-10-31
Use `bootupctl --auto` in alongside mode

cgwalters:
This just leverages logic from
https://github.com/coreos/bootupd/pull/555
which simplifies things.

### DIFF
--- a/lib/src/bootloader.rs
+++ b/lib/src/bootloader.rs
@@ -1,5 +1,4 @@
 use std::os::unix::prelude::PermissionsExt;
-use std::path::Path;
 
 use anyhow::{Context, Result};
 use camino::Utf8Path;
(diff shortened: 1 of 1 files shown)


lib/src/bootloader.rs
```diff
@@ -37,18 +26,7 @@ pub(crate) fn install_via_bootupd(
     // to only doing that.  This is only on x86_64 because that's the only arch that has multiple
     // components right now.
     // TODO: Add --component=auto which moves this logic into bootupd
```
jlebon, line 28: Seems like we can remove this now.

jlebon (approved):
LGTM (stating the obvious: this needs to wait until bootupd has propagated enough).

cgwalters:
> (stating the obvious: this needs to wait until bootupd has propagated enough).

Splitting things up into separate git repositories on different cycles...man, the pain :cry:

PR #167, opened 2023-11-01
cli/edit: Minor tweaks

jlebon:
cli/edit: Don't error out if host spec unchanged

This matches e.g. `kubectl edit`.

---

cli/edit: Default to interactive edit

This better matches `kubectl edit`.


cgwalters (approved):
Thank you!

PR #170, opened 2023-11-02
cli: Correctly no-op unchanged updates

cgwalters:
Regression from previous rework around staged diffs; we were incorectly re-staging the same thing.


vrothberg (approved):
LGTM

PR #179, opened 2023-11-03
  systemd: New bootc-fetch-apply-updates.{timer,service} 

cgwalters:
Let's ship a baseline systemd unit that can be enabled for automatic updates.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,15 @@ all-test:
 install:
 	install -D -m 0755 -t $(DESTDIR)$(prefix)/bin target/release/bootc
 	install -d $(DESTDIR)$(prefix)/lib/bootc/install
+	# Support installing pre-generated man pages shipped in source tarball, to avoid
+	# a dependency on pandoc downstream
+	if test -d man; then install -D -m 0644 -t $(DESTDIR)$(prefix)/share/man/man5 man/*.5; fi
 	if test -d man; then install -D -m 0644 -t $(DESTDIR)$(prefix)/share/man/man8 man/*.8; fi
+
+# These are not installed by default; one recommendation is to put them in a separate
+# sub-package or sub-component.
+install-systemd-auto:
+	install -D -m 0644 -t $(DESTDIR)/$(prefix)/lib/systemd/system systemd/*.service systemd/*.timer
 
 bin-archive: all
 	$(MAKE) install DESTDIR=tmp-install && tar --zstd -C tmp-install -cf target/bootc.tar.zst . && rm tmp-install -rf
--- a/manpages-md-extra/bootc-fetch-apply-updates.service.md
+++ b/manpages-md-extra/bootc-fetch-apply-updates.service.md
@@ -1,0 +1,27 @@
+# NAME
+
+bootc-fetch-apply-updates.service
+
+# DESCRIPTION
+
+This service causes `bootc` to perform the following steps:
+
+- Check the source registry for an updated container image
+- If one is found, download it
+- Reboot
+
+This service also comes with a companion `bootc-fetch-apply-updates.timer`
+systemd unit.  The current default systemd timer shipped in the upstream
+project is enabled for daily updates.
+
+However, it is fully expected that different operating systems
+and distributions choose different defaults.
+
+## Customizing updates
+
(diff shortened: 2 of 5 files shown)


systemd/bootc-apply.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Apply bootc updates
+Documentation=man:bootc(8)
+ConditionPathExists=/run/ostree-booted
+
+[Service]
+Type=oneshot
+ExecStart=bootc update --apply --quiet
```
vrothberg, line 8:
```suggestion
ExecStart=/usr/bin/bootc update --apply --quiet
```

systemd/bootc-apply.service
```diff
@@ -0,0 +1,9 @@
...
+
```
vrothberg, line 9:
```suggestion
```

systemd/bootc-apply.service
vrothberg: Unless there's a reason not to, I prefer absolute paths to be explicit.

systemd/bootc-apply.service
jlebon, line 1: Minor: maybe `bootc-apply-updates.{service,timer}`? Feels like `apply` isn't descriptive enough.

jlebon: WFM!

cgwalters:
OK per https://github.com/containers/bootc/pull/179#discussion_r1383761082 let's think about this one a bit more.  We should perhaps also have `bootc-fetch.service` as an explicitly separated unit?

While this project is still classified as "interfaces subject to change", these particular units will be a really important "UX touchpoint" and it'd be good to get them as close to right as we think early on.

This whole thread really intersects with https://github.com/containers/bootc/issues/2 which is a tension I've felt since the start.

vrothberg:
> OK per [#179 (comment)](https://github.com/containers/bootc/pull/179#discussion_r1383761082) let's think about this one a bit more. We should perhaps also have `bootc-fetch.service` as an explicitly separated unit?

I wonder when such a unit would be useful?  Splitting fetching updates from applying them seems special.

If there's a use case to check for an update prior to executing it, callers could do an `upgrade --check` and then fire the unit.

PR as is LGTM

jlebon:
> > OK per [#179 (comment)](https://github.com/containers/bootc/pull/179#discussion_r1383761082) let's think about this one a bit more. We should perhaps also have `bootc-fetch.service` as an explicitly separated unit?
> 
> I wonder when such a unit would be useful? Splitting fetching updates from applying them seems special.

With [finalization locking](https://github.com/ostreedev/ostree/pull/1841), you can do all the steps ahead of time (fetch and deploy) and minimize down time to just updating the bootloader and restarting.

cgwalters:
Right, for locking let's reference https://github.com/ostreedev/ostree/issues/3025

cgwalters:
Actually one thing I'm really strongly opinionated on is that we should default to respecting systemd *blocking* inhibitors; see https://github.com/coreos/rpm-ostree/pull/2862

It allows the use case of `ssh root@examplemachine systemd-inhibit --mode=block bash` to debug something reliably.

I am also thinking here that alongside it'd make sense to add an explicit `bootc-apply.service` (and CLI verb that starts it?) that can have arbitrary hooks added.

For example, zincati today has code to invoke the classic [notify via tty](https://github.com/coreos/zincati/pull/479) code.  It's not clear to me that we should duplicate that in bootc, but we could easily support it externally via code that adds a systemd-drop-in that adds `ExecStartPre` or so?

cgwalters:
OK, did https://github.com/ostreedev/ostree/pull/3090

rhatdan:
Is this ready to merge?

cgwalters:
> Actually one thing I'm really strongly opinionated on is that we should default to respecting systemd blocking inhibitors; see https://github.com/coreos/rpm-ostree/pull/2862

We still don't do this...definitely in the nice to have.

cgwalters:
We're also not using finalization locking here yet.  Not fatal, because we're doing an "all-in-one" unit that does everything in one go, so the case of handling out-of-cycle reboots is a bit moot.

PR #180, opened 2023-11-06
minor doc tweaks

vrothberg:


docs/install.md
```diff
@@ -47,7 +47,7 @@ other options.
 Here's an example:
 
 ```
-$ podman run --privileged --pid=host --net=none --security-opt label=type:unconfined_t <image> bootc install --target-no-signature-verification /path/to/disk
+$ podman run --privileged --pid=host --net=none -v /path/to/disk:/target-root --security-opt label=type:unconfined_t <image> bootc install --target-no-signature-verification /target-root
```
cgwalters, line 50: Hmm, avoiding the need to do this was a goal.  Now I agree binding it is clearer...but if we're going to do that shouldn't we prefer `--device`?  But having all host devices is already implied by `--privileged`.

Also calling it `/target-root` is confusing because in this case it's really a block device, not a root filesystem.

cgwalters (approved):

PR #182, opened 2023-11-06
install: Verify target image fetch by default

cgwalters:
install: Compute target image reference upfront

Prep for verifying it before we do an install.

---

install: Verify target image fetch by default

Now that we've dropped the `--net=none` by default, let's
avoid two major footguns by verifying the target image specification
by default.

- Forgetting to use `--target-no-signature-verification` (most people are going to need this in demos right now)
- When the target OS requires an authenticated pull, but one didn't embed the pull secret in the target OS

---

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -59,6 +59,26 @@ The `--pid=host --security-opt label=type:unconfined_t` today
 make it more convenient for bootc to perform some privileged
 operations; in the future these requirement may be dropped.
 
+### "day 2" updates, security and fetch configuration
+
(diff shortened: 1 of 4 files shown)


cgwalters:
> Nov 06 19:20:24 qemu0 kola-runext-install[1401]: ERROR Verifying fetch: Preparing import: Fetching manifest: containers-policy.json specifies a default of `insecureAcceptAnything`; refusing usage

Hooray, CI successfully failed

jmarrero (approved):
/lgtm

PR #191, opened 2023-11-09
docs/install: Various minor fixes

cgwalters:
Most notable here is `s/install_t/unconfined_t` which is important for the alongside install.

While we're here, fix various minor markdown formatting lints.


vrothberg (approved):
LGTM

PR #193, opened 2023-11-10
install: Enable bootprefix by default

cgwalters:
This is the last bit necessary to enable installations to systems without a separate `/boot` partition (as is true in the default RHEL 9 AMIs).

I was too chicken to enable it by default in ostree long ago; and now here we are.  Should definitely enable it in ostree, but let's do it here now.

### DIFF
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -466,6 +466,9 @@ async fn initialize_ostree_root_from_self(
     };
     for (k, v) in [
         ("sysroot.bootloader", bootloader),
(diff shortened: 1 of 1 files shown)


vrothberg (approved):
LGTM

vrothberg (commented):
Any chance the property can be tested?

cgwalters:
Yep will followup, thanks

cgwalters:
@vrothberg promised followup in https://github.com/containers/bootc/pull/204

PR #194, opened 2023-11-10
docs/install: Also document using `--rm`

cgwalters:
There's no reason to leak these containers by default.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -47,7 +47,7 @@ other options.
 Here's an example:
 
 ```sh
-podman run --privileged --pid=host --security-opt label=type:unconfined_t <image> bootc install --target-no-signature-verification /path/to/disk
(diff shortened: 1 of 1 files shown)


vrothberg (approved):
LGTM

Did you consider integrating this call into bootc itself?  Maybe this could be hidden behind a flag?  Maybe something like `bootc --image=<image> ARGS...` which would be translated into `podman run $flags-that-I-keep-forgetting <image> ARGS...`

cgwalters:
That'd require the person invoking the tool to already have bootc in their $PATH which will be true for some but not others.  There's no bootc in e.g. the current RHEL 9 AMI.  Also if I want to *script* an in-place reprovision, I'll actually need to encode the command into something like an Ansible/Terraform config which generates my cloud-init configuration (potentially even doing dev work on a Mac or Windows env).

PR #197, opened 2023-11-10
lib: apply clippy fix for needless borrowing

miabbott:
See: https://rust-lang.github.io/rust-clippy/master/index.html#/needless_borrow

Applied `cargo clippy --fix --lib -p bootc-lib`


cgwalters (approved):
Honestly it's this clippy lint that made me give up on running clippy regularly...99% of the time it's noise and I'd lean more towards adding a default ignore for it.

But this is obviously correct as is too

vrothberg (commented):
Late LGTM
Trying to review as much as Rust as possible to get up to speed

PR #199, opened 2023-11-10
install: Update to new bootupd uuid/EFI code

cgwalters:
This significantly simplifies this code even more; bootupd takes care of writing the UUID files in both places (like coreos-installer does).

It also calls out to `efibootmgr` by default.


cgwalters:
OK, this one should be good to review.

vrothberg (commented):
@cgwalters shall we update the image references to the new repository/namespace/name?

PR #203, opened 2023-11-11
install: Re-implement hostns to work around supermin

cgwalters:
I'm experimenting with doing `bootc install` in a supermin VM like we do in coreos-assembler, but hit upon the fact that supermin's `/sbin/init` implementation incorrectly does `chroot()` instead of a `pivot_root`:

https://github.com/libguestfs/supermin/blob/5230e2c3cd07e82bd6431e871e239f7056bf25ad/init/init.c#L288

The problem here then is doing `nsenter` just takes us into the original rootfs, and all the real content is in `/root`.

Re-implement this by re-executing our own binary and doing the `setns()` and `chroot()` dance internally, because we can't really do it with external binaries.

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -29,7 +29,7 @@ liboverdrop = "0.1.0"
 once_cell = "1.9"
 openssl = "^0.10"
 # TODO drop this in favor of rustix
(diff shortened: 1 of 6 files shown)


cgwalters:
Filed https://github.com/libguestfs/supermin/issues/19

PR #206, opened 2023-11-13
install: Add `kargs` to installation config

cgwalters:
We know we eventually want "day 2" management of kargs, but supporting "install time" kargs in a somewhat container-native way will let us more properly set up things like the `console=` karg only for cloud images for example.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -96,20 +96,31 @@ in that case you will need to specify `--skip-fetch-check`.
 
 ### Operating system install configuration required
 
-The container image **MUST** define its default install configuration.  For example,
-create `/usr/lib/bootc/install/00-exampleos.toml` with the contents:
+The container image **MUST** define its default install configuration.  A key choice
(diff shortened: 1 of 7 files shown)


cgwalters:
Pairs with https://github.com/CentOS/centos-boot-layered/pull/6

cgwalters:
OK, tested this and it works!

docs/install.md
```diff
@@ -96,21 +96,32 @@ in that case you will need to specify `--skip-fetch-check`.
 
 ### Operating system install configuration required
 
-The container image **MUST** define its default install configuration.  For example,
-create `/usr/lib/bootc/install/00-exampleos.toml` with the contents:
+The container image **MUST** define its default install configuration.  A key choice
+that bootc by default lives up to the operating system image is the root filesystem
```
jlebon, line 100: leaves

jlebon (approved):
Seems sane to me!

PR #212, opened 2023-12-01
install: Drop `commit` from SourceInfo

cgwalters:
This is dead code right now; dropping it helps lower our dependency on ostree.


jmarrero (approved):
lgtm, but no signed off line 🖊️

PR #215, opened 2023-12-03
WIP: Use `podman pull` to fetch containers

cgwalters:
Prep in https://github.com/containers/bootc/pull/214Move pull code into deploy

---

WIP: Use `podman pull` to fetch containers

See https://github.com/containers/bootc/issues/147#issuecomment-1828751377

With this bootc starts to really gain support for a different backend
than ostree.  Here we basically just fork off `podman pull` to
fetch container images into an *alternative root* in
`/ostree/container-storage`,
(Because otherwise basic things like `podman image prune` would
 delete the OS image)

This is quite distinct from our use of `skopeo` in the ostree-ext project
because suddenly now we gain support for things
implemented in the containers/storage library like `zstd:chunked` and
OCI crypt.

*However*...today we still need to generate a final flattened
filesystem tree (and an ostree commit) in order to maintain
compatibilty with stuff in rpm-ostree.  (A corrollary to this is
we're not booting into a `podman mount` overlayfs stack)
Related to this, we also need to handle SELinux labeling.

Hence, we implement "layer squashing", and then do some final
"postprocessing" on the resulting image matching the same logic
that's done in ostree-ext such as `etc -> usr/etc` and handling `/var`.

Note this also really wants
https://github.com/ostreedev/ostree/pull/3106
to avoid duplicating disk space.

---


cgwalters:
Demo:

```
$ env RUST_LOG=debug /run/hostsrv/src/github/containers/bootc/target/release/bootc switch --backend container quay.io/cgwalters/ostest
DEBUG Re-executing current process for _ostree_unshared
DEBUG Already in a mount namespace
DEBUG Current security context is unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023
DEBUG Copying self to temporary file for re-exec
DEBUG Label for /tmp/.tmpnua4Vn is system_u:object_r:install_exec_t:s0
DEBUG Created "/tmp/.tmpnua4Vn"
DEBUG Re-executing _bootc_selinuxfs_mounted="/tmp/.tmpnua4Vn" "/tmp/.tmpnua4Vn" "switch" "--backend" "container" "quay.io/cgwalters/ostest"
DEBUG Already in a mount namespace
DEBUG Current security context is unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023
DEBUG Removing temporary file
DEBUG Pulling docker://quay.io/cgwalters/ostest
Trying to pull quay.io/cgwalters/ostest:latest...
Copying blob 797644653c72 skipped: already exists  
Copying blob 797644653c72 skipped: already exists  
Copying blob ef5675472650 skipped: already exists  
...
Writing manifest to image destination
DEBUG Creating temporary container for c864f541834eaef553a6e49d5a7da8f91a7fec26a7fdd9ce492a004499a4eef6
DEBUG Mounting 6674154c085cda1abf3f17b28c1fede55e74f975ad1acb0ec2d03252c2de340e
DEBUG Merging layer: ostree/container-storage/overlay/1cc5fa9dff7d8c9ce68cb1a635aede85960085b7bf5c6bce2ff91527fa0fb5ac/diff
DEBUG Merging layer: ostree/container-storage/overlay/f16810ac59f484691c15a8f186df29e54ac378e801972e3ca8416896bc882814/diff
DEBUG Merging layer: ostree/container-storage/overlay/705dd7ffd5fea45eee3dbe85688091c696f918746576419d1a22dfa2d666e8b3/diff
...
DEBUG Writing ostree commit
829754790d5c947c4816e15c3e07a11db72c7e64928e5620fa32b9a36f6c8409
Queued for next boot: ostree-image-signed:docker://quay.io/cgwalters/ostest
  Version: 39.20231123.1
  Digest: sha256:f7afa32ff5a924a9660c7ebd0c567cd37d9d85c6598ea08784961e13477d24e8
$
```

Now, we can also expose every single `podman` command but operating on our internal root, like:
```
$ /run/hostsrv/src/github/containers/bootc/target/release/bootc internal-podman images
REPOSITORY                TAG         IMAGE ID      CREATED       SIZE
quay.io/cgwalters/ostest  latest      c864f541834e  27 hours ago  1.24 GB
$
```

And also for example, we can optimize pushes and pulls between the bootc storage and the default `containers-storage` in `/var/lib/containers`.  (Particularly if they're on the same filesystem); i.e. we want something like a handy `bootc push-to-podman` as well as `bootc pull-from-podman` as sugar for effectively `skopeo copy containers-storage:[overlay@/ostree/containers-storage]:foo containers-storage:foo`.

lib/src/podman_ostree.rs
```diff
@@ -0,0 +1,272 @@
...
+
+/// The common container storage model is to store blobs (layers) as unpacked directories,
+/// and use the Linux `overlayfs` to merge them dynamically.
+///
+/// However, today the `ostree-prepare-root` model as used by ostree expects a final flattened
+/// filesystem tree.  (That said, there is work in ostree to use composefs, which will add
+/// a huge amount of flexibility)
+///
+/// This function is an "eager" implementation of computing the filesystem tree, implementing
+/// the same algorithm as overlayfs, including processing whiteouts.
```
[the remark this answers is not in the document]
cgwalters: I know that `podman image mount` exists because we're using it in this code.

The answer to your question is in the paragraph right above this one, but I will expand it further.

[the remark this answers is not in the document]
cgwalters: OK there's a bit more text now in the `podman_ostree.rs` module doc.

lib/src/ostree_authfile.rs
```diff
@@ -0,0 +1,72 @@
+//! # Copy of the ostree authfile bits as they're not public
```
cgwalters, line 1: Let's do https://github.com/ostreedev/ostree-rs-ext/pull/636

cgwalters: OK, https://github.com/containers/bootc/pull/581 merged

jeckersb:
Ok, I've pushed my rebased and seemingly-working fork onto the original here so I can continue to iterate here instead of off in my own world.  I know there are things that are still half-done or hacked-around that needs cleaned up, but this is at least something people can look at and build and play around with.

lib/src/deploy.rs
```diff
@@ -71,8 +103,14 @@ impl ImageState {
...
         repo: &ostree::Repo,
     ) -> Result<Option<ostree_ext::oci_spec::image::ImageManifest>> {
-        ostree_container::store::query_image_commit(repo, &self.ostree_commit)
-            .map(|v| Some(v.manifest))
+        match self.backend {
+            Backend::OstreeContainer => {
+                ostree_container::store::query_image_commit(repo, &self.ostree_commit)
+                    .map(|v| Some(v.manifest))
+            }
+            // TODO: Figure out if we can get the OCI manifest from podman
```
miabbott, line 111: I don't think it is possible to fetch a manifest with a `podman` command; you'd need to fall back to `skopeo inspect` or something else

cgwalters: We can continue to use https://github.com/containers/containers-image-proxy-rs/ for this stuff, see [this API](https://docs.rs/containers-image-proxy/latest/containers_image_proxy/struct.ImageProxy.html#method.fetch_manifest) which is what's used by the ostree-container code today.

We can consider extending the proxy API too with other things we need, as it's a reliable and tested bridge between the Rust and Go sides.

cgwalters: Actually also, we could fork `skopeo copy` and not `podman pull` too.  That may help keep things clearer even in the short term.

lib/src/podman_ostree.rs
```diff
@@ -0,0 +1,296 @@
...
+        if needs_create {
+            let mut db = DirBuilder::new();
+            db.mode(src_meta.mode());
+            output
+                .create_dir_with(&pathbuf, &db)
+                .with_context(|| format!("Creating {pathbuf:?}"))?;
+        }
+        // Now recurse
+        merge_layer(layer, pathbuf, output, state)?;
+    } else if (src_meta.mode() & libc::S_IFMT) == libc::S_IFCHR && src_meta.rdev() == 0 {
```
giuseppe, line 119: are opaque directories handled (or should they)?

cgwalters: I think you're right, we need to do that.  Huh.  So I did [this search for Go code referencing trusted.overlay.opaque](https://github.com/search?q=lang%3Ago+trusted.overlay.opaque&type=code), it's pretty interesting all the stuff out there...including [this windows tar2ext4 code](https://github.com/microsoft/hcsshim/blob/8beabacfc2d21767a07c20f8dd5f9f3932dbf305/ext4/tar2ext4/tar2ext4.go).  There's some interesting hits on the Rust side too, like [nydus](https://github.com/dragonflyoss/nydus/blob/master/builder/src/core/tree.rs#L34).

lib/src/podman_ostree.rs
```diff
@@ -0,0 +1,296 @@
...
+//! the filesystem tree; this is how it handles garbage collection, inspects metadata, etc.  Parts
+//! of bootc rely on this too today.
+//!
+//! ## Disadvantages
+//!
+//! One notable disadvantage of this model is that we're storing file *references* twice,
+//! which means the ostree deduplication is pointless.  In theory this is fixable by going back
+//! and changing the containers-storage files, but...
+//!
+//! ## Medium term: Unify containers-storage and ostree with composefs
```
cgwalters, line 24: One thought here...we could bite the bullet and add Go code to this project, and vendor c/storage and c/image.  It'd give us the ability to even handle patches for those if need be that we release before waiting for podman or skopeo.

It would also help clear the way for us to do things like copy the podman code for quadlet files to do stuff like https://github.com/containers/podman/issues/22785 right away, again without depending on podman changes in the short term.

Notable tradeoffs either way.  But, the more I think about it the more I lean that direction.  Does anyone have strong feelings about this?

lib/src/status.rs
```diff
@@ -138,41 +153,62 @@ pub(crate) fn create_imagestatus(
 
 /// Given an OSTree deployment, parse out metadata into our spec.
 #[context("Reading deployment metadata")]
-fn boot_entry_from_deployment(
+async fn boot_entry_from_deployment(
```
jeckersb, line 156: This whole function should probably be reworked, it's really hard to follow.

lib/src/status.rs
```diff
@@ -138,41 +153,62 @@ pub(crate) fn create_imagestatus(
...
+                        let rootfs = &cap_std_ext::cap_std::fs::Dir::reopen_dir(
+                            &crate::utils::sysroot_fd_borrowed(sysroot),
+                        )?;
+                        ImageState::from(podman::podman_inspect(rootfs, &image.image).await?)
+                    }
+                    Backend::OstreeContainer => {
+                        ImageState::from(*ostree_container::store::query_image_commit(repo, &csum)?)
+                    }
+                };
+                //let cached = imgstate.cached_update.map(|cached| {
```
jeckersb, line 183: I ignored this while moving things around, needs to be addressed properly.

lib/src/status.rs
```diff
@@ -250,30 +286,50 @@ pub(crate) fn get_status(
...
-        .map(|d| boot_entry_from_deployment(sysroot, d))
-        .transpose()
-        .context("Booted deployment")?;
-    let rollback = deployments
-        .rollback
-        .as_ref()
-        .map(|d| boot_entry_from_deployment(sysroot, d))
-        .transpose()
-        .context("Rollback deployment")?;
+    let staged = if let Some(d) = deployments.staged.as_ref() {
```
jeckersb, line 289: I don't love how this changed, but I couldn't figure out a better way to do it since `await` snuck in here and it doesn't play as well with the chaining+closures.

lib/src/cli.rs
```diff
@@ -439,7 +454,7 @@ async fn upgrade(opts: UpgradeOpts) -> Result<()> {
             }
         }
     } else {
-        let fetched = crate::deploy::pull(sysroot, imgref, opts.quiet).await?;
+        let fetched = crate::deploy::pull(sysroot, spec.backend, imgref, opts.quiet).await?;
```
jeckersb, line 457: In the branch above here where we handle `--check`, it doesn't know how to store cached updates properly for the podman backend.  This in turn means we don't have a good way to represent that in `bootc status` (it will always just be `None` currently)

jeckersb:
Something else that needs fixed:

`ERROR Status: Computing status: Booted deployment: Reading deployment metadata: Reading manifest data from commit: Missing ostree.manifest-digest metadata on merge commit`

I get that if I do a `hack/Containerfile` build, `podman-bootc run` that, then `bootc switch --backend container quay.io/fedora/fedora-bootc:40`, then reboot and finally `bootc status`.

lib/src/status.rs
```diff
@@ -138,41 +153,62 @@ pub(crate) fn create_imagestatus(
...
+                        timestamp: imgstate.created,
+                        image_digest: imgstate.manifest_digest,
+                    }),
+                    cached,
+                )
+            } else {
+                // The deployment isn't using a container image
+                (None, None)
+            };
+            (image, cached, incompatible, get_image_backend(origin)?)
```
jeckersb, line 201: `get_image_backend(origin)?` should just be `backend` from above

lib/src/cli.rs
cgwalters: We have a lot of choice for how to represent this; it strongly relates to the question of image GC though. I think today, podman and c/storage generally handle this by creating containers which hold references to images.

jeckersb: This also feels like more reason re: [your previous comment](https://github.com/containers/bootc/pull/215#discussion_r1629415493) to switch to using skopeo instead of podman, since we could use `skopeo inspect` to only fetch the metadata in the `--check` case.  If I'm following correctly from [the way ostree does it](https://github.com/ostreedev/ostree-rs-ext/pull/537) today it's similar in that it's just saving the manifest/config.

cgwalters: Yep

lib/src/podman_ostree.rs
vrothberg: I'd rather stick with the tools. The tools are always released in sync to make sure that no change to c/storage, for instance, causes any unexpected regression. OCP ran into issues because of using different versions of c/storage in CRI-O and Podman.

vrothberg:
Trying to catch up: Which items are open in this PR?

jeckersb:
> Trying to catch up: Which items are open in this PR?

Colin and I discussed this a bit last week before the holiday, the first thing I want to land is a prep change that adds the machinery for multiple backends, which initially will have just one implemented backend (the current "OstreeContainer" backend).  This will force the backend API to get fleshed out, as well as help decouple the backend bits from everything else so I can (hopefully) spend less time rebasing this as things change around it.  I'm going to draft a proposed backend API in a new PR that I will link here, and we can debate the design over there.

Once that's in, we finish cleaning up the ideas here into the second backend implementation.  This PR as-is implements the backend via `podman pull`, however above discussions indicate we probably want to swap that out for `skopeo copy/inspect` instead.  I think ideally those skopeo bits would live either in a reimagined [containers-image-proxy-rs](https://github.com/containers/containers-image-proxy-rs/) or a new sibliing `containers-storage-proxy-rs` library or similar.  For the short-term, in bootc we'll probably just exec out to `skopeo` similar to how this draft execs `podman`.  Switching to `skopeo` lets us handle caching update metadata via `skopeo inspect` without pulling down the payload the same way the current ostree backend works.

There's also the opaque directory issue [noted above](https://github.com/containers/bootc/pull/215#discussion_r1631066841), which I haven't given any thought to.

I think that covers the known outstanding issues?

cgwalters:
> I dreamed of getting rid of skopeo from bootc images. That would drop ~30MB from the image.

Honestly what's probably the most practical thing is to vendor the skopeo (and while we're here, buildah) code into podman, with a `podman-all-tools` package that ships those as symbolic links or so...

cgwalters:
> Friendly ping. Are we on track to get this in before RHEL 10?

To be clear the thing that has me feeling a bit stuck is in which order we try to do things. We could switch over to basically forking `podman pull` or `skopeo copy` into the existing "plain overlay" backend and then post-processing it as this PR is doing.

What'd be more work is vendoring c/storage and trying to get patches in to do some things more intelligently. Even more work: Try to aim for a unified composefs-oriented storage.

This latter is what I personally find the most interesting and exciting because I care more about "sealing" then I do zstd:chunked basically.

OTOH there's other knock-on huge benefits to having the bootc image in c/storage such as being able to seamlessly push and build on it.

The thing that hurts my head is that if we do c/storage first before aiming for composefs then we will very likely need to do *another* storage transition later. But that's also true for the default `/var/lib/containers/storage`.

anish:
This would be great. We are unable to bootc switch directly from ECR (have not had time to debug why), but a podman login + pull followed by `--transport containers-storage` works just fine

cgwalters:
> We are unable to bootc switch directly from ECR (have not had time to debug why)

See https://bootc-dev.github.io/bootc/building/secrets.html#using-a-credential-helper

lib/src/podman.rs
```diff
@@ -1,12 +1,151 @@
...
+        .args(["mount", cid])
+        .output()
+        .await?;
+    if !st.status.success() {
+        anyhow::bail!("Failed to mount transient image: {st:?}");
+    }
+    Ok(newline_trim_vec_to_string(st.stdout)?.into())
+}
+
+pub(crate) async fn podman_pull(
```
cgwalters, line 59: Don't need this one anymore it ended up in `podstorage.rs`

lib/src/podman.rs
```diff
@@ -1,12 +1,151 @@
+//! # Helpers for interacting with podman
```
cgwalters, line 1: Probably most of this should go in `podstorage.rs`? Maybe. Some of it is generic podman stuff

lib/src/podman_ostree.rs
```diff
@@ -0,0 +1,296 @@
...
+//! That said, there is work in ostree to use composefs, which will add a huge amount of flexibility;
+//! we can generate an erofs blob dynamically with the target labels.
+//!
+//! Even more than that however the ostree core currently expects an ostree commit object to be backing
+//! the filesystem tree; this is how it handles garbage collection, inspects metadata, etc.  Parts
+//! of bootc rely on this too today.
+//!
+//! ## Disadvantages
+//!
+//! One notable disadvantage of this model is that we're storing file *references* twice,
```
cgwalters, line 20: Have to fix this for prod, really just reflinking (if available) or hardlinking (for ext4) would be enough.

lib/src/podman_ostree.rs
```diff
@@ -0,0 +1,296 @@
...
+    })
+    .await??;
+
+    tracing::debug!("Writing ostree commit");
+    let repo_fd = Arc::new(sysroot.repo().dfd_borrow().try_clone_to_owned()?);
+    let ostree_ref = image_commit_ostree_ref(imageid);
+    let mut cmd = sync_cmd_in_root(&squashed, "ostree")?;
+    cmd.args([
+        "--repo=/proc/self/fd/3",
+        "commit",
```
cgwalters, line 273: This is the bit that would need to be changed to reflink from the source.

lib/src/status.rs
```diff
@@ -109,6 +114,16 @@ pub(crate) fn try_deserialize_timestamp(t: &str) -> Option<chrono::DateTime<chro
     }
 }
 
+pub(crate) fn get_image_backend(origin: &glib::KeyFile) -> Result<Backend> {
```
cgwalters, line 117: Want to think about a better way to configure this. Maybe something like `bootc image --enable-storage-sync` (and `--disable-storage-sync`).

lib/src/podman_ostree.rs
cgwalters: In fact I think *most* of this file would actually be better implemented in the short term (if we're going to do a copy) via just changing the existing fetcher to pull from our own `containers-storage:` instance!

lib/src/podman_ostree.rs
cgwalters: cc https://github.com/containers/container-libs/issues/197

PR #217, opened 2023-12-05
cli: Add build-commit

### LINKED ISSUE #216: Add `bootc build commit`
This should take over the role of `ostree container commit`, however following up on https://github.com/ostreedev/ostree-rs-ext/pull/569 we can change our thoughts on `/var`.

jmarrero:
Closes #216

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -212,6 +212,8 @@ pub(crate) enum Opt {
     #[clap(subcommand)]
     #[cfg(feature = "install")]
     Install(InstallOpts),
+    /// Validate non supported files are not present.
+    BuildCommit,
     /// Execute the given command in the host mount namespace
     #[cfg(feature = "install")]
     #[clap(hide = true)]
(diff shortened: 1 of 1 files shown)


jmarrero:
mm I think I need to use the ostree-ext crate directly instead of just calling ostree.

sallyom (commented):
@jmarrero do you want to convert to PR?

cgwalters:
So one big thing we could do here in general is a lot of what happens today in the rpm-ostree postprocessing phase; for example, a big one is handling `/etc/passwd` and generating sysusers fragments.

Another interesting one is kernel handling; erroring out if you have multiple kernels, cleaning up the duplicate content the Fedora kernel packages drop in `/boot` etc.

However...lately I'm also thinking just for maximum compatibility we should start to do that stuff on the client too automatically.

jmarrero:
Did not get it our of Draft because of background chats with Colin, sorry for not updating sooner. I think we wanted, to make this bootc native code instead of calling back to ostree to reduce the dep on ostree and on top of that the logic with /var changes needs to be revised.

What do you guys think about:
1) check for stuff in /tmp and error out
2) relax the /var errors make it warning 
3)  any other check?

cgwalters:
Or maybe what would be best is something like a generalized "hooks" framework, i.e. we have `/usr/lib/bootc/container-commit.d` that contains drop-ins that can execute at build time.  That would make everything easily extensible...and today we have sysusers code in rpm-ostree that we could keep where it is, and have e.g.

```
cat /usr/lib/bootc/container-commit.d/10-rpmostree-sysusers
#!/bin/bash
exec rpm-ostree postprocess sysusers
```
or so?

jmarrero:
I guess that would work, but we would need also to scan the dropins and skip anything calling binaries not present on the system? Or just fail with a warning? Because in your example we would fail in an image that does not have rpm-ostree.

PR #220, opened 2023-12-10
mount: Use task infra to clean up error handling

cgwalters:
We were missing the "copy child stderr to our stderr" bits here on error.

Motivated by this bit failing with an overlayfs root.


jmarrero (approved):
lgtm

PR #222, opened 2023-12-13
mount: Be compatible with older util-linux

cgwalters:
This is the same as
https://github.com/coreos/bootupd/commit/1457c2004766b09ee979d5457cc13a1383fd01bc and makes things work on c9s.


jmarrero (approved):
lgtm

PR #225, opened 2023-12-14
install: Don't retarget /tmp or /var/tmp if they're not overlayfs

cgwalters:
It may be that we're involved via a container flow where e.g. `/tmp` is already "properly" set up as a tmpfs.

In that case we don't need to do a dance in retargeting.

xref https://github.com/osbuild/bootc-image-builder/issues/18#issuecomment-1855886913


ondrejbudai (approved):
I've just tested this commit in an osbuild environment and it works fine:

```
TRACE Already have /tmp with f_type=16914836
TRACE Already have /var/tmp with f_type=2435016766
```

Thanks!

vrothberg (approved):
LGTM

PR #226, opened 2023-12-15
install: Rename `install` -> `install to-disk`, peer with `to-filesystem`

cgwalters:
install: Rename `install` -> `install to-disk`, peer with `to-filesystem`

Take the current `install` verb and change it to be
`bootc install to-disk`, and then `install-to-filesystem` becomes
`bootc install to-filesystem` - they are obvious peers.

The main motivation here is that in the end many use cases will
want nontrivial filesystem customization, and while I like having
a fully opinionated builtin flow to write to a target disk,
we should really think of `to-filesystem` as a fully equal peer
of `to-disk`.

---

install-to-disk: Verify target is a block device

I saw someone get confused and think `bootc install` could work
on a filesystem.

---

install-to-filesystem: Verify target is a dir+mountpoint

Similarly to previous patch for `install to-disk`, verify
that the target is a directory *and* that it's a mountpoint (we
can't sanely support installing to a subdirectory of a filesystem).

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -136,7 +136,7 @@ jobs:
         run: |
           set -xeuo pipefail
           sudo podman run --rm -ti --privileged -v /:/target -v ./usr/bin/bootc:/usr/bin/bootc --pid=host --security-opt label=disable \
-            quay.io/centos-bootc/fedora-bootc-dev:eln bootc install-to-filesystem --target-no-signature-verification \
+            quay.io/centos-bootc/fedora-bootc-dev:eln bootc install to-filesystem --target-no-signature-verification \
             --karg=foo=bar --disable-selinux --replace=alongside /target
           ls -al /boot/loader/
           sudo grep foo=bar /boot/loader/entries/*.conf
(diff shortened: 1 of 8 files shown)


sallyom (approved):
LGTM

jerpeter1:
Possible (more simple? ) alternative:  Keep `bootc install`, and add an error if `bootc install` targets a non-device, non-mountpoint.

cgwalters:
> Possible (more simple? ) alternative: Keep bootc install, and add an error if bootc install targets a non-device, non-mountpoint.

There's a bit more rationale to why to make `install` a standalone verb in
https://github.com/osbuild/bootc-image-builder/issues/18#issuecomment-1857580685

rhatdan:
Should we keep `bootc install` to disk and
bootc install --file-system 

If we expect one to happen more often then the other?

cgwalters:
> Should we keep bootc install to disk and bootc install --file-system

That doesn't feel very different from the status quo though...and more importantly the reason there's a separate verb (i.e. CLI command) today is that the two paths have a few mutually-incompatible extra arguments (e.g. `install` has `--filesystem` but obviously can't do that with `to-filesystem`; it's already made), and `to-filesystem` has `--root-mount-spec` which doesn't make sense when targeting a blockdev.

rhatdan (approved):
LGTM

PR #227, opened 2023-12-15
install: Add tracing and error context around re-exec

cgwalters:
I hit a bug here and the error context and debug tracing would have been helpful.


jmarrero (approved):
lgtm

PR #228, opened 2023-12-15
install: Two smaller fixes

cgwalters:
install: Propagate skopeo stderr

We're going to need to consistently use a helper to get this right
in the future...basically most uses of `.output()` are wrong.

---

install: Fix exec helper to not swallow unexpected args

Our `exec-in-host-mount-namespace` internal helper was intended
to be a "passthrough" that accepts whatever child arguments.

However I misunderstood how to do this in clap.  Before this
change we get:

```
$ bootc blabla
ERROR Re-exec in host mountns: Missing command
```

i.e. the exec code kicks in because `external_subcommand` takes
over *everything* unknown.

Fix things to use `trailing_var_arg` which is intended for this case.

After this patch:

```
$ bootc blabla
error: unrecognized subcommand 'blabla'

Usage: bootc <COMMAND>

For more information, try '--help'.
```

---

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -167,9 +167,10 @@ pub(crate) enum Opt {
     /// Execute the given command in the host mount namespace
     #[cfg(feature = "install")]
     #[clap(hide = true)]
-    #[command(external_subcommand)]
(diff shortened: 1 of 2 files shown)


jmarrero (approved):
lgtm

PR #230, opened 2023-12-15
Don't enforce container sigpolicy by default

cgwalters:
(Draft for discussion, also this depends on https://github.com/containers/bootc/pull/229 )

--

Don't enforce container sigpolicy by default

Closes: https://github.com/containers/bootc/issues/218

Basically this effort has not been really successful and adds
more pain than it solves.  We need to have a solution that works
for podman too.

In many scenarios, TLS is sufficient - or at least, we're far
from the only thing that if fetched from a compromised server
would result in a compromised system (e.g. privileged containers).

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -136,7 +136,7 @@ jobs:
         run: |
           set -xeuo pipefail
           sudo podman run --rm -ti --privileged -v /:/target -v ./usr/bin/bootc:/usr/bin/bootc --pid=host --security-opt label=disable \
-            quay.io/centos-bootc/fedora-bootc-dev:eln bootc install to-filesystem --target-no-signature-verification \
+            quay.io/centos-bootc/fedora-bootc-dev:eln bootc install to-filesystem \
             --karg=foo=bar --disable-selinux --replace=alongside /target
           ls -al /boot/loader/
           sudo grep foo=bar /boot/loader/entries/*.conf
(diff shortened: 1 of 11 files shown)


sallyom (commented):
I agree enforcing signature verification by default is causing more pain than good at this stage (and afaict everybody using bootc currently adds the --target-no-signature-verification). There's [this rpm-ostree issue](https://github.com/coreos/rpm-ostree/issues/4272) and general confusion about how to sign & verify. It's still easy enough to configure enforcing signature verification. This should be opt-in for now and enforcing by default should include clear documentation about how to correctly sign and verify (GPG, sigstore, keys, keyless).

vrothberg (approved):
I am not fan of opt-in security (signature verification).  Why would bootc behave differently than other github.com/containers tools?  If /etc/containers/policy.json is configured, I expect all tools to adhere to it. So this change worries me.

cgwalters:
> Why would bootc behave differently than other github.com/containers tools? If /etc/containers/policy.json is configured, I expect all tools to adhere to it. So this change worries me.

That's exactly it - *currently* (before this change) bootc does explicitly behave differently (to try to enforce a policy is enabled by default).  After, it doesn't.

cgwalters:
OK passing CI, and we now also drop this new default value out of the spec/status struct.

vrothberg:
Thanks, @cgwalters !  I must have misread the code.

jmpolom:
> There's https://github.com/coreos/rpm-ostree/issues/4272 and general confusion about how to sign & verify.

As someone who's tried to walk the path to adopt hands off automated container signing in a CI workflow I can say a lot of these are purely documentation issues. The upstream cosign docs are not great really. Improvements on that end whether directly to the upstream docs or if there were examples in bootc docs would be very helpful to golden path this for users and also show that this _is_ possible today.

PR #232, opened 2023-12-15
Use `Task` generally over `Command`

cgwalters:
Especially in places where we call `.output()`, what we generally always want is to get stdout, but keep stderr going to the parent process stderr.  The `Command` API doesn't make this obvious to do, and worse forces every caller to manually check the exit status.

The `Task` API fixes both of these.

(Now, the Task API also defaults to printing a description,
 which we need suppress in many cases with `.quiet()`)


HuijingHei (approved):
LGTM

vrothberg (approved):
LGTM

PR #236, opened 2023-12-18
Fix clippy warnings

jeckersb:


jeckersb:
Related, I noticed CONTRIBUTING.md explicitly asks for `cargo fmt` and `cargo clippy` to be run on submissions, but that the CI workflow is non-gating for clippy.  Do we want to make it gating to avoid requiring additional fixup patches like this in the future?

cgwalters (approved):

PR #237, opened 2023-12-19
lib: `#![allow(clippy::needless_borrow)]`

cgwalters:
In my experience this is far and away the clippy lint that is:

- Most common to trigger (especially when refactoring code)
- The least valuable to fix; there's no performance or correctness concerns really

### DIFF
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -12,6 +12,10 @@
 #![cfg_attr(feature = "dox", feature(doc_cfg))]
 #![deny(clippy::dbg_macro)]
 #![deny(clippy::todo)]
+// These two are in my experience the lints which are most likely
+// to trigger, and among the least valuable to fix.
+#![allow(clippy::needless_borrow)]
+#![allow(clippy::needless_borrows_for_generic_args)]
 
 pub mod cli;
 pub(crate) mod deploy;


jeckersb:
There's also `clippy::needless_borrows_for_generic_args` which is similar and also has an instance fixed by #236, so we might want to include both.

cgwalters:
> There's also clippy::needless_borrows_for_generic_args which is similar and also has an instance fixed by https://github.com/containers/bootc/pull/236, so we might want to include both.

Makes sense probably to add (as a followup to not rerun CI I think), but I don't think you're correct that there is an instance in that PR.

jeckersb:
> > There's also clippy::needless_borrows_for_generic_args which is similar and also has an instance fixed by #236, so we might want to include both.
> 
> Makes sense probably to add (as a followup to not rerun CI I think), but I don't think you're correct that there is an instance in that PR.

It's this one:

```
error: the borrowed expression implements the required traits
  --> lib/src/mount.rs:28:23
   |
28 |     let o = Task::new(&desc, "findmnt")
   |                       ^^^^^ help: change this to: `desc`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrows_for_generic_args
   = note: `-D clippy::needless-borrows-for-generic-args` implied by `-D clippy::style`
   = help: to override `-D clippy::style` add `#[allow(clippy::needless_borrows_for_generic_args)]`
```

cgwalters:
Ahh I figured it out, I was on 1.73, it looks like in 1.74 they split this lint into two.

jmarrero (approved):

cgwalters:
> Ahh I figured it out, I was on 1.73, it looks like in 1.74 they split this lint into two.

Stuff like this BTW is another reason not to hard block on "style" clippy lints; requiring every contributor to match the exact Rust version run in CI for clippy in the general case gets annoying.

PR #239, opened 2023-12-19
docs: Move manpages to separate toplevel

cgwalters:
I belatedly noticed that the previous change effectively "corrupted" the docs site by trying to include these generated man pages.

I think we can figure out how to integrate them better, but for now move them to a separate toplevel.

While we're here add one link from the main docs to the manpages.


dougsland:
LGTM

cgwalters:
I still haven't quite grokked how to build/view the docs locally, there's a big mess of ruby involved.

PR #243, opened 2023-12-19
Man improvements

cgwalters:
xtask: Fix man2markdown to use new location

We moved these in an earlier commit.

---

cli: Improve man pages

Add some more information into the man pages.

---


vrothberg (approved):
LGTM

PR #244, opened 2023-12-20
Add `HostType`, change `status` to always succeed

cgwalters:
This depends on https://github.com/containers/bootc/pull/241

---

Add `HostType`, change `status` to always succeed

Closes: https://github.com/containers/bootc/discussions/242

Basically it'd be useful for folks to be able to run `bootc status --json`
and have that always work - if we're not on a bootc-booted system
then we just output null data.

Specifically we drop the `isContainer` field and replace it
with an optional non-exhaustive enumeration.  For anything we don't
know about we just output `None` i.e. `null` in JSON.

---

### DIFF
--- a/lib/src/spec.rs
+++ b/lib/src/spec.rs
@@ -7,6 +7,8 @@ use crate::k8sapitypes;
 
 const API_VERSION: &str = "org.containers.bootc/v1alpha1";
 const KIND: &str = "BootcHost";
+/// The default object name we use; there's only one.
+pub(crate) const OBJECT_NAME: &str = "host";
 
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
 #[serde(rename_all = "camelCase")]
@@ -94,6 +96,16 @@ pub struct BootEntry {
     pub ostree: Option<BootEntryOstree>,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
+/// The detected type of running system.  Note that this is not exhaustive
+/// and new variants may be added in the future.
+pub enum HostType {
+    /// The current system is deployed in a bootc compatible way.
+    BootcHost,
+}
+
 /// The status of the host system
 #[derive(Debug, Clone, Serialize, Default, Deserialize, PartialEq, Eq, JsonSchema)]
 #[serde(rename_all = "camelCase")]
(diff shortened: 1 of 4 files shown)


cgwalters:
For reference on a non-bootc enabled host the output looks like

```
note: The format of this API is not yet stable
apiVersion: org.containers.bootc/v1alpha1
kind: BootcHost
metadata:
  name: host
spec:
  image: null
status:
  staged: null
  booted: null
  rollback: null
  type: null
```

With the important bit here being `type: null`.

Whereas on a bootc-enabled system it will look like e.g.:

```
apiVersion: org.containers.bootc/v1alpha1
kind: BootcHost
metadata:
  name: host
spec:
  image:
    image: quay.io/example/someos:latest
    transport: registry
status:
  # (omitted stuff)
  type: bootcHost
```

Though one corner case right now is on an ostree-but-not-bootc system (e.g. current FCOS,Fedora IoT) we show e.g.

```
apiVersion: org.containers.bootc/v1alpha1
kind: BootcHost
metadata:
  name: host
spec:
  image: null
status:
  staged: null
  booted:
    image: null
    incompatible: false
    pinned: false
    ostree:
      checksum: 380d2669f1726477cc6f28e07e0a7200855705d00f65a92f888a480ddd133d73
      deploySerial: 0
  rollback: null
  type: bootcHost
```

Which is probably wrong.  Will fix that.

beav:
This looks good to me. The logic would be something like:

* if bootc executable does not exist, its not bootc managed
* if bootc executable exists, execute and read `["status"]["type"]`. If it's null, its not bootc managed
* If you got here, its bootc managed. Read the rest of the status output for useful information

Does that sound right?

jeckersb:
>     * if bootc executable exists, execute and read `["status"]["type"]`. If it's null, its not bootc managed

Probably better to match on this being "bootcHost" instead of non-null, since under the current state this could also return "bootcContainer" (though you shouldn't see that from a "general observability" perspective).  But the more general concern is that the enumeration is non-exhaustive, so in the future could contain new states like (totally contrived example) bootcTransitioning where the system is in the process of becoming managed by bootc but not quite completed.  You probably don't want to consider that to be "bootc managed" yet.

beav:
> > ```
> > * if bootc executable exists, execute and read `["status"]["type"]`. If it's null, its not bootc managed
> > ```
> 
> Probably better to match on this being "bootcHost" instead of non-null

works for me

lib/src/spec.rs
```diff
@@ -94,6 +96,18 @@ pub struct BootEntry {
...
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
+/// The detected type of running system.  Note that this is not exhaustive
+/// and new variants may be added in the future.
+pub enum HostType {
+    /// The current system is deployed in a bootc compatible way.
+    BootcHost,
+    /// The current system is running as a container that is bootable.
+    BootcContainer,
```
cgwalters, line 108: The `cC` here "stutters" a bit i.e. it's "containerContainer" but I am not sure of a better term.

cgwalters: Hmmmm....yes, I think it is more elegant, but also more work to parse for the default/simple use cases.  Unsure.

cgwalters: Maybe what's really simpler is to drop the container detection out entirely.  It's not totally clear to me that it has a use case.  Yeah...let's do that.

vrothberg (approved):
LGTM

PR #246, opened 2024-01-02
cli: Document that check doesn't fetch layers

cgwalters:
Closes: https://github.com/containers/bootc/discussions/245


dougsland:
LGTM

PR #249, opened 2024-01-03
docs: Extend rpm-ostree relationship a bit

cgwalters:
Based on a question from a chat.

### DIFF
--- a/docs/relationship.md
+++ b/docs/relationship.md
@@ -38,16 +38,25 @@ Perhaps in the future we may actually support some kind of `Pod` analogue for re
 
 ## Relationship with rpm-ostree
 
-Today rpm-ostree directly links to `ostree-rs-ext`, and hence
-gains all the same container functionality.  This will likely
-continue.  For example, with rpm-ostree (or, perhaps re-framed as
-"dnf image"), it will continue to work to e.g. `dnf install`
-(i.e. `rpm-ostree install`) on the *client side* system.  However, `bootc upgrade` would
-(should) then error out as it will not understand how to upgrade
-the system.
+Today both bootc and rpm-ostree use the [ostree project](https://github.com/ostreedev/ostree-rs-ext)
+as a backing model.  Hence, when using a container source,
+`rpm-ostree upgrade` and `bootc upgrade` are effectively equivalent;
+you can use either command.
 
-rpm-ostree also has significant other features such as
-`rpm-ostree kargs` etc.
+However with rpm-ostree (or, perhaps re-framed as
+"dnf image"), it will continue to work to e.g. `dnf install`
+(i.e. `rpm-ostree install`) on the *client side* system.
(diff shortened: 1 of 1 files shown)


docs/relationship.md
```diff
@@ -38,16 +38,25 @@ Perhaps in the future we may actually support some kind of `Pod` analogue for re
...
-(i.e. `rpm-ostree install`) on the *client side* system.  However, `bootc upgrade` would
-(should) then error out as it will not understand how to upgrade
-the system.
+(i.e. `rpm-ostree install`) on the *client side* system.
+In addition there are other client-side mutation commands such as
+`rpm-ostree initramfs --enable`.
+
+However, as soon as you do something like this `bootc upgrade`
+then error out as it will not understand how to upgrade
+the system.  The bootc project currently takes a relatively
```
miabbott, line 54:
```suggestion
However, as soon as you mutate the system like this, `bootc upgrade`
will error out as it will not understand how to upgrade
the system.  The bootc project currently takes a relatively
```

Small rewording for clarity

rhatdan: s/will error/errors/g

docs/relationship.md
```diff
@@ -38,16 +38,25 @@ Perhaps in the future we may actually support some kind of `Pod` analogue for re
...
+hard stance that system state should come from a container image.
 
-rpm-ostree also has significant other features such as
-`rpm-ostree kargs` etc.
+The way kernel argument work also uses ostree on the backend
+in both cases, so using e.g. `rpm-ostree kargs` will also work
+on a system updating via bootc.
```
miabbott, line 59: I'm not fully understanding what is being said here...it seems to conflict with what was previously said about client-side mutation.  We want users to define kargs as part of the container image, correct?

cgwalters: Most of the time, yes.  But there are also valid use cases for machine-specific kernel arguments, so that continues to work.

jmarrero (approved):
lgtm

docs/relationship.md
```diff
@@ -38,16 +38,25 @@ Perhaps in the future we may actually support some kind of `Pod` analogue for re
...
 "dnf image"), it will continue to work to e.g. `dnf install`
-(i.e. `rpm-ostree install`) on the *client side* system.  However, `bootc upgrade` would
-(should) then error out as it will not understand how to upgrade
-the system.
+(i.e. `rpm-ostree install`) on the *client side* system.
+In addition there are other client-side mutation commands such as
+`rpm-ostree initramfs --enable`.
+
+However, as soon as you mutate the system in this way, `bootc upgrade`
+will error out as it will not understand how to upgrade
```
beav, line 53: does `bootc status` detect this state?

cgwalters: Yes, it's what the `incompatible: true` field represents today.

PR #252, opened 2024-01-05
docs: Add relationship-particles.md

cgwalters:
This original vision document from Lennart is great. Let's lay out a vision for why the bootc project exists and is different from this.

### DIFF
--- a/docs/relationship-particles.md
+++ b/docs/relationship-particles.md
@@ -1,0 +1,392 @@
+---
+nav_order: 6
+---
+
+# Relationship with systemd "particles"
+
+There is an excellent [vision blog entry](https://0pointer.net/blog/fitting-everything-together.html)
+that puts together a coherent picture for how a systemd (and [uapi-group.org](https://uapi-group.org/))
+oriented Linux based operating system can be put together, and the rationale for doing so.
+
+The "bootc vision" aligns with parts of this, but differs in emphasis and also some important technical details - and some of the emphasis and details have high level ramifications.  Simply stated: related but different.
+
+## System emphasis
+
+The "particle" proposal mentions that the desktop case is most
+interesting; the bootc belief is that servers are equally
+important and interesting.  In practice, this is not a real point
+of differentation, because the systemd project has done an excellent
+job in catering to all use cases (desktop, embedded, server) etc.
+
+An important aspect related to this is that the bootc project exists and must
+interact with many ecosystems, from "systemd-oriented Linux" to Android and
+Kubernetes. Hence, we would not explicitly compare with just ChromeOS, but also
+with e.g. [Kairos](https://kairos.io/) and many others.
+
+## Design goals
+
+Many of the toplevel design goals do overall align. It is clear that e.g.
+[Discoverable Disk Images](https://uapi-group.org/specifications/specs/discoverable_disk_image/)
+and [OCI images](https://github.com/opencontainers/image-spec) align on managing
+systems in an image-oriented fashion.
+
+### A difference on goal 11
+
+Goal 11 states:
+
+> Things should not require explicit installation. i.e. every image should be a live image. For installation it should be sufficient to dd an OS image onto disk.
+
+The `bootc install` approach is explicitly intending to support things such
+as e.g. static IP addresses provisioned via kernel arguments at install time;
+it is not a goal for installations to be equivalent to `dd`.  The bootc creator has experience with systems that install this way, and it creates practical problems in nontrivial scenarios such as ["Advanced Format"](https://en.wikipedia.org/wiki/Advanced_Format) disk drives, etc.
+
+### New Goal: An explicit alignment with cloud-native
+
+The bootc project has an explicit goal to to take formats, cues and inspiration
+from the container and cloud-native ecosystem.  More on this in several sections below.
+
+### New Goal: Continued explicit support for "unlocked" systems
+
+A strong emphasis of the particle approach is "sealed" systems that chain from Secure Boot.
+bootc aims to support the same.  And in practice, nothing in "particles" strictly
+requires Secure Boot etc.
+
+However, bootc has a stronger emphasis on continuing to support "unlocked"
+systems into the forseeable future in which key (even root level) operating system
+changes can be that are outside of an explicit signed state and feel
+*equally* first class, not just "developer system extensions".
+
+Or stated more simply, it will be explicitly supported to create bootc-based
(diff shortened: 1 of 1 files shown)


docs/relationship-particles.md
```diff
@@ -0,0 +1,386 @@
...
+job in catering to all use cases (desktop, embedded, server) etc.
+
+An important aspect related to this is that the bootc project exists and must
+interact with many ecosystems, from "systemd-oriented Linux" to Android and
+Kubernetes. Hence, we would not explicitly compare with just ChromeOS, but also
+with e.g. [Kairos](https://kairos.io/) and many others.
+
+## Design goals
+
+Many of the toplevel design goals do overall align.  It is clear that e.g. Discoverable Disk Images and container image (OCI) align on managing systems in an image-oriented fashion.
```
sallyom, line 28: add (DDI) and replace `container image` with `Open Container Initiative` - and link both `Discoverable Disk Images` and Open Container Initiative here.

cgwalters: Good suggestion, done

docs/relationship-particles.md
```diff
@@ -0,0 +1,389 @@
...
+
+## Boot Counting + Assessment
+
+This topic relates to the previous one; because of multiple bootloaders, there is not one single approach.  The systemd [automatic boot assessment](https://systemd.io/AUTOMATIC_BOOT_ASSESSMENT) is good where it can be used, but we also will support e.g. Android bootloaders.
+
+## Picking the Newest Version
+
+Because the storage of images is not just files or partitions, bootc will not exposeto the user/administrator a semantic of `strvercmp` or package-manager oriented versioning semantics.  Instead, the implementation of "latest" will be implemented in a more Kubernetes-oriented fashion of having "local" API objects with spec and status.  This makes it easy and obvious for higher level management (e.g. cluster)
```
miabbott, line 348: nit: 's/exposeto/expose to/`

docs/relationship-particles.md
```diff
@@ -0,0 +1,389 @@
...
+
+This is a key point of bootc: we aim for operating systems and distributions to ship their own bootc-compatible base images that can be used as a default derivation source.  These images are just OCI images that will follow simple rules (as mentioned above, the kernel is found in `/usr/lib/modules/$kver/vmlinuz`) for example for the extra state to boot.
+
+However in order to enable "sealed" systems (using signed composefs digests), the container build system will need support for this.  But, it is a goal to standardize the composefs metadata needed alongside the OCI, and to support this in the broader container ecosystem of tools (e.g. docker, podman) as well as bootc.
+
+## Final words
+
+This document is obviously very heavily inspired by [the original blog](https://0pointer.net/blog/fitting-everything-together.html).
+
+A point of diverge is that a goal of the bootc project *is* to strongly influence the existing operating systems and distributions and help them migrate their customers into an image-based world - and to make practical compromises in order to aid that goal.
```
miabbott, line 386: nit: `s/diverge/divergence/`

docs/relationship-particles.md
```diff
@@ -0,0 +1,392 @@
...
+that puts together a coherent picture for how a systemd (and [uapi-group.org](https://uapi-group.org/))
+oriented Linux based operating system can be put together, and the rationale for doing so.
+
+The "bootc vision" aligns with parts of this, but differs in emphasis and also some important technical details - and some of the emphasis and details have high level ramifications.  Simply stated: related but different.
+
+## System emphasis
+
+The "particle" proposal mentions that the desktop case is most
+interesting; the bootc belief is that servers are equally
+important and interesting.  In practice, this is not a real point
```
jeckersb, line 17: I've just finished reading the particle paper after this tipped me off to it (so thanks for educating me!).  I think this is an oversimplified take on what the paper is saying.  The key point I took away is that the desktop case is a superset of the complications that arise in server/embedded (maybe your point here is that you disagree with that assertion?).  So if you concentrate on the desktop case, then it follows that you also cover 100% of the server case as well (and then some).

ericcurtin (approved):
This reads great to me! There is some things that could be expanded on, like how OTA transport protocols differ between traditional ostree and bootc.

But this is super detailed already, I enjoyed reading it, it's a nice overview.

I know there is interest in the community in reducing the number of bytes that go over the air during an upgrade. So that's just one area in particular that springs to mind.

docs/relationship-particles.md
cgwalters: > (maybe your point here is that you disagree with that assertion?).

Yes, this.  There are some things that are hard on servers that do not exist on desktops in isolation; particularly "distributed systems" interactions.

docs/relationship-particles.md
```diff
@@ -0,0 +1,392 @@
...
+a bootloader that directly understands how to start a container image, for example.
+We've gotten as far as running a Linux userspace in the initial RAM disk, and the
+physical root filesystem is mounted.
+
+Here, we circle back to [composefs](https://github.com/containers/composefs).  One can
+think of composefs as effectively a way to manage something like dm-verity, but using
+files.
+
+What bootc builds on top of that is to target a specfic container image rootfs
+that is part of the "physical" root.  Today, this is implemented again using ostree, via the `ostree=` kernel commandline argument.  In the future, it is likely to be a `bootc.image`.
```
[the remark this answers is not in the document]
cgwalters: > I do try and lean against using kargs these days, in a world where you use a UKI secured thing, there's no difference between using /proc/cmdline and a file in /etc in the initrd. It sounds like we are leaning the same way here.

Yes, this is a good point.  In ostree today we go to some effort actually to avoid changing the initramfs when only userspace changes - and kernel args are really the only way for us to reliably pass state between generic bootloader setups and the initramfs.

But you're right I think in a sealed setup we should indeed not use kargs.

alexlarsson:
Overall I think this looks good to me. Maybe we should mention the optional approach of a transient /etc. I think that makes a lot of sense for many types of sealed systems.

rhatdan:
If we decide to add transient /etc info, lets do it in a different PR.

ericcurtin:
Question on this...

As of today is rpm-ostree a build time dependency of bootc?

Like how is the initrd built server side?

A naive search of "dracut" in the bootc repos, yield no results

ericcurtin:
I should have more clear:

As of today is rpm-ostree a build time dependency of a bootc image?

Thanks for the link @cgwalters , the answer is yes from reading https://containers.github.io/bootc/bootc-images/#building-bootc-compatible-base-images

but I didn't word my question right.

ericcurtin:
Ability to use everyday dnf at compose-time:

https://containers.github.io/bootc/bootc-images/#deriving-from-existing-base-images

rather than just in a temporary way via `usroverlay` is a really neat feature.

PR #253, opened 2024-01-05
labeler: Fix syntax

cgwalters:
I broke it earlier.


jmarrero (approved):

cgwalters:
Hm, it still errored out, but I think that's because the labeler action doesn't checkout, it runs on git main.

PR #256, opened 2024-01-11
upgrade: fix --apply when the desired deployment is already staged

jeckersb:
Previously, running:

`bootc update; bootc update --apply`

would result in the system not rebooting to apply the previously-staged deployment.

Closes: https://github.com/containers/bootc/issues/234

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -356,6 +356,10 @@ async fn upgrade(opts: UpgradeOpts) -> Result<()> {
             .unwrap_or_default();
         if staged_unchanged {
             println!("Staged update present, not changed.");
+
+            if opts.apply {
+                crate::reboot::reboot()?;
+            }
         } else if booted_unchanged {
             println!("No update available.")
         } else {


lib/src/cli.rs
```diff
@@ -356,6 +356,10 @@ async fn upgrade(opts: UpgradeOpts) -> Result<()> {
             .unwrap_or_default();
         if staged_unchanged {
             println!("Staged update present, not changed.");
+
+            if opts.apply {
```
cgwalters, line 360: Hmm....I need to think about this.  Is it that simple?  Yeah, I think that will work.   I'm OK with this though I do wonder if we should basically set `changed=true` in this case instead and just fall through.

If we do that, then maybe the `changed` variable is misnamed and it's really just `bool have_staged`.

(BTW we should probably drop `--touch-if-changed` since I think it's better done by querying the status API)

cgwalters (approved):

lib/src/cli.rs
jeckersb: FWIW I had the same reaction once I wrapped my head around things (is it this simple??).  Agreed with everything else you said though.  My only other thought was the bit in the help on `--apply` re: "In the future this command will detect the case where no kernel changes are queued, and perform a userspace-only restart." which will require some additional rearranging of things.  Maybe we just leave it for now and then it'll shake out in that future reorganization.

PR #257, opened 2024-01-12
cli: Drop `upgrade --touch-if-changed`

cgwalters:
We expect people to be parsing our spec/status via e.g. JSON.


jeckersb (approved):
lgtm

PR #258, opened 2024-01-12
hack/Containerfile: New file for local container builds

cgwalters:
This is a relatively efficient container-oriented build flow suitable for local iteration.

### DIFF
--- a/hack/Containerfile
+++ b/hack/Containerfile
@@ -1,0 +1,13 @@
+# This container build is just a demo effectively; it shows how one might
+# build bootc in a container flow, using Fedora ELN as the target.
+FROM quay.io/centos-bootc/fedora-bootc:eln as build
+RUN dnf config-manager --set-enabled eln-crb && dnf -y install cargo ostree-devel openssl-devel && dnf clean all
+COPY . /build
+WORKDIR /build
+# See https://www.reddit.com/r/rust/comments/126xeyx/exploring_the_problem_of_faster_cargo_docker/
+# We aren't using the full recommendations there, just the simple bits.
+RUN --mount=type=cache,target=/build/target --mount=type=cache,target=/var/roothome make bin-archive && mkdir -p /out && cp target/bootc.tar.zst /out
+
+FROM quay.io/centos-bootc/fedora-bootc:eln
+COPY --from=build /out/bootc.tar.zst /tmp
+RUN tar -C / --zstd -xvf /tmp/bootc.tar.zst && rm -vf /tmp/*


.dockerignore
```diff
@@ -1,2 +1,6 @@
 .cosa
+_kola_temp
+docs
+ci
```
jmarrero, line 4: I think this is breaking:
https://github.com/containers/bootc/blob/fad73f26a49eec58b3a150dd3579e9ed7d3a92d6/ci/Dockerfile.ci#L7

jmarrero (changes requested):

.dockerignore
cgwalters: Ah right, thanks for debugging this!  It's not actually important so dropped

cgwalters:
OK exciting...CI just fell over with 

```
error[E0308]: mismatched types
  --> /github/home/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rustix-0.38.29/src/backend/linux_raw/net/netdevice.rs:23:78
   |
23 |     unsafe { ifreq.ifr_ifrn.ifrn_name[..if_name_bytes.len()].copy_from_slice(if_name_bytes) };
   |                                                              --------------- ^^^^^^^^^^^^^ expected `&[i8]`, found `&[u8]`
   |                                                              |
   |                                                              arguments to this method are incorrect
   |
   = note: expected reference `&[i8]`
              found reference `&[u8]`
note: method defined here
  --> /builddir/build/BUILD/rustc-1.75.0-src/library/core/src/slice/mod.rs:3661:12

error[E0604]: only `u8` can be cast as `char`, not `i8`
  --> /github/home/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rustix-0.38.29/src/backend/linux_raw/net/netdevice.rs:47:22
   |
47 |             .map(|v| *v as char)
   |                      ^^^^^^^^^^ invalid cast
   |
help: try casting from `u8` instead
  --> /github/home/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rustix-0.38.29/src/backend/linux_raw/net/netdevice.rs:47:22
   |
47 |             .map(|v| *v as char)
   |                      ^^^^^^^^^^

Some errors have detailed explanations: E0308, E0604.
For more information about an error, try `rustc --explain E0308`.
error: could not compile `rustix` (lib) due to 2 previous errors
Error: Process completed with exit code 101.
```

which seems clearly unrelated to this PR...hmmm....I wonder if this is another obscure incremental compilation bug.

cgwalters:
Ah, this was https://github.com/bytecodealliance/rustix/commit/f546573ca5e66c0361f819068e213ade95afa1f7 but we aren't yet checking in `Cargo.lock`...perhaps it is time to do so.
/retest

PR #260, opened 2024-01-15
install: Support `to-disk --via-loopback`

cgwalters:
This is such a small amount of additional code that I think there's no good reason not to support it directly here.


cgwalters:
Motivated by working on https://github.com/containers/podman/issues/21243

runcom (commented):
LGTM

Issue #259, opened 2024-01-15
Broken link in docs (index.md)


larsks:
`docs/index.md` contains the following:

```
### Base images

Many users will be more interested in base (container) images.

For pre-built base images:

* [Fedora CoreOS](https://quay.io/repository/fedora/fedora-coreos) can be used as a base image; you will need to [enable bootc](https://github.com/coreos/rpm-ostree/blob/main/docs/bootc.md) there.
* There is also an in-development [centos-boot](https://github.com/centos/centos-boot) project.
```

The URL `https://github.com/coreos/rpm-ostree/blob/main/docs/bootc.md` leads to a 404 error.

PR #261, opened 2024-01-16
docs: Fix FCOS description

cgwalters:
We now have bootc packaged, so let's just tell people to use that.


vrothberg (approved):
LGTM

vrothberg:
Fixes: #259

PR #262, opened 2024-01-16
lsm: exit early if the process already has install_t

ondrejbudai:
There's no need to perform any additional steps if the bootc process already has install_t.

### DIFF
--- a/lib/src/lsm.rs
+++ b/lib/src/lsm.rs
@@ -84,6 +84,10 @@ impl Drop for SetEnforceGuard {
 #[context("Ensuring selinux install_t type")]
 #[cfg(feature = "install")]
 pub(crate) fn selinux_ensure_install_or_setenforce() -> Result<Option<SetEnforceGuard>> {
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
Thanks!

lib/src/lsm.rs
```diff
@@ -84,6 +84,10 @@ impl Drop for SetEnforceGuard {
 #[context("Ensuring selinux install_t type")]
 #[cfg(feature = "install")]
 pub(crate) fn selinux_ensure_install_or_setenforce() -> Result<Option<SetEnforceGuard>> {
+    // If the process already has install_t, exit early
+    if self_has_install_t()? {
+        return Ok(None);
+    }
     selinux_ensure_install()?;
     let current = std::fs::read_to_string("/proc/self/attr/current")
```
cgwalters, line 92: (Totally optional followup, we could hoist this to a `const`)

PR #263, opened 2024-01-16
install: add --source-imgref

ondrejbudai:
blockdev: fix recognizing parent loop devices

lsblk returns TYPE == loop for loop devices, modify the code
so loop devices are correctly recognized.

---

install: make the source digest optional

This is a prep work for the following commit. It will implement pulling
the container image from an explicitly given container reference. Thus,
there's no need to always fetch the digest because the caller is being
explicit about the container image, so let's make the field optional.

---

install: add --source-imgref

bootc install and install-to-filesystem currently rely on the fact that they
run inside a podman container. That's quite inconvenient for using bootc
for osbuild, because osbuild already run everything in a container.
While having a container in a container is surely possible, it gets quite
messy.

Instead of going this route, this commit implements a new --source-imgref
argument. --source-imgref accepts a container image reference (the same one
that skopeo uses). When --source-imgref is used, bootc doesn't escape the
container to fetch the container image from host's container storage. Instead,
the container image given by --source-imgref is used.

Even when running in this mode, bootc needs to run in a container created from
the same container image that is passed using --source-imgref. However, this
isn't a problem to do in osbuild. This really just removes the need for bootc
to escape the container to the host mount namespace.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -234,3 +234,18 @@ At the current time, leftover data in `/` is **NOT** automatically cleaned up.  
 be useful, because it allows the new image to automatically import data from the previous
 host system!  For example, things like SSH keys or container images can be copied
 and then deleted from the original.
+
+### Using `bootc install to-filesystem --source-imgref <imgref>`
+
+By default, `bootc install` has to be run inside a podman container. With this assumption,
+it can escape the container, find the source container image (including its layers) in
+the podman's container storage and use it to create the image.
+
+When `--source-imgref <imgref>` is given, `bootc` no longer assumes that it runs inside podman.
+Instead, the given container image reference (see [containers-transports(5)](https://github.com/containers/image/blob/main/docs/containers-transports.5.md)
+for accepted formats) is used to fetch the image. Note that `bootc install` still has to be
+run inside a chroot created from the container image. However, this allows users to use
+a different sandboxing tool (e.g. [bubblewrap](https://github.com/containers/bubblewrap)).
+
+This argument is mainly useful for 3rd-party tooling for building disk images from bootable
+containers (e.g. based on [osbuild](https://github.com/osbuild/osbuild)).   
--- a/lib/src/blockdev.rs
+++ b/lib/src/blockdev.rs
@@ -197,7 +197,7 @@ pub(crate) fn find_parent_devices(device: &str) -> Result<Vec<String>> {
         let kind = dev
             .get("TYPE")
             .with_context(|| format!("device in hierarchy of {device} missing TYPE"))?;
-        if kind == "disk" {
+        if kind == "disk" || kind == "loop" {
             parents.push(name.clone());
         } else if kind == "mpath" {
             parents.push(name.clone());
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -222,7 +222,7 @@ pub(crate) struct SourceInfo {
     /// Image reference we'll pull from (today always containers-storage: type)
     pub(crate) imageref: ostree_container::ImageReference,
     /// The digest to use for pulls
-    pub(crate) digest: String,
+    pub(crate) digest: Option<String>,
     /// Whether or not SELinux appears to be enabled in the source commit
     pub(crate) selinux: bool,
 }
@@ -384,7 +384,7 @@ impl SourceInfo {
         let selinux = crate::lsm::xattrs_have_selinux(&xattrs);
         Ok(Self {
             imageref,
-            digest,
+            digest: Some(digest),
             selinux,
         })
     }
@@ -572,8 +572,12 @@ async fn initialize_ostree_root_from_self(
     let mut temporary_dir = None;
     let src_imageref = if state.skopeo_supports_containers_storage {
         // We always use exactly the digest of the running image to ensure predictability.
-        let spec =
-            crate::utils::digested_pullspec(&state.source.imageref.name, &state.source.digest);
+        let digest = state
+            .source
+            .digest
+            .as_ref()
+            .ok_or_else(|| anyhow::anyhow!("Missing container image digest"))?;
+        let spec = crate::utils::digested_pullspec(&state.source.imageref.name, digest);
         ostree_container::ImageReference {
             transport: ostree_container::Transport::ContainerStorage,
             name: spec,
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -103,6 +103,18 @@ pub(crate) struct InstallTargetOpts {
 }
 
 #[derive(clap::Args, Debug, Clone, Serialize, Deserialize)]
+pub(crate) struct InstallSourceOpts {
+    /// Install the system from an explicitly given source.
+    ///
+    /// By default, bootc install and install-to-filesystem assumes that it runs in a podman container, and
+    /// it takes the container image to install from the podman's container registry.
+    /// If --source-imgref is given, bootc uses it as the installation source, instead of the behaviour explained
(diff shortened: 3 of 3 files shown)


lib/src/blockdev.rs
```diff
@@ -197,7 +197,7 @@ pub(crate) fn find_parent_devices(device: &str) -> Result<Vec<String>> {
         let kind = dev
             .get("TYPE")
             .with_context(|| format!("device in hierarchy of {device} missing TYPE"))?;
-        if kind == "disk" {
+        if kind == "disk" || kind == "loop" {
```
cgwalters, line 200: Ah yes.  It would probably make sense to try to use a shared crate for dealing with block devices at some point.

lib/src/skopeo.rs
```diff
@@ -0,0 +1,23 @@
...
+use std::process::Command;
+
+#[derive(Deserialize)]
+#[serde(rename_all = "PascalCase")]
+pub(crate) struct Inspect {
+    pub(crate) digest: String,
+}
+
+/// Given an image ID, return its manifest digest
+pub(crate) fn oci_archive_digest(imageref: &str) -> Result<String> {
```
cgwalters, line 12: This isn't specific to oci archives right?  If we want it to be, let's make this a `&Utf8Path`, and then explicitly specify `dir:` below say.

If we want to keep it general...we also already use https://docs.rs/containers-image-proxy/latest/containers_image_proxy/ which is a handy Rust client for talking to skopeo, but I don't really in the end have a problem with being practical and just forking it either.

lib/src/install.rs
```diff
@@ -562,28 +614,37 @@ async fn initialize_ostree_root_from_self(
...
+                    &state.source.digest,
+                );
+                ostree_container::ImageReference {
+                    transport: ostree_container::Transport::ContainerStorage,
+                    name: spec,
+                }
+            } else {
+                let td = tempfile::tempdir_in("/var/tmp")?;
+                let path: &Utf8Path = td.path().try_into().unwrap();
+                let r = copy_to_oci(&state.source.imageref, path)?;
```
cgwalters, line 634: (Unrelated to this change, but I would probably be fine just hard dropping this hacky "copy to oci" code that I don't think we need anymore since c9s has a new enough skopeo and that's fine from my PoV)

Totally optional of course...just reminded by the seeing the code again here.  If you want I can also do a prep PR removing it (or feel free to yourself), it seems like it would simplify this change.

lib/src/install.rs
```diff
@@ -368,6 +394,31 @@ impl SourceInfo {
...
+            digest,
+            Some(HostMountnsInfo {
+                skopeo_supports_containers_storage,
+            }),
+        )
+    }
+
+    #[context("Gathering source info from an external source")]
+    pub(crate) fn from_imageref(imageref: &str) -> Result<Self> {
+        let digest = crate::skopeo::oci_archive_digest(imageref)?;
```
cgwalters, line 411: Today, the ostree-container stack very much handles just being passed a non-digested pull spec; we'll do the digest resolution there. An alternative here is to make `digest` an `Option<T>` instead.  I have a mild preference for this, but I'm also fine with the code as is.

cgwalters (commented):
Cool, thanks for this!  Overall mostly LGTM, just some nits and optionals.

lib/src/skopeo.rs
cgwalters: (Forgot to mention, see below where we could also probably drop this code)

lib/src/install.rs
cgwalters: Ended up doing this in https://github.com/containers/bootc/pull/265 - but I still don't have a really strong opinion - we can merge this before that one if you prefer too.

vrothberg (commented):
I'd love to see docs on that as well.  Which `containers-transports` are supported?  I feel that the name `--source` is so generic that different users/personas will have a very different interpretation/feeling for that a source may be.

lib/src/skopeo.rs
ondrejbudai: Dropped (see below).

lib/src/install.rs
ondrejbudai: No objections from my side. #265 has a failing test suite currently, though. I think that rebasing #265 on top of this PR might be easier, so I prefer landing this one first. I must admit that it's a tad bit selfish view, though. :angel:

lib/src/install.rs
ondrejbudai: I hope I understood your comment correctly. I made `digest` an `Option<T>`. This loses a bit of "null-safety" (handled gracefully, ofc), but it allowed me to drop the whole `skopeo` crate. I have a backup of the old code if we change our midns.

ondrejbudai:
> I'd love to see docs on that as well. Which `containers-transports` are supported? I feel that the name `--source` is so generic that different users/personas will have a very different interpretation/feeling for that a source may be.

I renamed `--source` to `--source-imgref`, and added some documentation about it. Does it work for you?

docs/install.md
```diff
@@ -234,3 +234,18 @@ At the current time, leftover data in `/` is **NOT** automatically cleaned up.
 be useful, because it allows the new image to automatically import data from the previous
 host system!  For example, things like SSH keys or container images can be copied
 and then deleted from the original.
+
+### Using `bootc install to-filesystem --source-imgref <imgref>`
```
cgwalters, line 238: I'm good with docs...

lib/src/install.rs
```diff
@@ -102,6 +102,18 @@ pub(crate) struct InstallTargetOpts {
...
 
+#[derive(clap::Args, Debug, Clone, Serialize, Deserialize)]
+pub(crate) struct InstallSourceOpts {
+    /// Install the system from an explicitly given source.
+    ///
+    /// By default, bootc install and install-to-filesystem assumes that it runs in a podman container, and
+    /// it takes the container image to install from the podman's container registry.
+    /// If --source-imgref is given, bootc uses it as the installation source, instead of the behaviour explained
+    /// in the previous paragraph. See skopeo(1) for accepted formats.
+    #[clap(long, hide = true)]
```
cgwalters, line 113: But documenting seems to clash with `hide = true`.  Either it's something we support ~forever or it's a hidden thing that could be removed in theory.  I guess given the state, we should probably lift `hide = true` right?

cgwalters (approved):
LGTM, just one final nit, feel free to merge after addressing!

lib/src/install.rs
ondrejbudai: I removed `hide = true`.

docs/install.md
```diff
@@ -234,3 +234,18 @@ At the current time, leftover data in `/` is **NOT** automatically cleaned up.
...
+
+By default, `bootc install` has to be run inside a podman container. With this assumption,
+it can escape the container, find the source container image (including its layers) in
+the podman's container storage and use it to create the image.
+
+When `--source-imgref <imgref>` is given, `bootc` no longer assumes that it runs inside podman.
+Instead, the given container image reference (see [skopeo(1)](https://github.com/containers/skopeo/blob/main/docs/skopeo.1.md)
```
vrothberg, line 245: Can we point to https://github.com/containers/image/blob/main/docs/containers-transports.5.md instead?  Are all of the transports supported?

vrothberg: Thanks!  After quick glimpse at the skopeo-proxy code it seems it supports all transports.  @cgwalters, do you confirm?

cgwalters: Yep, it does

cgwalters (approved):

PR #265, opened 2024-01-17
install: Drop support for old skopeo

cgwalters:
Let's just hard require a skopeo that can fetch from `containers-storage`. Motivated by https://github.com/containers/bootc/pull/263 which was moving this code around.

### DIFF
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -245,14 +245,7 @@ pub(crate) struct SourceInfo {
     /// Whether or not SELinux appears to be enabled in the source commit
     pub(crate) selinux: bool,
     /// Whether the source is available in the host mount namespace
-    pub(crate) in_host_mountns: Option<HostMountnsInfo>,
-}
-
-/// Information about the host mount namespace
-#[derive(Debug, Clone)]
-pub(crate) struct HostMountnsInfo {
-    /// True if the skoepo on host supports containers-storage:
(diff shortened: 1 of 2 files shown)


lib/src/install.rs
```diff
@@ -927,6 +890,9 @@ async fn prepare_install(
 
     let skopeo_supports_containers_storage = skopeo_supports_containers_storage()
         .context("Failed to run skopeo (it currently must be installed in the host root)")?;
+    if !skopeo_supports_containers_storage {
+        anyhow::bail!("skopeo is too old");
```
vrothberg, line 894: I think  the `skopeo_supports_containers_storage` should error out with a message indicating the current / expected version of Skeop. "too old" will put the user on a journey figuring out the requirement.

vrothberg (approved):
The test claims that skopeo is too old.

cgwalters:
> The test claims that skopeo is too old.

Ah yes, ubuntu.  I will look at excising this test.

cgwalters:
OK, I did some tweaking of the GH/ubuntu CI to build skopeo from git, which will also conveniently give us some integration testing as things change there.

rhatdan:
LGTM

rhatdan:
@vrothberg PTAL

cgwalters:
I think the comments here are basically addressed.

PR #267, opened 2024-01-17
install: Add `--copy-etc`

cgwalters:
This allows injection of arbitrary config files from an external source into the target root.

This is pretty low tech...I'd really like to also support structured, cleanly "day 2" updatable configmaps, etc.

But there is simply no getting away from the generally wanting the ability to inject arbitrary machine-local external state today. It's the lowest common denominitator that applies across many use cases.

We're agnostic to *how* the data is provided; that could be fetched from cloud instance metadata, the hypervisor, a USB stick, config state provided for bootc-image-builder, etc.

Just one technical implementation point, we do handle SELinux labeling here in a consistent way at least.


cgwalters:
(OK yeah, need to wean this project off of FCOS for testing...gets into the messy topic of what other framework)

lib/src/install.rs
```diff
@@ -1058,6 +1093,63 @@ async fn prepare_install(
...
+                let mut db = DirBuilder::new();
+                db.mode(meta.mode());
+                let label = Labeler::new(sepolicy, path, meta.mode())?;
+                dest.create_dir_with(&path, &db)
+                    .with_context(|| format!("Creating {path:?}"))?;
+                drop(label);
+            }
+            r += copy_unmanaged_etc(sepolicy, src, dest, path)?;
+        } else {
+            dest.remove_file_optional(&path)?;
```
cgwalters, line 1132: Or, like Ignition should we default to failing when there is an extant file there?  That gets annoying though because suddenly one needs an extra override flag for when it's truly necessary.

cgwalters:
That said...hmm.  Maybe even without a registry, we should really still track this filesystem state in a somewhat more rigorous way by keeping each individual provided path as an extra hidden layer (ostree commit/ref), something like `bootc-<staterootname>-etc-<name>` or so.  Or more ideally in the future creating local composefs.  The more I think about this the more I like it because it means that we can at least always factory reset - and we can even remove things individually.

And now that we effectively have 3 sources for `/etc` content:

- upstream OS
- initial provisioning state
- current state

We can diff all three of these.  How exactly we make this visible gets tricky...and does still beg the question of supporting more useful metadata for them, and remote fetches.

cgwalters:
One technical difference here I am seeing vs systemd confext is that we definitely need to support configuring systemd units via this mechanism.

cgwalters:
FWIW, I rebased this and cleaned it up a bit.  I'm still not totally sure about it though.

PR #268, opened 2024-01-17
lsm: Deduplicate and comment a bit of the install_t code

cgwalters:
Just a followup to previous changes.


ondrejbudai (approved):
Thanks!

PR #269, opened 2024-01-17
Two systemd unit fixes

cgwalters:
build: Install systemd units by default

I was being too chicken; there's no reason not to install our
systemd units by default.  They're not enabled by default,
but we do want to encourage their use.

---

Fix manpage for service

We need `-s` for the pandoc invocation to render this right.

---


vrothberg (approved):
LGTM

PR #270, opened 2024-01-18
contrib/packaging: Updates for manpages and systemd unit

cgwalters:
The COPR build is failing due to this.
While we're here, also switch to autochangelog/autorelease.

Personally I think spec files upstream makes little sense, but COPR wants it.  Also for the files section,
it is tempting to copy/paste https://src.fedoraproject.org/rpms/rpm-ostree/blob/rawhide/f/rpm-ostree.spec#_191 here too.


vrothberg (approved):
LGTM

PR #272, opened 2024-01-18
install: Add `print-configuration`

cgwalters:
Even when an external process is creating the disk, we want it to be able to honor the root filesystem type specified in the container.


vrothberg (approved):
LGTM

runcom (approved):
lgtm

PR #273, opened 2024-01-19
docs/usage: New file

cgwalters:
Document offline USB upgrades.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -1,0 +1,30 @@
+---
+nav_order: 3
+---
+
+# Managing bootc systems
+
+## Performing offline updates via USB
+
+In a usage scenario where the operating system update is in a fully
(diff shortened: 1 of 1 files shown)


vrothberg (approved):
LGTM
Also a really cool feature!

cgwalters:
Just noting here for ourselves...right now we have this super simplistic model where everything is in a container image, and that's it.

But in the future, we will need a more sophisticated mirroring setup to add things like https://github.com/containers/bootc/issues/22 and https://github.com/containers/bootc/issues/128 - [like OpenShift](https://docs.openshift.com/container-platform/4.14/installing/disconnected_install/installing-mirroring-disconnected.html), we may end up needing something like `bootc image copy registry:quay.io/exampleos/someos:latest oci:example.oci`.

PR #274, opened 2024-01-19
switch: Be idempotent

cgwalters:
Instead of erroring out if the spec is unchanged, just print a notice.  This better matches Kubernetes default style.


cgwalters:
Looks like generic prow issues

sallyom (approved):
lgtm!

PR #275, opened 2024-01-19
install: Don't unshare if we have an external source

cgwalters:
This was being done in osbuild work to use `bootc install`, so we might as well make it automatic.  The environment variable was not intended as more than a debugging aid.


mvo5 (approved):
Thank you!

PR #276, opened 2024-01-19
install: Add more error contexts

cgwalters:
I forgot to bind mount `-v /:/target` and got an unfriendly bare
`ERROR No such file or directory (os error 2)`

Add more error context.

### DIFF
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -1080,6 +1080,7 @@ fn installation_complete() {
 }
 
 /// Implementation of the `bootc install to-disk` CLI command.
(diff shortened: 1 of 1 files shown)


vrothberg (approved):
LGTM

Interesting to see the differences between languages.  Go usually adds paths to error messages.

cgwalters:
Yeah, creating error messages like that requires heap allocations in general, and Rust doesn't do that by default.

cgwalters:
/override ci/prow/kola-run
Unrelated

PR #278, opened 2024-01-24
install: Only switch to loopback after re-exec

cgwalters:
blockdev: Add some debugging around loopback

---

install: Only switch to loopback after re-exec

I noticed we were leaking a loopback device, and the reason is
because the "re-exec self for selinux" dance.

We should really try to move that re-exec way earlier because
it's a big hazard for stuff like this right now.

This is a simple fix though that just moves the switch to
allocating the loopback device until after we've done the install
prep (including that re-exec).

---

### DIFF
--- a/lib/src/blockdev.rs
+++ b/lib/src/blockdev.rs
@@ -89,6 +89,7 @@ impl LoopbackDevice {
             .quiet()
             .read()?;
         let dev = Utf8PathBuf::from(dev.trim());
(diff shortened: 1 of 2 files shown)


cgwalters:
(ok now correctly resolved the merge conflict with https://github.com/containers/bootc/pull/277 )

jeckersb (approved):
lgtm

PR #282, opened 2024-01-24
README: provide an example to generate a booting image

mvo5:
To help people get an impression of `bootc` this commit adds an example how to install bootc to a disk-image and boot it with qemu.

Arguable this example is not complete as there is no way for the user to login in right now and maybe some example how to use the host would be nice but this might be a useful start.

[draft as it's mostly here to get a discussion going if this is useful and if so if it should be in the README or somewhere else, also we might want to discuss if we use the new --via-loopback instead of this manual setup and if we can put this into the auto-testing infra]

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,18 @@ which is used to boot.  At runtime on a target system, the base userspace is
 *not* itself running in a container by default.  For example, assuming
 systemd is in use, systemd acts as pid1 as usual - there's no "outer" process.
 
+# Example
+
+To try bootc today run:
+```
+$ truncate -s 10G test-disk.img
+$ sudo losetup -Pf test-disk.img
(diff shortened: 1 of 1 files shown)


mvo5:
Given that https://github.com/containers/bootc/pull/281 is already merged, I suppose this can be closed. I would want  to ask however if we can make this part of the README still maybe? To give people a quick success when they want to try it out?

cgwalters:
Remember bootc is intending to be a generic, distro/OS independent project; it's intended to be a *component* of the distro/OS (like podman); we can't in general be "the" frontend.  The container image you reference already has some docs which talk about stuff like this, so I think in general we should just point people at that.  Closing but without prejudice, I know this is a tricky topic and we should figure out how to generally include generic instructions here and link to specific downstreams.

PR #283, opened 2024-01-24
Port to ostree-ext 0.13, cap-std 3

cgwalters:
Main changes are

- Configuration object is no longer an `Option<T>`
- The metadata trait extensions now must be from cap-std


jeckersb (approved):
lgtm other than the bot being angry about missing signoff

PR #284, opened 2024-01-24
Fix install_t again

cgwalters:
lsm: Make setenforce 0 fallback require `BOOTC_SETENFORCE0_FALLBACK`

We shouldn't perform global system mutation without an opt-in.
As painful as it is.

---

lsm: Test if we have install_t capability

Hardcoding `install_t` is a bit ugly; maybe at some point
things change so that `spc_t` has `install_t` privileges.

Let's do a runtime check if we can set an invalid label; if
so then we're good.

---

lsm: Make a not-`nosuid` `/tmp`

This was the thing that was breaking our `unconfined_t` -> `install_t`
transition; the host `/tmp` is `nosuid`.  It simplifies things
here to just make our own, so do that.

---

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -293,7 +293,9 @@ pub(crate) async fn prepare_for_write() -> Result<()> {
     }
     ensure_self_unshared_mount_namespace().await?;
     if crate::lsm::selinux_enabled()? {
-        crate::lsm::selinux_ensure_install()?;
+        if !crate::lsm::selinux_ensure_install()? {
+            tracing::warn!("Do not have install_t capabilities");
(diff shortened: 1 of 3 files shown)


cgwalters:
I realized we were silently falling back to `setenforce 0` - this was motivated by trying to stop doing that by default.

cgwalters:
Another important thing in this change is that instead of trying to just check the security context is `install_t` even we actually *test* that it works by setting an unmapped security context.  So that's much more robust.

jeckersb (approved):
LGTM, I didn't know about the `/proc/1/root` trick, interesting!

PR #285, opened 2024-01-26
install: Automatically enable --generic-image when --via-loopback

cgwalters:
It's a bit confusing because `to-disk` is intended primarily for *per machine` installs.  However, when used with `--via-loopback` it can be used to make disk images which are probably intended to be generic.

For now, let's just automatically flip that on.


vrothberg (approved):

mvo5 (approved):
Thank you! I like tihs!

lib/src/install.rs
```diff
@@ -1093,13 +1093,18 @@ fn installation_complete() {
...
+pub(crate) async fn install_to_disk(mut opts: InstallToDiskOpts) -> Result<()> {
     let mut block_opts = opts.block_opts;
     let target_blockdev_meta = block_opts
         .device
         .metadata()
         .with_context(|| format!("Querying {}", &block_opts.device))?;
     if opts.via_loopback {
+        if !opts.config_opts.generic_image {
+            eprintln!("Automatically enabling --generic-image when installing via loopback");
+            std::thread::sleep(std::time::Duration::from_secs(2));
```
mvo5, line 1105: Just to double check, this sleep is here so that a user can quickly `ctrl-c` in case that is not what they want?

PR #286, opened 2024-01-26
install: Optionally use host mounted `/var/lib/containers`

cgwalters:
I just keep hitting the host skopeo requirement in corner cases; it's annoying because *otherwise* the container is self-sufficient. Change our installation instructions to add a `/var/lib/containers` bind mount.

For the time being of course we continue to support forking off `skopeo` on the host.

One thing I still want to investigate is dropping some requirements here and switch to *dynamically* setting up the mount points inside the container as is mentioned in https://brauner.io/2023/02/28/mounting-into-mount-namespaces.html but this currently requires relatively new host kernels.

As far as test coverage, this changes the Github Action that uses ubuntu and needed to build a newer skopeo to stop doing that, and in fact we explicitly *remove* skopeo to verify it's not being used in the install process.

I didn't change the other install tests to verify they keep working.

Closes: https://github.com/containers/bootc/issues/81

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,39 +116,13 @@ jobs:
         run: sudo tar -C / -xvf bootc.tar.zst
       - name: Integration tests
         run: bootc internal-tests run-container-integration
-  build-skopeo-ubuntu:
-    name: "Build skopeo git main for ubuntu"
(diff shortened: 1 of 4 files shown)


jmarrero (approved):
lgtm

PR #289, opened 2024-01-29
install: Generalize `root-fs-type` into `install.filesystem.root.type`

cgwalters:
Keep (but soft-deprecate) the existing `root-fs-type`, and add a more general set of tables in `install.filesystem.root`, with `type` as a field underneath that.

This somewhat resembles the [Image Builder blueprint](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/composing_a_customized_rhel_system_image/creating-system-images-with-composer-command-line-interface_composing-a-customized-rhel-system-image#composer-blueprint-format_creating-system-images-with-composer-command-line-interface) design.

In particular, this aims to leave space for
https://github.com/containers/bootc/issues/287
where we'd add e.g.

```
[install.filesystem.root]
extra = "5G"
```

for size specification.

Another obvious extension would be `options` to pass through options to `mkfs.$fs`; not clear to me we totally want to go there, but we clearly need something a bit more general.

### DIFF
--- a/lib/src/install/config.rs
+++ b/lib/src/install/config.rs
@@ -14,31 +14,114 @@ pub(crate) struct InstallConfigurationToplevel {
     pub(crate) install: Option<InstallConfiguration>,
 }
 
+/// Configuration for a filesystem
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+#[serde(deny_unknown_fields)]
+pub(crate) struct RootFS {
+    #[serde(rename = "type")]
+    pub(crate) fstype: Option<super::baseline::Filesystem>,
+}
+
+/// This structure should only define "system" or "basic" filesystems; we are
+/// not trying to generalize this into e.g. supporting `/var` or other ones.
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+#[serde(deny_unknown_fields)]
+pub(crate) struct BasicFilesystems {
+    pub(crate) root: Option<RootFS>,
+    // TODO allow configuration of these other filesystems too
+    // pub(crate) xbootldr: Option<FilesystemCustomization>,
+    // pub(crate) esp: Option<FilesystemCustomization>,
+}
+
 /// The serialized [install] section
 #[derive(Debug, Clone, Serialize, Deserialize, Default)]
 #[serde(rename = "install", rename_all = "kebab-case", deny_unknown_fields)]
 pub(crate) struct InstallConfiguration {
     /// Root filesystem type
     pub(crate) root_fs_type: Option<super::baseline::Filesystem>,
+    pub(crate) filesystem: Option<BasicFilesystems>,
     /// Kernel arguments, applied at installation time
     #[serde(skip_serializing_if = "Option::is_none")]
     pub(crate) kargs: Option<Vec<String>>,
 }
 
-impl InstallConfiguration {
-    /// Apply any values in other, overriding any existing values in `self`.
-    fn merge(&mut self, other: Self) {
-        fn mergeopt<T>(s: &mut Option<T>, o: Option<T>) {
-            if let Some(o) = o {
-                *s = Some(o);
+fn merge_basic<T>(s: &mut Option<T>, o: Option<T>) {
+    if let Some(o) = o {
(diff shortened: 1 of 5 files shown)


lib/src/install/config.rs
```diff
@@ -14,32 +14,120 @@ pub(crate) struct InstallConfigurationToplevel {
...
+    ///   both are set, then the latter takes precedence
+    pub(crate) fn canonicalize(&mut self) {
+        // New canonical form wins.
+        if let Some(rootfs_type) = self
+            .filesystem
+            .as_ref()
+            .and_then(|r| r.root.as_ref())
+            .and_then(|f| f.fstype.as_ref())
+        {
+            self.root_fs_type = Some(rootfs_type.clone())
```
jeckersb, line 119: clippy nit: the `Filesystem` enum is `Copy`, so drop the `clone()` and just make a copy with `Some(*rootfs_type)`.  Or maybe better drop `as_ref()` on `fstype` above, then it becomes just `Some(rootfs_type)`.

lib/src/install/config.rs
```diff
@@ -14,32 +14,120 @@ pub(crate) struct InstallConfigurationToplevel {
...
+        } else if let Some(rootfs) = self.root_fs_type.as_ref() {
+            let fs = self.filesystem.get_or_insert_with(|| Default::default());
+            let root = fs.root.get_or_insert_with(|| Default::default());
+            root.fstype = Some(rootfs.clone());
```
jeckersb, line 123: same as above, `Some(*rootfs)` instead to copy, or drop `as_ref()`

lib/src/install/config.rs
```diff
@@ -14,32 +14,120 @@ pub(crate) struct InstallConfigurationToplevel {
...
+        if let Some(rootfs_type) = self
+            .filesystem
+            .as_ref()
+            .and_then(|r| r.root.as_ref())
+            .and_then(|f| f.fstype.as_ref())
+        {
+            self.root_fs_type = Some(rootfs_type.clone())
+        } else if let Some(rootfs) = self.root_fs_type.as_ref() {
+            let fs = self.filesystem.get_or_insert_with(|| Default::default());
+            let root = fs.root.get_or_insert_with(|| Default::default());
```
jeckersb, line 122: These two can become `get_or_insert_with(Default::default)` to remove the unneeded closure

lib/src/install/config.rs
```diff
@@ -14,32 +14,120 @@ pub(crate) struct InstallConfigurationToplevel {
...
+    /// to ensure they're the same.
+    ///
+    /// - install.root-fs-type is synchronized with install.filesystems.root.type; if
+    ///   both are set, then the latter takes precedence
+    pub(crate) fn canonicalize(&mut self) {
+        // New canonical form wins.
+        if let Some(rootfs_type) = self
+            .filesystem
+            .as_ref()
+            .and_then(|r| r.root.as_ref())
```
jeckersb, line 116: this chain up to this line is the same as the new `filesystem_root` function below, so just call that instead

lib/src/install/config.rs
```diff
@@ -92,11 +182,20 @@ root-fs-type = "xfs"
     let other = InstallConfigurationToplevel {
         install: Some(InstallConfiguration {
             root_fs_type: Some(Filesystem::Ext4),
+            filesystem: None,
             kargs: None,
         }),
     };
     install.merge(other.install.unwrap());
-    assert_eq!(install.root_fs_type.unwrap(), Filesystem::Ext4);
```
jeckersb, line 99: This was probably fine as-is?  More of the `Copy` dance from before.

lib/src/install/config.rs
```diff
@@ -130,3 +230,35 @@ kargs = ["console=ttyS0", "foo=bar"]
         )
     )
 }
+
+#[test]
+fn test_parse_filesystems() {
```
jeckersb, line 235: :+1:

jeckersb (changes requested):
Looks great, just a few minor style/simplification nits

jeckersb (approved):
:+1:

PR #290, opened 2024-01-29
install: Ensure we label `/` (and `/boot`)

cgwalters:
This came out of a discussion with bootc-image-builder, which has this issue right now:
https://github.com/osbuild/bootc-image-builder/issues/149

As I noted in that issue, I think it's basically been working here because we always write to a real fresh filesystem, but let's be very explicit.

There's a notable tricky bootstrapping we're solving here around "what's the label of `/`" because we know we are running the target OS as a container image already.


jmarrero (approved):
lgtm

PR #293, opened 2024-01-31
install: Add BOOTC_SKIP_SELINUX_HOST_CHECK

cgwalters:
We're going to investigate supporting installing SELinux-enabled targets from a SELinux-disabled host.  This environment variable will allow bypassing the check.

xref https://github.com/ostreedev/ostree/pull/3151


mvo5 (approved):
Thank you!

PR #294, opened 2024-02-01
lsm: Forcibly relabel all initial state

cgwalters:
This is attempting to handle the "install selinux-enabled target from selinux-disabled host". We're kind of papering over effectively ostree design bugs here as in this case libostree itself should be setting the labels, but doing that is a bit hard and awkward right now.


bcl:
This seems like a good idea to me, but I'm not sure I have a good enough grasp of all the moving parts to ack it :) Also, is there any testing that would cover this, or does the current test suite passing inherently prove it doesn't break things?
Not sure what's up with the failing test, those look like rust errors.

cgwalters:
> Not sure what's up with the failing test, those look like rust errors.

Just to explain this a bit...the `bootc install` is a cargo feature that can be disabled.  Basically some operating systems/distros may have their own installer and don't want to support `bootc` carrying its own.

(That said increasingly I'm thinking at least `bootc install-to-filesystem` should always be on...but `bootc install-to-disk` is way more opinionated and something that one might want to replace)

cgwalters:
Actually forgot I'd started on this one here.  This is obsoleted by https://github.com/containers/bootc/pull/397

PR #295, opened 2024-02-01
Update bootc-fetch-apply-updates.timer with comments for RandomizedDe…

mrguitar:
…laySec=

RandomizedDelaySec= is the best kept secret of timers. Given that such few people know about it and it's very useful for large deployments, I think it makes sense to add it as a comment.

Thoughts about providing this out of the box as a comment? I'm not proposing we turn it on by default.

### DIFF
--- a/systemd/bootc-fetch-apply-updates.timer
+++ b/systemd/bootc-fetch-apply-updates.timer
@@ -7,6 +7,8 @@ ConditionPathExists=/run/ostree-booted
 OnBootSec=1h
 # This time is relatively arbitrary and obviously expected to be overridden/changed
 OnUnitInactiveSec=8h
+# When deploying a large number of systems, it may be beneficial to increase this value to help with load on the registry.
+RandomizedDelaySec=2h
 
 [Install]
 WantedBy=timers.target


jeckersb:
/ok-to-test

:+1: from me, reminds me of messing with cfengine splaytime many moons ago!

cgwalters (approved):
I'm in favor of turning it on by default actually...there's not really a strong downside right?

mrguitar:
@jeckersb exactly! I always describe this as "splay"!

@cgwalters, I can't think of a downside. Maybe the comment should change to this if we enable it: "When deploying a large number of systems, it may be beneficial to increase this value to help with load on the registry."

I *think* 2h is a sane default, but I just put that as an example. Should it change?

mrguitar:
ok, I tried to make that change, but since I suck at git it's stuck on DCO. :) This is what PM help looks like. lol

cgwalters:
See https://docs.pi-hole.net/guides/github/how-to-signoff/

TL;DR do `git commit --amend --signoff`

cgwalters:
I force pushed an update to this:

- squashed the commits
- Tweaked the commit message now that it's not a comment

PR #299, opened 2024-02-02
Add suggestion for `LABEL containers.bootc 1`

cgwalters:
As we aim to slowly move away from ostree, let's add a bootc label that we expect to be used.  For now, this is just a warning.

I will change rpm-ostree to add this by default if we take this.

But this one is also an important step for if we ever add the flow of simply deriving from an existing application base image.

### DIFF
--- a/docs/bootc-images.md
+++ b/docs/bootc-images.md
@@ -50,6 +50,16 @@ The requirement for both methods is that your initial treefile/manifest
 However, the ostree usage is an implementation detail
 and the requirement on this will be lifted in the future.
 
+## Standard metadata for bootc compatible images
+
+It is strongly recommended to do:
+
+```dockerfile
+LABEL containers.bootc 1
+```
+
+This will signal that this image is intended to be usable with `bootc`.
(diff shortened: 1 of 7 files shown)


jmarrero (approved):
lgtm

vrothberg (approved):
LGTM

cgwalters:
OK we learned from https://github.com/containers/bootc/issues/297 that our usage of `eprintln!` is problematic, so I changed this to use the systemd journal.

In the general case we're probably going to need to grow a logging abstraction a bit like what rpm-ostree has where we can write to a mix of stdout (when running as a systemd unit directly e.g. as part of `bootc-fetch-apply-updates.service` *or* when running under podman w/out systemd) or to the journal (when running outside of a unit, but on a booted host, which is the current default interactive case, but cc https://github.com/containers/bootc/issues/2 )

cgwalters:
Also did https://github.com/ostreedev/ostree-rs-ext/pull/595 related to this

PR #301, opened 2024-02-05
deploy: Log filtered content warning to journal, not stderr

cgwalters:
This one is really a "soft warning"...it's not a serious problem to have stuff in `/run`, it's just ignored.

At the same time, we want to persist it so that someone debugging things later can find the information.


jeckersb (approved):
lgtm

PR #304, opened 2024-02-08
install: Enable composefs by default, add a `disable_composefs` option

cgwalters:
Let's enable this by default now, and allow disabling it if desired.

### DIFF
--- a/manpages-md-extra/bootc-install-config.md
+++ b/manpages-md-extra/bootc-install-config.md
@@ -18,9 +18,10 @@ that can be overridden in a derived container image.
 
 This is the only defined toplevel table.
 
-The `install`` section supports two subfields:
+The `install`` section supports the following subfields:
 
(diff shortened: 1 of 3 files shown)


manpages-md-extra/bootc-install-config.md
```diff
@@ -18,9 +18,10 @@ that can be overridden in a derived container image.
 
 This is the only defined toplevel table.
 
-The `install`` section supports two subfields:
+The `install`` section supports the following subfields:
 
 - `filesystem`: See below.
+- `disable_composefs`: A boolean, which will use the legacy ostree mode.
```
ondrejbudai, line 24: I think this needs more explanation, and maybe even a better name. Where exactly is being `composefs` disabled? In the kernel? In the filesystem layout in the root mount namespace? Might be also worth explaining what's the legacy ostree mode.

PR #318, opened 2024-02-09
install: Drop setting immutable bit on physical /

cgwalters:
I've gone back and forth on this one over time; lately my thought is that conceptually `/` is stuff that bootc/ostree doesn't actually own, and so we shouldn't add this bit.

For a use case of e.g. testing out bootc via an "alongside" install, we would want to enable rolling back to the non-bootc state (at some point in the future when we e.g. save the original `/boot`) and the immutable bit would be a thing we'd then need to undo.

The readonly bind mount over `/sysroot` should be sufficient for our use cases.


jlebon:
This matches https://github.com/coreos/coreos-assembler/pull/3693 too.

PR #327, opened 2024-02-09
hack: Allow injecting stuff into target/dev-rootfs

cgwalters:
Specifically, I want to hack on both e.g. ostree and bootc at the same time.  With this, I can do (from ostree):

```bash
$ make install DESTDIR=/path/to/bootc/target/dev-rootfs
```

Then from this repo, running a container build will get me both.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,3 @@
 .cosa
 target
+!target/dev-rootfs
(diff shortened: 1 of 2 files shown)


jmarrero (approved):
lgtm

PR #338, opened 2024-02-12
Update to ostree-ext 0.13.2

cgwalters:
I decided to just do a release anyways, so we don't need to track git.

This (effectively) reverts commit 79295cedaf0135c9e9dd5505575921d8a46d45ba.


jmarrero (approved):
lgtm

PR #339, opened 2024-02-13
Don't output sigverify+transport for default

cgwalters:
Same motivation as https://github.com/ostreedev/ostree-rs-ext/pull/604

Because fetching from a registry with the default sigverify is the 90% case default, when we see this (and are formatting for human consumption), use the "alternate" formatting to just display the image name.

This makes the ostree stuff *much* less in the user's face in the default path.


ckyrouac:
lgtm, I'm hesitant to click merge though.

PR #340, opened 2024-02-13
Make `--disable-selinux` work on hosts that have `selinux=0`

### LINKED ISSUE #303: bootc install to-disk --disable-selinux does not disable selinux when the host has selinux disabled
The purpose of the `--disable-selinux` command is to enable the creation of an image with selinux enabled from a host with selinux disabled. On my Fedora 39 machine, when I set selinux to disabled or permissive, [this selinux_enabled check](https://github.com/containers/bootc/blob/8969314f224754b6bb282afc21a3235c08298bb4/lib/src/lsm.rs#L27) still returns true. This results in an image with broken selinux that won't fully boot. Would checking the output of getenforce be more reliable?

ckyrouac:
Prior to this fix the selinux_enabled function would always return true, even if selinux was set to disabled or permissive.

fixed #303

### DIFF
--- a/lib/src/lsm.rs
+++ b/lib/src/lsm.rs
@@ -24,8 +24,9 @@ const SELF_CURRENT: &str = "/proc/self/attr/current";
 
 #[context("Querying selinux availability")]
 pub(crate) fn selinux_enabled() -> Result<bool> {
-    let filesystems = std::fs::read_to_string("/proc/filesystems")?;
-    Ok(filesystems.contains("selinuxfs\n"))
+    Path::new("/proc/1/root/sys/fs/selinux/enforce")
+        .try_exists()
+        .map_err(Into::into)
 }
 
 /// Get the current process SELinux security context
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -732,7 +732,10 @@ pub(crate) fn reexecute_self_for_selinux_if_needed(
     if srcdata.selinux {
         let host_selinux = crate::lsm::selinux_enabled()?;
         tracing::debug!("Target has SELinux, host={host_selinux}");
-        if host_selinux {
+        if override_disable_selinux {
+            ret_did_override = true;
+            println!("notice: Target has SELinux enabled, overriding to disable")
+        } else if host_selinux {
             // /sys/fs/selinuxfs is not normally mounted, so we do that now.
             // Because SELinux enablement status is cached process-wide and was very likely
             // already queried by something else (e.g. glib's constructor), we would also need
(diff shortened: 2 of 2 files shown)


lib/src/lsm.rs
```diff
@@ -24,8 +24,13 @@ const SELF_CURRENT: &str = "/proc/self/attr/current";
 
 #[context("Querying selinux availability")]
 pub(crate) fn selinux_enabled() -> Result<bool> {
-    let filesystems = std::fs::read_to_string("/proc/filesystems")?;
-    Ok(filesystems.contains("selinuxfs\n"))
+    let path = "/proc/1/root/sys/fs/selinux/enforce";
+    if Path::new(path).exists() {
+        let enabled = std::fs::read_to_string(path)?;
+        return Ok(enabled.eq("1"));
```
cgwalters, line 30: It's a bit more idiomatic to do `enabled == "1"`

However...I think we actually *should* work even if `enforce == 0`.  So this check can probably just boil down to

```
Path::new("/proc/1/root/sys/fs/selinux/enforce").try_exists().map_err(Into::into)
```
or so right?

cgwalters (commented):
For testing coverage...a bit tricky right now as we'd need to boot a system with your case of `selinux=0` and run an install from it.  We have a legacy here of using the coreos "kola" testing framework, and I am hoping that we can eventually align with what's happening in
https://gitlab.com/bootc-org/tests

I'm OK without an integration test here for this for now, it's not like we're likely to regress on normal pull requests.

lib/src/lsm.rs
cgwalters: > However...I think we actually should work even if enforce == 0

To elaborate slightly, this is the current case in Anaconda - it boots with `enforcing=0`.  Our existing install logic should work just fine in this case.

It's really the case of `selinux=0` (fully disabled at runtime) that we need a fallback behavior.

ckyrouac: ah that makes sense, if the labels exist then bootc should work. For some reason that is not the case right now (i.e. the labels are broken on the image when installing from a host with selinux set to permissive), I guess that is what @bcrochet is working on?

I fixed the first commit with your suggestions. I also added a second commit to prioritize the `--disable-selinux` flag. It seems wrong to silently ignore the flag if the host has selinux enabled. I'm probably missing something though. Flipping those if statements makes `install to-disk --disable-selinuc` work from a host with selinux set to permissive though.

cgwalters:
Looks generally good, just two nits:

- need to run `cargo fmt`
- Can you add some basic text into each commit message and also add a "topic prefix"?  (xref https://cbea.ms/git-commit/ and we try to keep somewhat "Linux kernel style") here too

So the second commit for example:

> install: Make --disable-selinux always override host state
>
> If the user explicitly disables SELinux, we should always honor that and not
> care about the host state.

or so?

cgwalters (approved):
Nice, thanks!

ckyrouac:
thanks for the reviews!

PR #341, opened 2024-02-14
install: Mount `/boot` readonly by default

cgwalters:
As we want to support enabling `root.transient` in some images, this means that things like `apt|dnf install foo` literally just works out of the box.

However...we have a looming danger around things like kernels.  Typically the package installation scripts for those aren't going to handle this correctly.

Let's mount `/boot` readonly by default, as we have been doing in Fedora CoreOS and derivatives for a while.

Now I'm not totally happy with this because ultimately I think this should be configurable by the OS, not hardcoded in bootc.  We have some thought to put in to exactly how that's exposed.

But for now let's set the precedent here.

### DIFF
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -362,6 +362,15 @@ impl MountSpec {
             self.source, self.target, self.fstype, options
         )
     }
+
+    /// Append a mount option
+    pub(crate) fn push_option(&mut self, opt: &str) {
+        let options = self.options.get_or_insert_with(Default::default);
+        if !options.is_empty() {
+            options.push(',');
+        }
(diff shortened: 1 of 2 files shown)


cgwalters:
Ah we got bit by https://github.com/containers/bootc/issues/57 ...I enabled auto-merge letting it be blocked on a review, but that got reverted again.

mvo5 (commented):
I can give this a retro :+1: - I actually read it before it got merged and it looks (and looked) good but as I'm not super fluent in rust for a real +1 I would have checked it out and played a bit first.

cgwalters:
Thanks!  Post-merge review is still good.

PR #342, opened 2024-02-14
install: Drop unimplemented root_options

cgwalters:
install: Drop unimplemented root_options

I added this in 8078700c433b41ae80564def5190cdbe08315108
but it was never implemented.

Drop it to avoid confusion.

---

manpages-md: Regenerate

(TODO: add a periodic GH action that does this)

---


jeckersb (approved):
lgtm!

PR #344, opened 2024-02-14
Add selinuxfs to be mounted in the container

bcrochet:
As a follow-on to #302, we want to also mount the selinuxfs special filesystem if the host also has that filesystem mounted.

Related #303

### DIFF
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -51,6 +51,12 @@ const BOOT: &str = "boot";
 const RUN_BOOTC: &str = "/run/bootc";
 /// This is an ext4 special directory we need to ignore.
 const LOST_AND_FOUND: &str = "lost+found";
+/// The mount path for selinux
+#[cfg(feature = "install")]
+const SELINUXFS: &str = "/sys/fs/selinux";
+/// The mount path for uefi
+#[cfg(feature = "install")]
+const EFIVARFS: &str = "/sys/firmware/efi/efivars";
 pub(crate) const ARCH_USES_EFI: bool = cfg!(any(target_arch = "x86_64", target_arch = "aarch64"));
 
 /// Kernel argument used to specify we want the rootfs mounted read-write by default
@@ -750,7 +756,7 @@ pub(crate) fn reexecute_self_for_selinux_if_needed(
             // already queried by something else (e.g. glib's constructor), we would also need
             // to re-exec.  But, selinux_ensure_install does that unconditionally right now too,
             // so let's just fall through to that.
-            crate::lsm::container_setup_selinux()?;
+            setup_sys_mount("selinuxfs", SELINUXFS)?;
             // This will re-execute the current process (once).
             g = crate::lsm::selinux_ensure_install_or_setenforce()?;
         } else if std::env::var_os(skip_check_envvar).is_some() {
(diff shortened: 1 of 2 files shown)


lib/src/install.rs
```diff
@@ -865,37 +865,42 @@ pub(crate) fn setup_tmp_mounts() -> Result<()> {
...
 
-    // First of all, does the container already have the mount?
-    let path = Utf8Path::new(root_efivars);
-    if path.try_exists()? {
-        tracing::debug!("Check if efivarfs already mount");
-        let inspect = crate::mount::inspect_filesystem(path);
-        if inspect.is_ok() {
-            tracing::trace!("Already have efivarfs {root_efivars}");
-            return Ok(());
+        // First of all, does the container already have the mount?
```
cgwalters, line 885: So I like the change conceptually to checking for a mountpoint...but unfortunately the `is_mountpoint` we have here right now is not 100% reliable.

Perhaps the simplest thing really is to check if the directory is non-empty here too?

cgwalters (changes requested):

lib/src/install.rs
```diff
@@ -865,37 +865,42 @@ pub(crate) fn setup_tmp_mounts() -> Result<()> {
 #[context("Ensuring sys mounts")]
 pub(crate) fn setup_sys_mounts() -> Result<()> {
```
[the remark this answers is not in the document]
bcrochet: Should this just become a function that takes a /sys subpath instead of looping? That seems to make more sense now that I look at it... Or as you said, we can just drop the other code that does this, and just make sure this is called ahead of where that was done before.

lib/src/install.rs
bcrochet: Will do.

lib/src/install.rs
```diff
@@ -750,7 +756,7 @@ pub(crate) fn reexecute_self_for_selinux_if_needed(
             // already queried by something else (e.g. glib's constructor), we would also need
             // to re-exec.  But, selinux_ensure_install does that unconditionally right now too,
             // so let's just fall through to that.
-            crate::lsm::container_setup_selinux()?;
+            let _ = setup_sys_mount("selinuxfs", SELINUXFS)?;
```
cgwalters, line 759: The `let _ ` here looks unnecessary.

lib/src/lsm.rs
```diff
@@ -123,6 +123,7 @@ pub(crate) fn selinux_ensure_install_or_setenforce() -> Result<Option<SetEnforce
 
 /// Ensure that /sys/fs/selinux is mounted, and ensure we're running
 /// as install_t.
+#[allow(dead_code)]
```
cgwalters, line 126: We can delete this function instead of marking it dead, no?

lib/src/install.rs
```diff
@@ -1002,7 +1001,7 @@ async fn prepare_install(
         super::cli::ensure_self_unshared_mount_namespace().await?;
     }
 
-    setup_sys_mounts()?;
+    let _ = setup_sys_mount("efivarfs", EFIVARFS)?;
```
cgwalters, line 1004: Ditto here.

lib/src/install.rs
```diff
@@ -871,39 +877,32 @@ pub(crate) fn setup_tmp_mounts() -> Result<()> {
...
-        tracing::debug!("Check if efivarfs already mount");
-        let inspect = crate::mount::inspect_filesystem(path);
-        if inspect.is_ok() {
-            tracing::trace!("Already have efivarfs {root_efivars}");
-            return Ok(());
-        }
+    // Check that the path that should be mounted is even populated.
+    // Since we are dealing with /sys mounts here, if it's populated,
+    // we can be at least a little certain that it's mounted.
+    if Path::new(fspath).try_exists()? && !std::fs::read_dir(fspath)?.next().is_none() {
```
cgwalters, line 897: Minor nit, I think `std::fs::read_dir(fspath)?.next().is_some()` is easier to read (avoids double negation)

cgwalters (changes requested):

cgwalters (approved):

PR #349, opened 2024-02-15
ci: Target c9s in more places

cgwalters:
There's some instability in eln right now, but more importantly I think our baseline target should be c9s because that acts as the real lower bound for what we want to support.


jmarrero (approved):
lgtm

PR #350, opened 2024-02-15
xtask: Fixes for `cargo xtask package`

cgwalters:
The release process has drifted with xtask; I forget exactly why but I ended up with `.zstd`, not `.zst` in the tarballs and I've been hand-hacking that manually.

Fix things up so that `cargo xtask package` generates the source snapshot and the vendor tarball named exactly how we release them now.


jmarrero (approved):
lgtm

PR #366, opened 2024-02-27
Cargo.lock: Adjust for nix bump

cgwalters:
For some reason the changes that dependabot made here recently still result in local `Cargo.lock` changes for me.


cgwalters:
OK fun, I don't know what's going on with that test that fell over, will debug async

PR #368, opened 2024-02-27
privtests: Show stderr on failure

cgwalters:
This test fell over, let's show what stderr actually had instead.


cgwalters:
Oh this is because something in the eln base image changed and

```
grep -Ee 'passwd|group' /etc/nsswitch.conf 
passwd:     files systemd
group:      files [SUCCESS=merge] altfiles [SUCCESS=merge] systemd
netgroup:   files
```

which is clearly wrong and breaks `runuser -u bin`.

cgwalters:
xref https://github.com/coreos/fedora-coreos-tracker/issues/1680

jeckersb (approved):
:+1:

PR #369, opened 2024-02-28
tests: Another missing `assert!` change to dump stderr

cgwalters:
I missed that there were two here; followup to previous commit.


jeckersb (approved):
I missed it too!

PR #373, opened 2024-03-04
test: add integration test

henrywang:
This PR includes integration test and and CI triggered by each PR.
Integration test includes RPM build and bootc install+upgrade. bootc inside RHEL/CS9-dev container image will be replaced by new built RPM, then the bootc install and upgrade will be tested on AWS ec2 for both x86_64 and aarch64 instances. Integration test is run by TMT and Testing Farm.

TMT + Testing Farm results:

- CS9-dev image with x86_64: http://artifacts.osci.redhat.com/testing-farm/55d81810-cbe7-4d93-b2fc-ea757c09635b/
- CS9-dev image with aarch64: http://artifacts.osci.redhat.com/testing-farm/a0510666-9e2f-4ac9-861d-c31866038038/

Some credentials need to be added in this repo. Anyone can help on this? Thanks.


tests/integration/install-upgrade.sh
```diff
@@ -0,0 +1,178 @@
+#!/bin/bash
+set -exuo pipefail
```
cgwalters, line 2: So this is currently a copy of the code at https://gitlab.com/bootc-org/tests/bootc-workflow-test right?  Can we include that as a git submodule instead, that gets updated with renovate?

henrywang: Not exactly, I'd like to add more test in the future, like `bootc` arguments tests. So I decided to add a new file here.

cgwalters: @henrywang I sent you an invite

cgwalters: > New feature or bug fix PR can update and run test in the same PR. And I'd like to add git submodule in bootc-workflow-test repo in the future.

IOW you want to have bootc-workflow-test consume this repository as a git submodule? I think that makes sense...

cgwalters:
Looks like a syntax error?

>  Error: The template is not valid. .github/workflows/integration.yml (Line: 42, Col: 12): Error reading JToken from JsonReader. Path '', line 0, position 0.,.github/workflows/integration.yml (Line: 43, Col: 12): Error reading JToken from JsonReader. Path '', line 0, position 0.,.github/workflows/integration.yml (Line: 44, Col: 17): Error reading JToken

tests/integration/install-upgrade.sh
henrywang: > IOW you want to have bootc-workflow-test consume this repository as a git submodule? I think that makes sense...

Yes, upstream always first.

The workflow I would like to do:
1. most of test code will be saved in upstream repo and QE spend most of their efforts on PRs of upstream repo.
2. bootc-workflow-test `(it will be bootc and bootc image downstream test repo)` will consume upstream repo as `git submodule`. The git update will be triggered by new bootc RHEL `brew` build. That will avoid code gap between upstream`(always latest)` and downstream. I know it can't 100% fix the gap issue, but 
3. then downstream will run daily regular test.

henrywang:
> Looks like a syntax error?
> 
> > Error: The template is not valid. .github/workflows/integration.yml (Line: 42, Col: 12): Error reading JToken from JsonReader. Path '', line 0, position 0.,.github/workflows/integration.yml (Line: 43, Col: 12): Error reading JToken from JsonReader. Path '', line 0, position 0.,.github/workflows/integration.yml (Line: 44, Col: 17): Error reading JToken

`GITHUB_TOKEN` permission issue fixed according to https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#get-a-pull-request.

@cgwalters May I have permission to add some github secrets? After secrets added, permission can be removed. Thanks!

cgwalters:
> @cgwalters May I have permission to add some github secrets? After secrets added, permission can be removed. Thanks!

Backing up a level...this seems like something that may even be better done at an organization level - if we ended up wanting to share testing farm bits across other repositories.  cc @cevich

cevich:
I'm not opposed in principal.  My concern is over security, and I'm not 100% on how/where/when github action secrets are exposed.  Since it would expose every org. secret to every repo., is it possible someone could craft a repo-level workflow that would leak a secret?  In other words, I speculate (not verified) there are some containers-org repos. that RH isn't 100% in control over.

cgwalters:
> Since it would expose every org. secret to every repo., is it possible someone could craft a repo-level workflow that would leak a secret? In other words, I speculate (not verified) there are some containers-org repos. that RH isn't 100% in control over.

Yes, this is why any repository with secrets needs to carefully watch PRs that affect workflows.  It's difficult and error prone.  The Testing Farm bits currently require manual approval.

But my question wasn't about adding these secrets org wide, my question was more about a github.com/containers SOP (or even, a SOP for a subset of them) - does the Cirrus setup involve any repo secrets?

cevich:
>  does the Cirrus setup involve any repo secrets?

Cirrus-CI supports both, org-level and repo-level.  The org-level secrets are dangerous for the same reason as I outlined above.  The repo-level secrets are only as safe as the (per-repo) "anybody can decrypt" setting.  When that's turned off, it has similar "require approval" button behavior.  That became a MAJOR headache on the high-traffic repos, so we keep it turned on.  That makes org-wide Cirrus-CI secrets even more dangerous.

So my SOP is to avoid anything but very inconsequential secrets in cirrus, except for repos. where that "anybody can decrypt" button can be left/turned off.   Otherwise, for high-traffic repos, I'd prefer to keep all secret stuffs nestled away in the github settings. or other safe places.

henrywang:
@cgwalters Running testing farm needs TF_API_KEY secret available inside the forked repo. So the `pull_request_target` trigger has to be used in this case. To protect the secrets, this workflow has a PR sender permission checking at first job. Only collaborator with repo write or admin permission can run this workflow.

`pull_request_target` only runs in the pull request's target branch, and not the pull request's branch itself. This PR can't trigger this workflow included in this PR.

.github/workflows/integration.yml
```diff
@@ -0,0 +1,120 @@
...
+  statuses: write
+
+# Running testing farm needs TF_API_KEY secret available inside the forked repo.
+# So the pull_request_target trigger has to be used in this case. To protect the
+# secrets this workflow has a PR sender permission checking at first job. Only
+# collaborator with repo write or admin permission can run this workflow.
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
```
cgwalters, line 16: I noticed the test isn't actually running on this PR...did you test this manually in some way?

Or do we need to merge and then test on a new PR?

henrywang: > Or do we need to merge and then test on a new PR?

Correct. `pull_request_target` only runs in the pull request's target branch, and not the pull request's branch itself. This PR can't trigger this workflow included in this PR. Merge this PR first, then I'll send another PR to test CI and test case.

cgwalters:
ISTM that testing farm could do the same thing here by being a proper app instead of a raw action?  I guess to say this another way I know there's been an investment in Cirrus (and Prow, and GH actions) on other...wait, I see podman is running testing farm tests, but it looks like indirectly via packit?  Is that the better flow?

PR #375, opened 2024-03-06
blockdev: Use --direct-io=on for losetup

cgwalters:
This is what other projects do; it avoids double buffering and is generally a good idea.


jeckersb:
Ah gotcha.  My English parser is buggy :laughing:

mvo5:
Yeah, the code in https://github.com/util-linux/util-linux/blob/master/sys-utils/losetup.c#L687 also indicates the default is "off". But this is super nice, we will try this for osbuild too.

PR #377, opened 2024-03-06
test: update integration README

henrywang:
Use `--tls-verify=false` in podman build, don't need CERT_URL secret any more.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -79,8 +79,6 @@ jobs:
           git_ref: ${{ needs.pr-info.outputs.ref }}
           arch: ${{ matrix.arch }}
           tmt_context: "arch=${{ matrix.arch }}"
-          update_pull_request_status: true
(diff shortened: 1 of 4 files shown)


cgwalters:
Looks like that run hit > Maximum test time '40m' exceeded...hmm, maybe the pull was slow?  I did a blind attempt to add a pull before the run (with retries) to more cleanly separate network issues from install issues.

henrywang:
@cgwalters There're two test status for one testing farm case (github job status and testing farm status). This does not make sense. Github job status update can't be disabled, so testing farm PR job status update has to be disabled.

BTW: I filed two `testing-farm-as-github-action` github action bugs https://github.com/sclorg/testing-farm-as-github-action/issues/129 and https://github.com/sclorg/testing-farm-as-github-action/issues/130.

henrywang:
@cgwalters Can we use centos stream 9 image to test `runuser -u bin`? That might workaround this failure.

PR #379, opened 2024-03-06
ci: Pin eln image

cgwalters:
Until the fix for https://github.com/coreos/fedora-coreos-tracker/issues/1680 finally propagates.


bcrochet (approved):
/lgtm

PR #381, opened 2024-03-07
cli: add container lint

### LINKED ISSUE #216: Add `bootc build commit`

prestist:
Add an entrypoint that basically everyone can start
adding to their builds which performs some basic
static analysis for known problems.

---

(original description follows)

We know this does not fix #216 yet, we are working this in time-slots as a part of mobbing to address this. In the next update we plan on having the logic to check for one kernel complete. Then we can debate if we want the lint to be in this code base, or if we want it to be moved somewhere else.

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -19,6 +19,7 @@ use std::os::unix::process::CommandExt;
 use std::process::Command;
 
 use crate::deploy::RequiredHostSpec;
+use crate::lints;
 use crate::spec::Host;
 use crate::spec::ImageReference;
 use crate::utils::sigpolicy_from_opts;
@@ -141,6 +142,14 @@ pub(crate) struct ManOpts {
     #[clap(long)]
     /// Output to this directory
     pub(crate) directory: Utf8PathBuf,
+}
+
+/// Subcommands which can be executed as part of a container build.
+#[derive(Debug, clap::Subcommand, PartialEq, Eq)]
+pub(crate) enum ContainerOpts {
+    /// Perform relatively inexpensive static analysis checks as part of a container
+    /// build.
+    Lint,
 }
 
 /// Hidden, internal only options
@@ -292,6 +301,9 @@ pub(crate) enum Opt {
     #[clap(subcommand)]
     #[cfg(feature = "install")]
     Install(InstallOpts),
+    /// Operations which can be executed as part of a container build.
+    #[clap(subcommand)]
+    Container(ContainerOpts),
     /// Execute the given command in the host mount namespace
     #[cfg(feature = "install")]
     #[clap(hide = true)]
@@ -662,6 +674,19 @@ async fn run_from_opt(opt: Opt) -> Result<()> {
         Opt::Rollback(opts) => rollback(opts).await,
         Opt::Edit(opts) => edit(opts).await,
         Opt::UsrOverlay => usroverlay().await,
+        Opt::Container(opts) => match opts {
+            ContainerOpts::Lint => {
+                if !ostree_ext::container_utils::is_ostree_container()? {
+                    anyhow::bail!(
+                        "Not in a ostree container, this command only verifies ostree containers."
+                    );
+                }
+
+                let root = cap_std::fs::Dir::open_ambient_dir("/", cap_std::ambient_authority())?;
+                lints::lint(&root)?;
+                Ok(())
+            }
+        },
         #[cfg(feature = "install")]
         Opt::Install(opts) => match opts {
             InstallOpts::ToDisk(opts) => crate::install::install_to_disk(opts).await,
--- a/docs/src/man/bootc-container-lint.md
+++ b/docs/src/man/bootc-container-lint.md
@@ -1,0 +1,23 @@
+# NAME
+
+bootc-container-lint - Perform relatively inexpensive static analysis
+checks as part of a container build
+
+# SYNOPSIS
+
+**bootc container lint** \[**-h**\|**\--help**\]
+
+# DESCRIPTION
+
+Perform relatively inexpensive static analysis checks as part of a
+container build
+
+# OPTIONS
+
+**-h**, **\--help**
+
+:   Print help
+
+# VERSION
+
+v0.1.11
--- a/docs/src/man/bootc-container.md
+++ b/docs/src/man/bootc-container.md
@@ -1,0 +1,33 @@
+# NAME
+
+bootc-container - Operations which can be executed as part of a
+container build
+
+# SYNOPSIS
+
+**bootc container** \[**-h**\|**\--help**\] \<*subcommands*\>
+
+# DESCRIPTION
+
+Operations which can be executed as part of a container build
+
+# OPTIONS
+
+**-h**, **\--help**
+
+:   Print help
+
+# SUBCOMMANDS
+
+bootc-container-lint(8)
+
+:   Perform relatively inexpensive static analysis checks as part of a
+    container build
+
+bootc-container-help(8)
+
+:   Print this message or the help of the given subcommand(s)
+
+# VERSION
+
+v0.1.11
--- a/docs/src/man/bootc.md
+++ b/docs/src/man/bootc.md
@@ -62,6 +62,10 @@ bootc-install(8)
 
 :   Install the running container to a target
 
+bootc-container(8)
+
+:   Operations which can be executed as part of a container build
+
 bootc-help(8)
 
 :   Print this message or the help of the given subcommand(s)
--- a/hack/Containerfile
+++ b/hack/Containerfile
@@ -16,3 +16,4 @@ FROM $base
 COPY --from=build /out/bootc.tar.zst /tmp
 COPY --from=build /build/target/dev-rootfs/ /
 RUN tar -C / --zstd -xvf /tmp/bootc.tar.zst && rm -vf /tmp/*
+RUN bootc container lint
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -21,6 +21,7 @@ pub mod cli;
 pub(crate) mod deploy;
 pub(crate) mod generator;
 pub(crate) mod journal;
+mod lints;
 mod lsm;
 pub(crate) mod metadata;
 mod reboot;
--- a/lib/src/lints.rs
+++ b/lib/src/lints.rs
@@ -1,0 +1,72 @@
+//! # Implementation of container build lints
+//!
+//! This module implements `bootc container lint`.
+
+use anyhow::Result;
+use cap_std::fs::Dir;
+use cap_std_ext::cap_std;
+use cap_std_ext::dirext::CapStdExtDirExt as _;
+use fn_error_context::context;
+
+/// check for the existence of the /var/run directory
+/// if it exists we need to check that it links to /run if not error
+/// if it does not exist error.
+#[context("Linting")]
+pub(crate) fn lint(root: &Dir) -> Result<()> {
+    let lints = [check_var_run, check_kernel];
+    for lint in lints {
+        lint(&root)?;
+    }
+    println!("Checks passed: {}", lints.len());
+    Ok(())
+}
+
+fn check_var_run(root: &Dir) -> Result<()> {
+    if let Some(meta) = root.symlink_metadata_optional("var/run")? {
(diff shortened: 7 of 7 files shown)


lib/src/cli.rs
```diff
@@ -12,6 +12,7 @@ use ostree_container::store::PrepareResult;
 use ostree_ext::container as ostree_container;
 use ostree_ext::keyfileext::KeyFileExt;
 use ostree_ext::ostree;
+use tracing::field::debug;
```
cgwalters, line 15: This looks strange, I think you meant to use `debug!` which is a macro, see below

lib/src/cli.rs
```diff
@@ -512,6 +515,27 @@ async fn usroverlay() -> Result<()> {
...
+/// async fn lint() -> Result<()> {
+fn lint() -> Result<()> {
+    // TODO: Check if 1 or more folder under the /lib/modules/ directory
+    println!("Checking if we are in the container");
+    if ostree_ext::container_utils::is_ostree_container()? {
+        let modules_path = cap_std::fs::Dir::open_ambient_dir("lib/modules", cap_std::ambient_authority())?;
+        for entry in modules_path.entries()? {
+            let dir = entry?;
+            let fileName= dir.file_name();
+            debug(fileName);
```
cgwalters, line 528: First nit: naming scheme in Rust is snake_case and not camelCase, i.e. should be `file_name` or `filename` or just `name`.

I think you want either `tracing::debug!` here (suitable to leave in persistently, will only show up with `RUST_LOG=debug`) or just a plain `dbg!` (to print something while testing/developing, but can't go in committed code).

lib/src/cli.rs
```diff
@@ -512,6 +515,27 @@ async fn usroverlay() -> Result<()> {
...
+
+        }
+        return Ok(());
+    }
+    anyhow::bail!(
```
cgwalters, line 533: This is OK as is, but I tend to prefer doing "early return" style to avoid rightward drift.  So like this:

```
if !is_ostree_container()? {
   anyhow::bail!(...)
}

...expected case here...
}

File: lib/src/cli.rs [hunk 4]
```diff
@@ -512,6 +515,27 @@ async fn usroverlay() -> Result<()> {
         .into());
 }
 
+/// Implementation of `bootc build commit`
+/// async fn lint() -> Result<()> {
+fn lint() -> Result<()> {
+    // TODO: Check if 1 or more folder under the /lib/modules/ directory
+    println!("Checking if we are in the container");
+    if ostree_ext::container_utils::is_ostree_container()? {
+        let modules_path = cap_std::fs::Dir::open_ambient_dir("lib/modules", cap_std::ambient_authority())?;
```
cgwalters, line 524: This is a super minor nit, but canonically `lib` :arrow_right: `usr/lib` now, so slightly cleaner to use that.

cgwalters (changes requested):
Thanks for working on this!

Mostly nits.

lib/src/cli.rs
```diff
@@ -512,6 +515,27 @@ async fn usroverlay() -> Result<()> {
...
+        for entry in modules_path.entries()? {
```
cgwalters, line 525: Also I forgot until looking that https://docs.rs/ostree-ext/latest/ostree_ext/bootabletree/fn.find_kernel_dir_fs.html exists...so we could just call that here, which is probably better as it ensure we use the exact same logic ostree is using.

cgwalters:
Also it'd be great to have a test for this, which could be done in a variety of ways there's currently 3 distinct frameworks in use here (GHA, tmt, Prow+kola), which I'd like to take down to just the first two probably, except Prow is the most container-native which is what this test would be oriented towards.  But basically building a derived image with two kernels and running the lint in it would make sense.

Another longer term thing to consider here is things like machine-readable output for lints; they're like compiler errors, with the same long term need for tooling that can do more to render than just scrape stdout.

lib/src/cli.rs
prestist: Hehheh, yeah that would be from us stumbling through pressing `ctl .` removed :)

lib/src/cli.rs
prestist: 100%, we have done this as well. left it at the bottom as we were still not solid on the logic that was needed for verifying everything. Good catch :)

lib/src/cli.rs
prestist: Ah ok, thank you!

lib/src/cli.rs
prestist: Wow this made our lives easy thank you!

prestist:
> Also it'd be great to have a test for this, which could be done in a variety of ways there's currently 3 distinct frameworks in use here (GHA, tmt, Prow+kola), which I'd like to take down to just the first two probably, except Prow is the most container-native which is what this test would be oriented towards. But basically building a derived image with two kernels and running the lint in it would make sense.
> 
Thank you for that idea, we had not even gotten to the testing stage yet, I think this makes sense and we will look into doing this in our next mob session!

> Another longer term thing to consider here is things like machine-readable output for lints; they're like compiler errors, with the same long term need for tooling that can do more to render than just scrape stdout.

Ah so similar to how the context.paths work in butane/ignition?

cgwalters:
> Ah so similar to how the context.paths work in butane/ignition?

I'm not sure, can you give an example?  Or do you mean just how Ignition is inherently JSON?

My thought was like https://doc.rust-lang.org/rustc/json.html (could obviously be much, much simpler than that of course).

cgwalters:
See https://github.com/containers/bootc/issues/410 for another good lint.

cgwalters (commented):
Thanks for working on this!  I think this is a good start; basically we can polish it up a bit and ship it I think.

cgwalters:
https://github.com/containers/bootc/issues/439 is perhaps another thing we could try to lint against.  (That said I'm increasingly thinking we should just make everything else be `usr_t`.)

cgwalters:
One thing that came up related to https://github.com/containers/bootc/issues/128 is we found out that people are trying to do `RUN podman pull <image>` - this is another thing that we could at least detect at build time.

lib/src/cli.rs
```diff
@@ -213,6 +214,8 @@ pub(crate) enum Opt {
     #[clap(subcommand)]
     #[cfg(feature = "install")]
     Install(InstallOpts),
+    /// Validate non supported files are not present.
+    BuildLint,
```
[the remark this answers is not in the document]
jmarrero: I refactored this to be:

`bootc container --lint`

wdyt?

jmarrero:
What about the land the initial kernel lint and then add the other lints on other PRs?

prestist:
@jmarrero, I think that makes sense. I dont want to see this outstanding for much longer. That way we add value, and can add more as time allows. I need to rebase this thing before I add the doc's that colin wanted me to create though. After that the pr should be healthy.

jmarrero:
@cgwalters how do you feel about this atm? 

If you wanted to have the container subcommand I have an alternative here:

https://github.com/jmarrero/bootc/commit/a0de33de8d84d74c8ec617eadc5323e6d18846cb

that runs via:

`bootc container --lint` instead of `bootc build-lint`

Is there one you think is better still, another option you prefer?

lib/src/cli.rs
```diff
@@ -610,6 +619,47 @@ async fn usroverlay() -> Result<()> {
...
+    if !ostree_ext::container_utils::is_ostree_container()? {
+        anyhow::bail!("Not in a ostree container, this command only verifies ostree containers.");
+    }
+
+    // check for the existance of the /var/run directory
+    // if it exists we need to check that it links to /run if not error
+    // if it does not exist error.
+
+    let root = cap_std::fs::Dir::open_ambient_dir("/", cap_std::ambient_authority())?;
+    let varRunLint = find_var_run_symlink_fs(&root)?;
```
cgwalters, line 635: Rust naming is snake_case not camelCase.  But in this case we're not doing anything with the return value, so we can just drop it.

Let's have this explicitly run each lint as its own function from a standard array, something like:

```
   let lints = [check_var_run, check_kernel];
   for lint in lints {
     lint()?;
   }
```

Specifically I think we can simplify `check_var_run()` to something like:

```
fn check_var_run(root: &Dir) -> Result<()> {
   if let Some(meta) = root.symlink_metadata_optional("var/run")? {
       if !meta.is_symlink()) {
          anyhow::bail!("Not a symlink: var/run");
       }
   }
   Ok(())
}
```

for example?

lib/src/cli.rs
```diff
@@ -610,6 +619,47 @@ async fn usroverlay() -> Result<()> {
...
 
+/// Implementation of `bootc build commit`
+/// async fn lint() -> Result<()> {
+#[context("linting")]
+fn lint() -> Result<()> {
+    if !ostree_ext::container_utils::is_ostree_container()? {
+        anyhow::bail!("Not in a ostree container, this command only verifies ostree containers.");
+    }
+
+    // check for the existance of the /var/run directory
```
cgwalters, line 630: The comment is slightly misplaced I think, as someone might think it was related to opening the rootfs instead of the lint check.

lib/src/cli.rs
```diff
@@ -730,3 +781,24 @@ fn test_parse_generator() {
...
 }
+
+#[test]
+fn test_linting() {
+    // linting should only occur in side of a container.
+    match ostree_ext::container_utils::is_ostree_container() {
+        Ok(result) => {
+            if !result {
+                let expected_error_message =
+                    "Not in a ostree container, this command only verifies ostree containers.";
```
cgwalters, line 792: Hmm well, this will break an IMO theoretically valid use case of doing a build+test inside a bootc container...

Let's switch to having coverage for the kernel and var/run bits which we can easily do in unit tests I think.

lib/src/privtests.rs
```diff
@@ -196,6 +195,37 @@ fn verify_selinux_recurse(root: &Dir, path: &mut PathBuf, warn: bool) -> Result<
     Ok(())
 }
 
+#[context("Container tests")]
+fn test_build_lint(image: &str) -> Result<()> {
```
cgwalters, line 199: Thanks for adding this test!  I'm working to clean up our testing framework a bit and you've hit some existing tech debt in that area.  I think we can get by with just unit tests to start.

cgwalters (changes requested):
Thanks so much for working on this!

lib/src/cli.rs
cgwalters: On the lint execution side, see https://github.com/coreos/rpm-ostree/blob/4f6dfd089c3686bbe1c9160332ce5758524c5865/rust/src/composepost.rs#L426 which is a *very* similar pattern of calling multiple functions.

prestist: Ah, ok I think we have done this now. Thank you; it looks much better.

lib/src/privtests.rs
prestist: Ok, we have removed it.

lib/src/cli.rs
prestist: Noted! and we will attack this shortly

cgwalters: > How can I manipulate this data without messing with the real filesystem?

That's one of the awesome things about cap-std, by passing in the rootfs into the lint functions, our unit tests can mock up a rootfs and be *very confident* that the lint checks are not escaping out and interacting with the user's workstation root!

I've updated the code to do that now!

cgwalters:
I went ahead and did some further updates here:

- Squashed the commits
- Moved the linting code into its own module
- Added targeted unit tests in that module
- Adjusted the documentation comments and ran `cargo xtask man2markdown` to generate manpages

WDYT?

prestist:
> I went ahead and did some further updates here:
> 
>     * Squashed the commits
> 
>     * Moved the linting code into its own module
> 
>     * Added targeted unit tests in that module
> 
>     * Changed our own container build to use it just to dogfood
> 
>     * Adjusted the documentation comments and ran `cargo xtask man2markdown` to generate manpages
> 
> 
> WDYT?

Im good to merge; Thank you for the assist! and all the help up to this point! LGTM.

cgwalters (approved):

PR #382, opened 2024-03-08
install: Add `to-existing-root`

cgwalters:
This is just mild sugar on top of the existing `to-filesystem` right now.  But, I think we can do more with this later.

Partially addresses https://github.com/containers/bootc/issues/380


rhatdan:
LGTM

cgwalters:
I think that c9s-dev-integration was a testing farm flake; restarted, let's see.

jmarrero (approved):
lgtm

PR #384, opened 2024-03-08
install: Support gathering more info for host root (including LVM)

cgwalters:
Builds on top of https://github.com/containers/bootc/pull/382
---

install: Support gathering more info for host root (including LVM)

Teach `install to-existing-root` how to gather kernel arguments
and information we need from `/proc/cmdline` (such as `rd.lvm.lv`).
In this case too, we need to adjust the args to use `-v /dev:/dev`
because we need the stuff udev generates from the real host root.

With these things together, I can do a bootc alongside install
targeting a Fedora Server instance.

Closes: https://github.com/containers/bootc/issues/175

---

### DIFF
--- a/docs/src/bootc-install.md
+++ b/docs/src/bootc-install.md
@@ -238,7 +238,7 @@ support the root storage setup already initialized.
 The core command should look like this (root/elevated permission required):
 
 ```bash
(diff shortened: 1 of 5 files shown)


cgwalters:
I had this as draft as it was building on the other PR, but maybe it's simpler to just merge them together now.

jmarrero (approved):
lgtm

PR #385, opened 2024-03-11
test: update ami image id

henrywang:
This PR is to update ami id to latest according to compose update.

### DIFF
--- a/tests/integration/playbooks/deploy-aws.yaml
+++ b/tests/integration/playbooks/deploy-aws.yaml
@@ -11,11 +11,11 @@
     ami_id: ""
     ami:
       x86_64:
-        rhel-9-4: ami-047cbb4b91f923934
-        centos-stream-9: ami-087dde0e824141404
+        rhel-9-4: ami-044442bf4326c9933
+        centos-stream-9: ami-08f2fe20b72b2ffa7
       aarch64:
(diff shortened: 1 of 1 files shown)


henrywang:
Hi @cgwalters! Test is runing into an error `Failed to invoke skopeo proxy method FinishPipe`. This error should be the same as issue https://github.com/coreos/rpm-ostree/issues/4567.
I retired twice and can't make it green.

cgwalters:
It's always been a race condition AFAICS, but unfortunate we're seeing it here too.

cgwalters:
As far as these AMI updates, I think we should probably configure renovate to do it.

henrywang:
> As far as these AMI updates, I think we should probably configure renovate to do it.

This ami id will be used by at least two repos(`bootc` and `bootc-workflow-test`), I'll save ami id into `AWS Systems Manager (SSM) Parameter Store`. There's a CI in https://github.com/virt-s1/bootc-workflow-test/blob/main/.github/workflows/upload_rhel9_ami.yml. The CI includes a fully test for new ami. So I'll update this CI to deploy test passed ami id to `AWS Systems Manager (SSM) Parameter Store` instead of update the ami id in playbook.

cgwalters (approved):

PR #389, opened 2024-03-12
install: manually label {/etc/fstab,tmpfile.d/bootc-root-ssh.conf}

mvo5:
Right now bootc supports an experimental install from a non-selinux host when using the `BOOTC_SKIP_SELINUX_HOST_CHECK=1` option.

This is nice and works relatively well. However files written during the install like /etc/fstab or the tmpfiles.dfile in /etc/tmpfile.d/bootc-root-ssh.conf must be labeled too.

This commit adds a (rather crude) manual way to do this.

Closes https://github.com/containers/bootc/issues/362

### DIFF
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -633,6 +633,7 @@ async fn initialize_ostree_root_from_self(
     let root = rootfs_dir
         .open_dir(path.as_str())
         .context("Opening deployment dir")?;
+    let root_path = &rootfs.join(&path.as_str());
     let mut f = {
         let mut opts = cap_std::fs::OpenOptions::new();
         root.open_with("etc/fstab", opts.append(true).write(true).create(true))
@@ -644,8 +645,16 @@ async fn initialize_ostree_root_from_self(
     }
     f.flush()?;
 
+    let fstab_path = root_path.join("etc/fstab");
+    state.lsm_label(&fstab_path, "/etc/fstab".into(), false)?;
+
     if let Some(contents) = state.root_ssh_authorized_keys.as_deref() {
-        osconfig::inject_root_ssh_authorized_keys(&root, contents)?;
+        osconfig::inject_root_ssh_authorized_keys(
+            &root,
+            &root_path,
+            |target, path, recurse| state.lsm_label(target, path, recurse),
+            contents,
+        )?;
     }
 
     let uname = rustix::system::uname();
--- a/lib/src/install/osconfig.rs
+++ b/lib/src/install/osconfig.rs
@@ -8,7 +8,15 @@ const ETC_TMPFILES: &str = "etc/tmpfiles.d";
 const ROOT_SSH_TMPFILE: &str = "bootc-root-ssh.conf";
 
 #[context("Injecting root authorized_keys")]
-pub(crate) fn inject_root_ssh_authorized_keys(root: &Dir, contents: &str) -> Result<()> {
+pub(crate) fn inject_root_ssh_authorized_keys<F>(
+    root: &Dir,
+    root_path: &Utf8Path,
+    lsm_label_fn: F,
+    contents: &str,
+) -> Result<()>
+where
+    F: Fn(&Utf8Path, &Utf8Path, bool) -> Result<()>,
+{
     // While not documented right now, this one looks like it does not newline wrap
     let b64_encoded = ostree_ext::glib::base64_encode(contents.as_bytes());
     // See the example in https://systemd.io/CREDENTIALS/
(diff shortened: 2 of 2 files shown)


lib/src/install.rs
```diff
@@ -644,8 +644,16 @@ async fn initialize_ostree_root_from_self(
...
     f.flush()?;
 
+    let fstab_path = rootfs.join("etc/fstab");
+    state.lsm_label(&fstab_path, "/etc/fstab".into(), false)?;
+
     if let Some(contents) = state.root_ssh_authorized_keys.as_deref() {
-        osconfig::inject_root_ssh_authorized_keys(&root, contents)?;
+        let tmpf_path = osconfig::inject_root_ssh_authorized_keys(&root, contents)?;
+        state.lsm_label(
+            &rootfs.join(tmpf_path.clone()),
```
mvo5, line 653: This is a bit silly (sorry!) I think:
```rust
-            &rootfs.join(tmpf_path.clone()),
+            &rootfs.join(&tmpf_path),
```   
better (so, rusty in rust).

But the tests fails right now with:
```
DEBUG exec: "matchpathcon" "-n" "/etc/fstab"
DEBUG Label for /target/etc/fstab is system_u:object_r:etc_t:s0
DEBUG exec: "chcon" "-h" "-h" "system_u:object_r:etc_t:s0" "/target/etc/fstab"
Injected: etc/tmpfiles.d/bootc-root-ssh.conf
DEBUG exec: "matchpathcon" "-n" "/etc/tmpfiles.d/bootc-root-ssh.conf"
DEBUG Label for /target/etc/tmpfiles.d/bootc-root-ssh.conf is system_u:object_r:etc_t:s0
DEBUG exec: "chcon" "-h" "-h" "system_u:object_r:etc_t:s0" "/target/etc/tmpfiles.d/bootc-root-ssh.conf"
chcon: cannot access '/target/etc/tmpfiles.d/bootc-root-ssh.conf': No such file or directory
ERROR Installing to filesystem: Creating ostree deployment: Labeling /etc/tmpfiles.d/bootc-root-ssh.conf: Task chcon failed: ExitStatus(unix_wait_status(256))
```
which I need to figure out still (it almost identical to the fstab call above which is puzzling)

jeckersb: I think you want to create the labels relative to `root` and not `rootfs`, since `root` is pointing into the ostree deployment (similar to immediately above with `root.open_with("etc/fstab", ...)`)

jeckersb: Something along the lines of...
```
-    let fstab_path = rootfs.join("etc/fstab");
+    let fstab_path = Utf8PathBuf::try_from(root.canonicalize("etc/fstab")?)?;
```
and then similar below for labeling the ssh keys

lib/src/install.rs
```diff
@@ -644,8 +644,16 @@ async fn initialize_ostree_root_from_self(
     }
     f.flush()?;
 
+    let fstab_path = rootfs.join("etc/fstab");
+    state.lsm_label(&fstab_path, "/etc/fstab".into(), false)?;
+
     if let Some(contents) = state.root_ssh_authorized_keys.as_deref() {
-        osconfig::inject_root_ssh_authorized_keys(&root, contents)?;
+        let tmpf_path = osconfig::inject_root_ssh_authorized_keys(&root, contents)?;
+        state.lsm_label(
```
cgwalters, line 652: I think it'd be better to pass the `state` into this API call and have it perform the labeling, instead of having it return the path.

cgwalters (commented):
Thanks, clearly a next step here is to add a CI check that walks the filesystem tree for unlabeled files, should be an easy add.  I can look at that.

This said, I find myself thinking we may just need to pull out the big hammer of doing a force relabel of `/etc` and `/var` in this case.

lib/src/install.rs
mvo5: I remember now why I did it the other way, because I was too lazy to update the test /o\ - sorry, I will look at this.

lib/src/install/osconfig.rs
```diff
@@ -18,6 +23,14 @@ pub(crate) fn inject_root_ssh_authorized_keys(root: &Dir, contents: &str) -> Res
     root.create_dir_all(tmpfiles_dir)?;
     let target = tmpfiles_dir.join(ROOT_SSH_TMPFILE);
     root.atomic_write(&target, &tmpfiles_content)?;
+
+    let as_path = Utf8Path::new(ETC_TMPFILES).join(ROOT_SSH_TMPFILE);
+    state.lsm_label(
+        &Utf8PathBuf::from("/").join(root_path).join(&as_path),
+        &Utf8PathBuf::from("/").join(&as_path),
```
cgwalters, line 30: I think this is just `root_path.join(as_path)` and `as_path` respectively, right?  They're already absolute paths I believe.

mvo5: The second one is actually relative, I discovered while writing the updated test. :)

mvo5: Hm, I now see one Ci failure:
```
 failed to execute cmd-kola: exit status 1
{"component":"entrypoint","error":"wrapped process failed: exit status 1","file":"k8s.io/test-infra/prow/entrypoint/run.go:84","func":"k8s.io/test-infra/prow/entrypoint.Options.internalRun","level":"error","msg":"Error executing test process","severity":"error","time":"2024-03-15T12:50:22Z"}
error: failed to execute wrapped command: exit status 1 
```
but cannot see more details, I wonder if it's related or not to my changes :/

cgwalters:
The prow/kola job failure is unrelated, should be fixed by https://github.com/coreos/coreos-assembler/pull/3760/commits/836f8ac6a230a1a5906b0bf97051715e9c9e842a

mvo5:
> The prow/kola job failure is unrelated, should be fixed by [coreos/coreos-assembler@836f8ac](https://github.com/coreos/coreos-assembler/commit/836f8ac6a230a1a5906b0bf97051715e9c9e842a)

Thank you! This should be okay to review now, I am not super familiar with rust so not sure if I overcomplicated things (or otherwise did non-idiomatic stuff) but it seems okay (I am really wondering if there is a simpler way to pass "function pointers", happy to look more into this if it's not a good/idiomatic solution).

cgwalters (approved):
Thanks, let's go ahead with this as it is an improvement.  There is a more complex approach I have in mind as a followup though.

PR #392, opened 2024-03-14
docs: clarify that `--karg` can be passed multiple times

mvo5:
It is not clear from the current man-page that `--karg` can be passed multiple times. From reading the code it seems to be the case but because it's not obvious (to me) I also added a small testcase to the CI to ensure that I'm not misreading things.

I added a small comment to the docs that it can be given multiple times.

[draft to double check in CI that it actually works]

P.S. it would be nice to have an easy way to run the CI test locally, especially since they won't run automatically and someone needs to "ack" them first here on GH

### DIFF
--- a/docs/src/man/bootc-install-to-disk.md
+++ b/docs/src/man/bootc-install-to-disk.md
@@ -96,7 +96,9 @@ disabled but where the target does have SELinux enabled.
 
 **\--karg**=*KARG*
 
-:   Add a kernel argument
+:   Add a kernel argument. This option can be provided multiple times.
+
+Example: \--karg=nosmt \--karg=console=ttyS0,114800n8
 
 **\--root-ssh-authorized-keys**=*ROOT_SSH_AUTHORIZED_KEYS*
 
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -133,8 +133,10 @@ pub(crate) struct InstallConfigOpts {
     #[serde(default)]
     pub(crate) disable_selinux: bool,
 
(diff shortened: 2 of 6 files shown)


cgwalters (changes requested):
Thanks!

docs/src/man/bootc-install-to-disk.md
```diff
@@ -96,7 +96,7 @@ disabled but where the target does have SELinux enabled.
 
 **\--karg**=*KARG*
 
-:   Add a kernel argument
+:   Add a kernel argument (can be given multiple times)
```
cgwalters, line 99: These are auto-generated right now, so we need to change the Rust code, and then run `cargo xtask manpages`.

It seems like the cargo functionality for this could be improved, because the Rust side definitely knows it can be passed multiple times.

cgwalters (changes requested):
Thanks for this!

lib/src/install.rs
```diff
@@ -133,12 +133,10 @@ pub(crate) struct InstallConfigOpts {
     #[serde(default)]
     pub(crate) disable_selinux: bool,
 
-    // Only occupy at most this much space (if no units are provided, GB is assumed).
-    // Using this option reserves space for partitions created dynamically on the
-    // next boot, or by subsequent tools.
-    //    pub(crate) size: Option<String>,
+    /// Add a kernel argument.  This option can be provided multiple times.
```
mvo5, line 136: Thank you @cgwalters !

PR #395, opened 2024-03-17
docs: Expand on filesystem /var

cgwalters:
Let's talk about the `/var` trap more.


henrywang:
Test failure is caused by two reasons:
1. testing farm can't deploy nested ARM machine to run test since Saturday. QE CI failed lots of tests this weekend with the same issue.
2. quay.io/fedora/fedora:40 image changed. Package `sudo` has not be installed (log: http://artifacts.osci.redhat.com/testing-farm/a7409509-aaaf-4632-b1cc-6459b2e9ce43/). I filed a bug https://bugzilla.redhat.com/show_bug.cgi?id=2269960 and mark it as testBlocker. The `sudo` should be installed from https://pagure.io/fedora-kickstarts/blob/f40/f/fedora-container-base.ks#_12. I don't know what happened in the new Fedora 40 image. But I have a PR https://github.com/containers/bootc/pull/396 to remove `sudo` to workaround this issue.

PR #398, opened 2024-03-18
test: add rollback and bootc upgrade fetch timer test

henrywang:
This PR will:

1. add rollback and bootc upgrade fetch timer test
2. To avoid hitting quay.io race limitation, add retry for podman push and podman build

### DIFF
--- a/tests/integration/playbooks/rollback.yaml
+++ b/tests/integration/playbooks/rollback.yaml
@@ -1,0 +1,75 @@
+---
+- hosts: guest
+  become: false
+  vars:
+    total_counter: "0"
+    failed_counter: "0"
+
+  tasks:
+    - name: rpm-ostree rollback
+      command: rpm-ostree rollback
+      become: true
+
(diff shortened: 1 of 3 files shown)


tests/integration/playbooks/rollback.yaml
```diff
@@ -0,0 +1,75 @@
+---
+- hosts: guest
+  become: false
+  vars:
+    total_counter: "0"
+    failed_counter: "0"
+
+  tasks:
+    - name: rpm-ostree rollback
+      command: rpm-ostree rollback
```
cgwalters, line 10: This is OK for now but I think in the future we should be pointing at https://github.com/containers/bootc/issues/383 - worth a cross-reference.

henrywang: I saw this issue last week. So I think I should add `rpm-ostree rollback` first. And it will be changed to `bootc rollback` in PR which added this feature.

PR #399, opened 2024-03-18
More image output cleanup

cgwalters:
deploy: Use alternative format for staging

Followup to e3db0108b194f420c014bdcbce0c3e705f3942f6
where this one was missed.  Copy-pasting that git log
for ease of consumption:

Same motivation as https://github.com/ostreedev/ostree-rs-ext/pull/604

Because fetching from a registry with the default sigverify is
the 90% case default, when we see this (and are formatting for
human consumption), use the "alternate" formatting to just
display the image name.

This makes the ostree stuff *much* less in the user's face
in the default path.

---

install: Clean up image output more

Ensure we use alternative formatting for the image reference
for consistency.

And move the image output (including digest) to the very
start where it's more centralized.

---

### DIFF
--- a/lib/src/deploy.rs
+++ b/lib/src/deploy.rs
@@ -265,7 +265,7 @@ pub(crate) async fn stage(
     )
     .await?;
     crate::deploy::cleanup(sysroot).await?;
-    println!("Queued for next boot: {}", spec.image);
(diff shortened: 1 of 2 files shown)


cgwalters:
@sallyom this will fix the issue you saw

jbtrystram (approved):
LGTM

PR #401, opened 2024-03-18
add `/usr/lib/bootc/kargs.d` support

lukewarmtemp:
Fixes https://github.com/containers/bootc/issues/255. Allows users to create files within `/usr/lib/bootc/kargs.d` with kernel arguments. These arguments can now be applied on a switch, upgrade, or edit.

General process:
- use ostree commit of fetched container image to return the file tree
- navigate to /usr/lib/bootc/kargs.d
- read each file within the directory
- push the contents of each file (kargs) into a vector containing all the desired kargs
- pass the kargs to the stage() function

Example Containerfile:
```Dockerfile
FROM quay.io/fedora/fedora-coreos:stable

RUN mkdir -p /usr/lib/bootc/kargs.d
RUN cat <<EOF >> /usr/lib/bootc/kargs.d/00-console.toml
kargs = ["console=ttyS0,114800n8"]
supported-arch = ["x86_64"]
EOF

RUN cat <<EOF >> /usr/lib/bootc/kargs.d/01-mitigations.toml
kargs = ["mitigations=on", "systemd.unified_cgroup_hierarchy=0"]
supported-arch = ["x86_64", "aarch64"]
EOF
```
Looking for some input as to how files within `/usr/lib/bootc/kargs.d` should be formatted (which will also affect how the contents of the file are parsed)

Signed-off-by: Luke Yang <luyang@redhat.com>

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -431,6 +431,7 @@ async fn upgrade(opts: UpgradeOpts) -> Result<()> {
         }
     } else {
         let fetched = crate::deploy::pull(sysroot, imgref, opts.quiet).await?;
+        let kargs = crate::kargs::get_kargs(repo, &booted_deployment, fetched.as_ref())?;
         let staged_digest = staged_image.as_ref().map(|s| s.image_digest.as_str());
         let fetched_digest = fetched.manifest_digest.as_str();
         tracing::debug!("staged: {staged_digest:?}");
@@ -452,7 +453,10 @@ async fn upgrade(opts: UpgradeOpts) -> Result<()> {
             println!("No update available.")
         } else {
             let osname = booted_deployment.osname();
-            crate::deploy::stage(sysroot, &osname, &fetched, &spec).await?;
+            let mut opts = ostree::SysrootDeployTreeOpts::default();
+            let kargs: Vec<&str> = kargs.iter().map(|s| s.as_str()).collect();
+            opts.override_kernel_argv = Some(kargs.as_slice());
+            crate::deploy::stage(sysroot, &osname, &fetched, &spec, Some(opts)).await?;
             changed = true;
             if let Some(prev) = booted_image.as_ref() {
                 if let Some(fetched_manifest) = fetched.get_manifest(repo)? {
@@ -525,6 +529,7 @@ async fn switch(opts: SwitchOpts) -> Result<()> {
     let new_spec = RequiredHostSpec::from_spec(&new_spec)?;
 
     let fetched = crate::deploy::pull(sysroot, &target, opts.quiet).await?;
+    let kargs = crate::kargs::get_kargs(repo, &booted_deployment, fetched.as_ref())?;
 
     if !opts.retain {
         // By default, we prune the previous ostree ref so it will go away after later upgrades
@@ -538,7 +543,10 @@ async fn switch(opts: SwitchOpts) -> Result<()> {
     }
 
     let stateroot = booted_deployment.osname();
-    crate::deploy::stage(sysroot, &stateroot, &fetched, &new_spec).await?;
+    let mut opts = ostree::SysrootDeployTreeOpts::default();
+    let kargs: Vec<&str> = kargs.iter().map(|s| s.as_str()).collect();
+    opts.override_kernel_argv = Some(kargs.as_slice());
+    crate::deploy::stage(sysroot, &stateroot, &fetched, &new_spec, Some(opts)).await?;
 
     Ok(())
 }
@@ -583,11 +591,16 @@ async fn edit(opts: EditOpts) -> Result<()> {
     }
 
     let fetched = crate::deploy::pull(sysroot, new_spec.image, opts.quiet).await?;
+    let repo = &sysroot.repo();
+    let kargs = crate::kargs::get_kargs(repo, &booted_deployment, fetched.as_ref())?;
 
     // TODO gc old layers here
 
     let stateroot = booted_deployment.osname();
-    crate::deploy::stage(sysroot, &stateroot, &fetched, &new_spec).await?;
+    let mut opts = ostree::SysrootDeployTreeOpts::default();
+    let kargs: Vec<&str> = kargs.iter().map(|s| s.as_str()).collect();
+    opts.override_kernel_argv = Some(kargs.as_slice());
+    crate::deploy::stage(sysroot, &stateroot, &fetched, &new_spec, Some(opts)).await?;
 
     Ok(())
 }
--- a/lib/src/deploy.rs
+++ b/lib/src/deploy.rs
@@ -278,8 +278,10 @@ async fn deploy(
     stateroot: &str,
     image: &ImageState,
     origin: &glib::KeyFile,
+    opts: Option<ostree::SysrootDeployTreeOpts<'_>>,
 ) -> Result<()> {
     let stateroot = Some(stateroot);
+    let opts = opts.unwrap_or_default();
     // Copy to move into thread
     let cancellable = gio::Cancellable::NONE;
     let _new_deployment = sysroot.stage_tree_with_options(
@@ -287,7 +289,7 @@ async fn deploy(
         image.ostree_commit.as_str(),
         Some(origin),
         merge_deployment,
-        &Default::default(),
+        &opts,
         cancellable,
     )?;
     Ok(())
@@ -312,6 +314,7 @@ pub(crate) async fn stage(
     stateroot: &str,
     image: &ImageState,
     spec: &RequiredHostSpec<'_>,
+    opts: Option<ostree::SysrootDeployTreeOpts<'_>>,
 ) -> Result<()> {
     let merge_deployment = sysroot.merge_deployment(Some(stateroot));
     let origin = origin_from_imageref(spec.image)?;
@@ -321,6 +324,7 @@ pub(crate) async fn stage(
         stateroot,
         image,
         &origin,
+        opts,
     )
     .await?;
     crate::deploy::cleanup(sysroot).await?;
--- a/lib/src/kargs.rs
+++ b/lib/src/kargs.rs
@@ -1,0 +1,168 @@
+use anyhow::Ok;
+use anyhow::Result;
+
+use crate::deploy::ImageState;
+use ostree::gio;
+use ostree_ext::ostree;
+use ostree_ext::ostree::Deployment;
+use ostree_ext::prelude::Cast;
+use ostree_ext::prelude::FileEnumeratorExt;
+use ostree_ext::prelude::FileExt;
+
+use serde::Deserialize;
+
+#[derive(Deserialize)]
+#[serde(rename_all = "kebab-case", deny_unknown_fields)]
+struct Config {
+    kargs: Vec<String>,
+    match_architectures: Option<Vec<String>>,
+}
+
+/// Compute the kernel arguments for the new deployment. This starts from the booted
+/// karg, but applies the diff between the bootc karg files in /usr/lib/bootc/kargs.d
+/// between the booted deployment and the new one.
+pub(crate) fn get_kargs(
+    repo: &ostree::Repo,
+    booted_deployment: &Deployment,
+    fetched: &ImageState,
+) -> Result<Vec<String>> {
+    let cancellable = gio::Cancellable::NONE;
+    let mut kargs: Vec<String> = vec![];
+    let sys_arch = std::env::consts::ARCH.to_string();
+
+    // Get the running kargs of the booted system
+    if let Some(bootconfig) = ostree::Deployment::bootconfig(booted_deployment) {
+        if let Some(options) = ostree::BootconfigParser::get(&bootconfig, "options") {
+            let options: Vec<&str> = options.split_whitespace().collect();
+            let mut options: Vec<String> = options.into_iter().map(|s| s.to_string()).collect();
+            kargs.append(&mut options);
+        }
+    };
+
+    // Get the kargs in kargs.d of the booted system
+    let mut existing_kargs: Vec<String> = vec![];
+    let fragments = liboverdrop::scan(&["/usr/lib"], "bootc/kargs.d", &["toml"], true);
+    for (_name, path) in fragments {
+        let s = std::fs::read_to_string(&path)?;
+        let mut parsed_kargs = parse_file(s.clone(), sys_arch.clone())?;
+        existing_kargs.append(&mut parsed_kargs);
+    }
+
+    // Get the kargs in kargs.d of the pending image
+    let mut remote_kargs: Vec<String> = vec![];
+    let (fetched_tree, _) = repo.read_commit(fetched.ostree_commit.as_str(), cancellable)?;
+    let fetched_tree = fetched_tree.resolve_relative_path("/usr/lib/bootc/kargs.d");
+    let fetched_tree = fetched_tree
+        .downcast::<ostree::RepoFile>()
+        .expect("downcast");
+    if !fetched_tree.query_exists(cancellable) {
+        // if the kargs.d directory does not exist in the fetched image, return the existing kargs
+        kargs.append(&mut existing_kargs);
+        return Ok(kargs);
+    }
+    let queryattrs = "standard::name,standard::type";
+    let queryflags = gio::FileQueryInfoFlags::NOFOLLOW_SYMLINKS;
+    let fetched_iter = fetched_tree.enumerate_children(queryattrs, queryflags, cancellable)?;
+    while let Some(fetched_info) = fetched_iter.next_file(cancellable)? {
+        // only read and parse the file if it is a toml file
+        let name = fetched_info.name();
+        if let Some(name) = name.to_str() {
+            if name.ends_with(".toml") {
+                let fetched_child = fetched_iter.child(&fetched_info);
+                let fetched_child = fetched_child
+                    .downcast::<ostree::RepoFile>()
+                    .expect("downcast");
+                fetched_child.ensure_resolved()?;
+                let fetched_contents_checksum = fetched_child.checksum();
+                let f =
+                    ostree::Repo::load_file(repo, fetched_contents_checksum.as_str(), cancellable)?;
+                let file_content = f.0;
+                let mut reader =
+                    ostree_ext::prelude::InputStreamExtManual::into_read(file_content.unwrap());
+                let s = std::io::read_to_string(&mut reader)?;
+                let mut parsed_kargs = parse_file(s.clone(), sys_arch.clone())?;
+                remote_kargs.append(&mut parsed_kargs);
+            }
+        }
+    }
+
+    // get the diff between the existing and remote kargs
+    let mut added_kargs: Vec<String> = remote_kargs
+        .clone()
+        .into_iter()
+        .filter(|item| !existing_kargs.contains(item))
+        .collect();
+    let removed_kargs: Vec<String> = existing_kargs
+        .clone()
+        .into_iter()
+        .filter(|item| !remote_kargs.contains(item))
+        .collect();
+
+    tracing::debug!(
+        "kargs: added={:?} removed={:?}",
+        &added_kargs,
+        removed_kargs
+    );
+
+    // apply the diff to the system kargs
+    kargs.retain(|x| !removed_kargs.contains(x));
+    kargs.append(&mut added_kargs);
+
+    Ok(kargs)
+}
+
+pub fn parse_file(file_content: String, sys_arch: String) -> Result<Vec<String>> {
+    let mut de: Config = toml::from_str(&file_content)?;
+    let mut parsed_kargs: Vec<String> = vec![];
+    // if arch specified, apply kargs only if the arch matches
+    // if arch not specified, apply kargs unconditionally
+    match de.match_architectures {
+        None => parsed_kargs = de.kargs,
+        Some(match_architectures) => {
+            if match_architectures.contains(&sys_arch) {
+                parsed_kargs.append(&mut de.kargs);
+            }
+        }
+    }
+    Ok(parsed_kargs)
+}
+
+#[test]
+/// Verify that kargs are only applied to supported architectures
+fn test_arch() {
+    // no arch specified, kargs ensure that kargs are applied unconditionally
+    let sys_arch = "x86_64".to_string();
+    let file_content = r##"kargs = ["console=tty0", "nosmt"]"##.to_string();
+    let parsed_kargs = parse_file(file_content.clone(), sys_arch.clone()).unwrap();
+    assert_eq!(parsed_kargs, ["console=tty0", "nosmt"]);
+    let sys_arch = "aarch64".to_string();
+    let parsed_kargs = parse_file(file_content.clone(), sys_arch.clone()).unwrap();
+    assert_eq!(parsed_kargs, ["console=tty0", "nosmt"]);
+
+    // one arch matches and one doesn't, ensure that kargs are only applied for the matching arch
+    let sys_arch = "aarch64".to_string();
+    let file_content = r##"kargs = ["console=tty0", "nosmt"]
+match-architectures = ["x86_64"]
+"##
+    .to_string();
+    let parsed_kargs = parse_file(file_content.clone(), sys_arch.clone()).unwrap();
+    assert_eq!(parsed_kargs, [] as [String; 0]);
+    let file_content = r##"kargs = ["console=tty0", "nosmt"]
+match-architectures = ["aarch64"]
+"##
+    .to_string();
+    let parsed_kargs = parse_file(file_content.clone(), sys_arch.clone()).unwrap();
+    assert_eq!(parsed_kargs, ["console=tty0", "nosmt"]);
+
+    // multiple arch specified, ensure that kargs are applied to both archs
+    let sys_arch = "x86_64".to_string();
+    let file_content = r##"kargs = ["console=tty0", "nosmt"]
+match-architectures = ["x86_64", "aarch64"]
+"##
+    .to_string();
+    let parsed_kargs = parse_file(file_content.clone(), sys_arch.clone()).unwrap();
+    assert_eq!(parsed_kargs, ["console=tty0", "nosmt"]);
+    std::env::set_var("ARCH", "aarch64");
+    let parsed_kargs = parse_file(file_content.clone(), sys_arch.clone()).unwrap();
+    assert_eq!(parsed_kargs, ["console=tty0", "nosmt"]);
+}
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -21,6 +21,7 @@ pub mod cli;
 pub(crate) mod deploy;
 pub(crate) mod generator;
 pub(crate) mod journal;
+pub(crate) mod kargs;
 mod lints;
 mod lsm;
 pub(crate) mod metadata;


lib/src/deploy.rs
```diff
@@ -340,6 +347,38 @@ pub(crate) fn switch_origin_inplace(root: &Dir, imgref: &ImageReference) -> Resu
     Ok(newest_deployment)
 }
 
+pub fn get_kargs(repo: &ostree::Repo, fetched: &ImageState) -> Result<Vec<String>> {
+    let cancellable = gio::Cancellable::NONE;
+    let (fetched_tree, _) = repo.read_commit(fetched.ostree_commit.as_str(), cancellable)?;
+    let fetched_tree = fetched_tree.resolve_relative_path("/usr/lib/bootc/kargs.d");
```
cgwalters, line 353: Yeah, the recursion here is unfortunate...not a blocker but perhaps it is an argument for driving this logic into ostree.

Or alternatively I guess we'd need to split up the ostree deployment phase into:

- check out filesystem tree
- compute data from it
- set up bootloader entry

Which I believe exists in the API.

lib/src/deploy.rs
```diff
@@ -340,6 +347,38 @@ pub(crate) fn switch_origin_inplace(root: &Dir, imgref: &ImageReference) -> Resu
...
+    let fetched_tree = fetched_tree.downcast::<ostree::RepoFile>().expect("downcast");
+    match fetched_tree.ensure_resolved() {
+        std::result::Result::Ok(()) => {}
+        Err(_) => {
```
cgwalters, line 357: I think it'd be better to use `query_exists()` here

lib/src/cli.rs
```diff
@@ -378,7 +379,10 @@ async fn upgrade(opts: UpgradeOpts) -> Result<()> {
             println!("No update available.")
         } else {
             let osname = booted_deployment.osname();
-            crate::deploy::stage(sysroot, &osname, &fetched, &spec).await?;
+            let mut opts = ostree::SysrootDeployTreeOpts::default();
+            let kargs: Vec<&str> = kargs.iter_mut().map(|s| { s.pop(); s.as_str() }).collect();
+            opts.override_kernel_argv = Some(kargs.as_slice());
```
cgwalters, line 384: The tricky part is...this will *entirely override* the kernel arguments, which means that any machine-specific kargs will get lost.  In the general case we need to support:

- kernel args owned by the image (as this PR is doing)
- kernel args injected as external state; this is what we're doing in coreos with the `console=` kargs for example, but we'll also support doing that via other installation tooling and image generation tooling

Tangential but what's up with the `s.pop()` there?  That looks odd...

cgwalters (commented):
Thanks for looking at this!

cgwalters (changes requested):
Marking as needs-work based on https://github.com/containers/bootc/pull/401/files#r1529248230

lib/src/cli.rs
lukewarmtemp: > Tangential but what's up with the s.pop() there? That looks odd...

Oh I added it in to get rid of the newline character at the end of each karg.

cgwalters: Ah, I think we want to [trim()](https://doc.rust-lang.org/std/primitive.str.html#method.trim) then.

lib/src/deploy.rs
```diff
@@ -340,6 +347,47 @@ pub(crate) fn switch_origin_inplace(root: &Dir, imgref: &ImageReference) -> Resu
...
+        kargs.push(s);
+    }
+
+    let existing_kargs = ostree::KernelArgs::new();
+    ostree::KernelArgs::append_proc_cmdline(
+        &existing_kargs,
+        cancellable,
+    )?;
+    let mut existing_kargs = existing_kargs.to_strv().iter().map(|s| { s.as_str().to_string() }).collect();
+    kargs.append(&mut existing_kargs);
```
lukewarmtemp, line 386: @cgwalters I use `append_proc_cmdline()` to get the existing kargs on the system, append them to the kargs owned by the image, and remove any duplicates. However, I have a few questions about the next step:

1. When both an external and container-owned karg is specified, but has different values, which one wins?
2. When I remove a previously defined karg in the container image and then upgrade, it still keeps the previously defined karg (since it is now part of the system and will be found by `append_proc_cmdline()`). Is this the desired behaviour or do we want to ensure that the karg is removed?
3. Along the same stream as question 2, if we switch to a container image with a unique karg and then proceed to switch to another container image with another unique karg, what is the desired behaviour? Do we remove the old container karg or do we keep both?

cgwalters: > When I remove a previously defined karg in the container image and then upgrade, it still keeps the previously defined karg (since it is now part of the system and will be found by append_proc_cmdline()). Is this the desired behaviour or do we want to ensure that the karg is removed?

This is exactly the problem right now and why it's not as easy as this.  That approach would create [hysteresis](https://blog.verbum.org/2020/08/22/immutable-%E2%86%92-reprovisionable-anti-hysteresis/) and it's very much the type of problem that bites people badly and image-based systems should be aiming to solve.

Any solution here is going to need to do something a bit more like what the OpenShift MCO does (and a bit like ostree's 3-way merge of `/etc`):

- Look at the previous image's `/usr/lib/bootc/kargs.d`
- Look at the new image's `/usr/lib/bootc/kargs.d`

Compute the difference between those two sets, and apply the diff to the new target.

cgwalters: Alternatively of course, we could try hard to have some *other* state file (besides the bootloader ones) that define "local system kargs".  But that brings back shades of `/etc/default/grub` which is just confusing...and we have things that directly write those karg files today (including of course `rpm-ostree kargs` etc.).

lib/src/deploy.rs
```diff
@@ -340,6 +347,64 @@ pub(crate) fn switch_origin_inplace(root: &Dir, imgref: &ImageReference) -> Resu
...
+        remote_kargs.push(s.trim().to_string());
+    }
+
+    // get the diff between the existing and remote kargs
+    let mut added_kargs: Vec<String> = remote_kargs.clone().into_iter().filter(|item| !existing_kargs.contains(item)).collect();
+    let removed_kargs: Vec<String> = existing_kargs.clone().into_iter().filter(|item| !remote_kargs.contains(item)).collect();
+
+    // apply the diff to the system kargs
+    kargs.retain(|x| !removed_kargs.contains(x));
+    kargs.append(&mut added_kargs);
```
lukewarmtemp, line 403: @cgwalters How's this with respect to addressing the hysteresis problem?

lib/src/deploy.rs
```diff
@@ -340,6 +347,64 @@ pub(crate) fn switch_origin_inplace(root: &Dir, imgref: &ImageReference) -> Resu
     Ok(newest_deployment)
 }
 
+pub fn get_kargs(repo: &ostree::Repo, fetched: &ImageState) -> Result<Vec<String>> {
+    let cancellable = gio::Cancellable::NONE;
+
+    // Get the running kernel commandline arguments
+    let kargs = ostree::KernelArgs::new();
+    ostree::KernelArgs::append_proc_cmdline(
```
cgwalters, line 355: Note that the existing ostree logic actually picks them up from the booted deployment's bootloader entry - which is subtly different from `/proc/cmdline`.  One difference there is that some things are actually injected or interpreted by grub (e.g.) and we don't want to re-inject them into the kernel commandline.

IOW, the starting point should be from calling `ostree_deployment_get_bootconfig` on the booted deployment.  https://github.com/ostreedev/ostree/blob/3c25a93bac257acbbf01e3aa64cff29f3bb84f4f/src/ostree/ot-admin-builtin-deploy.c#L186

lib/src/deploy.rs
```diff
@@ -340,6 +347,64 @@ pub(crate) fn switch_origin_inplace(root: &Dir, imgref: &ImageReference) -> Resu
...
+    while let Some(fetched_info) = fetched_iter.next_file(cancellable)? {
+        let fetched_child = fetched_iter.child(&fetched_info);
+        let fetched_child = fetched_child.downcast::<ostree::RepoFile>().expect("downcast");
+        fetched_child.ensure_resolved()?;
+        let fetched_contents_checksum = fetched_child.checksum();
+        let f = ostree::Repo::load_file(repo, fetched_contents_checksum.as_str(), cancellable)?;
+        let file_content = f.0;
+        let mut buffer = vec![];
+        let mut reader = ostree_ext::prelude::InputStreamExtManual::into_read(file_content.unwrap());
+        let _ = std::io::Read::read_to_end(&mut reader, &mut buffer);
```
cgwalters, line 392: Why are we ignoring errors here?

lib/src/deploy.rs
```diff
@@ -340,6 +347,64 @@ pub(crate) fn switch_origin_inplace(root: &Dir, imgref: &ImageReference) -> Resu
...
+        let s = std::string::String::from_utf8(buffer)?;
```
cgwalters, line 393: We can also directly use https://doc.rust-lang.org/std/io/fn.read_to_string.html

lib/src/deploy.rs
```diff
@@ -340,6 +347,64 @@ pub(crate) fn switch_origin_inplace(root: &Dir, imgref: &ImageReference) -> Resu
...
+    let kargs = ostree::KernelArgs::new();
+    ostree::KernelArgs::append_proc_cmdline(
+        &kargs,
+        cancellable,
+    )?;
+    let mut kargs: Vec<String> = kargs.to_strv().iter().map(|s| { s.as_str().to_string() }).collect();
+
+    // Get the kargs in kargs.d of the booted system
+    let mut existing_kargs = vec![];
+    let fragments = liboverdrop::scan(&["/usr/lib"], "bootc/kargs.d", &["toml"], true);
```
cgwalters, line 363: The file extension here is `.toml` but they're apparently just a single karg per file?

I think this calls for having a separate function to parse them, with its own unit tests etc.  (And we are going to have to bikeshed that format...)

This may be best in a dedicated `kargs.rs` mod.

lib/src/deploy.rs
```diff
@@ -340,6 +347,64 @@ pub(crate) fn switch_origin_inplace(root: &Dir, imgref: &ImageReference) -> Resu
...
+        let file_content = f.0;
+        let mut buffer = vec![];
+        let mut reader = ostree_ext::prelude::InputStreamExtManual::into_read(file_content.unwrap());
+        let _ = std::io::Read::read_to_end(&mut reader, &mut buffer);
+        let s = std::string::String::from_utf8(buffer)?;
+        remote_kargs.push(s.trim().to_string());
+    }
+
+    // get the diff between the existing and remote kargs
+    let mut added_kargs: Vec<String> = remote_kargs.clone().into_iter().filter(|item| !existing_kargs.contains(item)).collect();
```
cgwalters, line 398: It may make sense to deal with things in terms of https://docs.rs/ostree/latest/ostree/struct.KernelArgs.html which is actually already structured as a combination of "vector+map" (very minor but this filter is an O(N^2) because we're iterating a vector for each vector element).

lib/src/deploy.rs
cgwalters: Offhand yes...but, we are going to want some unit tests here.  A really nice thing about Rust is how easy it is to write unit tests (see below) - just mocking up a filesystem tree should be pretty easy.

(There's also integration tests but that can come later)

cgwalters:
Let's bikeshed some of the file format/semantics.

One angle: I think there are use cases for conditionally-applied kernel arguments.  One that very much comes up is the `console=` kernel argument being different on architectures and platforms - and dealing with that is painful for people deploying across platforms.  We could just say "You define your own build process which writes to `/usr/lib/bootc/kargs.d`" which would be fine...but, something like what systemd does with `.unit` files or perhaps with `.link` files could make sense.

IOW
```
[match]
architecture = x86_64

kargs = ["console=ttyS0,114800n8"]
```
maybe?

BTW, a whole thing running through all of this is that I think a number of bootc use cases are better off actually generating custom kernel binaries, where you don't need or want this at all; that's something to keep in mind.

lib/src/deploy.rs
lukewarmtemp: Should we move the `get_kargs()` function from `deploy.rs` to `kargs.rs` as well?

cgwalters: > Should we move the get_kargs() function from deploy.rs to kargs.rs as well?

Yeah I'd say so.

lib/src/deploy.rs
lukewarmtemp: Since the `KernelArgs` struct is basically a hashmap (at least from what I understand from the API), I guess we could possibly look at the keys of one `KernelArgs` struct and check if it exists in the second KernelArgs to get the diff, which should be O(N). However, it seems as though the API doesn't provide an easy way to access the `key-value` pairs, so I'm not sure if using this struct is the best option or if I missed something from the API.

cgwalters:
related to console https://github.com/coreos/fedora-coreos-config/blob/testing-devel/platforms.yaml

lukewarmtemp:
@cgwalters Do you know if there's there a way for Rust to detect what platform your system is running? I'm able to find and match against the arch of the system using `std::env::consts::ARCH`, but not sure if there's something similar for platform.

jeckersb:
> @cgwalters Do you know if there's there a way for Rust to detect what platform your system is running? I'm able to find and match against the arch of the system using `std::env::consts::ARCH`, but not sure if there's something similar for platform.

There's a bunch of things you can trigger conditional compilation on, such as [target_os](https://doc.rust-lang.org/reference/conditional-compilation.html#target_os), maybe that can cover what you need?

cgwalters:
By "platform" I meant e.g. AWS vs GCP vs OpenStack etc.  We're not going to build distinct bootc binaries for each of those, and runtime detection is "interesting".  What Fedora CoreOS and derivatives do is https://github.com/coreos/fedora-coreos-tracker/blob/main/internals/README-internals.md#ignitionplatformid

At a practical level though I think we will likely end up with distinct container builds per "platform" in this sense, so we probably don't need to do dynamic dispatch here.

That said, for base images that do include ignition they'll need to have `ignition.platform.id`  anyways...so in theory what we could support is a generic matching against an existing kernel argument (which would make this whole thing slightly recursive of course).

jeckersb:
> By "platform" I meant e.g. AWS vs GCP vs OpenStack etc. 

Ah gotcha.  That's what I get for not reading back far enough for context.  Carry on!

cgwalters:
It's a bit ugly but we could also consider matching what Cargo does for architecture conditionals... https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies

Maybe even lift the code.  Dunno.

I think my instinct here is that anything we do with kargs we should do with install configs too, as that's also a made up TOML thing but we may as well try to be consistent.

Specifically it makes sense to me to support "install only" kargs having architecture conditionals too.

cgwalters:
Thanks so much for working on this!  The code is looking good at a very high level.  We'll need to add documentation, but we can start on that once we're a bit more certain of the design.

I don't want to unnecessarily delay this, but I just keep honestly waffling on the file formats and location.  

We could consider placing these files in e.g. `/usr/lib/modules/$kver` which is our canonical location for kernel data.  Maybe `/usr/lib/modules/$kver/kargs.d` ?  Uncertain...

A *lot* of things would get simpler if one actually just included the kargs in the kernel image (but, doing so breaks any upstream secure boot signatures, and just in general changes the binary, which is why it's not typically done outside of custom embedded builds).

cgwalters:
Also, an integration test would really help.  But you don't need to block on writing it, I can help

lib/src/install/config.rs
```diff
@@ -319,3 +339,195 @@ block = ["tpm2-luks"]"##,
     // And verify passing a disallowed config is an error
     assert!(install.get_block_setup(Some(BlockSetup::Direct)).is_err());
 }
+
+#[test]
+/// Verify that kargs are only applied to supported architectures
+fn test_arch() {
+    // no arch specified, kargs ensure that kargs are applied unconditionally
+    std::env::set_var("ARCH", "x86_64");
```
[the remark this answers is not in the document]
lukewarmtemp: >Is something actually reading these?

It's part of the new changes I've implemented which allow for architecture matching on kargs. https://github.com/containers/bootc/pull/401/files#diff-e592ffb0075409e77a854e0502f4f5ffa7ac34d564042381036993cfa28c8309R107

>Which isn't relevant for these test cases, but actually is relevant for bootc in general

So to clarify, setting the `ARCH` enviornment variable is okay for this test case, but take care to not do so in the main `bootc` code?

lib/src/kargs.rs
```diff
@@ -0,0 +1,151 @@
...
+    booted_deployment: &Deployment,
+    fetched: &ImageState,
+) -> Result<Vec<String>> {
+    let cancellable = gio::Cancellable::NONE;
+    let mut kargs: Vec<String> = vec![];
+
+    // Get the running kargs of the booted system
+    if let Some(bootconfig) = ostree::Deployment::bootconfig(booted_deployment) {
+        if let Some(options) = ostree::BootconfigParser::get(&bootconfig, "options") {
+            let options: Vec<&str> = options.split_whitespace().collect();
```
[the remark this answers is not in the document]
lukewarmtemp: >shared kargs parser

Do you mind elaborating on this? What comes to mind is creating a function called ie. `parse_kargs()` that wraps around `.split_whitepace`, but I'm not sure if that's what you're getting at.

lib/src/install/config.rs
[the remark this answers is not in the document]
cgwalters: But `setenv()` won't affect the value of `std::env::consts::ARCH` right?

lib/src/kargs.rs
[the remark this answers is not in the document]
cgwalters: Let's use https://docs.rs/ostree/latest/ostree/struct.KernelArgs.html#method.from_string ?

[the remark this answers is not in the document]
lukewarmtemp: I think we had a previous conversation regarding the use of the KernelArgs struct: https://github.com/containers/bootc/pull/401#discussion_r1541557674. The reason we didn't want to use it is because there's no easy way to get all the `key-value` pairs for the kargs.

lib/src/install/config.rs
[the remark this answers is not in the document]
lukewarmtemp: Oh I see, yeah my tests were constructed such that it didn't actually test if `std::env::consts::ARCH` was being changed. I'll need to figure out the best way to test different architectures in this unit test.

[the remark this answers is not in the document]
lukewarmtemp: Work around by creating the `detected-arch` field in which a dummy string can be passed to the test: https://github.com/containers/bootc/pull/401/files#diff-e592ffb0075409e77a854e0502f4f5ffa7ac34d564042381036993cfa28c8309R52

lib/src/kargs.rs
```diff
@@ -0,0 +1,162 @@
...
+    let fragments = liboverdrop::scan(&["/usr/lib"], "bootc/kargs.d", &["toml"], true);
+    for (_name, path) in fragments {
+        let s = std::fs::read_to_string(&path)?;
+        let mut parsed_kargs = parse_file(s.clone(), sys_arch.clone())?;
+        existing_kargs.append(&mut parsed_kargs);
+    }
+
+    // Get the kargs in kargs.d of the remote image
+    let mut remote_kargs: Vec<String> = vec![];
+    let (fetched_tree, _) = repo.read_commit(fetched.ostree_commit.as_str(), cancellable)?;
```
cgwalters, line 53: BTW I just did https://github.com/ostreedev/ostree/pull/3252 which will make it *much* more straightforward to get a mounted root for a commit and also lead towards a more composefs-oriented future.

lib/src/install/config.rs
```diff
@@ -319,3 +345,189 @@ block = ["tpm2-luks"]"##,
     // And verify passing a disallowed config is an error
     assert!(install.get_block_setup(Some(BlockSetup::Direct)).is_err());
 }
+
+#[test]
+/// Verify that kargs are only applied to supported architectures
+fn test_arch() {
```
cgwalters, line 351: Can we split up this PR into two?  The install-time karg architecture matching is logically distinct from "day 2 changeable" kargs I think, and can go in separately.

lib/src/install/config.rs
```diff
@@ -161,19 +176,30 @@ pub(crate) fn load_config() -> Result<Option<InstallConfiguration>> {
...
+                install.detected_arch = std::env::consts::ARCH.to_string();
                 config.merge(install);
             }
         } else {
+            // If arch is specified and doesn't match the current arch, remove kargs
+            // We need to add this since the merge function isn't applied for the first config file
+            if let Some(ref mut install) = c.install {
+                if let Some(arch) = install.supported_arch.as_ref() {
+                    install.detected_arch = std::env::consts::ARCH.to_string();
+                    if !arch.contains(&install.detected_arch.to_string()) {
```
cgwalters, line 198: Minor nit, we're allocating a new string here unnecessarily.  Use e.g.
```
    if !arch.any(|v| v == install.detected_arch.as_str())
```

lib/src/install/config.rs
```diff
@@ -47,6 +47,11 @@ pub(crate) struct InstallConfiguration {
     /// Kernel arguments, applied at installation time
     #[serde(skip_serializing_if = "Option::is_none")]
     pub(crate) kargs: Option<Vec<String>>,
+    /// Architecture of the current system (inaccessible to users)
+    #[serde(skip_deserializing)]
+    pub(crate) detected_arch: String,
```
cgwalters, line 52: It's confusing that this is part of the structure, even if it's not supposed to be visible to users.  I think we can make it ambient data passed to the merge function instead, although I know that's more code churn.

lib/src/install/config.rs
```diff
@@ -100,9 +105,19 @@ impl Mergeable for InstallConfiguration {
         merge_basic(&mut self.block, other.block);
         self.filesystem.merge(other.filesystem);
         if let Some(other_kargs) = other.kargs {
-            self.kargs
-                .get_or_insert_with(Default::default)
-                .extend(other_kargs)
+            // if arch is specified, only apply kargs if it matches the current arch
+            // if arch is not specified, apply kargs unconditionally
+            if let Some(other_arch) = other.supported_arch {
+                if other_arch.contains(&other.detected_arch.to_string()) {
```
cgwalters, line 111: Same re `.any()` here

cgwalters:
> Any new thoughts relating to this?

I guess a downside of using the modules directory (aka `/usr/lib/modules/$kver`) is that anyone doing a derived build would need to be more careful to detect the *container* kernel version as is done for e.g. https://docs.fedoraproject.org/en-US/bootc/initramfs/#_regenerating_the_initrd and drop into that directory vs just a straight `COPY 20-mykargs.toml /usr/lib/bootc/kargs.d`.

So...I don't have anything better than `/usr/lib/bootc/kargs.d`.  I feel somewhat uncomfortable in inventing a new thing though.  One thing I'd like to do is try to standardize some of the bootc stuff in OCI upstream, and it'd be around exactly things like this.

cgwalters:
>  One thing I'd like to do is try to standardize some of the bootc stuff in OCI upstream, and it'd be around exactly things like this.

:arrow_right: https://groups.google.com/a/opencontainers.org/g/dev/c/OtLlhn4jxBg

lib/src/install/config.rs
lukewarmtemp: My only concern with passing ambient data to the merge function is the fact that we need to add another parameter to it. The merge function is used for other structs outside of `InstallConfiguration` (ie. `RootFS` and `BasicFilesystems`). However, if we're okay with changing all instances in which the `merge` function is called, I can make implement this change.

lib/src/install/config.rs
lukewarmtemp: https://github.com/containers/bootc/pull/567/files

lib/src/install/config.rs
lukewarmtemp: Fixed by passing an EnvProperties struct to the merge function: https://github.com/containers/bootc/pull/567/files#diff-e592ffb0075409e77a854e0502f4f5ffa7ac34d564042381036993cfa28c8309R11-R16

lib/src/kargs.rs
lukewarmtemp: Sorry, is there a rust binding for this?

cgwalters: Not yet (even if there was, we'd still need to do a release of the C library and wait for that to propagate).

lib/src/install/config.rs
cgwalters: That change is fine yes.  The allocation was coming from `h.to_string()` here.

lib/src/kargs.rs
lukewarmtemp: Should we just go ahead with the PR as is for now and then make amendments after the release occurs?

cgwalters: Yeah

lib/src/kargs.rs
```diff
@@ -0,0 +1,166 @@
...
+    let mut existing_kargs: Vec<String> = vec![];
+    let fragments = liboverdrop::scan(&["/usr/lib"], "bootc/kargs.d", &["toml"], true);
+    for (_name, path) in fragments {
+        let s = std::fs::read_to_string(&path)?;
+        let mut parsed_kargs = parse_file(s.clone(), sys_arch.clone())?;
+        existing_kargs.append(&mut parsed_kargs);
+    }
+
+    // Get the kargs in kargs.d of the remote image
+    let mut remote_kargs: Vec<String> = vec![];
```
cgwalters, line 52: I think "pending" or "next" would be a better term here than "remote" - we've already fetched it.

lib/src/kargs.rs
```diff
@@ -0,0 +1,166 @@
...
+    repo: &ostree::Repo,
+    booted_deployment: &Deployment,
+    fetched: &ImageState,
+) -> Result<Vec<String>> {
+    let cancellable = gio::Cancellable::NONE;
+    let mut kargs: Vec<String> = vec![];
+    let sys_arch = std::env::consts::ARCH.to_string();
+
+    // Get the running kargs of the booted system
+    if let Some(bootconfig) = ostree::Deployment::bootconfig(booted_deployment) {
```
cgwalters, line 34: This looks correct, though I would say we should probably add a convenience helper to this in ostree.

cgwalters (approved):
Thanks, I think this is generally looking good; the big gap I see is around integration testing.  I'd hoped to be able to give you better guidance and tooling here by now.  We'll continue to track that in https://github.com/containers/bootc/issues/543

Short term, I think we can probably merge, and e.g. @henrywang can look at updating the integration tests there.

We're also going to need changes to the docs.

Can you also take an action item to do a merge request to https://gitlab.com/fedora/bootc/examples say?

lukewarmtemp:
> the big gap I see is around integration testing

Oh sorry, I was going to look at this but it slipped my mind. I can probably have a crack at it to see if I can get it working. If I can't see to get it, I'll probably reach out to @henrywang.

> Can you also take an action item to do a merge request to https://gitlab.com/fedora/bootc/examples say?

Will do.

lukewarmtemp:
MR updating bootc examples: https://gitlab.com/fedora/bootc/examples/-/merge_requests/48

cgwalters:
I tested this out and hit:

```
OSTree:ERROR:src/libostree/ostree-repo-file.c:828:ostree_repo_file_tree_query_child: assertion failed: (self->tree_contents)
Bail out! OSTree:ERROR:src/libostree/ostree-repo-file.c:828:ostree_repo_file_tree_query_child: assertion failed: (self->tree_contents)
```

lib/src/kargs.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    // Get the kargs in kargs.d of the pending image
+    let mut remote_kargs: Vec<String> = vec![];
+    let (fetched_tree, _) = repo.read_commit(fetched.ostree_commit.as_str(), cancellable)?;
+    let fetched_tree = fetched_tree.resolve_relative_path("/usr/lib/bootc/kargs.d");
+    let fetched_tree = fetched_tree
+        .downcast::<ostree::RepoFile>()
+        .expect("downcast");
+    if !fetched_tree.query_exists(cancellable) {
+        return Ok(Default::default());
```
cgwalters, line 59: I also hit on the fact that my test machine didn't come up across an update, and I think this is the cause: this must be `return Ok(existing_kargs);` right?

lukewarmtemp: I think it might be
```
kargs.append(&mut existing_kargs);
return Ok(kargs);
```
Since the `kargs` vec contains the existing kargs on the running system.

PR #405, opened 2024-03-19
lsm/install: Use csum(), not name() from SELinux policy

cgwalters:
There's a bug in the ostree API here around nullability.

Closes https://github.com/containers/bootc/issues/403


mvo5 (approved):
Thank you!

PR #407, opened 2024-03-20
osconfig: install root ssh keys to /var/roothome

mvo5:
The current location via tmpfiles.d to install the root ssh keys seems to be not working on firstboot. There is an error that `/root/.ssh` does not exist from `systemd-tmpfiles-setup` and indeed the authorized_keys file is on in `/root/.ssh`.

It seems like everything else in the firstboot in
tmpfiles.d is refering to `/var/roothome` instead of `/root` and switching to this for the location of the ssh keys solves the issue.

I tested this via bib and a local container using the new `bootc install to-filesystem` pipeline :) So hopefully it should finally work.

### DIFF
--- a/lib/src/install/osconfig.rs
+++ b/lib/src/install/osconfig.rs
@@ -1,8 +1,10 @@
+use std::borrow::Cow;
 use std::io::Write;
 
-use anyhow::Result;
+use anyhow::{Context, Result};
+use camino::{Utf8Path, Utf8PathBuf};
 use cap_std::fs::Dir;
-use cap_std_ext::cap_std;
+use cap_std_ext::{cap_std, dirext::CapStdExtDirExt};
 use fn_error_context::context;
 use ostree_ext::ostree;
 
@@ -17,8 +19,22 @@ pub(crate) fn inject_root_ssh_authorized_keys(
 ) -> Result<()> {
     // While not documented right now, this one looks like it does not newline wrap
     let b64_encoded = ostree_ext::glib::base64_encode(contents.as_bytes());
+
+    // Eagerly resolve the path of /root in order to avoid tmpfiles.d clashes/problems.
+    // If it's local state (i.e. /root -> /var/roothome) then we resolve that symlink now.
+    let roothome_meta = root.symlink_metadata_optional("root")?;
+    let root_path = if roothome_meta.as_ref().filter(|m| m.is_symlink()).is_some() {
+        let path = root.read_link("root")?;
+        Utf8PathBuf::try_from(path)
+            .context("Reading /root symlink")
+            .map(Cow::Owned)?
+    } else {
+        Cow::Borrowed(Utf8Path::new("root"))
+    };
+
     // See the example in https://systemd.io/CREDENTIALS/
-    let tmpfiles_content = format!("f~ /root/.ssh/authorized_keys 600 root root - {b64_encoded}\n");
+    let tmpfiles_content =
+        format!("f~ /{root_path}/.ssh/authorized_keys 600 root root - {b64_encoded}\n");
 
     crate::lsm::ensure_dir_labeled(root, ETC_TMPFILES, None, 0o755.into(), sepolicy)?;
     let tmpfiles_dir = root.open_dir(ETC_TMPFILES)?;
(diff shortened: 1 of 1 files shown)


lib/src/install/osconfig.rs
```diff
@@ -18,7 +18,8 @@ pub(crate) fn inject_root_ssh_authorized_keys(
     // While not documented right now, this one looks like it does not newline wrap
     let b64_encoded = ostree_ext::glib::base64_encode(contents.as_bytes());
     // See the example in https://systemd.io/CREDENTIALS/
-    let tmpfiles_content = format!("f~ /root/.ssh/authorized_keys 600 root root - {b64_encoded}\n");
+    let tmpfiles_content =
+        format!("f~ /var/roothome/.ssh/authorized_keys 600 root root - {b64_encoded}\n");
```
cgwalters, line 22: OK this is a complex mess.  The thing is...I would actually like to support a case where `/root` is actually immutable state i.e. it's part of the base image, not in `/var`.

(The same for `/home`; for many headless server cases, persistent home directories are actually an anti-pattern.  We do have it set up this way for OCP nodes for example, but in e.g. GKE `/home` is a tmpfs because ssh to a given node is...again more of an anti-pattern)

I guess anyone who wants to do that could of course symlink `/var/roothome -> /root` or so, but...that has a few drawbacks, among them that we want to avoid "base OS code" in `/usr` inherently trusting local mutable state in `/var` as it would need to do if it traversed that symlink.

See https://www.chromium.org/chromium-os/chromiumos-design-docs/hardening-against-malicious-stateful-data/ for some examples of this.

So...mmm...I think the best fix here is probably to query the root and generate the target pointing directly at the end state or so?

cgwalters: On a distinct thread, one trap I think we have here is systemd *also* ships a default tmpfiles.d line that sources the SSH key from a credential if available...and I think we may be somewhat skating the "duplicate tmpfiles.d entries" problem there today.

Maybe in theory we should actually be injecting a "static credential" via...I guess the generic baseline would be a `systemd.set_credential` karg?

mvo5: Oh, thank you! That is important background. I will need to rethink that then, I did not consider the /root immutable case.

cgwalters: More commonly people will make `/root` and `/home` a `tmpfs`; the upside of that is that it works great when writing files via tmpfiles.d as this unit does, but it also means that any content in the container image is silently hidden, which can be confusing.

cgwalters: I pushed an update to do this (didn't test it beyond units) but what do you think?

lib/src/install/osconfig.rs
```diff
@@ -40,6 +56,35 @@ fn test_inject_root_ssh() -> Result<()> {
...
+    root.symlink("var/roothome", "root")?;
+    inject_root_ssh_authorized_keys(root, None, "ssh-ed25519 ABCDE example@demo\n").unwrap();
+
+    let content = root.read_to_string(format!("etc/tmpfiles.d/{ROOT_SSH_TMPFILE}"))?;
+    assert_eq!(
+        content,
+        "f~ /var/roothome/.ssh/authorized_keys 600 root root - c3NoLWVkMjU1MTkgQUJDREUgZXhhbXBsZUBkZW1vCg==\n"
+    );
+
+    root.create_dir("etc")?;
```
mvo5, line 69: This part of the test is failing right now and seems redundant with the next test, I was inclined to just:
```rust
@@ -60,17 +60,6 @@ fn test_inject_root_ssh() -> Result<()> {
     root.symlink("var/roothome", "root")?;
     inject_root_ssh_authorized_keys(root, None, "ssh-ed25519 ABCDE example@demo\n").unwrap();
 
-    let content = root.read_to_string(format!("etc/tmpfiles.d/{ROOT_SSH_TMPFILE}"))?;
-    assert_eq!(
-        content,
-        "f~ /var/roothome/.ssh/authorized_keys 600 root root - c3NoLWVkMjU1MTkgQUJDREUgZXhhbXBsZUBkZW1vCg==\n"
-    );
-
-    root.create_dir("etc")?;
-    root.remove_file("root")?;
-    root.create_dir("root")?;
-    inject_root_ssh_authorized_keys(root, None, "ssh-ed25519 ABCDE example@demo\n").unwrap();
-
     let content = root.read_to_string(format!("etc/tmpfiles.d/{ROOT_SSH_TMPFILE}"))?;
     assert_eq!(
         content,
```
but then didn't because maybe there is something subtle here that I miss (but I am not seeing it right now :) ?

lib/src/install/osconfig.rs
```diff
@@ -40,6 +56,35 @@ fn test_inject_root_ssh() -> Result<()> {
 
     // The code expects this to exist, reasonably so
     root.create_dir("etc")?;
+    // Test with a symlink
```
mvo5, line 59: Super nitpick: maybe worth renaming this helper now to `test_inject_root_ssh_root_symlinked` or something assuming this is what it is testing (i.e. that my other comment that parts of the test are redundant with `test_inject_root_ssh_dir` is true).

mvo5 (commented):
Thank you! This looks great and works well in my test. There is a tiny tweak in the tests needed it seems though.

lib/src/install/osconfig.rs
mvo5: P.S. I wonder if we should setup branch protection for this for "CI / tests (pull_request) ", it seems it does not block the merge right now even with unit tests failing.

cgwalters: > P.S. I wonder if we should setup branch protection for this for "CI / tests (pull_request) ", it seems it does not block the merge right now even with unit tests failing.

Yeah I changed the branch rules after dropping Prow but must have messed that up...will look

mvo5 (commented):
Thank you so much! :+1: for all the change that I did not make (which are all relevant ones ;)

PR #413, opened 2024-03-21
Correctly populate container environment rootless status

jeckersb:
Previously this was just always None via Default.

Also updated trace logging to show the entire container_info struct.
All of those fields are potentially useful, not just engine.

### DIFF
--- a/lib/src/containerenv.rs
+++ b/lib/src/containerenv.rs
@@ -48,6 +48,7 @@ pub(crate) fn get_container_execution_info(rootfs: &Dir) -> Result<ContainerExec
             "id" => r.id = v.to_string(),
             "image" => r.image = v.to_string(),
             "imageid" => r.imageid = v.to_string(),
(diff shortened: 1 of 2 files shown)


lib/src/containerenv.rs
```diff
@@ -48,6 +48,7 @@ pub(crate) fn get_container_execution_info(rootfs: &Dir) -> Result<ContainerExec
             "id" => r.id = v.to_string(),
             "image" => r.image = v.to_string(),
             "imageid" => r.imageid = v.to_string(),
+            "rootless" => r.rootless = Some(v.to_string()),
```
cgwalters, line 51: Argh...I swear I'd tested this before.  Hmmm...

cgwalters (approved):

lib/src/containerenv.rs
jeckersb: I'm just relieved there is a simple explanation instead of some crazy weird bug!

PR #414, opened 2024-03-21
Trace rootless

cgwalters:
This builds on https://github.com/containers/bootc/pull/413

---

This will catch rootless podman cases even more reliably (assuming
we were invoked with `--pid=host`).

### DIFF
--- a/lib/src/containerenv.rs
+++ b/lib/src/containerenv.rs
@@ -48,6 +48,7 @@ pub(crate) fn get_container_execution_info(rootfs: &Dir) -> Result<ContainerExec
             "id" => r.id = v.to_string(),
             "image" => r.image = v.to_string(),
             "imageid" => r.imageid = v.to_string(),
+            "rootless" => r.rootless = Some(v.to_string()),
             _ => {}
         }
     }
--- a/lib/src/containerenv.rs
+++ b/lib/src/containerenv.rs
@@ -8,7 +8,7 @@ use cap_std_ext::prelude::CapStdExtDirExt;
 use fn_error_context::context;
 
 /// Path is relative to container rootfs (assumed to be /)
-const PATH: &str = "run/.containerenv";
+pub(crate) const PATH: &str = "run/.containerenv";
 
 #[derive(Debug, Default)]
 pub(crate) struct ContainerExecutionInfo {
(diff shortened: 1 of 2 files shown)


jeckersb (approved):
Sanity checked this catches the case I was seeing:
```
⬢[jeckersb@toolbox ~]$ env BOOTC_INSTALL_LOG=trace podman-bootc run quay.io/jeckersb/centos-bootc-jeckersb:latest
TRACE starting
TRACE Preparing install
TRACE Verified uid 0 with CAP_SYS_ADMIN
TRACE OK: we're not pid 1
ERROR Installing to disk: /proc/1 is owned by 65534, not zero; this command must be run in the root user namespace (e.g. not rootless podman)
Error: unable to install bootc image: failed to create disk image: failed to run bootc install
```

jeckersb:
Test failure looks like an AWS flake?

```
+ greenprint 'Install centos-stream-9 bootc system'
++ date -Isecond
+ echo -e '\033[1;32m[2024-03-21T16:59:17-04:00] Install centos-stream-9 bootc system\033[0m'
[1;32m[2024-03-21T16:59:17-04:00] Install centos-stream-9 bootc system[0m
+ ansible-playbook -v -i /tmp/tmp.JzrW00HDAJ/inventory -e test_image_url=quay.io/redhat_emp1/centos-bootc-dev-test:j8qu playbooks/install.yaml
[WARNING]: Found both group and host with same name: guest
Using /var/ARTIFACTS/work-awsz91uao0l/plans/install-upgrade/aws/discover/default-0/tests/tests/integration/playbooks/ansible.cfg as config file

PLAY [guest] *******************************************************************

TASK [Gathering Facts] *********************************************************
fatal: [guest]: UNREACHABLE! => changed=false 
  msg: |-
    Data could not be sent to remote host "3.231.155.52". Make sure this host can be reached over ssh: ssh: connect to host 3.231.155.52 port 22: Connection timed out
  unreachable: true
```

cgwalters:
> Test failure looks like an AWS flake?

Yeah, right now we aren't always gating on these, though we probably should.  But...I would also like to change the testing framework to inject testing code into the OS image, instead of being ssh-push based as it is now.  The former is much more in line with the philosophy here :smi

PR #417, opened 2024-03-22
Add a systemd generator to fixup Anaconda's /etc/fstab

cgwalters:
Add a systemd generator to fixup Anaconda's /etc/fstab

This is a giant and hacky workaround for
https://github.com/ostreedev/ostree/issues/3193

The better real fix is probably in either systemd or anaconda
(more realistically both) but let's paper over things here for now.

Having code to run as a generator will likely be useful in the
future anyways.

---


lib/src/deploy.rs
```diff
@@ -377,3 +379,127 @@ fn test_switch_inplace() -> Result<()> {
     assert_eq!(replaced, target_deployment);
     Ok(())
 }
+
+/// A workaround for https://github.com/ostreedev/ostree/issues/3193
+/// as generated by anaconda.
+#[context("Updating /etc/fstab for anaconda+composefs")]
+pub(crate) fn fixup_etc_fstab(root: &Dir) -> Result<()> {
```
cgwalters, line 386: I have to say one really nice thing about pervasively using cap-std is that it makes it trivial to unit-test things like this in a super reliable fashion without accidentally trying to edit the system global `/etc/fstab` when running `cargo test` and also without going all the way to running containers.  (Which we should do too of course...)

cgwalters:
@henrywang we aren't testing anaconda in our integration tests yet, right?  Do you know offhand how hard that'd be to wire up?

A great thing to do here would be to change the tests to verify no systemd units fail across the board, which is the goal of this change.

henrywang:
> @henrywang we aren't testing anaconda in our integration tests yet, right? Do you know offhand how hard that'd be to wire up?
> 
> A great thing to do here would be to change the tests to verify no systemd units fail across the board, which is the goal of this change.

We have anaconda test in `virt-s1/bootc-workflow-test` with a daily run already. https://github.com/virt-s1/bootc-workflow-test/blob/main/anaconda.sh. Adding anaconda test into bootc repo is not hard.

henrywang:
> Ah OK. Yeah I think it could be an opt-in? Well...here's what I'd propose...let's land this change and then it will show up very quickly in the -dev images, and the workflow test runs against our -dev images and we can verify it works there?

Exactly. `-dev` image will be tested every day, or it can be manually triggered as well.

cgwalters:
> @cgwalters One more question. Does anaconda support bootc install already? Thanks.

No, that's https://github.com/rhinstaller/anaconda/discussions/5197

jeckersb (approved):
LGTM, love the tests!

henrywang:
@cgwalters Anaconda test passed with `bootc-202403251937.g0c250ea906-1.el9.x86_64
`. Test log: https://github.com/virt-s1/bootc-workflow-report/pull/205

cgwalters:
Yep cool, I no longer see the `systemd-remount-fs.service` failures in that -dev run.

runcom:
> The better real fix is probably in either systemd or anaconda
> (more realistically both)

@cgwalters are there tracker issues for this? just curious to follow and maybe help on this one to have a proper fix

champtar:
> I think the best fix is:
>     * Change anaconda not to write `/` into `/etc/fstab` for bootc/ostree systems by default

I agree, as setting ro if you are not using composefs will break the system, just commenting the '/' line works in both cases.
If you upgrade from a composefs enabled version to a composefs disabled version everything breaks.

champtar:
Doing some testing implementing a similar script and upgrading from really old versions,
to work this actually depends on having `rw` kargs as we are `Before=systemd-remount-fs.service` and `ostree-remount.service` is `After=systemd-remount-fs.service`, so without `rw` /etc is read-only and it fails

PR #418, opened 2024-03-22
feat(build): Add extra version information

### LINKED ISSUE #361: add --version argument

bcrochet:
Simple build-time code generation for now. Plan to add git info from the environment if it exists.

Fixes #361

### DIFF
--- a/lib/build.rs
+++ b/lib/build.rs
@@ -1,0 +1,22 @@
+// build.rs
+
+use std::env;
+use std::fs;
+use std::path::Path;
+
+fn main() {
+    let out_dir = env::var_os("OUT_DIR").unwrap();
+    let dest_path = Path::new(&out_dir).join("version.rs");
+    fs::write(
+        &dest_path,
+        "
+         #[allow(dead_code)]
+         #[allow(clippy::all)]
+         use clap::crate_version;
+         #[doc=r#\"Version string\"#]
+         pub const CLAP_LONG_VERSION: &str = crate_version!();
+        ",
+    )
+    .unwrap();
+    println!("cargo:rerun-if-changed=build.rs");
+}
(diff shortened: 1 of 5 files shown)


lib/build.rs
```diff
@@ -0,0 +1,22 @@
+// build.rs
+
+use std::env;
+use std::fs;
+use std::path::Path;
+
+fn main() {
+    let out_dir = env::var_os("OUT_DIR").unwrap();
+    let dest_path = Path::new(&out_dir).join("version.rs");
```
cgwalters, line 9: Just so I understand better...is this build.rs effectively a workaround for the constant generated by the clap tooling in https://docs.rs/clap/latest/clap/macro.crate_version.html not having a docstring?

Or basically can you just add a comment with *why* we need this `build.rs`?  Something like:

```
// Generate a source file with the crate version that includes a documentation string
// so that our #[deny(missing_docs)] works
```
or something like that?

(It seems like we could send a patch to clap to add a dummy docstring too?)

bcrochet: Doing this in order to then be able to pull in other information from the environment. In this way, we could have the version string reflect that it's a dev build, or do something different at release. It doesn't do that _now_. If you want me to go even simpler now and include the build.rs later when I get an actual use-case, I'm fine with that too.

cgwalters (approved):

lib/build.rs
cgwalters: I don't have a strong opinion, good with this as as is.  Thanks for working on this!

PR #420, opened 2024-03-22
install: Change no-SELinux -> SELinux to a warning && serialize to aleph

cgwalters:
install: Change no-SELinux -> SELinux to a warning

We believe we have almost all the labeling work here covered,
so degrade this to a warning.

---

install: Change SELinux state into enum, serialize to aleph

We want to support the "installing SELinux target from SELinux-disabled
host" - but in case we run into problems, let's serialize the state
of things at install time into the aleph data, for the same reason
we save other relevant environmental data like the kernel version.

---

### DIFF
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -1242,6 +1242,20 @@ pub(crate) async fn install_to_disk(mut opts: InstallToDiskOpts) -> Result<()> {
         loopback_dev.close()?;
     }
 
+    // At this point, all other threads should be gone.
+    if let Some(state) = Arc::into_inner(state) {
+        // If we had invoked `setenforce 0`, then let's re-enable it.
+        match state.selinux_state {
+            SELinuxFinalState::Enabled(Some(guard)) => {
+                guard.consume()?;
+            }
+            _ => {}
+        }
+    } else {
+        // This shouldn't happen...but we will make it not fatal right now
+        tracing::warn!("Failed to consume state Arc");
+    }
+
     installation_complete();
 
     Ok(())
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -288,9 +288,7 @@ pub(crate) struct SourceInfo {
 pub(crate) struct State {
     pub(crate) source: SourceInfo,
     /// Force SELinux off in target system
-    pub(crate) override_disable_selinux: bool,
-    #[allow(dead_code)]
-    pub(crate) setenforce_guard: Option<crate::lsm::SetEnforceGuard>,
+    pub(crate) selinux_state: SELinuxFinalState,
     #[allow(dead_code)]
     pub(crate) config_opts: InstallConfigOpts,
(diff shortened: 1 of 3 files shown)


cgwalters:
cc https://github.com/osbuild/osbuild/pull/1682 which motivated this

cgwalters:
Also @ckyrouac - this issue was also what you hit originally right way back?  I guess now that we have podman-bootc to streamline this it matters less, but I am curious if this direct support would help you now.

lib/src/install.rs
```diff
@@ -777,27 +778,53 @@ impl RootSetup {
     }
 }
 
+pub(crate) enum SELinuxFinalState {
+    /// Host and target both have SELinux, but user forced it off for target
+    ForceTargetDisabled,
+    /// Host and target both have SELinux
+    Enabled(Option<crate::lsm::SetEnforceGuard>),
```
jeckersb, line 785: This throws a `dead_code` warning with rust 1.77, see: https://github.com/rust-lang/rust/issues/122833

Can either `#[allow(dead_code)]` here, or alternatively I've verified you can change it to the struct-like variant and hint at it with underscore, so something like this works too:

```rust
Enabled { _guard: Option<crate::lsm::SetEnforcedGuard> }
```

cgwalters: Ah thanks, yeah my toolbox here is out of date.

Hmm...so I decided to fix this in a different way because this "execute code by side effect on `Drop`" was always a code smell, so I've changed it so we actually do process it in the normal path.

cgwalters: This all said, from time to time I've thought about writing a crate which basically forks off a helper child process that acts as "cleanup on parent process exit" for exactly things like temporary directories or other "external global" state - and it'd work really well for this; we'd rerun the `setenforce 1` even we panic'd etc.  (We use `panic=abort` here)

cgwalters:
(Or of course, maybe we just try to kill off the "setenforce 0" path because it *shouldn't* be necessary anymore...)

henrywang:
TF can't deploy aarch64 runner this weekend. I'll ping TF guys on Monday.

lib/src/install.rs
```diff
@@ -1218,6 +1242,20 @@ pub(crate) async fn install_to_disk(mut opts: InstallToDiskOpts) -> Result<()> {
         loopback_dev.close()?;
     }
 
+    // At this point, all other threads should be gone.
+    if let Some(state) = Arc::into_inner(state) {
+        // If we had invoked `setenforce 0`, then let's re-enable it.
+        match state.selinux_state {
```
jeckersb, line 1248: clippy prefers `if let` here since it's only destructuring a single pattern... but there's also other preexisting things it's complaining about, so I'll just let a few accumulate and do a separate PR to clean them up all at once.

jeckersb (approved):

PR #422, opened 2024-03-24
docs: Add a new "build guidance" section

cgwalters:
I originally was thinking these docs needed to live in downstream places but...it will be really helpful
to us to have generic recommended guidance here.


miabbott (approved):
Meaty details!  LGTM

PR #425, opened 2024-03-25
Fix various accumulated clippy lints

jeckersb:

### DIFF
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -1245,11 +1245,8 @@ pub(crate) async fn install_to_disk(mut opts: InstallToDiskOpts) -> Result<()> {
     // At this point, all other threads should be gone.
     if let Some(state) = Arc::into_inner(state) {
         // If we had invoked `setenforce 0`, then let's re-enable it.
-        match state.selinux_state {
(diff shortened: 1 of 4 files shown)


lib/src/install.rs
```diff
@@ -1245,11 +1245,8 @@ pub(crate) async fn install_to_disk(mut opts: InstallToDiskOpts) -> Result<()> {
     // At this point, all other threads should be gone.
     if let Some(state) = Arc::into_inner(state) {
         // If we had invoked `setenforce 0`, then let's re-enable it.
-        match state.selinux_state {
-            SELinuxFinalState::Enabled(Some(guard)) => {
-                guard.consume()?;
-            }
-            _ => {}
+        if let SELinuxFinalState::Enabled(Some(guard)) = state.selinux_state {
```
cgwalters, line 1248: While we're here...I think a further improvement is to make this a method on `State`, i.e. `impl State { fn consume(self) -> Result<()> }` which just calls this code (to better encapsulate things).

cgwalters (approved):

cgwalters:
Looks like random infra failures, no need to block on that.

Issue #380, opened 2024-03-07
Add `install alongside`


cgwalters:
Right now we have

`podman run --rm --privileged --pid=host -v /:/target --security-opt label=type:unconfined_t "${image}" bootc install-to-filesystem --karg=console=ttyS0,115200n8 --replace=alongside /target`

We can first consider simplifying like this:

`podman run --rm --privileged --pid=host --security-opt label=type:unconfined_t -v /:/target "${image}" bootc install to-existing-root`

or so, which would assume installing to `/target`.  This consolidates `--replace=alongside and /target` to just `to-existing-root`.

I don't see a reason to support installing alongside to *not* the rootfs.

Now, if we can switch to using [dynamic mounts](https://brauner.io/2023/02/28/mounting-into-mount-namespaces.html) then we can also drop the `-v /:/target` - I'd like to test this out as it could simplify a lot of our code.

## Handling kargs in alongside

When doing an alongside install, there are a ton of use cases for picking up things from the existing environment:

- root ssh keys
- console= kargs

Or in general, picking up "configuration".

I am a bit uncertain if we try to add a lot of sugar to bootc in the near term for this versus having higher level tools do it, but...

We could in `to-existing-root` also have something like `--import-kargs=console=*` or so to fetch the existing console kernel args (match by glob).

cgwalters:
Bigger picture though, while "alongside" installs were easy to implement (it's just small bits of code on top of `to-filesystem` I do think what we really want to encourage people to do is [takeover installs](https://github.com/containers/bootc/pull/78) because that ends up fixing several things:

- We don't "leak" the entire existing prior root
- We can switch filesystem types (including supporting initializing LUKS in the cloud, etc.)
-

cgwalters:
This all said I think in *practice* people should be automating these invocations, not typing them by hand.  It's already very doable by scripting in e.g. cloud-init and that's the intended design.

cgwalters:
> I read about dynamic mounts but not sure how this can allow access to / from within the privileged container without Podman requiring a change?

I believe we can get access to the root mount namespace at least because we have `--pid=host` so we inherently have `/proc/1/ns/mnt`.

rhatdan:
Ok, good idea.

cgwalters:
One part in https://github.com/containers/bootc/pull/382

cgwalters:
One thing I wanted to sanity check here is; do we have `open_tree` in current c9s?  And we appear to, as this works

```
[...]

PR #431, opened 2024-03-26
Stop if /dev is not a bind mount with loopback

### LINKED ISSUE #352: Requires `-v /dev:/dev` for reliable loopback
Sometimes I see:

```
DEBUG exec: "losetup" "--show" "-P" "--find" "/output/bootc.raw"
losetup: /output/bootc.raw: failed to set up loop device: No such file or directory
```

I think we may need to wait on udev on the host.

bcrochet:
At container start, /dev is snapshotted, so any new device files
don't get added unless /dev is bindmounted in. For now, check that
/dev is the same as the host's /dev via fsid. If they differ, it
means that /dev is not bindmounted.

Fixes #352

### DIFF
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -1177,6 +1177,9 @@ pub(crate) async fn install_to_disk(mut opts: InstallToDiskOpts) -> Result<()> {
                 block_opts.device
             );
         }
+        if !crate::mount::is_same_as_host(Utf8Path::new("/dev"))? {
+            anyhow::bail!("Loopback mounts (--via-loopback) require host devices (-v /dev:/dev)");
+        }
     } else if !target_blockdev_meta.file_type().is_block_device() {
         anyhow::bail!("Not a block device: {}", block_opts.device);
     }
--- a/lib/src/mount.rs
+++ b/lib/src/mount.rs
@@ -57,3 +57,25 @@ pub(crate) fn mount(dev: &str, target: &Utf8Path) -> Result<()> {
         [dev, target.as_str()],
     )
 }
+
+/// If the fsid of the passed path matches the fsid of the same path rooted
+/// at /proc/1/root, it is assumed that these are indeed the same mounted
+/// filesystem between container and host.
+/// Path should be absolute.
+#[context("Comparing filesystems at {path} and /proc/1/root/{path}")]
+pub(crate) fn is_same_as_host(path: &Utf8Path) -> Result<bool> {
+    // Add a leading '/' in case a relative path is passed
+    let path = Utf8Path::new("/").join(path);
+
+    // Using statvfs instead of fs, since rustix will translate the fsid field
+    // for us.
+    let devstat = rustix::fs::statvfs(path.as_std_path())?;
+    let hostpath = Utf8Path::new("/proc/1/root").join(path.strip_prefix("/")?);
+    let hostdevstat = rustix::fs::statvfs(hostpath.as_std_path())?;
+    tracing::trace!(
+        "base mount id {:?}, host mount id {:?}",
+        devstat.f_fsid,
+        hostdevstat.f_fsid
+    );
+    Ok(devstat.f_fsid == hostdevstat.f_fsid)
+}


bcrochet:
What's here now is the "quick and dirty" fix. If this is sufficient for the time being, I'd say it's good-to-go. I'm currently looking at cleaning up some of the mount code (there are lots of places that it's done now). Some deduplication is in order.

bcrochet:
> What's here now is the "quick and dirty" fix. If this is sufficient for the time being, I'd say it's good-to-go. I'm currently looking at cleaning up some of the mount code (there are lots of places that it's done now). Some deduplication is in order.

Moved around a little bit. I've added it as an additional commit. Will collapse before merging.

lib/src/mount.rs
```diff
@@ -57,3 +57,19 @@ pub(crate) fn mount(dev: &str, target: &Utf8Path) -> Result<()> {
...
 }
+
+// Ensure the `/dev` directory exists.
+fn ensure_mountpoint(target: &str) -> Result<()> {
+    std::fs::create_dir_all(target)?;
+    Ok(())
+}
+
+/// General mount that also checks that mount path exists
+pub(crate) fn ensure_mount(dev: &str, target: &str, fstype: &str) -> Result<()> {
```
cgwalters, line 68: Minor/optional: Function with especially >2 parameters all of the same type can be easy to call with the wrong arguments.  We could change `fstype` to be an `enum FilesystemType { Devtmpfs }` for example.

Another thing to do (which can be combined) is to use `&Utf8Path` here instead of `&str` which expresses semantically that it's a path.

lib/src/mount.rs
```diff
@@ -57,3 +57,19 @@ pub(crate) fn mount(dev: &str, target: &Utf8Path) -> Result<()> {
...
     )
 }
+
+// Ensure the `/dev` directory exists.
+fn ensure_mountpoint(target: &str) -> Result<()> {
+    std::fs::create_dir_all(target)?;
+    Ok(())
+}
+
+/// General mount that also checks that mount path exists
```
cgwalters, line 67:
```suggestion
/// Create the target directory if it doesn't exist, then mount the specified filesystem
```

(It doesn't "check", it "ensures" or so right?)

lib/src/mount.rs
```diff
@@ -57,3 +57,19 @@ pub(crate) fn mount(dev: &str, target: &Utf8Path) -> Result<()> {
         [dev, target.as_str()],
     )
 }
+
+// Ensure the `/dev` directory exists.
```
cgwalters, line 61: This docstring looks wrong (and function docstrings need `///`)

lib/src/mount.rs
```diff
@@ -57,3 +57,19 @@ pub(crate) fn mount(dev: &str, target: &Utf8Path) -> Result<()> {
...
+fn ensure_mountpoint(target: &str) -> Result<()> {
+    std::fs::create_dir_all(target)?;
+    Ok(())
```
cgwalters, line 64: I'm not totally sure this function carries its weight over simply inlining the `std::fs::create_dir_all(target)` into callers?

lib/src/mount.rs
```diff
@@ -57,3 +57,19 @@ pub(crate) fn mount(dev: &str, target: &Utf8Path) -> Result<()> {
...
+}
+
+/// General mount that also checks that mount path exists
+pub(crate) fn ensure_mount(dev: &str, target: &str, fstype: &str) -> Result<()> {
+    ensure_mountpoint(target)?;
```
cgwalters, line 69: I had actually expected this call to be handling *idempotence* - i.e. we don't do the mount if the target is already mounted with the expected type.

The current docs gained a `-v /dev:/dev` recommendation for this reason, and I think we don't want to overmount it.

lib/src/blockdev.rs
```diff
@@ -80,9 +85,17 @@ pub(crate) struct LoopbackDevice {
...
+fn ensure_dev_mounted() -> Result<()> {
+    crate::mount::ensure_mount(DEVTMPFS, DEV_MOUNT_POINT, DEVTMPFS)?;
+    Ok(())
+}
+
 impl LoopbackDevice {
     // Create a new loopback block device targeting the provided file path.
     pub(crate) fn new(path: &Path) -> Result<Self> {
+        // Ensure /dev exists and is mounted
+        ensure_dev_mounted()?;
```
cgwalters, line 98: Optional: I'm OK with this, although there's some action-at-a-distance here and arguably we should be doing this very early on in the callers instead (and we could add an assertion here that it's been done).

cgwalters (commented):
Thanks for working on this, our saga of filesystem mounts will probably continue for a while :smile:

lib/src/mount.rs
cgwalters: I went to look at [the rustix mount API](https://docs.rs/rustix/latest/rustix/mount/fn.mount.html) and saw a "Fs" parameter which I thought would be an enum like this, but it's actually just a generic bytestring (ultimately of course we don't want to *need* to change rustix/glibc when a new Linux filesystem appears so I guess it has to be just a string, but in this code nothing stops us from having an enum, feels like it would make sense)

bcrochet:
I got trigger-happy with pushing. I still need to address the idempotency part.

cgwalters:
Per discussion I think there are two paths here:

- Recommend `-v /dev:/dev` and for now *error out* if we're doing loopback and `/dev` is not the real system `/dev` (I think we could statx `/dev` versus `/proc/1/root/dev` and compare `stx_mnt_id` ?)

Other two paths are:

- The re-exec flow to force on `/dev` binding
- https://github.com/containers/bootc/issues/380#issuecomment-1986526854 to do dynamic remounts from host

bcrochet:
> Per discussion I think simplest may be:
> 
> * Recommend `-v /dev:/dev` and for now _error out_ if we're doing loopback and `/dev` is not the real system `/dev` (I think we could statx `/dev` versus `/proc/1/root/dev` and compare `stx_mnt_id` ?)
> 
> Other two paths are:
> 
> * The re-exec flow to force on `/dev` binding
> * [Add `install alongside` #380 (comment)](https://github.com/containers/bootc/issues/380#issuecomment-1986526854) to do dynamic remounts from host

Agreed. Let's start with the error out and suggestion of -v /dev:/dev.

cgwalters:
Comparing the mounts I think can be done via https://docs.rs/rustix/latest/rustix/fs/fn.statx.html

bcrochet:
So, this code works as expected, however for some reason once `-v /dev:/dev` is added, the losetup is still failing for me locally. Investigating...

lib/src/mount.rs
```diff
@@ -57,3 +57,24 @@ pub(crate) fn mount(dev: &str, target: &Utf8Path) -> Result<()> {
         [dev, target.as_str()],
     )
 }
+
+/// If the fsid of the passed path matches the fsid of the same path rooted
+/// at /proc/1/root, it is assumed that these are indeed the same mounted
+/// filesystem between container and host.
+/// Path *should* not include a leading '/'. It will be stripped.
```
cgwalters: I'm fine with this as is but it actually seems simpler to me to just let it be an absolute path (because in fact we are calling it that way just above :smile: ) and rely on the `strip_prefix` to make it relative for the hostpath.

jeckersb, line 64: The emphasis is misleading here and confused me the first time I read through it.  I think "should *not*" is preferable (or maybe "*should not*").

bcrochet:
> So, this code works as expected, however for some reason once `-v /dev:/dev` is added, the losetup is still failing for me locally. Investigating...

It works on a centos VM, but not on a fedora host directly. That's the only diff I have so far.

lib/src/install.rs
```diff
@@ -1177,6 +1177,9 @@ pub(crate) async fn install_to_disk(mut opts: InstallToDiskOpts) -> Result<()> {
                 block_opts.device
             );
         }
+        if !crate::mount::is_same_as_host(Utf8Path::new("/dev"))? {
+            anyhow::bail!("Add `-v /dev:/dev` to `podman` when using --via-loopback");
```
cgwalters, line 1181: How about "Loopback mounts (--via-loopback) require host devices (`-v /dev:/dev`)" ?

lib/src/mount.rs
```diff
@@ -57,3 +57,24 @@ pub(crate) fn mount(dev: &str, target: &Utf8Path) -> Result<()> {
         [dev, target.as_str()],
     )
 }
+
+/// If the fsid of the passed path matches the fsid of the same path rooted
+/// at /proc/1/root, it is assumed that these are indeed the same mounted
+/// filesystem between container and host.
+/// Path *should not* include a leading '/'. It will be stripped.
```
cgwalters, line 64: But we do pass a leading `/` in the single caller :smile: 

How about let's just say it should be absolute and rely on the `strip_prefix`  below.

bcrochet: Good point.

cgwalters (approved):

PR #441, opened 2024-03-26
filesystem: Explain more about /etc and /var

cgwalters:


jmarrero (approved):
lgtm

mvo5 (approved):
Thank you! This is really helpful!

docs/src/filesystem.md
```diff
@@ -22,22 +22,45 @@ The overall recommendation is to keep all operating system content in `/usr`.  S
...
-The `/etc` directory contains persistent state by default; however,
+The `/etc` directory contains mutable persistent state by default; however,
 it is suppported to enable the [`etc.transient` config option](https://ostreedev.github.io/ostree/man/ostree-prepare-root.html).
 
 When in persistent mode, it inherits the OSTree semantics of [performing a 3-way merge](https://ostreedev.github.io/ostree/atomic-upgrades/#assembling-a-new-deployment-directory)
-across upgrades.
+across upgrades.  In a nutshell:
+
+- The *new default* `/etc` is used as a base
+- The diff between current and previous `/etc` is applied to the new `/etc`
```
mvo5, line 32: Maybe an extra:
```
- Locally modified files in `/etc` different from the default `/usr/etc` (of the same deployment) will be retained
```
or something similar to give a hint to the reader what happens with local state ?

cgwalters: Thanks, I added that and also expanded a good bit more, cross-linking to the image building guidance.

PR #442, opened 2024-03-26
Add constraint on rust 1.76

jeckersb:
Renovate can't install 1.77 for some reason, and I'm not sure if it
can't install *any* version, or if something is just broken with 1.77
since it's relatively new.  So let's try 1.76 and see what happens.


cgwalters:
Very minor nit, style in this repo uses "topic prefixes", e.g. in this case I'd say "renovate: Add constraint on rust" or "ci/renovate: Add constraint on rust" for example.  Just from the bare subject one might think we were adding some constraint on the project as a whole.

cgwalters:
(Not worth respinning though)

PR #443, opened 2024-03-26
Revert "fix(deps): update all patch dependencies"

jeckersb:
Reverts containers/bootc#423

This is broken and didn't update Cargo.lock, roll it back until we figure out why renovate is failing, and then we'll have renovate re-do it correctly.


cgwalters:
Ahhhh that makes sense now.

PR #445, opened 2024-03-26
install: Add `block` to config, disable tpm2-luks unless opted-in

cgwalters:
tests: Use `..Default::default()` for install config

To make adding a new field not require touching all the tests.

---

install: Add `block` to config, disable tpm2-luks unless opted-in

This allows the container image builder more control over
`bootc install to-disk` in the installation config.  Per discussion in
https://github.com/containers/bootc/issues/421
this one definitely requires integration by the base image,
and not all of them will want it.

(Or if the do want LUKS, they may want more control over it)

The default value is `block: ["direct"]` which only enables
the simple filesystem install.

This change allows two different things:

`block: []`

With this, `bootc install to-disk` will just error out.  It's
a way to effectively disable it for those that want to use
an external installer always.

Another possibility is:

`block: ["direct", "tpm2-luks"]`

To explicitly re-enable the builtin tpm2-luks flow.

Or, one could do just `block: ["tpm2-luks"]` to enforce encrypted installs.

---

### DIFF
--- a/lib/src/install/config.rs
+++ b/lib/src/install/config.rs
@@ -6,6 +6,8 @@ use anyhow::{Context, Result};
 use fn_error_context::context;
 use serde::{Deserialize, Serialize};
 
+use super::baseline::BlockSetup;
+
 /// The toplevel config entry for installation configs stored
 /// in bootc/install (e.g. /etc/bootc/install/05-custom.toml)
 #[derive(Debug, Clone, Serialize, Deserialize, Default)]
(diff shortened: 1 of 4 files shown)


lib/src/install/config.rs
```diff
@@ -260,3 +282,40 @@ type = "xfs"
...
+
+    let c: InstallConfigurationToplevel = toml::from_str(
+        r##"[install]
+block = ["tpm2-luks"]"##,
+    )
+    .unwrap();
+    let mut install = c.install.unwrap();
+    install.canonicalize();
+    assert_eq!(install.block.as_ref().unwrap().len(), 1);
+    assert_eq!(install.get_block_setup(None).unwrap(), BlockSetup::Tpm2Luks);
```
jeckersb, line 320: Let's add one more check here to test the case of trying to pass `--block-setup` with a value disallowed by the config, something like?:

`assert!(install.get_block_setup(Some(BlockSetup::Direct)).is_err());`

jeckersb (changes requested):
Looks pretty good, just a couple minor nits/suggestions!

lib/src/install/config.rs
cgwalters: Thanks, nice suggestion!

jeckersb (approved):

PR #450, opened 2024-03-28
docs: Various updates and fixes

cgwalters:
I went through the docs and cleaned some things up, fixed some missing content, etc.

Also a patch to `xtask` to make it so that `cargo xtask manpages` works even when $PWD is in `docs`.

### DIFF
--- a/docs/src/man/bootc-rollback.md
+++ b/docs/src/man/bootc-rollback.md
@@ -1,0 +1,41 @@
+# NAME
+
+bootc-rollback - Change the bootloader entry ordering; the deployment
+under \`rollback\` will be queued for the next boot, and the current
+will become rollback. If there is a \`staged\` entry (an unapplied,
+queued upgrade) then it will be discarded
+
+# SYNOPSIS
+
+**bootc-rollback** \[**-h**\|**\--help**\] \[**-V**\|**\--version**\]
+
+# DESCRIPTION
+
+Change the bootloader entry ordering; the deployment under \`rollback\`
+will be queued for the next boot, and the current will become rollback.
+If there is a \`staged\` entry (an unapplied, queued upgrade) then it
+will be discarded.
+
+Note that absent any additional control logic, if there is an active
+agent doing automated upgrades (such as the default
+\`bootc-fetch-apply-updates.timer\` and associated \`.service\`) the
+change here may be reverted. Its recommended to only use this in concert
+with an agent that is in active control.
+
+A systemd journal message will be logged with
+\`MESSAGE_ID=26f3b1eb24464d12aa5e7b544a6b5468\` in order to detect a
+rollback invocation.
+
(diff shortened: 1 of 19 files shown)


docs/src/man/bootc-rollback.md
jeckersb, line 1: This needs to be added to `SUMMARY.md` so it shows up in the sidebar and is included.

docs/src/upgrades.md
```diff
@@ -18,21 +18,34 @@ There is also an opinionated `bootc-fetch-apply-updates.timer` and corresponding
...
 ## Rollback
 
-At the current time, bootc does not ship with an opinionated integrated
-rollback flow.  However, bootc always maintains (by default) a
-`rollback` container image that is accessible via `bootc status`.
+There is a  `bootc rollback` verb, and associated declarative interface
+accessible to tools via `bootc edit`.  This will swap the bootloader
+ordering to the previous boot entry.
+
+Man page: [bootc-rollback](man/bootc-rollback.md).
```
jeckersb, line 49: broken link; see file comment on `bootc-rollback.md`

docs/src/man/bootc-upgrade.md
```diff
@@ -22,7 +22,8 @@ upgrade \--apply\` verb which will automatically take action (rebooting)
 if the system has changed.
 
 However, in the future this is likely to change such that reboots
-outside of a \`bootc upgrade \--apply\` do \*not\* upgrade.
+outside of a \`bootc upgrade \--apply\` do \*not\* automatically apply
+the update in addition.
```
jeckersb, line 26: remove "in addition" here?  Confuses my head parser

docs/src/man/bootc-usr-overlay.md
jeckersb, line 1: This is an issue in the current version (nothing to do with this change) but maybe worth fixing here -- The sidebar refers to this as `bootc usroverlay` but everything here uses `bootc usr-overlay`.  I see there's a clap alias so either is valid, but I think it's better to update the sidebar in `SUMMARY.md` to use `bootc usr-overlay` so it's consistent across the docs.

jeckersb (changes requested):
Looks mostly good to me, just noted a couple of minor things

docs/src/man/bootc-rollback.md
cgwalters: Great catch!

jeckersb (approved):

cgwalters:
OK the `git rev-parse` failure is weird; took a stab at both propagating stderr to aid debugging, and I'm guessing it's the git "safe directory" stuff so I cargo culted a bit for that.

cgwalters:
Ah yes, my intuition about the safe directory bits was right :smile:

cgwalters:
(force merging since we had green CI before)

PR #457, opened 2024-04-04
docs: Add management services doc

ckyrouac:


docs/src/building/management-services.md
```diff
@@ -0,0 +1,49 @@
...
+
+# Typically when using a management service, it will determine when to upgrade the system.
+# So, disable bootc-fetch-apply-updates.timer if it is included in the base image.
+systemctl disable bootc-fetch-apply-updates.timer
+
+# Install the client from dnf, or some other method that applies for your client
+RUN dnf install management-client -y && dnf clean all
+
+# Bake the credentials for the management service into the image
+ARG activation_key=<INSERT KEY HERE>
```
cgwalters, line 16: I think a general best practice is to leave this empty, that way we can error out later if it's explicitly not set...I did that in the SSH key example.

Hmm...do you think we need this "activation key" bit in this doc?  Just talking about how the service basically takes over the bootc updates?

I'm good with it in there too.

docs/src/building/management-services.md
```diff
@@ -0,0 +1,49 @@
...
+
+# The existence of .run_next_boot acts as a flag to determine if the
+# registration is required to run when booting
+RUN touch /etc/management-client/.run_next_boot
+
+COPY <<"EOT" /usr/lib/systemd/system/management-client.service
+[Unit]
+Description=Run management client at boot
+After=network-online.target
+ConditionPathExists=/etc/management-client/.run_client_next_boot
```
cgwalters, line 26: Wait but wouldn't the agent want to run persistently?

docs/src/building/management-services.md
```diff
@@ -0,0 +1,49 @@
...
+
+# Link the service to run at startup
+RUN ln -s /usr/lib/systemd/system/management-client.service /etc/systemd/system/multi-user.target.wants/management-client.service
+
+# Store the credentials in a file to be used by the systemd service
+RUN echo -e "CLIENT_ACTIVATION_KEY=${activation_key}" > /etc/management-client/.credentials
+
+# Set the flag to enable the service to run one time
+# The systemd service will remove this file after the registration completes the first time
+RUN touch /etc/management-client/.run_next_boot
```
cgwalters, line 47: BTW this is a really minor nit unrelated to all things but one tricky bit with containerfiles by default is you end up with a lot of tiny layers unless you build with `--squash`, so I tend to combine `RUN` lines where it makes sense, and if it goes past 5 lines I usually move into a `build.sh` that gets copied in and it's just `RUN build.sh && rm -f build.sh`.

docs/src/building/management-services.md
```diff
@@ -0,0 +1,49 @@
...
+ExecStart=/usr/bin/management-client register --activation-key ${CLIENT_ACTIVATION_KEY}
+ExecStartPre=/bin/rm -f /etc/management-client/.run_next_boot
+ExecStop=/bin/rm -f /etc/management-client/.credentials
+
+[Install]
+WantedBy=multi-user.target
+EOT
+
+# Link the service to run at startup
+RUN ln -s /usr/lib/systemd/system/management-client.service /etc/systemd/system/multi-user.target.wants/management-client.service
```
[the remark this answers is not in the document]
ckyrouac: fixed

docs/src/building/management-services.md
ckyrouac: removed the <> chunk. I think most of these type of management services will require some type of authentication, and documented how to bake the key into the image is interesting.

docs/src/building/management-services.md
ckyrouac: I added docs to the top summary to explain why this is setup to run once.

docs/src/building/management-services.md
ckyrouac: Added info to the top explaining the containerfile is not optimized.

cgwalters (approved):
Thanks!

PR #462, opened 2024-04-05
install: Add `--skip-finalize`

cgwalters:
A generic use case for installing to an existing filesystem in the general case will be:

- `bootc install to-filesystem`
- Inject other configuration or data into `/etc` and/or `/var`

(Until such time as we offer an explicit API to do the latter)

Add a `--skip-finalize` which will avoid remounting the root. This also should work around qemu-user not implementing the FITRIM ioctl, which can be hit when doing cross-architecture installs.


henrywang:
Ansible galaxy server flaky issue has been fixed/workaround by PR #463

henrywang:
PR https://github.com/containers/bootc/pull/463 merged. I make a rebase to make test work in this PR.

mvo5:
Fwiw, the FITRIM support to qemu-user got merged Friday (https://www.mail-archive.com/qemu-devel@nongnu.org/msg1034880.html) so we could maybe ask for a backport into fedora (like we did for FIFREEZE/FITHAW).

cgwalters:
Yep makes sense, but wouldn't you agree we should still do this PR too?

jeckersb (approved):
Code looks good to me, I will approve with deferral to you guys on whether or not we really need it.  Seems like an overall improvement though that lets the user be explicit when it was otherwise implicit before (while still retaining the same behavior if not explicitly providing the flag)

jeckersb:
Oh auto-merge :man_shrugging: 

Well... ship it! :laughing:

mvo5:
> Yep makes sense, but wouldn't you agree we should still do this PR too?

Sorry, yes - it makes sense on it's own too. So thanks for this :)

PR #463, opened 2024-04-07
test: install ansible collection from tar.gz package

henrywang:
To avoid galaxy server unstable issue, like `504 gateway timeout`, `error when getting available versions of collection community.general`, download two required collection packages and install them locally. And those two collections are more general and do not have too much update often.

### DIFF
--- a/plans/install-upgrade.fmf
+++ b/plans/install-upgrade.fmf
@@ -12,7 +12,7 @@ prepare:
       - python3-devel
       - unzip
   - how: shell
-    script: ansible-galaxy collection install community.general ansible.posix
+    script: ansible-galaxy collection install https://ansible-collection.s3.amazonaws.com/ansible-posix-1.5.4.tar.gz https://ansible-collection.s3.amazonaws.com/community-general-8.5.0.tar.gz
 execute:
   how: tmt
 


plans/install-upgrade.fmf
```diff
@@ -12,7 +12,7 @@ prepare:
       - python3-devel
       - unzip
   - how: shell
-    script: ansible-galaxy collection install community.general ansible.posix
+    script: ansible-galaxy collection install https://ansible-collection.s3.amazonaws.com/ansible-posix-1.5.4.tar.gz https://ansible-collection.s3.amazonaws.com/community-general-8.5.0.tar.gz
```
cgwalters, line 15: OK, but it seems better to retry this task in a loop; this approach seems to be bypassing their metadata servers; e.g. what if they decided to store things in GCP or quay.io instead?

cgwalters (approved):

plans/install-upgrade.fmf
henrywang: I think saving to quay.io adds complex. Both GCP and S3 work for me.
BTW, I've tried to run ansible inside a container, but I dropped that for two reasons:
1. it adds a lot of complexity and test code is not easy to understand.
2. libvirt scenario blocked me.

PR #467, opened 2024-04-10
docs: Expand on HEALTHCHECK

cgwalters:
This one deserves a special entry.

### DIFF
--- a/docs/src/building/bootc-runtime.md
+++ b/docs/src/building/bootc-runtime.md
@@ -45,3 +45,19 @@ function at runtime.
 
 Ignored by bootc; typically you should configure individual services inside
 the bootc container to run as unprivileged users instead.
+
+### `HEALTHCHECK` (OCI: *no equivalent*)
+
+This is currently a Docker-specific metadata, and did not make it into the
+OCI standards.  (Note [podman healthchecks](https://developers.redhat.com/blog/2019/04/18/monitoring-container-vitality-and-availability-with-podman#))
+
+It is important to understand again is that there is no "outer container runtime" when a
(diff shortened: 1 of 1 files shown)


docs/src/building/bootc-runtime.md
```diff
@@ -44,4 +44,20 @@ function at runtime.
...
+This is currently a Docker-specific metadata, and did not make it into the
+OCI standards.  (Note [podman healthchecks](https://developers.redhat.com/blog/2019/04/18/monitoring-container-vitality-and-availability-with-podman#))
+
+It is important to understand again is that there is no "outer container runtime" when a
+bootc container is deployed on a host.  The system must perform health checking on itself (or have an external
+system do it).
+
+Relevant links:
+
+- [bootc rollback](man/bootc-rollback.md)
```
jeckersb, line 60: This link is broken; needs to be `../man/bootc-rollback.md` instead

jeckersb (changes requested):

docs/src/building/bootc-runtime.md
cgwalters: Thanks!  Out of curiosity how did you spot this?  Just by knowledge or did you do a local mdbook serve test?

Oh hey, there's even a https://github.com/Michael-F-Bryan/mdbook-linkcheck

jeckersb: Yeah for the docs stuff I always just do mdbook build and then look at it in firefox so I can see it rendered and sanity check the links work.

Nice find on the linkchecker, seems like it would catch stuff like that automatically!

jeckersb (approved):

PR #472, opened 2024-04-16
cli: Add --json support to --check

travier:
This will let higher level tools (Plasma Discover for example) more easily read the output of `bootc update --check --json`.

---

Needs https://github.com/ostreedev/ostree-rs-ext/pull/618

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -39,6 +39,10 @@ pub(crate) struct UpgradeOpts {
     #[clap(long, conflicts_with = "apply")]
     pub(crate) check: bool,
 
+    /// Output JSON
+    #[clap(long, conflicts_with = "apply", required_if_eq("check", "true"))]
+    pub(crate) json: bool,
+
     /// Restart or reboot into the new target image.
     ///
     /// Currently, this option always reboots.  In the future this command
@@ -433,7 +437,7 @@ async fn upgrade(opts: UpgradeOpts) -> Result<()> {
                 if let Some(previous_image) = booted_image.as_ref() {
                     let diff =
                         ostree_container::ManifestDiff::new(&previous_image.manifest, &r.manifest);
-                    diff.print();
+                    diff.print(opts.json);
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


cgwalters:
The original idea here is that the result of `--check` is cached, so you can just run `bootc status --json` to get that data after invoking `bootc upgrade --check`.  So I don't believe we need this, agreed?

travier:
Thanks for pointing that. Will have to give this a try then.

cgwalters:
However, right now we don't expose the diff in the status.  I think it would likely make sense to do that.  There's a bit of a messy thing here as we would need to probably change `cached_update` to be its own struct.  

Also this relates to https://github.com/ostreedev/ostree-rs-ext/pull/618 in that in the end we'd need to create a new `ContainerImageDiff` structure here anyways (because the API types for this crate need to be abstracted from those of ostree).

cgwalters:
> But does bootc offer an unprivileged path right now?

No.  Historically the polkit/unprivileged stuff in both rpm-ostree and things like PackageKit are big, complex beasts that add a lot of security-sensitive code paths. My thinking here is that for now such a thing is probably best as a separate project on top of bootc.  Or maybe we could add it as an optionally-installed secondary component here.

One thing I would say on this is that I also think many of the same concerns should apply to "workload" containers such as podman; ideally we share code there.  Access to the podman (and docker) remote API is equivalent to root in practice right now, note.   It may make sense to have a single component which can initiate upgrades for bootc as well as wrap e.g. https://docs.podman.io/en/latest/markdown/podman-auto-update.1.html (i.e. no ability to change system state, just do updates)

travier:
Creating a small daemon with a very limited, unprivileged DBus API would make sense. For GNOME Software/Plasma Discover, we mostly need "check for update" & "update" commands. And maybe something for "rebase/switch", but it would be hard to make this one unprivileged.

travier:
Might be needed to coordinate with the dnf team here. They already have a DBus daemon that does (a larger subset of) that.

cgwalters (changes requested):
> However, right now we don't expose the diff in the status.

This is what we should do.

cgwalters:
I believe https://github.com/containers/bootc/pull/472#issuecomment-2061219087 covers this, please reopen if anyone disagrees.

cgwalters:
https://github.com/containers/bootc/issues/932 tracks the diff

PR #478, opened 2024-04-18
docs: Add link to the bootable containers webpage

stefwalter:
This helps share the broader context and mission of the project when thinking of it beyond bootc.

### DIFF
--- a/docs/src/intro.md
+++ b/docs/src/intro.md
@@ -1,6 +1,7 @@
 # bootc
 
 Transactional, in-place operating system updates using OCI/Docker container images.
+bootc is the key component in a broader mission of [bootable containers](https://containers.github.io/bootable/).
 
 The original Docker container model of using "layers" to model
 applications has been extremely successful.  This project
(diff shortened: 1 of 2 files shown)


docs/src/intro.md
```diff
@@ -1,6 +1,7 @@
 # bootc
 
 Transactional, in-place operating system updates using OCI/Docker container images.
+bootc is the key component in a broader mission of [bootable containers](https://containers.github.io/bootable/).
```
jeckersb, line 4: Very minor nit, but it's a bit awkward to start the sentence with bootc as a lower-case 'b', but big 'B' seems equally wrong.  A quick scan elsewhere in the docs turned up https://containers.github.io/bootc/building/guidance.html#systemd-units which has a sentence beginning with "The bootc project".  I suspect maybe @cgwalters was trying to avoid the same thing.

cgwalters (approved):

docs/src/intro.md
cgwalters: Yeah agree, but...not worth a respin for this I think right now.

PR #479, opened 2024-04-19
install: Use partx to discover loopdev partitions

albertofaria:
When running `bootc install to-disk --via-loopback... ` under libkrun, BLKRRPART appears to have no effect. Switching to partx solves it.

### DIFF
--- a/lib/src/blockdev.rs
+++ b/lib/src/blockdev.rs
@@ -141,30 +141,6 @@ pub(crate) fn udev_settle() -> Result<()> {
     Ok(())
 }
 
-#[allow(unsafe_code)]
-pub(crate) fn reread_partition_table(file: &mut File, retry: bool) -> Result<()> {
-    let fd = file.as_raw_fd();
-    // Reread sometimes fails inexplicably.  Retry several times before
-    // giving up.
-    let max_tries = if retry { 20 } else { 1 };
-    for retries in (0..max_tries).rev() {
-        let result = unsafe { ioctl::blkrrpart(fd) };
-        match result {
-            Ok(_) => break,
-            Err(err) if retries == 0 && err == Errno::EINVAL => {
-                return Err(err)
-                    .context("couldn't reread partition table: device may not support partitions")
-            }
-            Err(err) if retries == 0 && err == Errno::EBUSY => {
-                return Err(err).context("couldn't reread partition table: device is in use")
-            }
-            Err(err) if retries == 0 => return Err(err).context("couldn't reread partition table"),
-            Err(_) => std::thread::sleep(std::time::Duration::from_millis(100)),
-        }
-    }
-    Ok(())
-}
-
 /// Parse key-value pairs from lsblk --pairs.
 /// Newer versions of lsblk support JSON but the one in CentOS 7 doesn't.
 fn split_lsblk_line(line: &str) -> HashMap<String, String> {
--- a/lib/src/install/baseline.rs
+++ b/lib/src/install/baseline.rs
@@ -269,14 +269,12 @@ pub(crate) fn install_create_rootfs(
     tracing::debug!("Created partition table");
 
     // Reread the partition table
-    {
-        let mut f = std::fs::OpenOptions::new()
-            .write(true)
-            .open(&device)
-            .with_context(|| format!("opening {device}"))?;
-        crate::blockdev::reread_partition_table(&mut f, true)
-            .context("Rereading partition table")?;
-    }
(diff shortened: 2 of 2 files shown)


cgwalters:
> When running bootc install to-disk --via-loopback... under libkrun,

Hmm.  I'm not opposed to this change but I feel like we probably want to do at least a little bit of root cause analysis...where is the loopback file mounted?

When you say "under libkrun"...is that like a container-as-VM environment, i.e. is systemd and especially udev running or not?  It must be.  Or wait, could this be related to the custom libkrun kernel?  Any theories at all for why this might fail would be appreciated.

On the tangentially related positive side; dropping this `ioctl` would be a step towards letting us migrate off the `nix` crate and fully use `rustix` only, which would be nice.
We're probably going to continue to fork external binaries for partitioning for the forseeable future, and depending on `util-linux-core` for partx is totally fine by me.

cgwalters:
Also, the direct equivalent of the ioctl here is `blockdev --rereadpt` - partx seems to be doing more.  I am curious if forking `blockdev --rereadpt` works?

albertofaria:
I'm running a bootc container under krun. The entrypoint/PID 1 is /bin/sh running a script that forks bootc.

`blockdev --rereadpt` also has no effect.

partx uses its own logic to detect partitions, then adds them to the kernel table using `ioctl(BLKPG)`.

I'm assuming this boils down to libkrunfw's kernel config, although it's unclear to me what the relevant options would be. I can dig deeper if needed.

Assuming partx's logic is as good as the kernel's, this patch should have the effect of making bootc compatible with more kernels.

cgwalters:
Hmm it might be this https://github.com/containers/libkrunfw/blob/f9bf3af4fc04da8e59c0b09483d0f0c2fdee106a/config-libkrunfw_x86_64#L987 - that option looks like it's very very legacy.

albertofaria:
I tried setting CONFIG_UEVENT_HELPER=n and unsetting CONFIG_UEVENT_HELPER_PATH, but I still get:

```
ERROR Installing to disk: Creating rootfs: Failed to find children after partitioning
```

albertofaria:
@cgwalters would you prefer that whatever kernel config issue is solved, or that bootc is able to tolerate this sort of issue?

cgwalters:
> I'm running a bootc container under [krun](https://github.com/containers/crun/blob/main/krun.1.md). The entrypoint/PID 1 is /bin/sh running a script that forks bootc.

Isn't this the real problem?  bootc is basically expecting systemd to be running as pid 1 because it's udev that will handle creating devices as triggered by the kernel normally (with devtmpfs as a baseline fallback, but I think that's not being covered for the loopback devices).

I am really uncertain about the value of trying to support non-systemd cases at install time.  What use case is there for that given that the *target* environment will have systemd anyways?

albertofaria:
I understand. I haven't measured it, but my intention was to run as few things as possible under krun to avoid additional delays.

I can fix things on my end, so I'm closing the MR.

cgwalters:
Don't get me wrong I'm open to negotiation here; if there's enough benefit to supporting non-systemd at install time then we can do it probably, but there really is the flip side where the install code path is already complex and fiddly enough...

BTW have you thought about the architecture we ended up at with https://github.com/containers/podman-bootc-cli where basically the install process is:

- run the container itself in privileged mode inside a podman-machine VM (i.e. systemd in VM)
- The install target is a file on the *host* mounted via virtiofs/9p

I think we could mimic this in the crunvm flow right?  (There's...also a whole lot going on with possibly trying to align podman-bootc to use krun, but also podman-machine...)

albertofaria:
> Don't get me wrong I'm open to negotiation here; if there's enough benefit to supporting non-systemd at install time then we can do it probably, but there really is the flip side where the install code path is already complex and fiddly enough...

crun-vm being the sole use case and it being able to work around it, I agree that it doesn't make sense to complicate bootc further right now.

> BTW have you thought about the architecture we ended up at with https://github.com/containers/podman-bootc-cli where basically the install process is:
> - run the container itself in privileged mode inside a podman-machine VM (i.e. systemd in VM)

crun-vm does this but with libkrun instead of podman-machine, meaning we can fairly easily extend its bootc container support to Docker/Kubernetes in the future. Also we can simply boot a VM to run bootc-install on demand, since that is very fast.

> - The install target is a file on the host mounted via virtiofs/9p

That's how it works currently: we launch the bootable container using krun to build the VM image into a dir shared with the host, krun terminates, and then we boot the final VM from the generated image using libvirt+qemu.

PR #480, opened 2024-04-20
install: Support being passed --filesystem with no install config

cgwalters:
This fixes a logical bug where for base images that don't have any install configuration at all, we errored out even if we were explicitly passed `--filesystem`.

This is just about making it easier to test base images that don't have an install config.


jeckersb (changes requested):
LGTM, one minor nit to appease our benevolent clippy overlord

jeckersb (approved):

Issue #461, opened 2024-04-05
support insecure registries


cgwalters:
I'd resisted doing this a lot, but there are enough local dev/test scenarios where it'd be useful to be able to fetch from an insecure registry.

This would probably be a simple new `insecure: true` flag in the host image spec which we end up passing down into the ostree-ext/skopeo stack.

jmarrero:
@ravanelli is going to look at this during the sprint. Let us know if that is ok.

cgwalters:
Awesome :tada: Happy to join a call and talk about it/mentor

ravanelli:
Thanks @cgwalters, I'm totally new to bootc and also to Rust. I was looking around and seems you mentioned this part: https://github.com/containers/bootc/blob/main/lib/src/spec.rs#L63 
Is it correct? 
If so, I will take the time to do the change and understand how to properly test it.

cgwalters:
That's signature verification which is different from TLS.  I stubbed this out to start

```
diff --git a/lib/src/deploy.rs b/lib/src/deploy.rs
index 3eb31a8..c35a001 100644
--- a/lib/src/deploy.rs
+++ b/lib/src/deploy.rs
@@ -121,6 +121,7 @@ pub(crate) async fn pull(
     quiet: bool,
 ) -> Result<Box<ImageState>> {
     let repo = &sysroot.repo();
+    // MODIFY HERE to eventually pass things down to the `skopeo` process with --src-tls-verify=false
     let ostree_imgref = &OstreeImageReference::from(imgref.clone());
     let mut imp = new_importer(repo, ostree_imgref).await?;
     let prep = match imp.prepare().await? {
diff --git a/lib/src/spec.rs b/lib/src/spec.rs
index 5f6df93..fbf2d44 100644
--- a/lib/src/spec.rs
+++ b/lib/src/spec.rs
@@ -74,6 +74,9 @@ pub struct ImageReference {
     /// Signature verification type
     #[serde(skip_serializing_if = "Option::is_none")]
     pub signature: Option<ImageSignature>,
+    /// Skip TLS and certificate verification; this is very insecure and
+    /// should only be used in testing environments.
+    pub disable_tls_verification: bool,
 }
 
 /// The status of the booted image
```

This will also likely require a change to...oh no wait we already have https://github.com/containers/containers-image-proxy-rs/blob/28155f45bf635edcbaf5b4e3540f3e3c54a13bd2/src/imageproxy.rs#L127

[...]

PR #483, opened 2024-04-22
Add support for insecure registries (Fix: #461)

ravanelli:
- Introduce 'insecure-disable-tls-verification' parameter for skipping TLS verification;
- Fix Issue: #461.

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -46,6 +46,11 @@ pub(crate) struct UpgradeOpts {
     /// a userspace-only restart.
     #[clap(long, conflicts_with = "check")]
     pub(crate) apply: bool,
+
+    /// Skip TLS and certificate verification.
+    /// This is very insecure and should only be used in testing environments
+    #[clap(long)]
+    pub(crate) insecure_disable_tls_verification: bool,
 }
 
 /// Perform an switch operation
@@ -62,6 +67,11 @@ pub(crate) struct SwitchOpts {
     /// This argument is deprecated and does nothing.
     #[clap(long, hide = true)]
     pub(crate) no_signature_verification: bool,
+
+    /// Skip TLS and certificate verification.
+    /// This is very insecure and should only be used in testing environments
+    #[clap(long)]
+    pub(crate) insecure_disable_tls_verification: bool,
 
     /// This is the inverse of the previous `--target-no-signature-verification` (which is now
     /// a no-op).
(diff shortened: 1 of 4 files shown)


ravanelli:
It still depends on https://github.com/containers/bootc/issues/461#issuecomment-2063710854

lib/src/cli.rs
```diff
@@ -46,6 +46,11 @@ pub(crate) struct UpgradeOpts {
     /// a userspace-only restart.
     #[clap(long, conflicts_with = "check")]
     pub(crate) apply: bool,
+
+    /// Skip TLS and certificate verification.
+    /// This is very insecure and should only be used in testing environments
+    #[clap(long)]
+    pub(crate) insecure_disable_tls_verification: bool,
```
cgwalters, line 53: I think this should always be part of the source spec; unlike `podman` I'm trying here to be declarative where we can.

This means that once you do a `bootc switch --insecure-disable-tls-verification`, it is "sticky" and applied automatically on subsequent `bootc upgrade`s.

So we can just drop this hunk I believe.

lib/src/cli.rs
```diff
@@ -103,6 +113,11 @@ pub(crate) struct EditOpts {
     /// Don't display progress
     #[clap(long)]
     pub(crate) quiet: bool,
+
+    /// Skip TLS and certificate verification.
+    /// This is very insecure and should only be used in testing environments
+    #[clap(long)]
+    pub(crate) insecure_disable_tls_verification: bool,
```
cgwalters, line 120: I don't think this one makes sense; edit just lets you edit the spec directly.

cgwalters (commented):
> It still depends on https://github.com/containers/bootc/issues/461#issuecomment-2063710854

OK yes...right.  This is harder than I thought because basically all of the existing logic is really assuming that it can serialize and deserialize the source image reference into a single string.

Which is basically not how docker (and podman) work; one is required to pass `--tls-verify=false` externally around.

Although...wait...actually, can we handle this by just telling the admin to add to the `/etc/containers/registries.conf.d` with the `[insecure]` flag?

cgwalters:
Just highlighting here
> Although...wait...actually, can we handle this by just telling the admin to add to the /etc/containers/registries.conf.d with the [insecure] flag?

Let's test this first

cgwalters:
Thanks for testing this @lukewarmtemp !  https://github.com/containers/bootc/pull/580

PR #488, opened 2024-04-25
install: Two informational patches

cgwalters:
install: Explicitly print which block setup we're using

This is an important thing.

---

install: Print more information when making filesystems

- Label is mandatory so change the API to always take it
- Print the label we're using as a description of the fs
- Also print the fs type as it's notable info

---

### DIFF
--- a/lib/src/install/baseline.rs
+++ b/lib/src/install/baseline.rs
@@ -57,6 +57,12 @@ pub(crate) enum BlockSetup {
     Tpm2Luks,
 }
 
+impl Display for BlockSetup {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.to_possible_value().unwrap().get_name().fmt(f)
+    }
+}
+
 impl Default for BlockSetup {
     fn default() -> Self {
         Self::Direct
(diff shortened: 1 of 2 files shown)


henrywang:
rhel94 test failure has been fixed by PR https://github.com/containers/bootc/pull/490

lib/src/install/baseline.rs
```diff
@@ -113,11 +119,11 @@ fn sgdisk_partition(
 fn mkfs<'a>(
     dev: &str,
     fs: Filesystem,
-    label: Option<&'_ str>,
+    label: &str,
     opts: impl IntoIterator<Item = &'a str>,
 ) -> Result<uuid::Uuid> {
     let u = uuid::Uuid::new_v4();
-    let mut t = Task::new("Creating filesystem", format!("mkfs.{fs}"));
+    let mut t = Task::new("Creating {label} filesystem ({fs}", format!("mkfs.{fs}"));
```
jeckersb, line 126: description needs wrapped in `format!` here, also missing closing ')' in description str

jeckersb (changes requested):

lib/src/install/baseline.rs
cgwalters: Thanks!  I had apparently forgotten to `git commit --amend`

cgwalters:
Man, the signoff thing...

lib/src/task.rs
```diff
@@ -9,9 +9,25 @@ use cap_std::fs::Dir;
...
+enum CmdVerbosity {
+    /// Nothing is output by default
+    Quiet,
+    /// Only the task description is output
+    Description,
+    /// The task description and the full command line are output
+    Verbose,
+}
+
+impl Default for CmdVerbosity {
```
jeckersb, line 22: Instead of manual `impl Default` here, better would be to derive it on the enum and add the `#[default]` attribute on `Description` - https://doc.rust-lang.org/std/default/trait.Default.html#enums

jeckersb (changes requested):

lib/src/task.rs
cgwalters: Huh wait did that always exist?  I had a feeling it was relatively recent and maybe that's why I'd internalized the manual impl.  Anyways done and changed one other place!

jeckersb (approved):

lib/src/task.rs
jeckersb: Looks like [since 1.56](https://github.com/rust-lang/rust/pull/86735/commits/c70147fd66e08962ab06adf12eb6a41bc1ea7f08)

PR #492, opened 2024-04-26
ci: Add concurrency cancellation

cgwalters:
Every time I (or someone else) force pushes to re-add the DCO we're wasting resources and/or just have added latency.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,10 @@ on:
 env:
   CARGO_TERM_COLOR: always
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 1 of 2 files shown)


cgwalters:
(That said the real problem here may be that GHA is out to lunch?)

jeckersb (approved):
For reference, [github docs about this](https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#concurrency) since I had to look this up to understand it

cgwalters:
OK so it seems like it's a testing farm backlog which is then holding open GHA runners which has caused a CI deadlock.  This PR should help.  Forcing merge.

cgwalters:
And I've now manually cancelled all prior outstanding integration test runs.

PR #497, opened 2024-04-26
docs: Move selinux into image info, add toplevel gotcha

cgwalters:
We had SELinux info but I was expecting it in the
"information about images" section, not the "building" section.  Move it, and link from the previous place.

Then add the gotcha about toplevel directories.


cgwalters:
Hooray the control/skip-ci label seems to work as intended!

cgwalters:
Though a messy thing here is some of those CI contexts are marked required...so we'll need to bypass branch protections for now.  Hard to deal with without some custom work for branch protection.

jeckersb (approved):

PR #501, opened 2024-04-30
users-and-groups: Document useradd httpd more

cgwalters:
This one is going to be a very common issue.

It is highly tempting to add some automatic merge logic for `/etc/passwd` to ostree, which would have different tradeoffs.


jmarrero (approved):
lgtm

PR #503, opened 2024-05-02
docs: Add some description of container storage

cgwalters:


jeckersb (approved):
Instead of repeatedly making you fix little nits, I just tacked some fixup commits onto the end so you don't have to do it yourself.  Let me know if you like that or if you'd prefer for me to just keep nagging you in comments :laughing: 

I really liked this one, lots of little things that I didn't really understand before and these docs help clarify how it all works!

cgwalters:
> Instead of repeatedly making you fix little nits, I just tacked some fixup commits onto the end so you don't have to do it yourself. Let me know if you like that or if you'd prefer for me to just keep nagging you in comments 😆

SGTM, thanks!

PR #508, opened 2024-05-03
Update MSRV to 1.75

jeckersb:
Currently the effective MSRV is 1.74.1 per `cargo msrv`.  We have 1.75
available in c9s, so let's target that at minimum.

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -6,7 +6,7 @@ name = "bootc-lib"
 readme = "README.md"
 repository = "https://github.com/containers/bootc"
 version = "0.1.10"
-rust-version = "1.64.0"
+rust-version = "1.75.0"
 build = "build.rs"
 
 include = ["/src", "LICENSE-APACHE", "LICENSE-MIT"]


cgwalters (commented):
I think for now our baseline target should be CentOS Stream 9.  It stays relatively up to date.  We could have a periodic job that polls the latest c9s and bumps our MSRV to whatever rust version it finds there.

Trying to actually *validate* the MSRV by installing that toolchain version sounds nice too in addition, but I think the larger problem is that the MSRV is just stale and automating bumping it is the general fix.

jeckersb:
> I think for now our baseline target should be CentOS Stream 9. It stays relatively up to date. We could have a periodic job that polls the latest c9s and bumps our MSRV to whatever rust version it finds there.
> 
> Trying to actually _validate_ the MSRV by installing that toolchain version sounds nice too in addition, but I think the larger problem is that the MSRV is just stale and automating bumping it is the general fix.

Yeah that's basically the conclusion I'm reaching too, as I run awkwardly into the wall trying to do the validation.  I'll prune this back to just do the manual version bump for now and then look into how the periodic jobs work.

cgwalters (approved):

PR #512, opened 2024-05-06
docs: Describe how to boot local builds

cgwalters:
We want to make this use case feel awesome; today it doesn't for multiple reasons.

We probably want a `bootc status --booted-image` which would print the digested pull spec to start.

Then one could at least `podman pull $(bootc status --booted-image)`.

But we clearly *also* want to streamline exporting from the local ostree storage back into a container, which has some downsides/bugs right now.

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -18,6 +18,7 @@
 
 - [Upgrade and rollback](upgrades.md)
 - [Offline/disconnected updates](offline-updates.md)
+- [Booting local builds](booting-local-builds.md)
 - [`man bootc`](man/bootc.md)
 - [`man bootc status`](man/bootc-status.md)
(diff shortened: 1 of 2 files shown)


jeckersb (approved):
LGTM.  I like `bootc status --booted-image` for ergonomics.  Hopefully I can get #215 sorted out soon once all of the excitement of the week is over, then that fixes a lot of this.

ericcurtin:
To workaround, no "bootc status --booted-image", I did this:

https://github.com/ericcurtin/dnf-bootc/blob/main/bootc.py

```
    def get_os_from_bootc_status(self):
        data = subprocess.run(["bootc", "status"], capture_output=True, text=True, check=True)
        data = yaml.safe_load(data.stdout)
        image = data['spec']['image']['image']
        return f"FROM {image}\n"
```

I notice there's a --booted option that didn't seem to make any difference to the output.

PR #524, opened 2024-05-13
deploy: Add number of layers to fetch progress

cgwalters:
We had a `{prefix}` in the template, but nothing was set.  Let's print the number of layers.

Now, in theory this should be a fix to ostree-ext, but doing so would require a Rust API change, and anyways in the future we want to switch to podman pulls anyways.

### DIFF
--- a/lib/src/deploy.rs
+++ b/lib/src/deploy.rs
@@ -113,6 +113,55 @@ pub(crate) fn check_bootc_label(config: &ostree_ext::oci_spec::image::ImageConfi
     }
 }
 
+/// Write container fetch progress to standard output.
+async fn handle_layer_progress_print(
+    mut layers: tokio::sync::mpsc::Receiver<ostree_container::store::ImportProgress>,
+    mut layer_bytes: tokio::sync::watch::Receiver<Option<ostree_container::store::LayerProgress>>,
+    total_layers: usize,
+    n_layers_fetched: &mut usize,
+) {
+    let style = indicatif::ProgressStyle::default_bar();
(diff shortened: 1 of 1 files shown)


lib/src/deploy.rs
```diff
@@ -113,6 +113,55 @@ pub(crate) fn check_bootc_label(config: &ostree_ext::oci_spec::image::ImageConfi
     }
 }
 
+/// Write container fetch progress to standard output.
+async fn handle_layer_progress_print(
```
cgwalters, line 117: See https://github.com/ostreedev/ostree-rs-ext/blob/9a4743a657ffe0435018d9720c6df80a486ca0f1/lib/src/cli.rs#L566 from which this forked to add the layer count.

henrywang:
Please ignore rhel 9.4 integration test failure. The new `rhel9-rhel_bootc:rhel-9.4` image sets `Compose ID` to ''. So the `/etc/yum.repo.d/rhel-9.repo` can't be configured without `Compose ID`. I'll update the test in another PR later to fix this issue.

jeckersb (approved):
LGTM, sanity checked it works as expected:

```
[root@ibm-p8-kvm-03-guest-02 ~]# ./bootc switch quay.io/centos-bootc/centos-bootc:stream9
layers already present: 0; layers needed: 66 (835.7 MB)
[24/66] 1.33 MiB [███████░░░░░░░░░░░░░] (0s) Fetching ostree chunk sha256:594c21b0ae2a (3.5 MB)                                                                                                                                                                                   
```

PR #529, opened 2024-05-14
install/to-disk: Drop separate /boot by default

cgwalters:
In order to simplify what we're doing here, let's drop the separate `/boot` aka XBOOTLDR partition by default for the `to-disk --block-setup=direct` path (the default).

We retain it when using `--block-setup=tpm2-luks` as it's required there.

Notably this kills off a hardcoded "ext4 for /boot" which is suboptimal for many reasons.

Longer term again I'd like to emphasize `install to-filesystem` with external installers, plus integrating external installation scripts as part of `bootc install to-disk`.

xref:
- https://github.com/containers/bootc/issues/499
- https://github.com/containers/bootc/issues/440


jmarrero (approved):
lgtm

PR #530, opened 2024-05-14
install/baseline: Make `/boot` the same fs type as `/`

cgwalters:
Let's not hardcode a fs type.


cgwalters:
Note this *textually* but not logically conflicts with https://github.com/containers/bootc/pull/529 - whichever merges first I'll rebase the other.

jmarrero (approved):
lgtm

PR #533, opened 2024-05-15
Revert "ci/docs: Add mdbook-linkcheck"

cgwalters:
This reverts commit 4f2a30ccbf9dbef9f3da043211f0d2058b207ba4. It broke the main docs site.

Closes: https://github.com/containers/bootc/issues/532


vrothberg (approved):
LGTM
Are the docs rebuilt for each PR so we can double-check?

cgwalters:
They're built on pushes to git main, yes.

BTW I tested this change by pushing to my fork and verified it works; anyone hacking on the docs should have this as a SOP.

jeckersb:
> They're built on pushes to git main, yes.
> 
> BTW I tested this change by pushing to my fork and verified it works; anyone hacking on the docs should have this as a SOP.

I see what happened here.  Previously, the html would land directly into `docs/book`, but since introducing linkcheck adds another backend, it goes into `docs/book/html` instead (and there's a `docs/book/linkcheck` where linkcheck does whatever it does).  I *did* test this locally, but I didn't notice that the directory changed so I just followed the extra `html` subdir without thinking much about it.

I just enabled the docs workflow on my fork, and tested this with the old PR and confirmed the deployed docs gives a 404, but if you append `/html` the docs are there correctly.  This is a one-line fix in the upload step to use the html subdir instead, which is now [working correctly on my fork](https://john.eckersberg.com/bootc/).  I'll put in an updated PR with the fixed version.

PR #534, opened 2024-05-15
docs/install: A few changes

cgwalters:
- Emphasize `to-filesystem` more, clarify `to-disk` is more "demo"; we definitely support "external installers" too with `to-filesystem`.
- Add machine-local state post-install section


vrothberg (approved):
LGTM

PR #544, opened 2024-05-17
docs: Set bin_name and disable version option for subcommands

PaulYuuu:
`bin_name` will be displayed in the `SYNOPSIS` section, with this, the
generated man pages have correct name and synopsis.

Also, remove `.title("bootc")` to let the subcommands have their own
titles.

Subcommands don't have their own versions, so disable the "version"
option.

Signed-off-by: Yihuang Yu <yihyu@redhat.com>


PaulYuuu:
Well, I don't know if I need to generate those markdown files, but let me generate them for preview.

cgwalters (approved):
Nice, thanks!

PR #547, opened 2024-05-17
two minor patches

cgwalters:
Colin Walters (2):
      hack: Also support --build-arg=base=<fedora>
      lib: Run `cargo fmt`


jeckersb (approved):
lgtm!

PR #549, opened 2024-05-18
install: A few minor patches

cgwalters:
install: Add a spinner for container deployment

This can take substantial time.

---

install: Be verbose for EFI creation and bootloader install

We previously changed the rootfs `mkfs` invocation to be
verbose; for consistency let's do the same for the EFI partition
as well as the bootloader run just so people can see
what's going on more.

---

bootloader: Drop --src-root /

It's the default.

---

### DIFF
--- a/lib/src/bootloader.rs
+++ b/lib/src/bootloader.rs
@@ -27,5 +27,8 @@ pub(crate) fn install_via_bootupd(
             device.as_str(),
             rootfs.as_str(),
         ]);
-    Task::new_and_run("Running bootupctl to install bootloader", "bootupctl", args)
+    Task::new("Running bootupctl to install bootloader", "bootupctl")
+        .args(args)
+        .verbose()
+        .run()
 }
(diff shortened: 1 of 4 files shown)


cgwalters:
I think we can ignore the failing tests here, the `cargo fmt` one is fixed in another PR, and I think the other rhel integration test fixes are in flight in another PR too.

jeckersb (approved):
Looks good!

cgwalters:
On Tue, May 21, 2024, at 2:07 PM, John Eckersberg wrote:
> 
> 👍 I sanity checked this on my system, it was too fast to actually see 
> the spinner but I verified it omits the `Deployment complete` output 
> and instead just puts out `done` as expected. I'm assuming if the 
> deployment was slower I would see the actual spinner 😄

If you're testing with podman-bootc you need
https://github.com/containers/podman-bootc/pull/37
and its followup in https://github.com/containers/podman-bootc/pull/41
to see the spinner.

PR #559, opened 2024-05-22
Add support for (weakly) "lifecycle bound" podman images

cgwalters:
This is a working PoC implementation of part of
https://github.com/containers/bootc/issues/128

Demo:

```
$ cat Containerfile
FROM localhost/bootc
COPY *.image /usr/share/containers/systemd
$ cat foo.image
[Container]
# bootc: bound
Image=quay.io/centos/centos:stream9
$ podman build -t localhost/testbootc .
$ podman-bootc run localhost/testbootc
...
[root@ibm-p8-kvm-03-guest-02 ~]# podman images
REPOSITORY             TAG         IMAGE ID      CREATED       SIZE
quay.io/centos/centos  stream9     75a875ea6cd8  43 hours ago  163 MB
[root@ibm-p8-kvm-03-guest-02 ~]#
```

---

Example user story:

- Admin can take the standard podman-systemd .image files they have and add a special marker
- When generating a disk image *and* at `bootc upgrade` time, bootc will pre-fetch these container images into the standard `/var/lib/containers/storage` location
- This means the default case avoids firstboot latency (see all the comments in podman-systemd about image pull timeouts)

However, the container images and containers can still be updated live if desired, and that's actually expected.  For example, I might update a version of an app *before* the base image's tag.

(a bit more in e.g. https://docs.fedoraproject.org/en-US/bootc/running-containers/#_lifecycling_and_updating_containers_separate )

---

notes:

- *edit* filed https://github.com/containers/podman/issues/22785
- Why not do this by default for all .image files? We could consider that, and having a way to exclude things instead. Either way we should clearly get out of the "magic comment" business and have a proper documented flag, but it'd require changes to podman
- This won't work with anaconda until we fix https://github.com/rhinstaller/anaconda/discussions/5197
- This further increases the problems we have with `/var` 
- bootc-image-builder errors out with: `Error: mkdir /etc/containers/networks: read-only file system` - need to fix podman to not try to create that directory

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -193,6 +193,7 @@ dependencies = [
  "openssl",
  "ostree-ext",
  "regex",
+ "rust-ini",
  "rustix",
  "schemars",
  "serde",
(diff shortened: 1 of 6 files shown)


cgwalters:
(draft since we need docs and CI tests, and it could use some cleanup)

cgwalters:
Why not do this by default for all `.image` files?  We could consider that, and having a way to exclude things instead.  Either way we should clearly get out of the "magic comment" business and have a proper documented flag, but it'd require changes to podman.

cgwalters:
Closing in favor of https://github.com/containers/bootc/pull/659

PR #567, opened 2024-05-27
Pass env properties to merge function

lukewarmtemp:
Required for https://github.com/containers/bootc/pull/401

Adds a `EnvProperties` struct to allow for conditional merging of kargs depending on whether the archiecture matches. Future properties such as `platform.id` can be added in the future.

Signed off: Luke Yang <luyang@redhat.com>


lib/src/install/config.rs
```diff
@@ -171,9 +192,18 @@ pub(crate) fn load_config() -> Result<Option<InstallConfiguration>> {
         if let Some(config) = config.as_mut() {
             if let Some(install) = c.install {
                 tracing::debug!("Merging install config: {install:?}");
-                config.merge(install);
+                config.merge(install, &env);
             }
         } else {
+            // If arch is specified and doesn't match the current arch, remove kargs
```
cgwalters, line 198: I think the goal of this PR was just to do the passing of the environment without functional change, but this still does the functional change of handling arch-specific install args right?

So the "prep change" here would also drop this and the tests and really everything related to actually putting content in that env function.

Alternatively, I'm also fine if we just roll together the new API alongside the prep work!  IOW you could also just rework the commit message to say something like:

```
install: Add support for architecture filtering for kargs

The use case here is that some kernel arguments are
architecture specific, such as the `console=` kargs.
This allows a generic base image to default to including
arch-specific values post install.
```

lib/src/install/config.rs
```diff
@@ -81,28 +89,43 @@ where
...
-        self.filesystem.merge(other.filesystem);
-        if let Some(other_kargs) = other.kargs {
-            self.kargs
-                .get_or_insert_with(Default::default)
-                .extend(other_kargs)
+    fn merge(&mut self, other: Self, env: &EnvProperties) {
+        // if arch is specified, only merge config if it matches the current arch
+        // if arch is not specified, merge config unconditionally
+        if let Some(other_arch) = other.match_architectures {
+            if other_arch.contains(&env.sys_arch) {
```
cgwalters, line 110: We can avoid duplicating the processing logic I think via this:

```
if other.match_architectures.map(|a| a.contains(&env.sys_arch)).unwrap_or(true) {
```

lib/src/install/config.rs
```diff
@@ -171,10 +197,20 @@ pub(crate) fn load_config() -> Result<Option<InstallConfiguration>> {
...
-                config.merge(install);
+                config.merge(install, &env);
             }
         } else {
-            config = c.install;
+            // Only set the config if it matches the current arch
+            // If no arch is specified, set the config unconditionally
+            if let Some(ref mut install) = c.install {
+                if let Some(arch) = install.match_architectures.as_ref() {
+                    if arch.contains(&env.sys_arch) {
```
cgwalters, line 207: I think this can boil down to:

```
            if let Some(ref mut install) = c.install {
                if install.match_architectures.as_ref().map(|a| a.contains(&env.sys_arch)).unwrap_or(true) {
```
                    if  {

cgwalters (commented):
Thanks!  This is looking pretty close, just a nit on code duplication, and would you mind updating the docs on this?

lib/src/install/config.rs
lukewarmtemp: Oh I didn't know you could do this, very cool :)

cgwalters: No worries; I feel like it took me a solid year of writing Rust often to really start to feel like I'd mastered the whole sub-language around combinators.  But it's cool when you get there :smile:

cgwalters (approved):

PR #576, opened 2024-05-31
tests: Drop `internal-testing-api`, move to tests-integration

cgwalters:
Previous work started moving our tests into an external binary; this is just cleaner because it can test things how a user would test.  Also, we started using `libtest-mimic` to have a "real" test scaffolding that e.g. allows selecting individual tests to run, etc.

Complete the picture here by moving the remaining bits into the tests-integration binary.  We now run the `tests-integration` binary in two ways in e.g. Github Actions:

- It's compiled directly on the Ubuntu runner, and orchestrates things itself
- It's built in our default container image (Fedora) but as an external `/usr/bin/bootc-integration-tests` binary

Also while we're here, drop the kola tests.


jmarrero (approved):
lot of changes here but seems ok to me.

PR #577, opened 2024-05-31
install: Explicitly label .ostree.cfs

cgwalters:
It'd be cleaner to have this in ostree itself, but...in general right now it's really a special install-time only thing.

It's also not actually strictly speaking *needed* - since we mount this in the initramfs before policy is loaded.  However, I want to turn on a CI check that verifies that all files are labeled, and this is the sole remaining exception.


jmarrero (approved):
lgtm

PR #580, opened 2024-06-03
docs: Describe how to configure insecure registries

cgwalters:
We don't have a `--tls-verify=false` flag because:

- It's a bad idea to make it too easy to do
- When you *do* do it, you want it to be persistent/global anyways so a global config file is right

Closes: https://github.com/containers/bootc/issues/461


prestist (approved):
Lgtm, maybe a nit on the PR? Im not sure but it looks like the 't' was dropped on your comment for 'righ'

cgwalters:
> Lgtm, maybe a nit on the PR? Im not sure but it looks like the 't' was dropped on your comment for 'righ'

Yeah you're righ, I edited the PR description but didn't force push because it'd rerun CI just for that.  Let's just live with my typo for ETERNITY.

PR #581, opened 2024-06-04
Update to ostree-ext 0.14.0, gvariant 0.5.0

cgwalters:
This pulls in at least the public authfile bits we want for the podman pull changes.  Also this transitively bumps our oci-spec dependency which lets us drop some imported "darling" crates.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -394,17 +394,15 @@ dependencies = [
 
 [[package]]
 name = "containers-image-proxy"
-version = "0.5.8"
+version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


cgwalters:
Weird...

>  size 256 of data subvolume is too small, minimum 4096 blocks

I don't think that can be related?  Tried restarting the test...offhand I think that error may mean the target blockdev was too small somehow?

jmarrero (approved):
lgtm

jeckersb:
> Weird...
> 
> > size 256 of data subvolume is too small, minimum 4096 blocks
> 
> I don't think that can be related? Tried restarting the test...offhand I think that error may mean the target blockdev was too small somehow?

Glad to know I'm not the only one who was confused by that one :laughing:

PR #582, opened 2024-06-04
Drop unused serde_with

cgwalters:
I noticed we had some old crates being pulled in, and this was the cause.  We aren't actually using it...so drop it.


jmarrero (approved):
lgtm

PR #584, opened 2024-06-04
ci: Fully test file labeling post-install

cgwalters:
Now that we've started to label the last holdout in the `.ostree.cfs` superblock, let's verify everything is properly labeled.


jmarrero (approved):
had to dig to https://github.com/cgwalters/bootc/blob/3b7e799620725e764527156141173e4e447a86dc/lib/src/privtests.rs#L173 to understand what is happening here.

lgtm

cgwalters:
That's exactly it!  Thanks for checking.

PR #586, opened 2024-06-05
tests/integration: install libzstd-devel

jeckersb:
Required to build zstd-sys - https://artifacts.osci.redhat.com/testing-farm/66cab0ec-1653-42ec-b482-a90001dbd65a/work-awskhuoovz0/plans/install-upgrade/aws/execute/data/guest/default-0/tests/integration/install-upgrade/rpm-build-1/output.txt

### DIFF
--- a/contrib/packaging/bootc.spec
+++ b/contrib/packaging/bootc.spec
@@ -20,6 +20,7 @@ BuildRequires: zlib-devel
 BuildRequires: ostree-devel
 BuildRequires: openssl-devel
 BuildRequires: systemd-devel
+BuildRequires: libzstd-devel
 
 %description
 %{summary}
--- a/tests/integration/mockbuild.sh
+++ b/tests/integration/mockbuild.sh
@@ -12,7 +12,7 @@ function redprint {
 }
 
 greenprint "📥 Install required packages"
-dnf install -y cargo zstd git openssl-devel ostree-devel rpm-build mock podman skopeo jq
+dnf install -y cargo zstd git libzstd-devel openssl-devel ostree-devel rpm-build mock podman skopeo jq
 cargo install cargo-vendor-filterer
 
 greenprint "⛏ Build archive"


jmarrero:
lgtm
however can you add to the description, why and what this fixes or preps for?

jmarrero (approved):

cgwalters:
This also relates to https://github.com/containers/bootc/commit/0455f808b12e587dbf7700a8468fb984b2c78b3e - in chatting with @henrywang I was hoping that we could rework the integration tests to always start from our `hack/Containerfile` to get closer to a single source of truth for the build starting point.

cgwalters:
I think it's because the framework is passing `--resultsdir` to mock, but the test harness isn't picking up the files written there.

cgwalters:
I'm not sure; maybe it's somehow a flake.  I tried restarting the jobs.

jeckersb:
Ah I think we need to update the buildrequires from the spec in contrib/, testing that now locally to see if that's sufficient.

jeckersb:
> Ah I think we need to update the buildrequires from the spec in contrib/, testing that now locally to see if that's sufficient.

Seems to fix it for me locally, added a new commit with that, we'll see how it shakes out in CI

cgwalters:
Ah right.  Hmm...I might lean a bit towards always fetching the spec from fedora instead.  This is a classic problem with specs/dpkg/arch etc in upstream, it just bitrots as it's not canonical.

(Actually though, in *theory* in Fedora we could make use of the "dynamic buildrequires" stuff to auto-determine from our source external dependencies we have, but not worth doing right now)

cgwalters:
It's actually kind of tempting to change ostree-ext to also just support forking off `/usr/bin/zstd`, I don't have any problem with taking a hard dep on it...

jeckersb:
Now it looks like quay issues:
```Error: authenticating creds for "quay.io": received unexpected HTTP status: 504 Gateway Time-out```

cgwalters:
I guess let's give the other jobs a little bit more time to see if they get farther.  But if not, we can just merge over red for those too.

cgwalters:
Ooh no it looks like https://github.com/containers/bootc/pull/581 did break more things:

>     [31mERROR[0m Installing to filesystem: Verifying fetch: Creating importer: Conflicting authentication options

cgwalters:
https://github.com/ostreedev/ostree-rs-ext/pull/639

PR #588, opened 2024-06-06
Makefile: Serialize tarballs reproducibly

cgwalters:
This ensures that when we rerun the container build without changing the source, we get the same thing.  Which then further ensures we don't invalidate any cached disk images, etc.


Makefile
```diff
@@ -1,5 +1,9 @@
 prefix ?= /usr
 
+SOURCE_DATE_EPOCH ?= $(shell git log -1 --pretty=%ct)
+# https://reproducible-builds.org/docs/archives/
```
jeckersb, line 4: :heart: 

This was a good read, I especially like this bit, given our recent discussions:

`Care must be taken to ensure that sort is called in the context of the C locale to avoid any surprises related to collation order.`

jeckersb (approved):

Makefile
cgwalters: Right; the doc mentions that GNU tar's `--sort=name` is always locale insensitive.

cgwalters:
Test was a generic timeout issue.

PR #593, opened 2024-06-10
tests-integration: Add basic local tmt flow

cgwalters:
Part of https://github.com/containers/bootc/issues/543

I'm not totally happy with this, but it does demonstrate the basic
wiring flow of:

- `cargo xtask test-tmt`

That will do a container build, make a disk image from it,
and run a "hello world" tmt test.

A lot more to do here including wiring up our existing tests
into this, and deduplicating with the other integration tests.

A key aspect too will be exploring workflows that e.g. expose
a registry locally.


cgwalters:
> cargo: error while loading shared libraries: libssh2.so.1: cannot open shared object file: No such file or directory

:arrow_right: https://bugzilla.redhat.com/show_bug.cgi?id=2291191

tests-integration/src/runvm.rs
```diff
@@ -0,0 +1,165 @@
...
+        serde_json::from_value(config.clone()).context("Parsing config")?;
+    let builder = config
+        .labels()
+        .as_ref()
+        .and_then(|l| l.get(BUILDER_ANNOTATION))
+        .ok_or_else(|| anyhow::anyhow!("Missing {BUILDER_ANNOTATION}"))?;
+    Ok(builder.to_owned())
+}
+
+#[context("Running bootc-image-builder")]
```
henrywang, line 60: Can we use `bootc install to-disk` to generate disk image?

cgwalters: Yeah, but I think down the line we'll migrate bib to be more of a SDK (cc https://gitlab.com/fedora/bootc/tracker/-/issues/2 ) and so we might as well start pulling a container here.

cgwalters:
Rebased :surfer:

henrywang:
We can land this PR first. I have another PR to run this on testing farm.

henrywang (approved):

PR #595, opened 2024-06-11
status: Use prepare_for_write instead of require_root

ckyrouac:
This prevents /sysroot from switching to being
persistently mounted as rw.

### DIFF
--- a/lib/src/status.rs
+++ b/lib/src/status.rs
@@ -305,7 +305,7 @@ pub(crate) async fn status(opts: super::cli::StatusOpts) -> Result<()> {
     let host = if !Utf8Path::new("/run/ostree-booted").try_exists()? {
         Default::default()
     } else {
-        crate::cli::require_root()?;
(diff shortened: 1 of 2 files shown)


cgwalters (approved):

cgwalters:
I'm a bit confused since the link for "
Integration Test / integration (x86_64, rhel-9-5) (pull_request_target) Failing after 49m " to testing farm takes me to something for Fedora 40?

cgwalters:
Ah, at least it looks like the integration tests are incorrectly *testing* for this behavior:

```
TASK [assert] ******************************************************************
fatal: [guest]: FAILED! => changed=false 
  assertion: result_sysroot_mount_status.stdout == "rw"
  evaluated_to: false
  msg: /sysroot is not mounted with rw permission
```

cgwalters (approved):

cgwalters:
Not sure why the install tests suddenly started falling over with ENOSPC...trying to debug

PR #596, opened 2024-06-11
hack: Add remote lldb utilities to hack dir

ckyrouac:
This is useful for remote debugging bootc running in a VM.

Until podman-bootc has support for forwarding arbitrary ports, this custom solution is needed.

This has been working well for me and I thought it might be useful to include in the docs for others to try if they're interested.

### DIFF
--- a/hack/lldb/Containerfile
+++ b/hack/lldb/Containerfile
@@ -1,0 +1,18 @@
+FROM quay.io/centos-bootc/centos-bootc-dev:stream9
+
+COPY ./etc/sysctl.conf /etc/sysctl.conf
+COPY ./etc/systemd/system/lldb-server.service /etc/systemd/system/lldb-server.service
+COPY ./etc/sudoers.d/wheel-nopasswd /etc/sudoers.d
+ARG sshpubkey
+
+RUN dnf -y install lldb firewalld vim && \
+    firewall-offline-cmd --add-port 1025-65535/tcp && \
+    systemctl enable lldb-server.service && \
+
+    # add test user
+    if test -z "$sshpubkey"; then echo "must provide sshpubkey"; exit 1; fi; \
+    useradd -G wheel test && \
+    mkdir -m 0700 -p /home/test/.ssh && \
+    echo $sshpubkey > /home/test/.ssh/authorized_keys && \
+    chmod 0600 /home/test/.ssh/authorized_keys && \
+    chown -R test: /home/test
--- a/hack/lldb/deploy.sh
+++ b/hack/lldb/deploy.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+# connect to the VM using https://libvirt.org/nss.html
+
+set -e
+
+# build the container image
+sudo podman build --build-arg "sshpubkey=$(cat ~/.ssh/id_rsa.pub)" -f Containerfile -t localhost/bootc-lldb .
+
+# build the disk image
+mkdir -p ~/.cache/bootc-dev/disks
+rm -f ~/.cache/bootc-dev/disks/lldb.raw
+truncate -s 10G ~/.cache/bootc-dev/disks/lldb.raw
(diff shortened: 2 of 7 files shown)


hack/lldb/deploy.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
cgwalters, line 1: Some overlap here with https://github.com/containers/bootc/pull/593 a bit and of course - a lot of overlap with https://github.com/containers/podman-bootc ...I was really thinking about trying to make that a more generic tool, but it gets tricky.

There's *so many* virtualization wrappers on Linux...

hack/lldb/etc/sysctl.conf
```diff
@@ -0,0 +1,2 @@
+net.ipv6.conf.all.disable_ipv6 = 1
```
cgwalters, line 1: Why out of curiosity?

hack/lldb/deploy.sh
```diff
@@ -0,0 +1,21 @@
...
+rm -f ~/.cache/bootc-dev/disks/lldb.raw
+truncate -s 10G ~/.cache/bootc-dev/disks/lldb.raw
+sudo podman run --pid=host --network=host --privileged --security-opt label=type:unconfined_t -v /var/lib/containers:/var/lib/containers -v ~/.cache/bootc-dev/disks:/output -v /dev:/dev localhost/bootc-lldb bootc install to-disk --via-loopback --generic-image --skip-fetch-check /output/lldb.raw
+
+# create a new VM in libvirt
+set +e
+virsh -c qemu:///system destroy bootc-lldb
+virsh -c qemu:///system undefine --nvram bootc-lldb
+set -e
+sudo virt-install --name bootc-lldb --cpu host --vcpus 8 --memory 8192 --import --disk ~/.cache/bootc-dev/disks/lldb.raw --os-variant rhel9-unknown
```
cgwalters, line 21: For this one we could use `--qemu-commandline` to pass the ssh key via credentials too right?  This is probably something to do via a patch to virt-install.

hack/lldb/Containerfile
```diff
@@ -0,0 +1,18 @@
+FROM quay.io/centos-bootc/centos-bootc-dev:stream9
+
+COPY ./etc/sysctl.conf /etc/sysctl.conf
+COPY ./etc/systemd/system/lldb-server.service /etc/systemd/system/lldb-server.service
+COPY ./etc/sudoers.d/wheel-nopasswd /etc/sudoers.d
+ARG sshpubkey
+
+RUN dnf -y install lldb firewalld vim && \
+    firewall-offline-cmd --add-port 1025-65535/tcp && \
```
cgwalters, line 9: Hmm, interesting that you're adding firewalld here only to poke a lot of holes in the firewall...any reason for that?

(In the near future we may want to actually add firewalld to our base images though)

cgwalters (commented):
Nice!  I have never tried using lldb...I'm usually a "printf debugging" person.

hack/lldb/deploy.sh
ckyrouac: agreed, eventually podman-bootc should provide everything needed for this, but in the meantime I just wanted to use lldb. The primary (only?) thing preventing using podman-bootc is the networking stuff.

hack/lldb/etc/sysctl.conf
ckyrouac: I had a weird issue where the lldb connection was trying to use the ipv6 address and failing, this fixed it. I have ipv6 turned off on my host.

hack/lldb/Containerfile
ckyrouac: firewalld is just really simple to use. I was surprised it's not in the base image. lldb uses the user defined port for the initial connection along with seemingly any random port in the 30000-* range.

hack/lldb/deploy.sh
ckyrouac: yep, I was playing around with the now defunct [included-ssh-pubkey](https://gitlab.consulting.redhat.com/redhat-cop/rhel-bootc-examples/-/commit/9f81d440d530245e2ba8d8a36a788f56e1be20dc) example and got it working so I just continued using it. Either way works for me.

cgwalters (approved):
We can always do fixes here as followups, so fine to me to merge as is.

PR #598, opened 2024-06-12
build: Install git too

cgwalters:
This must be a regression from the timestamp change: 91ed63caf196dd728fb7c7da10bdce40e2008504

Without this we fail to parse the timestamp and get errors; maybe something else changed.

Of course, we should use a non-Makefile language for this so we get proper error checking.  I may move some of the makefile bits into xtask.rs or so.


jeckersb (commented):
Do we need stuff from `git` or is `git-core` enough?  I usually mindlessly choose the latter since it pulls in less stuff.

cgwalters:
Good call, changed to use git-core

jeckersb (approved):

PR #600, opened 2024-06-13
upgrade: Add error message when staged deployment is incompatible

ckyrouac:
When running `rpm-ostree install <package>` without a reboot, `bootc upgrade` will fail. Prior to this, the error message was unclear of the cause.

fixes RHEL-36003

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -384,17 +384,26 @@ async fn upgrade(opts: UpgradeOpts) -> Result<()> {
         crate::status::get_status_require_booted(sysroot)?;
     let imgref = host.spec.image.as_ref();
     // If there's no specified image, let's be nice and check if the booted system is using rpm-ostree
-    if imgref.is_none()
-        && host
+    if imgref.is_none() {
+        let booted_incompatible = host
             .status
             .booted
             .as_ref()
-            .map_or(false, |b| b.incompatible)
-    {
-        return Err(anyhow::anyhow!(
-            "Booted deployment contains local rpm-ostree modifications; cannot upgrade via bootc. You can run `rpm-ostree reset` to undo the modifications."
-        ));
+            .map_or(false, |b| b.incompatible);
+
+        let staged_incompatible = host
+            .status
+            .staged
+            .as_ref()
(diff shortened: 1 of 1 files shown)


lib/src/cli.rs
```diff
@@ -384,17 +384,31 @@ async fn upgrade(opts: UpgradeOpts) -> Result<()> {
...
-        return Err(anyhow::anyhow!(
-            "Booted deployment contains local rpm-ostree modifications; cannot upgrade via bootc. You can run `rpm-ostree reset` to undo the modifications."
-        ));
+        {
+            return Err(anyhow::anyhow!(
+                "Booted deployment contains local rpm-ostree modifications; cannot upgrade via bootc. You can run `rpm-ostree reset` to undo the modifications."
+            ));
+        }
+
+        if host.status.staged.is_some()
```
cgwalters, line 399: We should be able to drop this part; I know Rust combinators can be quite obscure but it's basically what the `map_or(false)` is doing here...if the option is None, then the result will be `false`.

lib/src/cli.rs
```diff
@@ -384,17 +384,31 @@ async fn upgrade(opts: UpgradeOpts) -> Result<()> {
...
+            && host
+                .status
+                .staged
+                .as_ref()
+                .map_or(false, |b| b.incompatible)
```
cgwalters, line 404: But for a bigger cleanup, I think we can merge the logic for checking something like this:

```
let booted_incompatible = host
            .status
            .booted
            .as_ref()
            .map_or(false, |b| b.incompatible);
let staged_incompatible = host
                .status
                .staged
                .as_ref()
                .map_or(false, |b| b.incompatible);
if booted_incompatible || staged_incompatible {
   return Err...
}
```
or so?

lib/src/cli.rs
ckyrouac: :+1: thanks for the tip

lib/src/cli.rs
ckyrouac: sounds good, updated

cgwalters (approved):
Thanks!

PR #602, opened 2024-06-13
Human-readable bootc status

lukewarmtemp:
Fixes https://github.com/containers/bootc/issues/408

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -116,6 +116,11 @@ pub(crate) struct StatusOpts {
     /// Only display status for the booted deployment.
     #[clap(long)]
     pub(crate) booted: bool,
+    
+    /// Condense status into most important info, to reduce eye fatigue 
+    #[clap(long)]
+    pub(crate) pretty: bool,
+    
 }
 
 #[cfg(feature = "install")]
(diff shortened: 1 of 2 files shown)


cgwalters:
Thanks for working on this! I need to pull your PR and try it but just at a quick look, it's just a shorter YAML serialization right?

I had been thinking we'd output something more like `rpm-ostree status` that is explicitly not parsable by machines, just intended solely for human visual consumption. Other prior art is e.g. `systemctl status` which is partially structured but also uses ASCII art drawing characters, etc.

Before we try to write too much more code I think it'd be good to have a bike shed (maybe in the original issue) with some proposed example outputs that different people would like to see from this. Does that makes sense?

lib/src/cli.rs
```diff
@@ -116,6 +116,11 @@ pub(crate) struct StatusOpts {
     /// Only display status for the booted deployment.
     #[clap(long)]
     pub(crate) booted: bool,
+    
+    /// Condense status into most important info, to reduce eye fatigue
```
cgwalters, line 120: I think we can probably just go ahead and default to human-readable if we're writing to a tty by default.

That avoids the need to call it anything at all, and avoids "typing fatigue" from a potentially frequently used CLI option.

Arguably then, we could also simultaneously switch to JSON by default if we're not on a tty.

But, we probably do need a general option here like:

```
enum OutputFormat {
  HumanReadable,
  YAML,
  JSON
}
```
or so and 

```
pub(crate) format Option<OutputFormat>
```

or so.

cgwalters: https://github.com/containers/bootc/pull/639

jmarrero:
closing in favor of #738

PR #604, opened 2024-06-13
install: Move warning after state and possible re-exec

cgwalters:
I'm not sure if this fell out of a refactoring or not, but right now we end up emitting the warning here
twice in the case that we re-exec the current process in order to get the right SELinux label.

It's unfortunately a bit hard to truly minimize the set of things we do before `prepare_install`, so we'll just have to be mindful.


jmarrero (approved):
lgtm

PR #605, opened 2024-06-16
docs: Even more about how bootc is not a container at runtime

cgwalters:
This is definitely the `#1` confusion many people have to start;
it's very understandable!  Because in theory we *could* have
a runtime and a larger role, but it was an explicit choice
not to do that.

### DIFF
--- a/docs/src/building/bootc-runtime.md
+++ b/docs/src/building/bootc-runtime.md
@@ -5,11 +5,16 @@ Fundamentally, `bootc` reuses the [OCI image format](https://github.com/opencont
 as a way to transport serialized filesystem trees with included metadata such as a `version`
 label, etc.
 
-However, `bootc` generally ignores the [Container configuration](https://github.com/opencontainers/image-spec/blob/main/config.md)
-section at runtime today.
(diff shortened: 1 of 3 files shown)


docs/src/building/bootc-runtime.md
```diff
@@ -5,11 +5,16 @@ Fundamentally, `bootc` reuses the [OCI image format](https://github.com/opencont
 as a way to transport serialized filesystem trees with included metadata such as a `version`
 label, etc.
 
-However, `bootc` generally ignores the [Container configuration](https://github.com/opencontainers/image-spec/blob/main/config.md)
-section at runtime today.
+A bootc container operates in two basic modes.  First, when invoked by a container run time such as `podman` or `docker` (typically as part of a build process), the bootc container behaves exactly the same as any other container. For exmaple, although there is a kernel embedded in the container image, it is not executed - the host kernel is used.  There's no additional mount namespaces, etc.  Ultimately, the container runtime is in full control here.
```
jmarrero, line 8: exmaple -> example

jmarrero (changes requested):
small typo

docs/src/building/bootc-runtime.md
cgwalters: Thanks for catching that; funnily enough I did a search to find it and it turns out there was another exmaple of the same one.

jeckersb (approved):

jmarrero (approved):

PR #606, opened 2024-06-16
install: Fix install config kargs + to-filesystem

cgwalters:
I think this got broken in a refactoring; add
test coverage.  In general all the heavy
lifting should move out of `baseline.rs`; a
good way to do that is probably to take the
next step of making it its own crate that
doesn't depend on the bootc core logic perhaps.

Closes: https://github.com/containers/bootc/issues/570


jmarrero (approved):
lgtm

PR #607, opened 2024-06-16
install/baseline: Drop separate /dev mount

cgwalters:
Now that we hard require the host's `/dev`, let's
drop our duplicate `devtmpfs` copy. My main motivation here is I wanted to add some other debugging and trying to invoke e.g. `lsblk` on our temporary copy doesn't work.

With the combination of the host `/dev` plus our invocation of udev settling, we shouldn't hit races.


jmarrero (approved):
lgtm

PR #612, opened 2024-06-18
tests-integration: Optimize rebuilds

cgwalters:
- Pass `--timestamp` to podman build to squash timestamps in order to gain reproducibilty, working around https://github.com/containers/buildah/issues/5592
- Fix the xattr reading code to correctly skip trailing nils


jmarrero (approved):
/lgtm

Issue #611, opened 2024-06-18
Following `HACKING.md` instructions leads to unable to pull `localhost/bootc` error


omertuc:
Following [these HACKING.md instructions](https://github.com/containers/bootc/blob/7ec44f6a0b2a4482c2551ce000ff2a3c1094978a/HACKING.md?plain=1#L44-L52) leads me to run into this error:

```
Trying to pull localhost/bootc:latest...
Error: unable to install bootc image: failed to pull image: initializing source docker://localhost/bootc:latest: pinging container registry localhost: Get "https://localhost/v2/": dial tcp [::1]:443: connect: connection refused
```

Is there a mistake in the instructions? Or is my system somehow set up wrong?

Is podman-bootc pulling an image in [this](https://github.com/containers/podman-bootc/blob/82fa2d72710dca7a6e7f9b0e46963fea09cb9e7a/pkg/bootc/bootc_disk.go#L281-L282) manner even able to reach for the container images stored in podman's internal image storage following a `podman build`?

cgwalters:
Are you using Linux? You need to be using podman-machine for that flow currently.

omertuc:
I am using podman-machine as instructed

ckyrouac:
the image needs to exist in the podman-machine's root storage. The instructions seem to assume that `podman` is setup to always connect to podman-machine. You should be able to build it on the podman-machine via `podman -c podman-machine-default-root build -t localhost/bootc -f hack/Containerfile .` The image should be listed via `podman -c podman-machine-default-root images`.

cgwalters:
I do this via `CONTAINER_CONNECTION` https://github.com/cgwalters/homegit/commit/70c08dde56d18f4d7afc4cc8506ff447c289a84c

jeckersb:
> I do this via `CONTAINER_CONNECTION` [cgwalters/homegit@70c08dd](https://github.com/cgwalters/homegit/commit/70c08dde56d18f4d7afc4cc8506ff447c289a84c)

Beware that podman currently will [crash](https://github.com/containers/podman/pull/22997) if you have `CONTAINER_CONNECTION` set but the connection is invalid, for instance if the podman machine instance gets deleted.

omertuc:
>the image needs to exist in the podman-machine's root storage.

Ah I see now, never used podman machine so didn't realize it has its own container image storage and that using the machine needs to be specified with `-c` / `CONTAINER_CONNECTION`. That makes a lot of sense.

[...]

PR #614, opened 2024-06-19
HACKING.md: Clarify `CONTAINER_CONNECTION`

omertuc:
This commit adds a note to `HACKING.md` to clarify that the environment
variable `CONTAINER_CONNECTION` should be set to the name of the podman
machine.

As an inexperienced user of podman-machine, this detail was not obvious
to me (#611).


omertuc:
Forced pushed for DCO

cgwalters:
Minor procedural note; I try to use "topic prefix" style in git commits in my projects (like the Linux kernel and others) which would look more like:

> HACKING.md: Clarify `CONTAINER_CONNECTION`

Issue #496, opened 2024-04-26
ci: Change testing farm runs to not occupy a runner


cgwalters:
The github.com/containers organization is pretty large and active, but only has the default 20 Github-hosted runners available right now.

We have this repo hooked up to Testing Farm.  The problem here is that the 4 distinct TF runs we do each occupy a whole Github-hosted action runner virtual machine to basically poll a remote HTTP server, which is quite wasteful.

It looks to me like the TF action itself supports being confused to report status back to the PR, without holding a runner?  There's an `update_pull_request_status` flag...

Hmm, are we doing things this way because *we're* cloning the git repository here because we have the tests?

cc @henrywang

henrywang:
Yes, https://github.com/virt-s1/bootc-workflow-test is using `update_pull_request_status`. But that still needs a github action runner. For example, https://github.com/virt-s1/bootc-workflow-test/actions/runs/8845171999/job/24288556567.
Or we can use self hosted runner (container). The https://github.com/virt-s1/bootc-workflow-report repo already uses self hosted github action runner (container). For example, https://github.com/virt-s1/bootc-workflow-report/blob/c723ccca5482495d0860b6f187d971d0058d7560/.github/workflows/trigger-rhel-9-4.yml#L13.

cgwalters:
> Or we can use self hosted runner (container). 

Oh yes that makes *lots* of sense.  Hmm.  Actually...yeah, we should wire up something like this officially to the whole github.com/containers organization.  Just thinking through how it works, there's also https://github.com/redhat-actions/openshift-actions-runners and https://github.com/actions/actions-runner-controller which looks even better (I think, although it's not clear offhand if it supports using e.g. `container:` to configure the pod image as distinct from the runner hoster, if it doesn't that weakens things a lot).

henrywang:
Yeah, this's a really good solution for self hosted runner. Three reasons I didn't use this solution:

1. This solution needs a public k8s, I can't find a free one.
2. RHEL for Edge needs VM or bare metal server, not container. (RHEL for Edge does not use testing farm yet)
3. [openshift is not supported](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/quickstart-for-actions-runner-controller#prerequisites) 

I can spend some time on this solution and try something. That should be interesting.

thrix:
[...]

PR #615, opened 2024-06-19
test: add packit to run tmt integration test

henrywang:
This PR is a continuous setting of https://github.com/containers/bootc/pull/613. It adds [Packit](https://packit.dev/) to run tmt integration test.

`testing-farm:fedora-40-x86_64:tmt-integration-test` is from [Packit](https://packit.dev/).

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,0 +1,9 @@
+---
+jobs:
+  - job: tests
+    trigger: pull_request
+    targets:
+      - fedora-40-x86_64
+    tmt_plan: /build
+    skip_build: true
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
If I understand correctly, the packit CI here also doesn't hold up a GHA runner (cc https://github.com/containers/bootc/issues/496 ) right?

henrywang:
> If I understand correctly, the packit CI here also doesn't hold up a GHA runner (cc #496 ) right?

Right. Packit does not use github action runner.

PR #616, opened 2024-06-20
test: move tmt test log to tmt plan and enable aarch64 test

henrywang:
This PR includes:

1. move tmt test log to tmt plan data folder, the tmt integration test log can be archived in testing farm artifacts.
2. give tmt vm 20G disk instead of default 40G. Sometimes the deployed testing farm runner does not have big disk,
to avoid this issue, use smaller tmt vm disk instead
3. add aarch64 tmt integration test


henrywang:
Okey. The Packit aarch64 testing farm runner should come from AWS ec2 ARM bare metal server. The deployment for aarch64 runner is faster than deploying x86_64 runner. So we can use both x86_64 and aarch64 test.

plans/build.fmf
```diff
@@ -19,3 +19,7 @@ discover:
   how: fmf
   test:
     - /build-tmt/build-image
+finish:
+  how: shell
+  script:
+    - cp -r /var/tmp/tmt/run-001 $TMT_PLAN_DATA
```
cgwalters, line 25: For future readers can you add a comment here like `# move tmt test log to tmt plan data folder, the tmt integration test log can be archived in testing farm artifacts.` ?

cgwalters (approved):

plans/build.fmf
cgwalters: (not worth a whole new CI run for a comment of course, so merging now, but as part of further work)

PR #617, opened 2024-06-20
bootc switch `--stateroot` flag

omertuc:
Trying this patch by @cgwalters out, creating this draft PR for discussion on
what I've observed

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -88,6 +88,9 @@ pub(crate) struct SwitchOpts {
 
     /// Target image to use for the next boot.
     pub(crate) target: String,
+
+    #[clap(long)]
+    pub(crate) stateroot: Option<String>,
 }
 
 /// Options controlling rollback
@@ -573,7 +576,9 @@ async fn switch(opts: SwitchOpts) -> Result<()> {
         }
     }
 
-    let stateroot = booted_deployment.osname();
+    let osname = booted_deployment.osname();
+    let stateroot = opts.stateroot.as_deref().unwrap_or_else(|| osname.as_str());
+
     let mut opts = ostree::SysrootDeployTreeOpts::default();
     let kargs: Vec<&str> = kargs.iter().map(|s| s.as_str()).collect();
     opts.override_kernel_argv = Some(kargs.as_slice());
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -159,6 +159,9 @@ pub(crate) struct InstallConfigOpts {
     #[clap(long)]
     #[serde(default)]
     pub(crate) generic_image: bool,
+
+    #[clap(long)]
+    pub(crate) stateroot: Option<String>,
 }
 
 #[derive(Debug, Clone, clap::Parser, Serialize, Deserialize, PartialEq, Eq)]
@@ -546,8 +549,11 @@ async fn initialize_ostree_root_from_self(
     // Another implementation: https://github.com/coreos/coreos-assembler/blob/3cd3307904593b3a131b81567b13a4d0b6fe7c90/src/create_disk.sh#L295
     crate::lsm::ensure_dir_labeled(rootfs_dir, "", Some("/".into()), 0o755.into(), sepolicy)?;
 
-    // TODO: make configurable?
-    let stateroot = STATEROOT_DEFAULT;
+    let stateroot = state
+        .config_opts
+        .stateroot
+        .as_deref()
+        .unwrap_or(STATEROOT_DEFAULT);
     Task::new_and_run(
         "Initializing ostree layout",
         "ostree",


omertuc:
After shelling into the podman-bootc VM and running:

```
ostree admin stateroot-init foo
bootc switch --stateroot foo quay.io/fedora/fedora-bootc:40
```

I've tried to then reboot into the new stateroot, but the only grub option is the original CentOS stateroot.

I've also observed the following weird behavior:

Ran the above commands and then:

```
[root@localhost ~]# ostree admin status
  foo 51d7dcab77ab6a4e7235bbdbbe32f70f6f97b6a679f7a359bfea346ae2f18b0e.0 (staged)
    origin: <unknown origin type>
* default f96c05656ae82d9758be08bcf888d8f130fafaf9723967255bd76b51f26bd2e4.0
    origin: <unknown origin type>
[root@localhost ~]# ostree admin set-default 0
  foo 51d7dcab77ab6a4e7235bbdbbe32f70f6f97b6a679f7a359bfea346ae2f18b0e.0 (staged)
    origin: <unknown origin type>
* default f96c05656ae82d9758be08bcf888d8f130fafaf9723967255bd76b51f26bd2e4.0
    origin: <unknown origin type>
[root@localhost ~]# ostree admin set-default 1
[root@localhost ~]# ostree admin status
* default f96c05656ae82d9758be08bcf888d8f130fafaf9723967255bd76b51f26bd2e4.0
    origin: <unknown origin type>
```

The stateroot just disappears? That's a bit odd

cgwalters:
Let's split the install vs switch changes into separate PRs; the install one I think we can trivially land with just a simple integration test.

omertuc:
Split install CLI into https://github.com/containers/bootc/pull/622

omertuc:
Should we provide users with an option to skip the `--var` copy? Or maybe skipping should be the default?

cgwalters:
This one heavily relates to https://github.com/containers/bootc/issues/404

lib/src/cli.rs
```diff
@@ -687,18 +693,52 @@ async fn switch(opts: SwitchOpts) -> Result<()> {
...
     let new_spec = RequiredHostSpec::from_spec(&new_spec)?;
 
+    if old_stateroot != stateroot {
+        let init_result = sysroot.init_osname(&stateroot, cancellable);
+        match init_result {
+            Ok(_) => {
+                Task::new("Copying /var to new stateroot", "cp")
+                    .args([
+                        "--recursive",
+                        "--reflink=auto",
```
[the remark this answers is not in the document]
cgwalters: One subtle thing about this implementation is that as it is today because we're operating on the stateroot view of `/var` I don't think we'll see any mounts that happen externally today. For example if someone has `/var/lib/postgres` on a separate partition - which is a totally sane thing to do.

Speaking of, we almost certainly want to pass `--one-file-system` here just in case we somehow encounter a mount.

(Also this relates a bit to https://github.com/ostreedev/ostree/pull/3292 )

cgwalters:
I think my overall inclination here is to actually make this part of `bootc install`, and roll up fixing things related to https://github.com/containers/bootc/pull/137 at the same time.

Hence because it's `install` I think I'd argue we drop all state by default...*except* let's add a check for "you have `root` in current kargs but missing in new kargs". That would avoid probably the biggest footgun and help people understand that often they're going to need to allow-list in at least some state from the current root.

Something like `bootc install --stateroot=newroot --inherit-karg=root` or something?

Or I guess the best place to bikeshed this would be  https://github.com/containers/bootc/issues/404

lib/src/cli.rs
```diff
@@ -687,18 +693,52 @@ async fn switch(opts: SwitchOpts) -> Result<()> {
...
     let new_spec = {
         let mut new_spec = host.spec.clone();
         new_spec.image = Some(target.clone());
         new_spec
     };
 
-    if new_spec == host.spec {
-        println!("Image specification is unchanged.");
+    if new_spec == host.spec && old_stateroot == stateroot {
+        // TODO: Should we really be confusing users with terms like "stateroot"?
```
cgwalters, line 713: Yeah, it's a good question. My thinking here is it's going to be hard to hide...and we should lean into exposing it as a more first class verb.

lib/src/cli.rs
```diff
@@ -687,18 +693,52 @@ async fn switch(opts: SwitchOpts) -> Result<()> {
...
+        println!(
+            "The currently running deployment in stateroot {stateroot} is already using this image"
+        );
         return Ok(());
     }
     let new_spec = RequiredHostSpec::from_spec(&new_spec)?;
 
+    if old_stateroot != stateroot {
```
cgwalters, line 721: Related to the above I'm uncertain about doing this by default. It may make sense to have a `bootc stateroot init --clone` or so?

Definitely in some cases we explicitly *want* to ignore previous state.

lib/src/cli.rs
```diff
@@ -99,6 +100,11 @@ pub(crate) struct SwitchOpts {
 
     /// Target image to use for the next boot.
     pub(crate) target: String,
+
+    /// Make the switch into a custom stateroot. If the stateroot doesn't exist, it will be created
+    /// and `/var` of the current stateroot will be copied (copy-on-write) into it.
+    #[clap(long)]
+    pub(crate) stateroot: Option<String>,
```
cgwalters, line 107: In the interest of forward progress, how about adding `#[clap(hide = true))]` and calling this `experimental_clone_into_stateroot` so it can be an experimental feature for now while we decide on how we stabilize this?

cgwalters:
Thanks for starting this! My desktop was having some issues and I wanted to do a factory reset, so I had some motivation to push this one farther. I ended up doing it as part of the `install` flow which I think feels cleaner.

New PR in https://github.com/bootc-dev/bootc/pull/1389

PR #619, opened 2024-06-20
cli: Make sysroot lock automatically check root + setup mountns

cgwalters:
Previous to this change, we were always pairing together two functions:

- `prepare_for_write()`
- `get_locked_sysroot()`

Make the latter automatically call the former (and make it private to the cli code).

This will avoid bugs like https://github.com/containers/bootc/pull/595/commits/dcfc7521acaf21072032777b9683796ff710bdfe where we forgot to call the first one and didn't end up with an unshared mountns.

While we're here, drop an unnecessary `async` from these functions, and just in case make `prepare_for_write()` idempotent.

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -324,7 +324,7 @@ pub(crate) enum Opt {
 /// TODO use https://github.com/ostreedev/ostree/pull/2779 once
 /// we can depend on a new enough ostree
 #[context("Ensuring mountns")]
-pub(crate) async fn ensure_self_unshared_mount_namespace() -> Result<()> {
+pub(crate) fn ensure_self_unshared_mount_namespace() -> Result<()> {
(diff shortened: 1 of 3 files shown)


ckyrouac (approved):

cgwalters:
Can anyone spot what the actual error is in https://artifacts.osci.redhat.com/testing-farm/25531a16-1e9e-4781-b0e6-09970b9bd762/#artifacts-/plans/install-upgrade/to-disk ?

jeckersb:
> Can anyone spot what the actual error is in https://artifacts.osci.redhat.com/testing-farm/25531a16-1e9e-4781-b0e6-09970b9bd762/#artifacts-/plans/install-upgrade/to-disk ?

Unfortunately looks like it's been garbage collected by the time I got to it?  `testout.log` is now just a 404 :cry:

PR #620, opened 2024-06-20
cli: Add a new bootc image subcommand

cgwalters:
We have a basic `bootc image list` but more interesting is
`bootc image push` which defaults to copying the booted
image into the container storage.

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -199,24 +199,12 @@ pub(crate) enum ImageOpts {
     ///
     /// Note that these are distinct from images stored via e.g. `podman`.
     List,
-    /// Copy a container image from the bootc storage to a target.
-    ///
-    /// ## Copying the booted container image to containers-storage: (podman)
+    /// Copy a container image from the bootc storage to `containers-storage:`.
     ///
     /// The source and target are both optional; if both are left unspecified,
-    /// via a simple invocation of `bootc image push`, then the default is to
+    /// via a simple invocation of `bootc image copy-to-storage`, then the default is to
     /// push the currently booted image to `containers-storage` (as used by podman, etc.)
     /// and tagged with the image name `localhost/bootc`,
-    ///
-    /// ## Copying the booted container image to a remote registry
-    ///
-    /// Aside from the special case above, default transport is `registry`. This
-    /// means that an invocation of
-    ///
-    /// `bootc image push --target quay.io/example/someimage:latest` will push the
-    /// booted container image to the target registry. This will be done via an
-    /// invocation equivalent to `skopeo copy`, and hence the defaults for that apply.
-    /// For example, the default registry authentication file applies.
     ///
     /// ## Copying a non-default container image
     ///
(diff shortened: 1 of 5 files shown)


cgwalters:
```
$ bootc image list
docker://<your image here>
$ bootc image push
...
$ podman run --rm -ti localhost/bootc bash
# Now we're in the booted host as a container
```

This is aimed to close the biggest gap in https://containers.github.io/bootc/booting-local-builds.html - because then we can easily add something like `bootc switch --local` which would default to pulling from `localhost/bootc` in the container storage, closing the loop.

This builds on https://github.com/ostreedev/ostree-rs-ext/pull/642 which has notes about its inefficiencies.

cgwalters:
OK yes, this PR also hit the install flake (I added it to https://github.com/containers/bootc/issues/579 )

```
Creating root filesystem (xfs) on device /dev/loop0p1 (size=1M)
> mkfs.xfs -m uuid=790e772b-c617-47dc-af23-04ca799c8aad -L root /dev/loop0p1
size 256 of data subvolume is too small, minimum 4096 blocks
```

With the new size printing we can clearly see that the partition size is just 1M. I'm going to add some more logging to see if I can figure out more here.

cgwalters:
OK this one now has a test based on the framework added in https://github.com/containers/bootc/pull/625

lib/src/cli.rs
```diff
@@ -174,6 +175,31 @@ pub(crate) enum ContainerOpts {
     Lint,
 }
 
+/// Subcommands which operate on images.
+#[derive(Debug, clap::Subcommand, PartialEq, Eq)]
+pub(crate) enum ImageOpts {
+    List,
+    /// Perform relatively inexpensive static analysis checks as part of a container
```
vrothberg, line 182: Copy paste description?

vrothberg (commented):
I´d love to see docs (no blocker) and like to bikeshed a bit on the naming of `push`.  It feels more like a `write-to-storage` or something down these lines.

lib/src/cli.rs
cgwalters: Yes thanks, fixed!

cgwalters:
> I´d love to see docs (no blocker) 

I've updated that, though note that right now this is hidden from the CLI as an experimental feature so we can use it in our own CI, and stabilize things later.

> and like to bikeshed a bit on the naming of push. It feels more like a write-to-storage or something down these lines.

The default is indeed like `copy-to-storage` or so, but the functionality is more general than that in that we support e.g. `bootc image push --target quay.io/exampleos/someimage:latest`.

That said...it would actually simplify things perhaps if we *only* supported `bootc image copy-to-storage <tag>` because one can then of course directly use `podman` for them.

There's also the longer term bikeshed topic of (assuming we made it actually *efficient* and not a physical copy) if the default should actually be to keep the bootc image visible in containers-storage or so.

vrothberg:
> That said...it would actually simplify things perhaps if we _only_ supported `bootc image copy-to-storage <tag>` because one can then of course directly use `podman` for them.

Agreed, that sounds like a good idea :+1:

cgwalters:
OK it's now called `bootc image copy-to-storage` and I dropped the `--transport` option, so it only supports `containers-storage:`.

vrothberg (approved):
LGTM thank you!

cgwalters:
I dropped the conflicting logging change to kargs from here for now, will split it out separately.

PR #621, opened 2024-06-20
install: Add more information on target blockdev

cgwalters:
It's useful to see this; we're already somewhat verbose at install time.


jmarrero (approved):
lgtm

PR #622, opened 2024-06-21
install: Make stateroot configurable

omertuc:
This commit makes it so that the `bootc install` stateroot will be configurable (it defaults to `default`). For now this is a hidden CLI option until we decide whether we want to commit to this API.

In the future we also want to make the stateroot of `bootc switch` be configurable (https://github.com/containers/bootc/pull/617) so that users can install an image to a new stateroot while they already have an existing stateroot

Also removed the constant `STATEROOT_DEFAULT`, we're now simply taking it from the `ostree_ext` crate

### DIFF
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -51,8 +51,6 @@ use crate::store::Storage;
 use crate::task::Task;
 use crate::utils::sigpolicy_from_opts;
 
-/// The default "stateroot" or "osname"; see https://github.com/ostreedev/ostree/issues/2794
-const STATEROOT_DEFAULT: &str = "default";
 /// The toplevel boot directory
 const BOOT: &str = "boot";
 /// Directory for transient runtime state
@@ -171,6 +169,10 @@ pub(crate) struct InstallConfigOpts {
     #[clap(long, hide = true)]
     #[serde(default)]
     pub(crate) skip_bound_images: bool,
+
+    /// The stateroot name to use. Defaults to `default`.
+    #[clap(long)]
+    pub(crate) stateroot: Option<String>,
 }
 
 #[derive(Debug, Clone, clap::Parser, Serialize, Deserialize, PartialEq, Eq)]
@@ -567,8 +569,12 @@ async fn initialize_ostree_root(state: &State, root_setup: &RootSetup) -> Result
     // Another implementation: https://github.com/coreos/coreos-assembler/blob/3cd3307904593b3a131b81567b13a4d0b6fe7c90/src/create_disk.sh#L295
     crate::lsm::ensure_dir_labeled(rootfs_dir, "", Some("/".into()), 0o755.into(), sepolicy)?;
 
-    // TODO: make configurable?
-    let stateroot = STATEROOT_DEFAULT;
+    let stateroot = state
+        .config_opts
+        .stateroot
+        .as_deref()
+        .unwrap_or(ostree_ext::container::deploy::STATEROOT_DEFAULT);
+
     Task::new_and_run(
         "Initializing ostree layout",
         "ostree",
@@ -638,7 +644,11 @@ async fn install_container(
 ) -> Result<(ostree::Deployment, InstallAleph)> {
     let sepolicy = state.load_policy()?;
     let sepolicy = sepolicy.as_ref();
-    let stateroot = STATEROOT_DEFAULT;
+    let stateroot = state
+        .config_opts
+        .stateroot
+        .as_deref()
+        .unwrap_or(ostree_ext::container::deploy::STATEROOT_DEFAULT);
 
     let container_rootfs = &Dir::open_ambient_dir("/", cap_std::ambient_authority())?;
 
@@ -1099,7 +1109,9 @@ pub(crate) fn setup_sys_mount(fstype: &str, fspath: &str) -> Result<()> {
     Task::new(format!("Mounting {fstype} {fspath}"), "mount")
         .args(["-t", fstype, fstype, fspath])
         .quiet()
-        .run()
+        .run()?;
+
+    Ok(())
 }
 
 /// Verify that we can load the manifest of the target image
--- a/tests-integration/src/install.rs
+++ b/tests-integration/src/install.rs
@@ -25,6 +25,9 @@ pub(crate) const BASE_ARGS: &[&str] = &[
     "label=disable",
 ];
 
+// Arbitrary
+const NON_DEFAULT_STATEROOT: &str = "foo";
+
 /// Clear out and delete any ostree roots, leverage bootc hidden wipe-ostree command to get rid of
 /// otherwise hard to delete deployment files
 fn reset_root(sh: &Shell, image: &str) -> Result<()> {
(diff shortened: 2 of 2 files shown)


lib/src/install.rs
```diff
@@ -546,8 +548,11 @@ async fn initialize_ostree_root_from_self(
     // Another implementation: https://github.com/coreos/coreos-assembler/blob/3cd3307904593b3a131b81567b13a4d0b6fe7c90/src/create_disk.sh#L295
     crate::lsm::ensure_dir_labeled(rootfs_dir, "", Some("/".into()), 0o755.into(), sepolicy)?;
 
-    // TODO: make configurable?
-    let stateroot = STATEROOT_DEFAULT;
+    let stateroot = state
+        .config_opts
+        .stateroot
+        .as_deref()
+        .unwrap_or(STATEROOT_DEFAULT);
```
cgwalters, line 555: This is just `let stateroot = state.config_opts.stateroot.as_str()` now right?

cgwalters:
For tests, try adding a new case to https://github.com/containers/bootc/blob/b5f47ff8fe5121f7eade46aa53048edd75fd6116/tests-integration/src/install.rs#L68

lib/src/install.rs
omertuc: Yeah of course

lib/src/install.rs
```diff
@@ -159,6 +157,18 @@ pub(crate) struct InstallConfigOpts {
...
+
+    /// The stateroot name to use. Defaults to `default`.
+    #[clap(long, default_value = ostree_ext::container::deploy::STATEROOT_DEFAULT, hide = true)]
+    #[serde(default = "default_stateroot")]
+    pub(crate) stateroot: String,
+}
+
+// TODO: Is this necessary? Added so that the install_opts_serializable test passes, because that
+// test is doing deserialization, but will users ever really do deserialization themselves, and if
+// yes, should we help them by providing a default value?
```
cgwalters, line 169: Users won't deserialize, but the idea is we may do so internally as part of things like https://github.com/containers/bootc/pull/78 - that's why the serialization was added.

Basically for various reasons we end up wanting to re-exec the current process, and it seems more elegant to do so via `vec<string> -> clap -> serde -> json` instead of `vec<string> -> clap -> vec<string>` (or trickier, trying to skip back over what clap did and find the original `vec<string>`).

It may be simpler on the implementation to keep it as an `Option<String>`, then serde and clap will handle it the same way, and we do the `unwrap_or_else()` in the install code.

omertuc: >Basically for various reasons we end up wanting to re-exec the current process, and it seems more elegant to do so via vec<string> -> clap -> serde -> json instead of vec<string> -> clap -> vec<string> (or trickier, trying to skip back over what clap did and find the original vec<string>).

In that case, iiuc in the re-exec scenario we still don't need to provide a serde default for deserialization purposes because no user will hand-craft the serialized form for us to deserialize. We'll always be serializing clap-initialized structs, so only clap really needs to have a default. The clap default ensures all structs serialized and deserialized will already have this value initialized

But I've switched to Option<String> regardless to keep it simple

lib/src/install.rs
```diff
@@ -158,6 +156,10 @@ pub(crate) struct InstallConfigOpts {
     #[clap(long)]
     #[serde(default)]
     pub(crate) generic_image: bool,
+
+    /// The stateroot name to use. Defaults to `default`.
+    #[clap(long, hide = true)]
```
cgwalters, line 161: We should at least debate: any reason not to stabilize (i.e. not `hide = true`) this? This really I guess is just the question of whether we want to tie bootc to the ostree concept of "stateroot" as is basically forever.

I think it's unlikely to be valuable for us to try to do anything really different here, so I'd say stabilize.

cgwalters (approved):
Looks sane to me, could just use a test in `tests-integration/src/install.rs` or so.

omertuc:
> Looks sane to me, could just use a test in `tests-integration/src/install.rs` or so.
> 

Coming back to this, is this still the right place? Or should it be part of the e2e tmt stuff?

lib/src/install.rs
omertuc: >question of whether we want to tie bootc to the ostree concept of "stateroot" 

I guess in a future where that doesn't exist anymore, this flag can become a simple no-op

tests-integration/src/install.rs
```diff
@@ -23,18 +23,32 @@ pub(crate) const BASE_ARGS: &[&str] = &[
...
+
+    for stateroot in [DEFAULT_STATEROOT, NON_DEFAULT_STATEROOT] {
+        if !Path::new(&format!("/ostree/deploy/{stateroot}")).exists() {
+            continue;
+        }
+
+        if Path::new(&format!("/ostree/deploy/{stateroot}/deploy")).exists() {
+            cmd!(
+            sh,
+            // env var hack needed because cmd! doesn't interpolate inside single quotes
```
omertuc, line 44: Worth reviewing

lib/src/install.rs
```diff
@@ -638,7 +644,11 @@ async fn install_container(
 ) -> Result<(ostree::Deployment, InstallAleph)> {
     let sepolicy = state.load_policy()?;
     let sepolicy = sepolicy.as_ref();
-    let stateroot = STATEROOT_DEFAULT;
+    let stateroot = state
+        .config_opts
+        .stateroot
+        .as_deref()
+        .unwrap_or(ostree_ext::container::deploy::STATEROOT_DEFAULT);
```
omertuc, line 651: Worth reviewing, I wasn't sure whether to duplicate this or just mutate state with the resolved stateroot

tests-integration/src/install.rs
cgwalters: Yeah, there's a lack of a "list stateroots" API in ostree here that's biting us. And a lack of an API to prune.

Hmmm...maybe in the short term we can add a hidden `bootc install wipe-ostree` or something?

cgwalters: A distinct but related issue here is the test suite semi-intentionally doesn't *itself* depend on ostree-rs-ext because it creates logistical complications for us in CI scenarios so it's best if we rely on hidden CLI bits in bootc to start and then we can use our experience from CI to potentially "graduate" those CLIs to stable APIs for others later.

cgwalters:
> Coming back to this, is this still the right place? Or should it be part of the e2e tmt stuff?

the e2e-tmt flow is generally oriented around bootc as a *client* flow, which is currently pretty distinct from the install path. I think it's still the right place to test install code in that file.

tests-integration/src/install.rs
omertuc: >Worth reviewing: any better ideas?

My comment was referring particularly to the single quotes thing, but I opened #776 if we want to pursue your suggestion further

omertuc: Let me know if you prefer I add that subcommand in this PR instead

omertuc: Force pushed to use #777

lib/src/install.rs
cgwalters: Hmm...perhaps slightly cleaner as a `fn stateroot(&self) -> &str` on `State`?

cgwalters (approved):

PR #623, opened 2024-06-21
install: Support non EFI partition format

yoheiueda:
This patch enables partition layouts that do not include EFI partition and/or boot partition.

Without this patch, partition numbers becomes not consecutive, and subsequent `findpart` will fail.

### DIFF
--- a/lib/src/install/baseline.rs
+++ b/lib/src/install/baseline.rs
@@ -29,10 +29,8 @@ use super::RW_KARG;
 use crate::mount;
 use crate::task::Task;
 
-pub(crate) const BOOTPN: u32 = 3;
 // This ensures we end up under 512 to be small-sized.
 pub(crate) const BOOTPN_SIZE_MB: u32 = 510;
-pub(crate) const EFIPN: u32 = 2;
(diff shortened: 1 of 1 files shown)


lib/src/install/baseline.rs
```diff
@@ -250,18 +250,20 @@ pub(crate) fn install_create_rootfs(
...
             &mut sgdisk.cmd,
-            1,
+            partno,
             "0:+1M",
             "BIOS-BOOT",
             Some("21686148-6449-6E6F-744E-656564454649"),
         );
     } else if cfg!(target_arch = "aarch64") {
         // reserved
+        partno += 1;
```
cgwalters, line 263: (Tangential but at this point we can kill the reserved partition on aarch64, we don't have the same use case/problems that motivated it on CoreOS with having symmetry with x86_64 - makes sense in a followup)

cgwalters (approved):
Thanks!

lib/src/install/baseline.rs
cgwalters: :arrow_right: https://github.com/containers/bootc/pull/624

PR #626, opened 2024-06-22
tree: Add missing -v /dev:/dev in a few places

cgwalters:
We're really going to need to switch over to having the container do dynamic mounts; cc https://github.com/containers/bootc/issues/380#issuecomment-1983721453

Just noticed this missing in one place, and found others with a grep.

Right now we do operate without, but it can be racier.


jmarrero (approved):
lgtm

PR #636, opened 2024-06-25
Add support for ppc64le architecture

sacsant:
Running bootc install commands on ppc64le system bails out
 with unsupported architecture error.

 ERROR Installing to disk: Creating rootfs: Unsupported architecture: powerpc64

 This patch enables support for ppc64le arch.

### DIFF
--- a/lib/src/install/baseline.rs
+++ b/lib/src/install/baseline.rs
@@ -258,6 +258,16 @@ pub(crate) fn install_create_rootfs(
         );
     } else if cfg!(target_arch = "aarch64") {
         // Only UEFI here for now
+    } else if cfg!(target_arch = "powerpc64") {
+        // PowerPC-PReP-boot
+        partno += 1;
+        sgdisk_partition(
+            &mut sgdisk.cmd,
+            partno,
+            "0:+4M",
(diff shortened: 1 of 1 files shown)


lib/src/install/baseline.rs
```diff
@@ -256,6 +256,16 @@ pub(crate) fn install_create_rootfs(
...
             Some("21686148-6449-6E6F-744E-656564454649"),
         );
+    } else if cfg!(target_arch = "powerpc64") {
+        // BIOS-BOOT
+        partno += 1;
+        sgdisk_partition(
+            &mut sgdisk.cmd,
+            partno,
+            "0:+4M",
+            "BIOS-BOOT",
```
cgwalters, line 266: I wonder if we should try to match https://github.com/coreos/coreos-assembler/blob/38235218542c39a29255933d3cea64e91754e85f/src/create_disk.sh#L206 and make the label `PowerPC-PReP-boot`?

sacsant: Yes, good point. I have pushed the modified code.

vrothberg (approved):
LGTM

PR #638, opened 2024-06-25
Test: Run e2e test on Packit

henrywang:
This PR includes:
1. rename old integration test to e2e test
2. drop aws test, use libvirt test (for bootc install to-existing-root and bootc switch test) instead
3. drop rhel test
4. replace quay.io registry by local registry
5. use Packit to run e2e test
6. build bootc PRM package on copr. Support `x86_64`, `aarch64`, and `ppc64le` arch and `RHEL 9`, `CS9`, `Fedora-40/rawhide` distros.
7. the [`integration` workflow](https://github.com/containers/bootc/blob/main/.github/workflows/integration.yml) removed.


tests/e2e/bootc-install.sh
```diff
@@ -76,35 +51,64 @@ EOF
...
+    -days 365 \
+    -out "${TEMPDIR}/domain.crt" \
+    -subj "/C=US/ST=Denial/L=Stockholm/O=bootc/OU=bootc-test/CN=bootc-test/emailAddress=bootc-test@bootc-test.org"
+
+greenprint "Update CA Trust"
+sudo cp "${TEMPDIR}/domain.crt" "/etc/pki/ca-trust/source/anchors/${REGISTRY_IP}.crt"
+sudo update-ca-trust
+
+greenprint "Deploy local registry"
+podman run \
```
cgwalters, line 83: 90% of the commands in this file start with `sudo`...but this one doesn't. Is that intentional?

I've been caught out in the past by being confused in similar CI jobs mixing things that require `sudo` vs those that don't and wondering why my `podman build` isn't visible.

It seems to me it'd be simpler to just run everything here as root - and where we specifically want to lower privileges, we can do so via e.g. `systemd-run -P DynamicUser=yes` for example.

plans/e2e.fmf
```diff
@@ -0,0 +1,59 @@
...
+/to-existing-root:
+  summary: Run bootc install to-existing-root and bootc switch test locally (nested)
+  environment+:
+    TEST_CASE: to-existing-root
+  discover+:
+    test:
+      - /to-existing-root
+  adjust+:
+    - when: arch == ppc64le
+      enabled: false
```
cgwalters, line 48: This syntax is YAML sorcery to me...is there any description of what the `/` and `+` do in particular?

tests/e2e/playbooks/templates/user-data.j2
```diff
@@ -0,0 +1,14 @@
+#cloud-config
+users:
+  - default
+  - name: {{ ssh_user }}
+    groups: wheel
+    sudo: ALL=(ALL) NOPASSWD:ALL
+    lock_passwd: true
+    ssh_authorized_keys:
+      - {{ lookup('ansible.builtin.file', ssh_key_pub) }}
+power_state:
```
cgwalters, line 10: Interesting...wasn't aware of this one. COuld use a comment I think...something like:
`# Shut down after the testing script has executed` or so?

cgwalters (approved):

plans/e2e.fmf
henrywang: When configure tmt plan or test, it supports regular expression, starting with `/` to avoid regular expression, but just a string match. The `/` comes from plan or test name, for example:
```
➜ tmt test ls
/tests/e2e/e2e/to-disk
/tests/e2e/e2e/to-existing-root
/tests-integration/build-image
```

The `+` means adding addition configuration based on `top level` configuration. In this case, The top level has `adjust` configured already.

tests/e2e/bootc-install.sh
henrywang: Updated.

tests/e2e/playbooks/templates/user-data.j2
henrywang: Comment added.

xtask/src/xtask.rs
henrywang, line 1: Hi @cgwalters, I made some changes here to support update `bootc.spec` to work with copr build requirement. Any comment on this code change? Thanks.

henrywang:
There is outage in Packit testing farm runner for x86_64 runner. I'll re-run the x86_64 tests after outage fixed. Please don't merge this PR until all tests passed. Thanks.

xtask/src/xtask.rs
```diff
@@ -246,6 +247,41 @@ fn package(sh: &Shell) -> Result<()> {
     Ok(())
 }
 
+fn update_spec(sh: &Shell) -> Result<Utf8PathBuf> {
```
cgwalters, line 250: Hmm...but this overlaps with `package-srpm` right? I think in theory we could share code here...or even extract the spec from the srpm?

xtask/src/xtask.rs
cgwalters: Code overall looks fine :smile:

xtask/src/xtask.rs
cgwalters: This also relates to https://github.com/containers/bootc/pull/651 and the copr makefile...we've grown way too many copies of this logic.

henrywang:
Hi @cgwalters! I think we can merge this PR now.
Sometimes, the testing farm might report `Guest couldn't be provisioned: Artemis resource ended in 'error' state`. Ticket [`add more bare metal instance types to public ranch`](https://issues.redhat.com/browse/TFT-2691) to track this issue.

cgwalters (approved):
Needs a rebase, but otherwise looks sane to me!

cgwalters (approved):

xtask/src/xtask.rs
```diff
@@ -244,6 +245,41 @@ fn package(sh: &Shell) -> Result<()> {
...
+                writeln!(o, "# Replaced by cargo xtask spec")?;
+                writeln!(o, "Version: {v}")?;
+            } else if line.starts_with("Source0") {
+                writeln!(o, "Source0: {srcpath}")?;
+            } else if line.starts_with("Source1") {
+                writeln!(o, "Source1: {src_vendorpath}")?;
+            } else {
+                writeln!(o, "{}", line)?;
+            }
+        }
```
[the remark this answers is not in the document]
henrywang: Learning and learned! Thanks.

PR #639, opened 2024-06-25
status: Add explicit `--format` argument

cgwalters:
Prep for adding `--format=human` for example. But this is also useful for anything that explicitly wants to consume YAML today.

### DIFF
--- a/docs/src/man/bootc-status.md
+++ b/docs/src/man/bootc-status.md
@@ -4,7 +4,7 @@ bootc-status - Display status
 
 # SYNOPSIS
 
-**bootc status** \[**\--json**\] \[**\--booted**\]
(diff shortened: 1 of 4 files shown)


vrothberg (commented):
Code LGTM but I don't see docs.

Q: Shall we remove the --json flag before 1.0?

cgwalters:
> Code LGTM but I don't see docs.

Currently the man pages are auto-generated into tarball releases; there's a currently manual `cargo xtask man2markdown` which also auto-reflects the manpages into markdown stored in git. (It's confusing...and tempting to go to writing man pages ...manually)

> Q: Shall we remove the --json flag before 1.0?

I marked it hidden, but it's trivial to continue to support for now.

vrothberg (approved):
LGTM

PR #642, opened 2024-06-26
tree-wide: rustfmt with `group_imports = "StdExternalCrate"`

cgwalters:
We were inconsistent here; I prefer this `StdExternalCrate` style. Unfortunately it's not stabilized, but this is a manual run with current nightly rustfmt. On the positive side, rustfmt will preserve this ordering in the future.


vrothberg (approved):
LGTM

PR #645, opened 2024-06-27
kargs: Various cleanups

cgwalters:
No functional changes intended; just prep for further work.

---

kargs: parser cleanups

- Accept &str vs owned String, allowing dropping unnecessary `clone`
- It isn't parsing a file necessarily, just a string buffer, so rename
  it
- Drop unnecessary `pub`
- Add doc comment
- Add error context to callers to be clear *which* file we failed
  to parse

---

kargs: Add a few more unit tests

---

kargs: Parse booted kargs via cap-std, not liboverdrop

liboverdrop isn't helpful here because I don't think we want
or need right now to actually support things like having a file
in `/etc` or `/run` override the kargs. Especially for `/run`
there's no dynamic runtime state.

No functional changes intended, just prep for more work.

---

It *might* in theory be useful for someone to have `/etc`
overrides, but since we weren't actually doing that today, don't
pretend we might by using liboverdrop but not passing `/etc` to
it.

Using cap-std here makes it easier to unit test safely.
(Although, liboverdrop really should grow an optional feature
 to use cap-std)

---

kargs: Drop unnecessary cloning for sys_arch

Since the function accepts a `&str`, drop the unnecessary cloning.

---

kargs: Drop unnecessary `mut`

By passing ownership of the vector here we don't need to make
it `mut`able.

---

kargs: Drop unnecessary allocation and `mut`

We can just pass the iterator directly to `extend`.

---

kargs: Minor code reordering

No functional changes, but it's clearer to have the `remote_kargs`
definition below where we do the short-circuit.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -193,6 +193,7 @@ dependencies = [
  "serde_ignored",
  "serde_json",
  "serde_yaml",
+ "similar-asserts",
  "tempfile",
  "tokio",
  "tokio-util",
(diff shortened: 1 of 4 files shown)


lukewarmtemp (approved):
LGTM

PR #647, opened 2024-06-28
install: Pick up kargs.d kernel arguments too

cgwalters:
This was a rather important miss; we need to pick
up the kargs.d files when doing a `bootc install` too.

---


lukewarmtemp (approved):
LGTM

PR #648, opened 2024-06-28
kargs: A few more minor cleanups and test improvements

cgwalters:


lib/src/kargs.rs
```diff
@@ -63,25 +63,28 @@ fn get_kargs_from_ostree(
...
-                    ostree::Repo::load_file(repo, fetched_contents_checksum.as_str(), cancellable)?;
-                let file_content = f.0;
-                let mut reader =
-                    ostree_ext::prelude::InputStreamExtManual::into_read(file_content.unwrap());
-                let s = std::io::read_to_string(&mut reader)?;
-                let parsed_kargs =
-                    parse_kargs_toml(&s, sys_arch).with_context(|| format!("Parsing {name}"))?;
-                ret.extend(parsed_kargs);
-            }
+        let name = if let Some(name) = name.to_str() {
```
lukewarmtemp, line 77: I'm totally fine with this change, but I guess I just personally don't like having `else` statements only containing `continue` even if it helps with code readability. Once again, okay with this change.

lukewarmtemp (approved):
LGTM

lib/src/kargs.rs
cgwalters: It's somewhat idiomatic I think in Rust at least.

Although...actually for a while now there's been [let-else](https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html) which I think we can do a tree-wide conversion to.

cgwalters: :arrow_right:  https://github.com/containers/bootc/pull/650

PR #649, opened 2024-06-28
install: Only open / once

cgwalters:
We juggle multiple things we call in different places the "rootfs", in some the container's root, in others the target root.

Add a handy member variable of `State` for the container root and use it consistently. This improves clarity now, and is prep for later patches.


ckyrouac (approved):
lgtm, this will be handy

PR #650, opened 2024-06-28
tree-wide: Convert to let-else syntax

cgwalters:
Convert to [let else](https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html) where possible. I've been coding Rust long before it was stable, and my muscle memory has the expanded form.

But the short form is nicer.


jmarrero (approved):
lgtm

PR #651, opened 2024-06-28
copr: Use dnf builddep

cgwalters:
Yet *another* copy of our build dependencies here... The core problem right now is that we embed pre-built manpages as part of the "source"...which means we need all of our build dependencies there.

It's tempting to switch to not generating man pages.


cgwalters:
xref https://download.copr.fedorainfracloud.org/results/rhcontainerbot/bootc/srpm-builds/07689202/builder-live.log.gz

jmarrero (approved):
lgtm

PR #652, opened 2024-06-28
tree-wide: Fix minor clippy lints

cgwalters:
We had an unused variable...should probably start denying that in CI.

Fix a few other style things while we have the patient open.


jmarrero (approved):
lgtm

PR #654, opened 2024-06-28
install: Use separate pull stage for progress output

cgwalters:
The goal here is to get interactive progress on pulling, as that can be slow and also I'd like to get more information there.


cgwalters:
Draft since I'm getting tty output when I don't expect it, and I need to dig into what others are doing WRT indicatif in that space.

cgwalters:
Replaced by https://github.com/containers/bootc/pull/655

PR #655, opened 2024-06-29
install: Disable fsync() in repo when pulling && improved pull progress

cgwalters:
See individual patches. Motivated by https://github.com/containers/bootc/discussions/646

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -533,7 +533,7 @@ async fn upgrade(opts: UpgradeOpts) -> Result<()> {
             }
         }
     } else {
-        let fetched = crate::deploy::pull(sysroot, imgref, opts.quiet).await?;
+        let fetched = crate::deploy::pull(repo, imgref, opts.quiet).await?;
         let kargs = crate::kargs::get_kargs(repo, &booted_deployment, fetched.as_ref())?;
         let staged_digest = staged_image.as_ref().map(|s| s.image_digest.as_str());
         let fetched_digest = fetched.manifest_digest.as_str();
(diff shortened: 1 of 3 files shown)


cgwalters:
@henrywang Hmm,

```
fatal: [guest]: FAILED! => changed=false 
  assertion: result_booted_cachedupdate_digest.stdout == result_rollback_image_digest.stdout
```

Have you see this in other PRs? I think it'd be helpful to acquire the full systemd journal here, is that easily doable?

jeckersb:
Overall looks great to me.  The only thing is that it appears to cut off the `byte_bar` at 80 characters, which looks weird on a wider terminal and cuts off the majority of the message.  Ends up like:
```
 └ Fetching ██████████████████░░ 127.51 MiB/136.26 MiB (232.34 MiB/s) layer sha2
```
I'm not familiar with `indicatif` but I'll see if I can make it use more space when it's available.

cgwalters:
Yeah, I finally figured out that mutating the `{prefix}` were what was causing the "eating previous terminal output" behavior I saw in previous changes here, and I switched to setting `{msg}` which is how things are intended to work. But indeed it seems like the library isn't detecting the terminal width correctly, or something else is going wrong. Could add a `dbg!` with https://docs.rs/indicatif/latest/indicatif/trait.TermLike.html#tymethod.width ?

cgwalters:
Figured it out https://github.com/containers/podman-bootc/pull/55

jeckersb:
> Figured it out [containers/podman-bootc#55](https://github.com/containers/podman-bootc/pull/55)

Nice, I had gotten as far as:

https://github.com/console-rs/console/blob/master/src/unix_term.rs#L46-L64

And determined that `isatty` is correctly returning `1` but the `ioctl` is leaving the winsize rows and cols as `0` so it falls back to the default width of `80`.

jeckersb (approved):
Confirmed this works like a charm with the `podman-bootc` fix :tada:

PR #659, opened 2024-07-01
Retrieve bound images when staging new image

ckyrouac:
This is just a PoC using quadlet/systemd to pull "lifecycle" bound images. 

I have a bootc image pushed to `quay.io/ckyrouac/bootc-lldb-bound:latest` for testing. Any custom bootc image with quadlet image files located in `/etc/containers/systemd/bootc` should work.

Steps for testing:

1. Build a container image with these changes and start a VM: https://github.com/containers/bootc/blob/main/HACKING.md#creating-your-edit-compile-debug-cycle
2. `sudo bootc switch quay.io/ckyrouac/bootc-lldb-bound:latest`
3. Validate the `quay.io/ckyrouac/bootc-lldb` image was pulled via `sudo podman images`
4. Validate the new image is staged via `sudo bootc status`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -195,6 +195,7 @@ dependencies = [
  "serde_yaml",
  "similar-asserts",
  "tempfile",
+ "tini",
  "tokio",
  "tokio-util",
  "toml",
@@ -2036,6 +2037,12 @@ checksum = "d050e60b33d41c19108b32cea32164033a9013fe3b46cbd4457559bfbf77afaa"
 dependencies = [
  "num_cpus",
 ]
+
+[[package]]
+name = "tini"
+version = "1.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e004df4c5f0805eb5f55883204a514cfa43a6d924741be29e871753a53d5565a"
 
 [[package]]
 name = "tinyvec"
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -30,7 +30,7 @@ liboverdrop = "0.1.0"
 libsystemd = "0.7"
 openssl = "^0.10.64"
 # TODO drop this in favor of rustix
-nix = { version = "0.29", features = ["ioctl", "sched"] }
+nix = { version = "0.29", features = ["ioctl", "sched" ] }
 regex = "1.10.4"
 rustix = { "version" = "0.38.34", features = ["thread", "fs", "system", "process"] }
 schemars = { version = "0.8.17", features = ["chrono"] }
@@ -45,6 +45,7 @@ tempfile = "3.10.1"
 toml = "0.8.12"
 xshell = { version = "0.2.6", optional = true }
 uuid = { version = "1.8.0", features = ["v4"] }
+tini = "1.3.0"
 
 [dev-dependencies]
 similar-asserts = { version = "1.5.0" }
--- a/lib/src/boundimage.rs
+++ b/lib/src/boundimage.rs
@@ -1,0 +1,290 @@
+use crate::task::Task;
+use anyhow::{Context, Result};
+use camino::Utf8Path;
+use cap_std_ext::cap_std::fs::Dir;
+use cap_std_ext::dirext::CapStdExtDirExt;
+use fn_error_context::context;
+use ostree_ext::ostree::Deployment;
+use ostree_ext::sysroot::SysrootLock;
+use rustix::fd::BorrowedFd;
+use rustix::fs::{OFlags, ResolveFlags};
+use std::fs::File;
+use std::io::Read;
+use std::os::unix::io::AsFd;
+
+const BOUND_IMAGE_DIR: &str = "usr/lib/bootc-experimental/bound-images.d";
+
+// Access the file descriptor for a sysroot
+#[allow(unsafe_code)]
+pub(crate) fn sysroot_fd(sysroot: &ostree_ext::ostree::Sysroot) -> BorrowedFd {
+    unsafe { BorrowedFd::borrow_raw(sysroot.fd()) }
+}
+
+pub(crate) fn pull_bound_images(sysroot: &SysrootLock, deployment: &Deployment) -> Result<()> {
+    let sysroot_fd = sysroot_fd(&sysroot);
+    let sysroot_fd = Dir::reopen_dir(&sysroot_fd)?;
+    let deployment_root_path = sysroot.deployment_dirpath(&deployment);
+    let deployment_root = &sysroot_fd.open_dir(&deployment_root_path)?;
+
+    let bound_images = parse_spec_dir(&deployment_root, BOUND_IMAGE_DIR)?;
+    pull_images(deployment_root, bound_images)?;
+
+    Ok(())
+}
+
+#[context("parse bound image spec dir")]
+fn parse_spec_dir(root: &Dir, spec_dir: &str) -> Result<Vec<BoundImage>> {
+    let Some(bound_images_dir) = root.open_dir_optional(spec_dir)? else {
+        return Ok(Default::default());
+    };
+
+    let mut bound_images = Vec::new();
+
+    for entry in bound_images_dir
+        .entries()
+        .context("Unable to read entries")?
+    {
+        //validate entry is a symlink with correct extension
+        let entry = entry?;
+        let file_name = entry.file_name();
+        let file_name = if let Some(n) = file_name.to_str() {
+            n
+        } else {
+            anyhow::bail!("Invalid non-UTF8 filename: {file_name:?} in {}", spec_dir);
+        };
+
+        if !entry.file_type()?.is_symlink() {
+            anyhow::bail!("Not a symlink: {file_name}");
+        }
+
+        //parse the file contents
+        let path = Utf8Path::new(spec_dir).join(file_name);
+        let mut file: File = rustix::fs::openat2(
+            root.as_fd(),
+            path.as_std_path(),
+            OFlags::CLOEXEC | OFlags::RDONLY,
+            rustix::fs::Mode::empty(),
+            ResolveFlags::IN_ROOT,
+        )
+        .context("Unable to openat")?
+        .into();
+
+        let mut file_contents = String::new();
+        file.read_to_string(&mut file_contents)
+            .context("Unable to read file contents")?;
+
+        let file_ini = tini::Ini::from_string(&file_contents).context("Parse to ini")?;
+        let file_extension = Utf8Path::new(file_name).extension();
+        let bound_image = match file_extension {
+            Some("image") => parse_image_file(file_name, &file_ini),
+            Some("container") => parse_container_file(file_name, &file_ini),
+            _ => anyhow::bail!("Invalid file extension: {file_name}"),
+        }?;
+
+        bound_images.push(bound_image);
+    }
+
+    Ok(bound_images)
+}
+
+#[context("parse image file {file_name}")]
+fn parse_image_file(file_name: &str, file_contents: &tini::Ini) -> Result<BoundImage> {
+    let image: String = file_contents
+        .get("Image", "Image")
+        .ok_or_else(|| anyhow::anyhow!("Missing Image field in {file_name}"))?;
+
+    //TODO: auth_files have some semi-complicated edge cases that we need to handle,
+    //      so for now let's bail out if we see one since the existence of an authfile
+    //      will most likely result in a failure to pull the image
+    let auth_file: Option<String> = file_contents.get("Image", "AuthFile");
+    if auth_file.is_some() {
+        anyhow::bail!("AuthFile is not supported by bound bootc images");
+    }
+
+    let bound_image = BoundImage::new(image.to_string(), None)?;
+    Ok(bound_image)
+}
+
+#[context("parse container file {file_name}")]
+fn parse_container_file(file_name: &str, file_contents: &tini::Ini) -> Result<BoundImage> {
+    let image: String = file_contents
+        .get("Container", "Image")
+        .ok_or_else(|| anyhow::anyhow!("Missing Image field in {file_name}"))?;
+
+    let bound_image = BoundImage::new(image.to_string(), None)?;
+    Ok(bound_image)
+}
+
+#[context("pull bound images")]
+fn pull_images(_deployment_root: &Dir, bound_images: Vec<BoundImage>) -> Result<()> {
+    //TODO: do this in parallel
+    for bound_image in bound_images {
+        let mut task = Task::new("Pulling bound image", "/usr/bin/podman")
+            .arg("pull")
+            .arg(&bound_image.image);
+        if let Some(auth_file) = &bound_image.auth_file {
+            task = task.arg("--authfile").arg(auth_file);
+        }
+        task.run()?;
+    }
+
+    Ok(())
+}
+
+#[derive(PartialEq, Eq)]
+struct BoundImage {
+    image: String,
+    auth_file: Option<String>,
+}
+
+impl BoundImage {
+    fn new(image: String, auth_file: Option<String>) -> Result<BoundImage> {
+        validate_spec_value(&image).context("Invalid image value")?;
+
+        if let Some(auth_file) = &auth_file {
+            validate_spec_value(auth_file).context("Invalid auth_file value")?;
+        }
+
+        Ok(BoundImage { image, auth_file })
+    }
+}
+
+fn validate_spec_value(value: &String) -> Result<()> {
+    let mut number_of_percents = 0;
+    for char in value.chars() {
+        if char == '%' {
+            number_of_percents += 1;
+        } else if number_of_percents % 2 != 0 {
+            anyhow::bail!("Systemd specifiers are not supported by bound bootc images: {value}");
+        } else {
+            number_of_percents = 0;
+        }
+    }
+
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use cap_std_ext::cap_std;
+    use std::io::Write;
+
+    #[test]
+    fn test_parse_spec_dir() -> Result<()> {
+        const CONTAINER_IMAGE_DIR: &'static str = "usr/share/containers/systemd";
+
+        // Empty dir should return an empty vector
+        let td = &cap_std_ext::cap_tempfile::TempDir::new(cap_std::ambient_authority())?;
+        let images = parse_spec_dir(td, &BOUND_IMAGE_DIR).unwrap();
+        assert_eq!(images.len(), 0);
+
+        td.create_dir_all(BOUND_IMAGE_DIR).unwrap();
+        td.create_dir_all(CONTAINER_IMAGE_DIR).unwrap();
+        let images = parse_spec_dir(td, &BOUND_IMAGE_DIR).unwrap();
+        assert_eq!(images.len(), 0);
+
+        // Should return BoundImages
+        let mut foo_file = td
+            .create(format!("{CONTAINER_IMAGE_DIR}/foo.image"))
+            .unwrap();
+        foo_file.write_all(b"[Image]\n").unwrap();
+        foo_file.write_all(b"Image=quay.io/foo/foo:latest").unwrap();
+        td.symlink_contents(
+            format!("/{CONTAINER_IMAGE_DIR}/foo.image"),
+            format!("{BOUND_IMAGE_DIR}/foo.image"),
+        )
+        .unwrap();
+
+        let mut bar_file = td
+            .create(format!("{CONTAINER_IMAGE_DIR}/bar.image"))
+            .unwrap();
+        bar_file.write_all(b"[Image]\n").unwrap();
+        bar_file.write_all(b"Image=quay.io/bar/bar:latest").unwrap();
+        td.symlink_contents(
+            format!("/{CONTAINER_IMAGE_DIR}/bar.image"),
+            format!("{BOUND_IMAGE_DIR}/bar.image"),
+        )
+        .unwrap();
+
+        let mut images = parse_spec_dir(td, &BOUND_IMAGE_DIR).unwrap();
+        images.sort_by(|a, b| a.image.as_str().cmp(&b.image.as_str()));
+        assert_eq!(images.len(), 2);
+        assert_eq!(images[0].image, "quay.io/bar/bar:latest");
+        assert_eq!(images[1].image, "quay.io/foo/foo:latest");
+
+        // Invalid symlink should return an error
+        td.symlink("./blah", format!("{BOUND_IMAGE_DIR}/blah.image"))
+            .unwrap();
+        assert!(parse_spec_dir(td, &BOUND_IMAGE_DIR).is_err());
+
+        // Invalid image contents should return an error
+        let mut error_file = td.create("error.image").unwrap();
+        error_file.write_all(b"[Image]\n").unwrap();
+        td.symlink_contents("/error.image", format!("{BOUND_IMAGE_DIR}/error.image"))
+            .unwrap();
+        assert!(parse_spec_dir(td, &BOUND_IMAGE_DIR).is_err());
+
+        Ok(())
+    }
+
+    #[test]
+    fn test_validate_spec_value() -> Result<()> {
+        //should not return an error with no % characters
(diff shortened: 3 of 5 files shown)


lib/src/task.rs
```diff
@@ -76,6 +74,11 @@ impl Task {
         self
     }
 
+    pub(crate) fn env(mut self, k: &String, v: &String) -> Self {
```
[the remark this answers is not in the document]
cgwalters: - String
- &String (borrowed)
- &str - slice of borrowed string

lib/src/boundimage.rs
```diff
@@ -0,0 +1,129 @@
...
+const QUADLET_BINARY: &'static str = "/usr/lib/systemd/system-generators/podman-system-generator";
+const SYSTEMD_DIR: &'static str = "/etc/systemd/system";
+
+pub(crate) struct BoundImageManager {
+    quadlet_unit_dir: String,
+    units: Vec<String>,
+}
+
+impl BoundImageManager {
+    pub(crate) fn new(deployment: Deployment, sysroot: &SysrootLock) -> BoundImageManager {
```
cgwalters, line 23: Should likely be `&Deployment` here (and borrow with `&` at the call site)

lib/src/deploy.rs
```diff
@@ -278,20 +278,21 @@ async fn deploy(
     image: &ImageState,
     origin: &glib::KeyFile,
     opts: Option<ostree::SysrootDeployTreeOpts<'_>>,
-) -> Result<()> {
+) -> Result<Deployment> {
```
cgwalters, line 281: We can split this out as its own prep PR that can land immediately.

vrothberg:
How do we deal with the following scenario?

The new image ships with a new authfile and the `.image` files reference both, authfiles on the old image and the new image.

cgwalters:
> The new image ships with a new authfile and the .image files reference both, authfiles on the old image and the new image.

To be clear this is a specific instance of the general question of what happens between overlap/conflict between options in the old and new .image file right? e.g. they could also conflict on `CertDir` or even the arbitrary `PodmanArgs`.

Or actually the most obvious conflict: what happens when the `.image` file references an image by digest (a valid thing to do!), and the digests are different? We *must* handle this scenario.

One approach would be to change quadlet to support generating unique names for the units (a simple approach is to hash the input, i.e. we end up with `2c26b46b68ffc68ff99b453c1d30413413422d706483bfa0f98a5e886266e7ae-image.service` and not `foo-image.service`; ugly but viable here).

But in general let's remember here - the goal is that the image is present. And it must already be present for the current root (barring some explicit admin action to prune)!

Perhaps the simplest thing here is to add a prefix `bootc-pending-` to the generated unit names?

ckyrouac:
> The new image ships with a new authfile and the .image files reference both, authfiles on the old image and the new image.

Ah, if I understand this right, there is a scenario where a new bootc image has a new bound .image file that references a AuthFile, e.g. `AuthFile=%E/registry/newauth.json`. We can't parse the AuthFile location due to the systemd specifier and it doesn't exist on the existing root. 

So I think that leaves two options
1. run all of this in the new root, i.e. chroot new-image -> quadlet -> systemd -> podman pull
2. create a new "bootc-image" spec file that we can reliably parse against the new root to locate the Authfile, CertDir, etc. but still run `podman pull` from the existing root

Option 1 would ideally be implemented after https://github.com/containers/bootc/issues/640 is in place. There is also a possibility of a conflict between podman and the storage format, e.g. podman 6 for some reason changes the storage format of `/var/lib/containers` making it incompatible with the existing root without some migration. Pretty unlikely and rare though, and the migration would need to happen eventually anyways.

Option 2 could still result in a failed pull if the "bootc-image" spec changes in the new image.

I lean towards option 1. I'm not sure of all the implications of running in the new image's root though.

ckyrouac:
Started playing around with trying to run systemd in a chroot or a container, I don't think that will be possible without a bunch of hacks (unless I'm missing some hidden systemd features). I'll start working on reading a separate spec file.

cgwalters:
I would lean towards erroring out if we detect any specifiers (scanning for any `%X` that's not `%%` I think would be right).

Taking `%E` as an example...these are currently always going to be system images, so it will always resolve to `/etc`.

I could *imagine* someone doing something with `%w` or `%W` (os version/variant related) but...eh.

If we do that, then 

> Started playing around with trying to run systemd in a chroot or a container, I don't think that will be possible without a bunch of hacks (unless I'm missing some hidden systemd features). 

That's actually the default when we just run a base bootc image today, right? It's what you get when you `podman run quay.io/centos-bootc/centos-bootc:stream9`. There's a bit of internal homework that we should do though to set things up in the same way as `ostree-prepare-root` does in the initramfs (e.g. use composefs).

In theory we could switch to by default mounting each deployment in that way? Some cost/benefit there for sure. Today for example the previous deployment roots are just plain hardlinked dirtrees in the legacy ostree model.

cgwalters:
Although...it would be reasonable to want to add a new image and a pull secret for that image as a single step, so we likely should default to resolving any specified filesystem paths relative to the target root. This doesn't actually require us to run the target code in a container image though; we could resolve the paths and prepend the deployment root (or, change the logic so we open file descriptors for them, and pass them via the `/proc/self/fd/N` magic link).

ckyrouac:
> That's actually the default when we just run a base bootc image today, right? It's what you get when you podman run quay.io/centos-bootc/centos-bootc:stream9. There's a bit of internal homework that we should do though to set things up in the same way as ostree-prepare-root does in the initramfs (e.g. use composefs).

Ah, I was trying to use systemctl in a simple chroot but maybe running systemd directly in a container might work. Something like,

`podman run quay.io/centos-bootc/centos-bootc:stream9 /usr/lib/systemd/systemd --unit=bootc-bound-images.target`

> I would lean towards erroring out if we detect any specifiers (scanning for any %X that's not %% I think would be right).

That seems reasonable to me. I'm not sure how common it is to use the specifiers and it should be a relatively easy thing for a user to replace in the image.

ckyrouac:
Unless anyone objects, I'll continue with running in the existing root, restricting the use of specifiers, and expanding file paths to the new root. This should be the simplest/quickest thing to implement. If users complain about the lack of specifiers then we can add support for them by doing the container implementation.

vrothberg:
I do not want to derail the direction but I see a lot of corner cases being discussed which may cause pain in the future.

I think that we won't run into those issues by introducing a new file `/etc/bootc/pinned_images.conf` with a simple syntax and something a user is able to comprehend.

I believe we need two fields:
* Image: the image to pull
* Authfile: optional authfile.  If the authfile exists on the new image, it'll take precedence over the one of the host.

I find this _much_ simpler to document, implement and maintain.  I am not married to the idea but want to to motivate to reconsider.

vrothberg:
@ckyrouac WDYT?

ckyrouac:
I'll start working on using a separate image spec file. I'll try to make sure the spec is easy to adapt to new fields in the future.

lib/src/deploy.rs
cgwalters: WDYT about doing that now? I think we could also add the code using `openat` to read the target root.

cgwalters: I also did https://github.com/containers/bootc/pull/674/commits/009c08adb4f427cdcde2add48f5683817ef9c19f#diff-b28d7b8052d2d63505b324be5a3affca036ff7060cba6fc61f4519e29ca5ca9eR35 while looking at code related to this for a different reason...although - and this is subtle - today [this readme describes why cap-std doesn't use RESOLVE_IN_ROOT](https://github.com/bytecodealliance/cap-std/?tab=readme-ov-file#similar-crates) ...so we'd need to add a variant that looks like https://github.com/containers/composefs/pull/286/files#diff-2c1a4b8055c87401bf768c30392660303de5e336af37c1cec34e3eeaee55e6deR85 that drops down to the raw rustix apis.

(In general for us, it actually would be nice potentially to add to cap-std a variant that uses `RESOLVE_IN_ROOT`...I'll chime in on https://github.com/bytecodealliance/cap-std/issues/353 about that)

ckyrouac: just opened #675 to address the first part. I'm not quite sure where to put the openat code other than in the BoundImage functions of this PR.

I also just pushed my latest WIP changes to this PR's branch. I've done some initial testing of the latest changes, it seems to be basically working. I still need to add concurrency to the pulls, tests, docs, etc. but I wanted to get the current state out there for an initial review.

lib/src/boundimage.rs
```diff
@@ -0,0 +1,171 @@
...
+use ostree_ext::sysroot::SysrootLock;
+use regex::Regex;
+use rustix::fs::{OFlags, ResolveFlags};
+use std::fs;
+use std::fs::File;
+use std::io::Read;
+use std::os::unix::io::AsFd;
+use std::path::Path;
+
+const BOUND_IMAGE_DIR: &'static str = "usr/lib/bootc/bound-images.d";
```
cgwalters, line 15: We discussed making this something like `usr/lib/bootc-experimental/bound-images.d` so we can allow experimenting with it without committing immediately to the interface.

lib/src/boundimage.rs
```diff
@@ -0,0 +1,171 @@
...
+
+pub(crate) struct BoundImageManager {
+    deployment_root: String, // full path to the deployment root
+    spec_dir: String,        //full path to the bound image spec dir
+}
+
+impl BoundImageManager {
+    pub(crate) fn new(deployment: &Deployment, sysroot: &SysrootLock) -> Result<BoundImageManager> {
```
cgwalters, line 23: Usually we have the sysroot go first in arguments

lib/src/boundimage.rs
```diff
@@ -0,0 +1,171 @@
...
+
+        if let Some(auth_file) = &auth_file {
+            validate_spec_value(auth_file).context("Invalid auth_file value")?;
+        }
+
+        Ok(BoundImage { image, auth_file })
+    }
+}
+
+fn validate_spec_value(value: &String) -> Result<()> {
```
cgwalters, line 164: This one seems very unit-testable

lib/src/boundimage.rs
```diff
@@ -0,0 +1,171 @@
...
+use rustix::fs::{OFlags, ResolveFlags};
+use std::fs;
+use std::fs::File;
+use std::io::Read;
+use std::os::unix::io::AsFd;
+use std::path::Path;
+
+const BOUND_IMAGE_DIR: &'static str = "usr/lib/bootc/bound-images.d";
+
+pub(crate) struct BoundImageManager {
```
cgwalters, line 17: I'm generally OK with this, however...AFAICS we basically create the struct to call a single method on it (`run`) right?

Do we need the struct at all then? In Rust (unlike e.g. [Java which forces everything in a class](http://steve-yegge.blogspot.com/2006/03/execution-in-kingdom-of-nouns.html) or Python which inherited some of the OOP (but also supports free functions)), Rust has a lot more of the sense of "modules are namespaces, and free functions are totally OK".

For example the [std::fs module](https://doc.rust-lang.org/std/fs/index.html) has a ton of pure free functions.

In our own [kargs.rs](https://github.com/containers/bootc/blob/main/lib/src/kargs.rs) - it's just free functions, and I think personally that's fine for what it does.

In this case I think it'd be equally clear if the sole "external" entrypoint was something like

```
pub(crate) fn pull_bound_images(sysroot: &SysrootLock, deployment: &Deployment) -> Result<()>
```
or so, right?

The unit of "encapsulation" here is a module, not a struct; the default `fn` is module private.

But again, I don't have a *really* strong opinion here, this is mainly a style thing.

lib/src/boundimage.rs
```diff
@@ -0,0 +1,171 @@
...
+
+            let fd = rustix::fs::openat2(
+                root_fd,
+                file_path,
+                OFlags::empty(),
+                rustix::fs::Mode::empty(),
+                ResolveFlags::IN_ROOT,
+            )?;
+
+            let mut file = { File::from(fd) };
```
cgwalters, line 80: Minor style thing, I think this would be clearer as `let fd: File = rustix::fs::openat2()?.into()` alongside the above.
(Dealing with `OwnedFd` vs `File` is a weird part of dropping down to the rustix level)

lib/src/boundimage.rs
```diff
@@ -0,0 +1,171 @@
...
+            }?;
+
+            bound_images.push(bound_image);
+        }
+
+        Ok(bound_images)
+    }
+
+    #[context("parse image file {file_name}")]
+    fn parse_image_file(&self, file_name: &str, file_contents: &ini::Ini) -> Result<BoundImage> {
```
cgwalters, line 99: Could probably unit test this one too

lib/src/boundimage.rs
```diff
@@ -0,0 +1,171 @@
...
+        let image = if let Some(img) = file_contents.get_from(Some("Image"), "Image") {
+            img
+        } else {
+            anyhow::bail!("Missing Image field in {file_name}");
+        };
```
cgwalters, line 104: This is 100% fine as is, but slightly more idiomatic I think as:

```
let image = file_contents.get_from(Some("Image"), "Image").ok_or_else(|| anyhow!("Missing ..."))?;
```

lib/src/boundimage.rs
```diff
@@ -0,0 +1,171 @@
...
+
+        let bound_image = BoundImage::new(image.to_string(), None)?;
+        Ok(bound_image)
+    }
+
+    #[context("pull bound images")]
+    fn pull_images(&self, bound_images: Vec<BoundImage>) -> Result<()> {
+        //TODO: do this in parallel
```
cgwalters, line 132: If we did do that it'd open up interesting questions around how we display progress bars. Are we doing that today via podman? I think the `Task` bits may default stdin to `/dev/null`? If we're not then we should probably explicitly provide e.g. `-q`.

One thing I'm thinking here is perhaps we go ahead and generate a systemd unit for each container to pull (exactly like quadlet would do)?

But, clearly this is not a blocker for landing this PoC work.

cgwalters (commented):
Thanks! This is looking like good progress!

The other thing to figure out adding here is an integration test...let's chat about that.

ckyrouac:
Thanks for the thorough review! It's very helpful to get to know rust style/idioms. I pushed all the suggested style related changes and will start on some unit tests now. For integration tests, I was planning to try to write some with the new tmt/nushell framework, if that makes sense?

cgwalters:
> For integration tests, I was planning to try to write some with the new tmt/nushell framework, if that makes sense?

I think so yep! Though so far one tangential thing is those tests tend not to pull much from the network, but I think we can just bite that bullet.

lib/src/boundimage.rs
```diff
@@ -103,14 +109,14 @@ fn parse_container_file(file_name: &str, file_contents: &ini::Ini) -> Result<Bou
...
-fn pull_images(deployment_root: String, bound_images: Vec<BoundImage>) -> Result<()> {
+fn pull_images(_deployment_root: &Dir, bound_images: Vec<BoundImage>) -> Result<()> {
     //TODO: do this in parallel
     for bound_image in bound_images {
         let mut task = Task::new("Pulling bound image", "/usr/bin/podman")
             .arg("pull")
             .arg(&bound_image.image);
         if let Some(auth_file) = &bound_image.auth_file {
-            task = task.arg("--authfile").arg(format!("/{deployment_root}/{auth_file}"));
+            task = task.arg("--authfile").arg(auth_file);
```
cgwalters, line 119: We had a live chat about this...dealing with the authfiles is tricky, because at this current point in the flow, we may not have done the "etc merge" and so the `etc` that we see in the root will only have config from the image.

We may need to explicitly do a dance like "look for the auth file in the new root, fall back to current root"

cgwalters:
Hmm cargo-deny is giving:

```
 error[rejected]: failed to satisfy license requirements
  ┌─ tiny-keccak 2.0.2 (registry+https://github.com/rust-lang/crates.io-index):4:12
  │
4 │ license = "CC0-1.0"
  │            ^^^^^^^
  │            │
  │            license expression retrieved via Cargo.toml `license`
  │            rejected: not explicitly allowed
  │
  = CC0-1.0 - Creative Commons Zero v1.0 Universal:
  =   - FSF Free/Libre
  = tiny-keccak v2.0.2
    └── const-random-macro v0.1.16
        └── const-random v0.1.18
            └── dlv-list v0.5.2
                └── ordered-multimap v0.7.3
                    └── rust-ini v0.21.0
                        └── bootc-lib v0.1.13
                            └── bootc v0.1.9
```

There's a few new dependency crates here. I *think* we can allowlist CC-1.0, but I need to check. Hmm wait..like *why* is `dlv-list` pulling in a "compile time random number generator"? (digs) Oh I see, it's [this code](https://github.com/sgodwincs/dlv-list-rs/blob/a7c7da926f51a8a26ddc6ad417b8ef159982f35e/src/lib.rs#L2254) which we don't even run because it's only for no-std environments. Blah. There's definitely some days where the Rust dependency chains feel *too* fine grained...

cgwalters:
> I think we can allowlist CC-1.0, but I need to check. 

Unfortunately...this one appears to be marginal. See https://docs.fedoraproject.org/en-US/legal/allowed-licenses/ 

Honestly, it's pretty annoying that someone hyper-optimized the ini parser to the point of having a specialized linked-list data structure that ends up transitively pulling in a custom implementation of Keccak (sha-3 family) for a use case we don't need. 

I can understand e.g. [hyper-optimizing JSON parsing](https://github.com/serde-rs/json-benchmark) because yeah, sometimes you end up with gigabytes of JSON in the real world. But INI files? I don't think so.

[Looking around](https://lib.rs/search?q=ini) there's also [tini](https://docs.rs/tini/latest/) tini/) - the code looks good to me. Spot checking the maintenance status too by looking at PRs I see e.g. [this PR](https://github.com/pinecrew/tini/pull/42) by a fellow RHT employee. Let's switch to this crate?

ckyrouac:
bleh, thanks for digging into that. I'll switch to tini.

ckyrouac:
alright - added more unit tests and more cleanup. I think we can get this in so people can start some initial testing with it. I'll continue working on the install path and integration tests separately.

One significant change I made is to error out if an `AuthFile` field is found in the `.image` spec. The presence of an AuthFile most likely means the image requires it to be pulled, so we won't be able to pull the image until we add support for it. We could also just ignore the AuthFile, print a warning, and continue trying to pull the image.

Cargo.lock
```diff
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.21.0"
+version = "0.22.0"
```
cgwalters: I think we have a full `cargo update` here, I'll rebase this just on https://github.com/containers/bootc/pull/669

lib/src/boundimage.rs
```diff
@@ -0,0 +1,296 @@
...
+            format!("/{CONTAINER_IMAGE_DIR}/foo.image"),
+            format!("{BOUND_IMAGE_DIR}/foo.image"),
+        )
+        .unwrap();
+
+        let mut bar_file = td
+            .create(format!("{CONTAINER_IMAGE_DIR}/bar.image"))
+            .unwrap();
+        bar_file.write_all(b"[Image]\n").unwrap();
+        bar_file.write_all(b"Image=quay.io/bar/bar:latest").unwrap();
```
cgwalters, line 209: This is totally fine as is, but would be shorter with e.g.:

```
td.write(format!("{CONTAINER_IMAGE_DIR}/bar.image"), "[Image]\nImage=quay.io/bar/bar:latest").unwrap();
```

lib/src/boundimage.rs
```diff
@@ -0,0 +1,296 @@
...
+    }
+}
+
+fn validate_spec_value(value: &String) -> Result<()> {
+    let mut number_of_percents = 0;
+    for char in value.chars() {
+        if char == '%' {
+            number_of_percents += 1;
+        } else if number_of_percents % 2 != 0 {
+            anyhow::bail!("Systemd specifiers are not supported by bound bootc images: {value}");
```
cgwalters, line 164: Hmm, I don't think this is quite right, because if one has *two* specifiers then there'd still be an even number of `%`. We want to check for `%` not followed by another `%` right?

I recently did some similar code over here: https://github.com/containers/composefs/blob/45e6179993aa8a057d27946e26f5f7306b6a1779/rust/composefs/src/dumpfile.rs#L116

Actually even more strongly than that, we do want to *parse* the string here and an `%%` in the input should be evaluated to a single `%`, so it should be much more like an "unescape" operation than a validation.

lib/src/boundimage.rs
cgwalters: Should also use `&str` here

lib/src/boundimage.rs
```diff
@@ -0,0 +1,296 @@
...
+            anyhow::bail!("Invalid non-UTF8 filename: {file_name:?} in {}", spec_dir);
+        };
+
+        if !entry.file_type()?.is_symlink() {
+            anyhow::bail!("Not a symlink: {file_name}");
+        }
+
+        //parse the file contents
+        let path = Utf8Path::new(spec_dir).join(file_name);
+        let mut file: File = rustix::fs::openat2(
```
cgwalters, line 62: Side note, I think we can use https://github.com/coreos/cap-std-ext/pull/54 once that lands

lib/Cargo.toml
```diff
@@ -30,7 +30,7 @@ liboverdrop = "0.1.0"
 libsystemd = "0.7"
 openssl = "^0.10.64"
 # TODO drop this in favor of rustix
-nix = { version = "0.29", features = ["ioctl", "sched"] }
+nix = { version = "0.29", features = ["ioctl", "sched", "fs"] }
```
cgwalters, line 33: I think this change may have only been needed in an earlier PR version? I dropped it...as the TODO says ideally we port to rustix.

cgwalters (approved):
Some nits but I think we can get this in and gain experience and iterate from there!

cgwalters:
Ah hmm a few currently gating clippy lints:

```

Run cargo clippy -- -D clippy::correctness -D clippy::suspicious
   Compiling bootc-lib v0.1.13 (/__w/bootc/bootc/lib)
    Checking libtest-mimic v0.7.3
    Checking xtask v0.1.0 (/__w/bootc/bootc/xtask)
    Checking tests-integration v0.1.0 (/__w/bootc/bootc/tests-integration)
warning: constants have by default a `'static` lifetime
  --> lib/src/boundimage.rs:15:25
   |
15 | const BOUND_IMAGE_DIR: &'static str = "usr/lib/bootc-experimental/bound-images.d";
   |                        -^^^^^^^---- help: consider removing `'static`: `&str`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_static_lifetimes
   = note: `#[warn(clippy::redundant_static_lifetimes)]` on by default

warning: this boolean expression can be simplified
   --> lib/src/boundimage.rs:100:8
    |
100 |     if !auth_file.is_none() {
    |        ^^^^^^^^^^^^^^^^^^^^ help: try: `auth_file.is_some()`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#nonminimal_bool
    = note: `#[warn(clippy::nonminimal_bool)]` on by default

error: you are implementing `Ord` explicitly but have derived `PartialOrd`
   --> lib/src/boundimage.rs:152:1
    |
152 | / impl Ord for BoundImage {
153 | |     fn cmp(&self, other: &Self) -> std::cmp::Ordering {
154 | |         self.image.cmp(&other.image)
155 | |     }
156 | | }
    | |_^
    |
note: `PartialOrd` implemented here
   --> lib/src/boundimage.rs:134:25
    |
134 | #[derive(PartialEq, Eq, PartialOrd)]
    |                         ^^^^^^^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#derive_ord_xor_partial_ord
    = note: `-D clippy::derive-ord-xor-partial-ord` implied by `-D clippy::correctness`
    = help: to override `-D clippy::correctness` add `#[allow(clippy::derive_ord_xor_partial_ord)]`
    = note: this error originates in the derive macro `PartialOrd` (in Nightly builds, run with -Z macro-backtrace for more info)

warning: `bootc-lib` (lib) generated 2 warnings
error: could not compile `bootc-lib` (lib) due to 1 previous error; 2 warnings emitted
Error: Process completed with exit code 101.
```

I pushed fixes for those.

PR #660, opened 2024-07-01
copr: Also install git

cgwalters:
This whole "install all the build tools to make a source rpm" flow rather stinks...

One thing that would help is for me (or someone else) to package cargo-vendor-filterer at some point.

### DIFF
--- a/.copr/Makefile
+++ b/.copr/Makefile
@@ -1,5 +1,5 @@
 srpm:
-	dnf -y install dnf-utils
+	dnf -y install dnf-utils git
 	dnf -y builddep bootc
 	# similar to https://github.com/actions/checkout/issues/760, but for COPR
 	git config --global --add safe.directory '*'


vrothberg (approved):
LGTM

cgwalters:
OK well, that improved things but maddeningly enough it's [still broken](https://download.copr.fedorainfracloud.org/results/rhcontainerbot/bootc/srpm-builds/07698092/builder-live.log.gz) and the reason appears to be that the `make srpm` process runs in a f39 root. *Why* is it using f39? Well, there's all sorts of mock/copr magic going on there. The thing is, this would all be *so much more obvious* if building the SRPM just ran in a standard container image.

```

You can reproduce this build on your computer by running:

  sudo dnf install copr-rpmbuild
  /usr/bin/copr-rpmbuild --verbose --drop-resultdir --srpm --task-url https://copr.fedorainfracloud.org/backend/get-srpm-build-task/7698092

```

I mean yeah but, why can't I reproduce it with `podman run`?

OK anyways trying https://github.com/containers/bootc/pull/664

PR #661, opened 2024-07-01
status: Add --format-version

cgwalters:
In preparation for us changing the default output of `bootc status`
in the future (cc https://github.com/containers/bootc/issues/518 )
add `--format-version` that people can start using now to explicitly
request the current version.

It's possible that instead of a hard break we still support
outputting the current format for a while.

---

docs: Regenerate manpage markdown

---


vrothberg (approved):
LGTM

PR #663, opened 2024-07-02
lint: Verify we can parse the kargs.d files

cgwalters:
This verifies the syntax of all included `kargs.d` files. I hit this when I did a build in an image without `/usr/lib/bootc/kargs.d` and did a
`COPY foo.toml /usr/lib/bootc/kargs.d`
which created a regular file named `kargs.d` there.


vrothberg (approved):
LGTM

PR #664, opened 2024-07-02
copr: Use included spec

cgwalters:
For some reason the copr process uses a f39 root, and we don't have our zstd dependency there.

Hopefully this will the last fix...


jeckersb (approved):

cgwalters:
And it finally works https://copr.fedorainfracloud.org/coprs/rhcontainerbot/bootc/build/7699489/

PR #665, opened 2024-07-03
install: Use zipl to install bootloader on s390x

yoheiueda:
This is a part of s390x enablement https://github.com/containers/bootc/issues/569

Currently, the `zipl` command is called via libostree to install bootloader on s390x, but it does not work correctly on a loop device.

This patch changes `bootc` to execute the `zipl` command directly with appropriate command line options for installation on a loop device.

### DIFF
--- a/lib/src/bootloader.rs
+++ b/lib/src/bootloader.rs
@@ -1,4 +1,4 @@
-use anyhow::Result;
+use anyhow::{anyhow, bail, Context, Result};
 use camino::{Utf8Path, Utf8PathBuf};
 use fn_error_context::context;
 
@@ -50,4 +50,81 @@ pub(crate) fn install_via_bootupd(
         .args(args)
         .verbose()
         .run()
+}
+
+#[context("Installing bootloader using zipl")]
+pub(crate) fn install_via_zipl(device: &PartitionTable, boot_uuid: &str) -> Result<()> {
+    // Identify the target boot partition from UUID
+    let fs = crate::mount::inspect_filesystem_by_uuid(boot_uuid)?;
+    let boot_dir = Utf8Path::new(&fs.target);
+    let maj_min = fs.maj_min;
+
+    // Ensure that the found partition is a part of the target device
+    let device_path = device.path();
+
+    let partitions = crate::blockdev::list_dev(device_path)?
+        .children
+        .with_context(|| format!("no partition found on {device_path}"))?;
+    let boot_part = partitions
+        .iter()
+        .find(|part| part.maj_min.as_deref() == Some(maj_min.as_str()))
+        .with_context(|| format!("partition device {maj_min} is not on {device_path}"))?;
+    let boot_part_offset = boot_part.start.unwrap_or(0);
+
+    // Find exactly one BLS configuration under /boot/loader/entries
+    // TODO: utilize the BLS parser in ostree
+    let bls_dir = boot_dir.join("boot/loader/entries");
+    let bls_entry = bls_dir
+        .read_dir_utf8()?
+        .try_fold(None, |acc, e| -> Result<_> {
+            let e = e?;
+            let name = Utf8Path::new(e.file_name());
+            if let Some("conf") = name.extension() {
+                if acc.is_some() {
+                    bail!("more than one BLS configurations under {bls_dir}");
+                }
+                Ok(Some(e.path().to_owned()))
+            } else {
+                Ok(None)
+            }
+        })?
+        .with_context(|| format!("no BLS configuration under {bls_dir}"))?;
+
+    let bls_path = bls_dir.join(bls_entry);
+    let bls_conf =
+        std::fs::read_to_string(&bls_path).with_context(|| format!("reading {bls_path}"))?;
+
+    let mut kernel = None;
+    let mut initrd = None;
+    let mut options = None;
+
+    for line in bls_conf.lines() {
+        match line.split_once(char::is_whitespace) {
+            Some(("linux", val)) => kernel = Some(val.trim().trim_start_matches('/')),
+            Some(("initrd", val)) => initrd = Some(val.trim().trim_start_matches('/')),
+            Some(("options", val)) => options = Some(val.trim()),
+            _ => (),
+        }
+    }
+
+    let kernel = kernel.ok_or_else(|| anyhow!("missing 'linux' key in default BLS config"))?;
+    let initrd = initrd.ok_or_else(|| anyhow!("missing 'initrd' key in default BLS config"))?;
+    let options = options.ok_or_else(|| anyhow!("missing 'options' key in default BLS config"))?;
+
+    let image = boot_dir.join(kernel).canonicalize_utf8()?;
+    let ramdisk = boot_dir.join(initrd).canonicalize_utf8()?;
+
+    // Execute the zipl command to install bootloader
+    let zipl_desc = format!("running zipl to install bootloader on {device_path}");
+    let zipl_task = Task::new(&zipl_desc, "zipl")
+        .args(["--target", boot_dir.as_str()])
+        .args(["--image", image.as_str()])
+        .args(["--ramdisk", ramdisk.as_str()])
+        .args(["--parameters", options])
+        .args(["--targetbase", device_path.as_str()])
+        .args(["--targettype", "SCSI"])
+        .args(["--targetblocksize", "512"])
+        .args(["--targetoffset", &boot_part_offset.to_string()])
+        .args(["--add-files", "--verbose"]);
+    zipl_task.verbose().run().context(zipl_desc)
 }
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -1688,7 +1688,9 @@ fn test_gather_root_args() {
     // A basic filesystem using a UUID
     let inspect = Filesystem {
         source: "/dev/vda4".into(),
+        target: "/".into(),
         fstype: "xfs".into(),
+        maj_min: "252:4".into(),
         options: "rw".into(),
         uuid: Some("965eb3c7-5a3f-470d-aaa2-1bcf04334bc6".into()),
     };
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -573,15 +573,9 @@ async fn initialize_ostree_root_from_self(
         crate::lsm::ensure_dir_labeled(rootfs_dir, "boot", None, 0o755.into(), sepolicy)?;
     }
 
-    // Default to avoiding grub2-mkconfig etc., but we need to use zipl on s390x.
-    // TODO: Lower this logic into ostree proper.
-    let bootloader = if cfg!(target_arch = "s390x") {
-        "zipl"
-    } else {
-        "none"
-    };
     for (k, v) in [
-        ("sysroot.bootloader", bootloader),
+        // Default to avoiding grub2-mkconfig etc.
+        ("sysroot.bootloader", "none"),
         // Always flip this one on because we need to support alongside installs
         // to systems without a separate boot partition.
         ("sysroot.bootprefix", "true"),
@@ -1239,12 +1233,17 @@ async fn install_to_filesystem_impl(state: &State, rootfs: &mut RootSetup) -> Re
             })
             .context("Writing aleph version")?;
     }
+    if cfg!(target_arch = "s390x") {
+        // TODO: Integrate s390x support into install_via_bootupd
+        crate::bootloader::install_via_zipl(&rootfs.device_info, boot_uuid)?;
+    } else {
+        crate::bootloader::install_via_bootupd(
+            &rootfs.device_info,
+            &rootfs.rootfs,
+            &state.config_opts,
+        )?;
+    }
 
-    crate::bootloader::install_via_bootupd(
-        &rootfs.device_info,
-        &rootfs.rootfs,
-        &state.config_opts,
-    )?;
     tracing::debug!("Installed bootloader");
 
     // Finalize mounted filesystems
--- a/lib/src/mount.rs
+++ b/lib/src/mount.rs
@@ -12,6 +12,9 @@ use crate::task::Task;
 pub(crate) struct Filesystem {
     // Note if you add an entry to this list, you need to change the --output invocation below too
     pub(crate) source: String,
+    pub(crate) target: String,
+    #[serde(rename = "maj:min")]
+    pub(crate) maj_min: String,
     pub(crate) fstype: String,
     pub(crate) options: String,
     pub(crate) uuid: Option<String>,
@@ -29,7 +32,7 @@ fn run_findmnt(args: &[&str], path: &str) -> Result<Filesystem> {
             "-J",
             "-v",
             // If you change this you probably also want to change the Filesystem struct above
-            "--output=SOURCE,FSTYPE,OPTIONS,UUID",
+            "--output=SOURCE,TARGET,MAJ:MIN,FSTYPE,OPTIONS,UUID",
         ])
         .args(args)
         .arg(path)
@@ -47,6 +50,12 @@ fn run_findmnt(args: &[&str], path: &str) -> Result<Filesystem> {
 /// if the target is not the mount root.
 pub(crate) fn inspect_filesystem(path: &Utf8Path) -> Result<Filesystem> {
     run_findmnt(&["--mountpoint"], path.as_str())
+}
+
+#[context("Inspecting filesystem by UUID {uuid}")]
+/// Inspect a filesystem by partition UUID
+pub(crate) fn inspect_filesystem_by_uuid(uuid: &str) -> Result<Filesystem> {
+    run_findmnt(&["--source"], &(format!("UUID={uuid}")))
 }
 
 /// Mount a device to the target path.


lib/src/bootloader.rs
```diff
@@ -26,3 +26,113 @@ pub(crate) fn install_via_bootupd(
...
+        target: String,
+    }
+
+    #[derive(serde::Deserialize)]
+    struct Findmnt {
+        filesystems: Vec<Filesystem>,
+    }
+
+    let findmnt_desc = format!("Running findmnt to identify the boot partition with UUID={boot_uuid}");
+    let findmnt_task = Task::new(findmnt_desc, "findmnt");
```
cgwalters, line 50: Can you extend the existing `run_findmnt()` API we have internally for this?

lib/src/bootloader.rs
```diff
@@ -26,3 +26,113 @@ pub(crate) fn install_via_bootupd(
...
+                        "--uniq", "--first-only", "--nofsroot",
+                        "--source", &(format!("UUID={boot_uuid}"))];
+    let findmnt_out  = findmnt_task.args(findmnt_args).quiet().read()?;
+    let findmnt_json = serde_json::from_str::<Findmnt>(&findmnt_out).context("parsing findmnt output")?;
+
+    let boot_fs = findmnt_json.filesystems.get(0).context("finding boot filesystem")?;
+    let boot_dev = Utf8Path::new(&boot_fs.source);
+    let boot_dir = Utf8Path::new(&boot_fs.target);
+
+    // Run sfdisk to determine the boot partition offset
```
cgwalters, line 61: I think we can get this data from our existing `lsblk` wrapper in `blockdev.rs` probably.

lib/src/bootloader.rs
```diff
@@ -26,3 +26,113 @@ pub(crate) fn install_via_bootupd(
...
+    let sfdisk_json = serde_json::from_str::<Sfdisk>(&sfdisk_out).context("parsing findmnt output")?;
+
+    let block_size  = sfdisk_json.partitiontable.sectorsize;
+    let partitions  = sfdisk_json.partitiontable.partitions;
+    let boot_part   = partitions.iter().find(|p| p.node == boot_dev).context("finding boot partition")?;
+    let boot_offset = boot_part.start;
+
+    // Find the default BLS conf under /boot/loader/entries
+
+    let bls_dir = boot_dir.join("boot/loader/entries");
```
cgwalters, line 93: I think this code looks like it originally came from coreos-installer? This is fine for now, but it is a bit of a duplication as ostree already has a BLS parser too...and I think here because (unlike coreos-installer) we are hard linked to ostree today we could actually reference the `ostree::Deployment` object which has an associated boot loader config object we could use.

lib/src/bootloader.rs
```diff
@@ -26,3 +26,113 @@ pub(crate) fn install_via_bootupd(
...
+    let ls_desc = format!("running ls to find the default BLS configuration");
+    let ls_task = Task::new(ls_desc, "ls");
+    let ls_args = ["--sort=version", "--reverse", bls_dir.as_str()];
+    let ls_out = ls_task.args(ls_args).quiet().read()?;
+
+    let bls_name = ls_out.lines().find(|filename| filename.ends_with(".conf")).context("finding BLS conf")?;
+    let bls_path = bls_dir.join(bls_name);
+    let bls_conf = std::fs::read_to_string(&bls_path).context(format!("reading {bls_path:?}"))?;
+
+    let mut image = "";
```
cgwalters, line 104: I prefer the coreos-installer version of this where we use `Option<T>`: https://github.com/coreos/coreos-installer/blob/803b911dabac28b958a89a6b3ee01ce2df65f99a/src/s390x/zipl.rs#L202 versus representing "unset" as the empty string.

lib/src/bootloader.rs
```diff
@@ -26,3 +26,113 @@ pub(crate) fn install_via_bootupd(
...
+    for line in bls_conf.lines() {
+        match line.split_once(char::is_whitespace) {
+            Some(("linux",   val)) => image      = val.trim(),
+            Some(("initrd",  val)) => ramdisk    = val.trim(),
+            Some(("options", val)) => parameters = val.trim(),
+            _ => (),
+        }
+    };
+
+    let image_path   = boot_dir.join(image.trim_start_matches('/')).canonicalize_utf8()?;
```
cgwalters, line 117: Did something specifically motivate the `.canonicalize_utf8()` here out of curiosity?

lib/src/install.rs
```diff
@@ -1229,7 +1223,12 @@ async fn install_to_filesystem_impl(state: &State, rootfs: &mut RootSetup) -> Re
             .context("Writing aleph version")?;
     }
 
-    crate::bootloader::install_via_bootupd(&rootfs.device, &rootfs.rootfs, &state.config_opts)?;
+    if cfg!(target_arch = "s390x") {
+        // TODO: Integrate s390x support into install_via_bootupd
```
cgwalters, line 1227: Yeah...that would definitely be better in the longer term I think, but this is totally fine to do here for now. See also https://github.com/coreos/bootupd/issues/432

cgwalters (commented):
Thanks so much for working on this! I have some style/structural nits but I think we're heading in the right direction.

lib/src/bootloader.rs
yoheiueda: Sure. I'll create a new function in mount.rs.

lib/src/bootloader.rs
yoheiueda: It looks like we can use the function in blockdev.rs. I'll update the code.

cgwalters: We can probably change it to run `lsblk -O -J` to basically scrape all the data by default instead of the explicit current allowlist.

lib/src/bootloader.rs
yoheiueda: Actually, I referenced the logic in the zipl source code.
https://github.com/ibm-s390-linux/s390-tools/blob/b6c30404bcceca1f7328abab4df2be8535667564/zipl/src/scan.c#L1113

Due to my lack of knowledge about ostree, I'd like to use the Rust code here for now, and will add a TODO comment.

lib/src/bootloader.rs
yoheiueda: OK. I'll fix the code.

lib/src/bootloader.rs
yoheiueda: `canonicalize_utf8` is just a cosmetic, nice-to-have thing here.

When the target filesystem is `/boot`, `image_path` becomes something like `/run/bootc/mounts/rootfs/boot/boot/ostree/...`. `boot` directory appears twice in the path, and the second `boot` directory is actually a symlink to `.`.

This path representation is OK for zipl to work correctly, but looks a little bit strange. So I use `canonicalize_utf8` to remove the duplicated `boot` from the path.

cgwalters:
@yoheiueda let me know if you want me to take a pass at updating this PR

yoheiueda:
@cgwalters I am working on this. I will update the patch this week.

BTW, `lsblk` does not show an partition offset in a disk in my environment. Do you have any idea how to get partition offset info with `lsblk`?

cgwalters:
> BTW, lsblk does not show an partition offset in a disk in my environment. Do you have any idea how to get partition offset info with lsblk?

Just briefly looking at `lsblk -J -O /dev/vda` in a virt environment I see a `start` attribute; does that not exist for dasd?

yoheiueda:
lsblk from util-linux 2.37.4 (RHEL 9.4) does not have `start` attribute, but lsblk from util-linux 2.40.1 (Fedora 40) does have the attribute.

https://github.com/util-linux/util-linux/commit/856457bfcfaa64c90f02f250779a3f171535296f

It looks like the `start` attribute has been introduced from 2.38.

@cgwalters do you think it is safe to assume that the newer lsblk?  I think the commit needs to be backported to RHEL eventually if we use this command.

jlebon:
It might be in `sfdisk -J` since that one actually reads the GPT header.

You can also look at the `start` file in sysfs directly too.

cgwalters:
> @cgwalters do you think it is safe to assume that the newer lsblk? I think the commit needs to be backported to RHEL eventually if we use this command.

We need to support whatever is in C9S for code we land today as a baseline. So we could ask for that to be backported.

But, I think it's easiest for now to go with what jlebon suggested, and maybe also queue a backport request.

yoheiueda:
@cgwalters @jlebon thank you for your suggestions. I will use `/sys/dev/<maj:min>/start` to get a partition offset.

lib/src/bootloader.rs
```diff
@@ -26,3 +26,77 @@ pub(crate) fn install_via_bootupd(
...
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => Ok(0), // partition-less disk
+        Err(e) => Err(anyhow!(e)),
+    }
+    .with_context(|| format!("reading {sysfs_start_path:?}"))?;
+
+    // Find the default BLS conf under /boot/loader/entries
+    // TODO: utilize the BLS parser in ostree
+    let bls_dir = boot_dir.join("boot/loader/entries");
+    let ls_desc = format!("running ls to find the default BLS configuration under {bls_dir}");
+    let ls_task = Task::new(ls_desc, "ls").args(["--sort=version", "--reverse", bls_dir.as_str()]);
```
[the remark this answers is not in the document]
yoheiueda: The intention to use the `ls` command was to use [`strverscmp`](https://man7.org/linux/man-pages/man3/strverscmp.3.html) for filename comparison. But, I couldn't find `strverscmp` in Rust, so I used the `ls` command with the `--sort=version` option.

ostree and zipl use this function to find a default BLF configuration file.
https://github.com/ostreedev/ostree/blob/4b96359e106a832dacdec9f3ef19dc730783e7a3/src/libostree/ostree-sysroot.c#L585
https://github.com/ibm-s390-linux/s390-tools/blob/d7b3196286f78445ab7843ba000faf326a8673fd/zipl/src/scan.c#L742

It looks like we can call libc's strverscmp via the libc crate. I'll change the patch to call the libc function.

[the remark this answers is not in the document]
cgwalters: Ah yes I see.

> It looks like we can call libc's strverscmp via the libc crate. I'll change the patch to call the libc function.

I'm not opposed to that but sadly it'd require use of `unsafe{}` (and a conversion from Rust `String` or `&str` to `CString`).

As far as sorting these entries specifically...well, at *install time* we currently have a hard constraint that there is *exactly one* bootloader entry (and deployment), so we could just safely ignore the sorting problem and verify there is exactly one `.conf` file.

But if we want to handle the sorting *and* the version comparison...so this gets into a pretty embarrassing bug I hit recently, see https://github.com/ostreedev/ostree/issues/2961

(Note the finding that `zipl` does use the `version` field, but due to the grub incompatiblity we effectively need to sort both)

Anyways though again I think it's simplest to just require exactly one BLS entry for now.

lib/src/bootloader.rs
```diff
@@ -26,3 +26,94 @@ pub(crate) fn install_via_bootupd(
...
+    crate::blockdev::list_dev(device)?
+        .children
+        .with_context(|| format!("no partition found on {device:?}"))?
+        .iter()
+        .find(|dev| dev.maj_min == Some(maj_min.clone()))
+        .with_context(|| format!("partition device {:?} is not on {:?}", maj_min, device))?;
+
+    // Get an offset of the target partition by reading /sys/dev/block/<major:minor>/start
+    let sysfs_start_path = Utf8Path::new("/sys/dev/block").join(&maj_min).join("start");
+    let boot_offset = match std::fs::read_to_string(sysfs_start_path.clone()) {
```
cgwalters, line 47: Minor nit, this can just borrow via `&sysfs_start_path` instead of copying.

lib/src/bootloader.rs
```diff
@@ -26,3 +26,94 @@ pub(crate) fn install_via_bootupd(
...
+        Ok(v) => v.trim().parse::<i64>().map_err(|e| anyhow!(e)),
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => Ok(0), // partition-less disk
```
cgwalters, line 49: Fine as is, but - if you want (and again this is optional) we have a nice handy helper for this in https://docs.rs/cap-std-ext/latest/cap_std_ext/dirext/trait.CapStdExtDirExt.html#tymethod.open_optional

(Using this would require opening the rootfs as a `Dir` or accepting it as an option, see e.g. https://github.com/containers/bootc/pull/649/commits/ab1149d6cbaa94dbbed24b2747a10ae5d34b350b )

lib/src/bootloader.rs
```diff
@@ -26,3 +26,94 @@ pub(crate) fn install_via_bootupd(
...
+            }
+            Ok(acc)
+        })?;
+    let bls_name = entries
+        .iter()
+        .max_by(|a, b| {
+            // TODO: Remove this unsafe code when a Rust version of strverscmp is available
+            #[link(name = "c")]
+            extern "C" {
+                fn strverscmp(s1: *const libc::c_char, s2: *const libc::c_char) -> libc::c_int;
```
cgwalters, line 73: I thought this was a bit unusual and now I realize why we need to do it this way, it's because only GNU libc has `strverscmp` and Rust https://docs.rs/libc/latest/libc doesn't always expose those GNU-only APIs.

cgwalters (commented):
This is overall looking good!

lib/src/bootloader.rs
```diff
@@ -26,3 +26,94 @@ pub(crate) fn install_via_bootupd(
...
+    let fs = crate::mount::inspect_filesystem_by_uuid(boot_uuid)?;
+    let boot_dir = Utf8Path::new(&fs.target);
+    let maj_min = fs.maj_min;
+
+    // Ensure that the found partition is a part of the target device
+    crate::blockdev::list_dev(device)?
+        .children
+        .with_context(|| format!("no partition found on {device:?}"))?
+        .iter()
+        .find(|dev| dev.maj_min == Some(maj_min.clone()))
```
cgwalters, line 42: Minor nit, we're cloning the string for each iteration here. I think you may have done this because of a mismatch between `&String` and `String`? The best fix is to coerce both to `&str`:

`.find(|dev| dev.maj_min.as_deref() == Some(maj_min.as_str())`
I think will work.

lib/src/bootloader.rs
```diff
@@ -26,3 +26,94 @@ pub(crate) fn install_via_bootupd(
...
+        .with_context(|| format!("partition device {:?} is not on {:?}", maj_min, device))?;
```
cgwalters, line 43: Two minor nits:

- I don't believe we need the `Debug` format (`{:?}`) we can just use the `Display` format via `{}`
- In addition Rust has for quite a while supported "inline" format, like this: `format!("partition device {maj_min} is not on {device}")` which is a nice legibility improvement

lib/src/bootloader.rs
```diff
@@ -26,3 +26,94 @@ pub(crate) fn install_via_bootupd(
...
+
+    // Get an offset of the target partition by reading /sys/dev/block/<major:minor>/start
+    let sysfs_start_path = Utf8Path::new("/sys/dev/block").join(&maj_min).join("start");
+    let boot_offset = match std::fs::read_to_string(sysfs_start_path.clone()) {
+        Ok(v) => v.trim().parse::<i64>().map_err(|e| anyhow!(e)),
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => Ok(0), // partition-less disk
+        Err(e) => Err(anyhow!(e)),
+    }
+    .with_context(|| format!("reading {sysfs_start_path:?}"))?;
```
cgwalters, line 52: Also shouldn't need `:?` (here and a variety of other places)

Further explanatory info: Often in Rust codebases that are operating on `std::path::Path` which is *not* statically known to be UTF-8, you'll see use of `:?` because it forces conversion to a human-readable debug.

A big part of the value of using [camino](https://docs.rs/camino/latest/camino/) is in things like this - we can safely treat the pathnames as strings too where it's convenient.

lib/src/bootloader.rs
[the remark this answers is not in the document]
yoheiueda: OK. Thank you for the explanation.

Let's check exactly one `.conf` file here. It's simplest.

cgwalters:
xref https://github.com/containers/bootc/pull/667#issuecomment-2227005320

Specifically on this PR, we all the preparatory changes should be there now in the blockdev bits - do you want to rebase this one on https://github.com/containers/bootc/pull/680 ?

Note specifically for example https://github.com/containers/bootc/pull/680/commits/61feba48b956c43ad210a74c5acacc9d40cd1412 I think is cleaner.

yoheiueda:
@cgwalters thank you for the new PR. It looks good. 

I'll rebase this PR on #680.

lib/src/bootloader.rs
yoheiueda: done

lib/src/bootloader.rs
yoheiueda: done

lib/src/bootloader.rs
yoheiueda: Switch to use `backfill_start`

lib/src/bootloader.rs
yoheiueda: No longer use `strverscmp`

lib/src/bootloader.rs
yoheiueda: done

lib/src/bootloader.rs
yoheiueda: done

yoheiueda:
@cgwalters Ive rebased this PR on #688

cgwalters (approved):

cgwalters:
(I resolved the trivial conflict with https://github.com/containers/bootc/pull/667 )

PR #668, opened 2024-07-05
docs/filesystem: document state overlays

jlebon:
This works fine now and is another alternative to solving the `/opt` problem.


jlebon:
xref: https://github.com/ostreedev/ostree-rs-ext/issues/607#issuecomment-1973476187
xref: https://github.com/coreos/fedora-coreos-tracker/issues/1681#issuecomment-2211137520

cgwalters (approved):

PR #674, opened 2024-07-11
kargs: More cleanups

cgwalters:
deploy: Centralize karg computation

I happened to be looking at this code for an unrelated reason.
We had 3 copies of the function calling karg computation.
But that can just be pushed to a much lower level, right
before we tell ostree what to do.

---

kargs: Use merge deployment

This is in general for us right now a no-op, but is closer
to technically correct. In ostree the "merge deployment"
concept started out as "deployment we use for the original /etc"
which *may be different* from the booted deployment. For example,
when one wants to do a "factory reset", we want to start
from a fresh /etc.

The other important case here is when we're doing a fresh
install (or into a new stateroot) there may be no merge
deployment at all! It wouldn't be correct to look at `/`.

We only sidestep this issue right now because the install
logic bypasses this to directly gather kargs...and doesn't
use the same `deploy` API as inplace updates. But we
want to get closer to doing things that way.

---

kargs: Use type inference

No need for explicit types in most of these places.

---

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -541,7 +541,6 @@ async fn upgrade(opts: UpgradeOpts) -> Result<()> {
         }
     } else {
         let fetched = crate::deploy::pull(repo, imgref, opts.quiet).await?;
-        let kargs = crate::kargs::get_kargs(repo, &booted_deployment, fetched.as_ref())?;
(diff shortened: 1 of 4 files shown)


cgwalters:
Fixed textual conflict with https://github.com/containers/bootc/pull/675

jeckersb (approved):

PR #678, opened 2024-07-12
test: Error fix

henrywang:
This PR fixed two errors:

1. do not check cachedUpdate.imageDigest in rollback test. The cachedUpdate.imageDigest is always null. We have wget package checking for rollback successful or not.
2. do not delete image before copy it. Fix "ERROR Pushing image: Export: Copying image: No such image" in integration test.

### DIFF
--- a/tests/booted/002-test-image-pushpull-upgrade.nu
+++ b/tests/booted/002-test-image-pushpull-upgrade.nu
@@ -31,7 +31,6 @@ def initial_build [] {
     let td = mktemp -d
     cd $td
 
-    do --ignore-errors { podman image rm localhost/bootc o+e>| ignore }
     bootc image copy-to-storage
     let img = podman image inspect localhost/bootc | from json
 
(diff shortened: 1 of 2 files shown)


henrywang:
Hi @cgwalters, I can't figure out the root cause of integration test failure. I removed `do --ignore-errors { podman image rm localhost/bootc o+e>| ignore }`, but still has error. Could you please help on this error? Thanks!
```
Copying local image docker://localhost/bootc to containers-storage:localhost/bootc ...
ERROR Pushing image: Export: Copying image: No such image
```

cgwalters:
>  I removed do --ignore-errors { podman image rm localhost/bootc o+e>| ignore }, but still has error. 

Hmm...yeah, this is one of those downsides of nushell for sure, there are a few IMO obscure/weird syntaxes. (The other one I find strange is [error make](https://www.nushell.sh/book/control_flow.html#error-make))

But...this works for me? Hmm...will dig a bit

cgwalters (approved):

cgwalters:
> But...this works for me? Hmm...will dig a bit

I didn't get to the bottom of this, but if it somehow helps it's fine by me to just delete that line because we'll end up replacing the image anyways.

PR #680, opened 2024-07-13
blockdev+install updates

cgwalters:
This series is prep for 

- https://github.com/containers/bootc/pull/665
- https://github.com/containers/bootc/pull/667

---

blockdev: Get more properties

- Add `partlabel` in prep for ppc64le bootloader bits
- Add `maj_min` and `start` which we want for s390x
- Clearly move the filesystesm-related bits to a separate
  section

---

blockdev: Backfill `start` if missing

This way we continue to support the version of util-linux in C9S
today.

---

blockdev: Use `-b` for lsblk, hard require size property

I noticed that the JSON format inconsistently uses human-readable
sizes for some properties, but not others. `size` happens to
be one of the ones were it outputs human readable. But I think
we will often want to operate on the raw byte value on our
own, so let's tell lsblk to just give us raw numbers and
do formatting where needed.

While we're here, just hard require the `size` property.
I don't think any block devices can be missing this.

---

install: Gather blockdev info early in filesystem phase

The bootloader logic in general is going to need to
query the block layout. But especially for ppc64le and s390x
we'll need this data.

Gather it early on as global state so it's accessible
to the entire `install to-filesystem` phase. Note that
we shouldn't be mutating the blockdev setup in
`to-filesystem`.

---

### DIFF
--- a/lib/src/blockdev.rs
+++ b/lib/src/blockdev.rs
@@ -29,7 +29,7 @@ pub(crate) struct Device {
     pub(crate) model: Option<String>,
     pub(crate) partlabel: Option<String>,
     pub(crate) children: Option<Vec<Device>>,
-    pub(crate) size: Option<String>,
+    pub(crate) size: u64,
     #[serde(rename = "maj:min")]
     pub(crate) maj_min: Option<String>,
     // NOTE this one is not available on older util-linux, and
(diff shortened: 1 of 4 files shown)


jmarrero (approved):
lgtm

PR #684, opened 2024-07-15
build-sys: Add `make validate-rust`

cgwalters:
The `make validate-rust` reproduces locally the same checks we run in CI as gating by default; hook it up to the existing `make validate`.


prestist (approved):
lgtm

PR #686, opened 2024-07-16
packit: Adjust arch targets; build on s390x

cgwalters:
- Add s390x because it's a target too
- Drop other arches for non-primary targets just because the chance they break is basically nil


jeckersb (approved):
LGTM

PR #687, opened 2024-07-16
boundimage: Replace %% with % in bound image spec

ckyrouac:
The previous iteration would detect %% in a bound image spec field but it would not transform them into %. This is the only systemd specifier that is allowed in the bound image specs, so they need to be transformed into their parsed value.

### DIFF
--- a/lib/src/boundimage.rs
+++ b/lib/src/boundimage.rs
@@ -139,29 +139,37 @@ struct BoundImage {
 
 impl BoundImage {
     fn new(image: String, auth_file: Option<String>) -> Result<BoundImage> {
-        validate_spec_value(&image).context("Invalid image value")?;
+        let image = parse_spec_value(&image).context("Invalid image value")?;
 
-        if let Some(auth_file) = &auth_file {
-            validate_spec_value(auth_file).context("Invalid auth_file value")?;
-        }
+        let auth_file = if let Some(auth_file) = &auth_file {
+            Some(parse_spec_value(auth_file).context("Invalid auth_file value")?)
+        } else {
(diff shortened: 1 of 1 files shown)


lib/src/boundimage.rs
```diff
@@ -139,29 +139,36 @@ struct BoundImage {
...
 
         Ok(BoundImage { image, auth_file })
     }
 }
 
-fn validate_spec_value(value: &String) -> Result<()> {
-    let mut number_of_percents = 0;
+fn parse_spec_value(value: &String) -> Result<String> {
+    let mut parsed_value = String::new();
+    let mut prev_is_percent = false;
```
cgwalters, line 156: I think this is almost right - if I'm thinking through this correctly there's just a corner case where if the string ends in a trailing `%` I believe we will silently swallow it and not error out.

How about this code:

```
fn parse_spec_value(value: &str) -> Result<String> {
    let mut it = s.chars();
    let mut ret = String::new();
    while let Some(c) = it.next() {
        if c != '%' {
            ret.push(c);
            continue;
        }
        let c = it.next().ok_or_else(|| anyhow!("Unterminated %"))?;
        match c {
            '%' => { ret.push('%'); },
            _ => anyhow::bail!("Systemd specifiers are not supported by bound bootc images: {value}");
       }
    Ok(r)
}
```
or so?

ckyrouac: ah thanks! I was struggling with iterators but this clears it up. I was missing that it.next() within the while will advance the iterator.

cgwalters (approved):

PR #688, opened 2024-07-16
install: Use sfdisk, not lsblk

cgwalters:
This works around an issue where `lsblk` ends up parsing data cached from udev in `/run/udev`, but we don't have that mounted by default.

Adding a new mount point hard requirement is logistically complicated right now - we will eventually switch
to dynamic mounts with `open_tree` (cc
https://github.com/containers/bootc/issues/380 )
but this is a relatively straightforward workaround.

### DIFF
--- a/lib/src/blockdev.rs
+++ b/lib/src/blockdev.rs
@@ -124,6 +124,58 @@ pub(crate) fn list() -> Result<Vec<Device>> {
     list_impl(None)
 }
 
+#[derive(Debug, Deserialize)]
+struct SfDiskOutput {
+    partitiontable: PartitionTable,
+}
+
+#[derive(Debug, Deserialize)]
+#[allow(dead_code)]
+pub(crate) struct Partition {
+    pub(crate) node: String,
+    pub(crate) start: u64,
+    pub(crate) size: u64,
+    #[serde(rename = "type")]
+    pub(crate) parttype: String,
+    pub(crate) uuid: Option<String>,
+    pub(crate) name: Option<String>,
+}
+
+#[derive(Debug, Deserialize)]
+#[allow(dead_code)]
+pub(crate) struct PartitionTable {
+    pub(crate) label: String,
+    pub(crate) id: String,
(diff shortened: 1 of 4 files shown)


cgwalters:
```
 > bootupctl backend install --write-uuid --device /dev/loop0 /run/bootc/mounts/rootfs
cp: error writing '/run/bootc/mounts/rootfs/boot/efi/EFI/centos/shimx64-centos.efi': No space left on device
cp: error writing '/run/bootc/mounts/rootfs/boot/efi/EFI/centos/shimx64.efi': No space left on device
cp: error writing '/run/bootc/mounts/rootfs/boot/efi/EFI/centos/mmx64.efi': No space left on device
cp: error writing '/run/bootc/mounts/rootfs/boot/efi/EFI/centos/grubx64.efi': No space left on device
cp: error writing '/run/bootc/mounts/rootfs/boot/efi/EFI/BOOT/BOOTX64.EFI': No space left on device
cp: error writing '/run/bootc/mounts/rootfs/boot/efi/EFI/BOOT/fbx64.efi': No space left on device
error: boot data installation failed: installing component EFI: Failed to copy
ERROR Installing to disk: Installing bootloader: Task Running bootupctl to install bootloader failed: ExitStatus(unix_wait_status(256))
```

Hmm that's a new one. It's a bit baffling, looking at the space on the host before:

```
+ df -h /
Filesystem      Size  Used Avail Use% Mounted on
/dev/root        72G   34G   38G  48% /
```
And the rootfs size is fine:
```
 Creating root filesystem (xfs) on device /dev/loop0p3 (size=9.5 GB)
```
And the container image size is fine:
```
 layers already present: 0; layers needed: 74 (1.7 GB)
```

jlebon (approved):
Looks sane to me.

(https://github.com/util-linux/util-linux/pull/3099 seems related here; --> https://github.com/util-linux/util-linux/issues/2047 --> https://github.com/coreos/fedora-coreos-config/pull/2181#issuecomment-1398780538).

PR #689, opened 2024-07-16
boundimage: Use new RootDir API

cgwalters:
- This is an abstraction we want to use in multiple places
- It has shared convenience helpers like `read_to_string` which reduce code here


cgwalters:
This uses https://github.com/coreos/cap-std-ext/pull/54

ckyrouac (approved):

PR #695, opened 2024-07-16
WIP prep for multiple backends

jeckersb:


jeckersb:
Rough draft of how I envision this might work.  One notable thing is that this version does not use async at all, however I expect that will probably become necessary when we add future backends (it started to creep in over in #215).  Things were starting to get unwieldy here with async because of things like `Deployment` being both [`!Send`](https://docs.rs/ostree/0.19.1/ostree/struct.Deployment.html#impl-Send-for-Deployment) and [`!Sync`](https://docs.rs/ostree/0.19.1/ostree/struct.Deployment.html#impl-Sync-for-Deployment) so it'd require a bit more iterating on to get right.  But the existing backend doesn't currently need it so this works for a minimal example.

lib/src/backend/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+    fn backend<'a>(&self) -> Result<Box<dyn BackendImpl>> {
+        if let Some(origin) = self.origin().as_ref() {
+            origin.backend()
+        } else {
+            Err(anyhow!("Deployment has no origin"))
+        }
+    }
+}
+
+impl Backend for &glib::KeyFile {
```
cgwalters, line 47: This part feels weird to me...why the indirection? Why not just inline this in the `impl Backend for ostree::Deployment`?

lib/src/backend/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+pub(crate) trait BackendImpl {
+    fn imagestatus(
+        &self,
+        sysroot: &SysrootLock,
+        deployment: &ostree::Deployment,
+        image: OstreeImageReference,
+    ) -> Result<(Option<ImageStatus>, Option<ImageStatus>)>;
+}
+
+impl Backend for crate::spec::Backend {
```
cgwalters, line 27: The duplication of `Backend` here feels odd...WDYT about having something like:

```
pub(crate) fn load(ty: crate::spec::Backend) -> Result<Box<dyn BackendImpl>> {
  match ty {
          crate::spec::Backend::OstreeContainer => {
                Ok(Box::new(ostree_container::OstreeContainerBackend))
            }
  }
}
```

or so?

IOW it's only `Deployment` that has an `impl Backend`?

lib/src/backend/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+    fn backend(&self) -> Result<Box<dyn BackendImpl>>;
+}
+
+pub(crate) trait BackendImpl {
+    fn imagestatus(
+        &self,
+        sysroot: &SysrootLock,
+        deployment: &ostree::Deployment,
+        image: OstreeImageReference,
+    ) -> Result<(Option<ImageStatus>, Option<ImageStatus>)>;
```
cgwalters, line 24: Not a big deal but we probably want a struct here for the return value, just looking at it it's not clear that one is the current and one is the cached status. (Can we directly return the spec value?)

jeckersb: I've added a `backend::CachedImageStatus` struct to wrap this.  Re: spec value, I didn't want to directly return a full `BootEntry` since it contains ostree ("frontend"?) specific things that aren't relevant to the backend.  Felt like a clean break that the backend only cares about the image/cached_update, and the ostree parts are shared for `incompatible`, `pinned`, and `ostree`.

lib/src/backend/mod.rs
cgwalters: > the only thing we're going to have to guide which backend to use will be whatever is provided via the spec::Backend

Right...though we could "resolve" that early and pass down the backend as an argument into lower layers.

Here's one way I'm thinking about it, we could have an abstraction over `ostree::Sysroot` - like `bootc::Storage` that has an `Option<bootc:ContainerStorage>` or something, and if that's not-none then we use it.

jeckersb:
Notable changes since last version:

- General renaming of things from "backend" to "store" or some variation thereof, mostly to align with `ostree_ext::container::store`.  Plus I think "image store" is more reflective of what this actually is.
- Added store::Storage struct which wraps `SysrootLock` and a `ContainerImageStoreImpl` re: [previous feedback](https://github.com/containers/bootc/pull/695#discussion_r1684666213).  This also is `Deref` for `SysrootLock` so we don't have to change that absolutely everywhere.
- `spec::BootEntry` now has a `store` field so we can expose it in the host status
- Added `store::load` re: [previous feedback](https://github.com/containers/bootc/pull/695#discussion_r1680067878) and removed store implementations for things other than `Deployment`.  Folded the `origin` bits under `Deployment`.

lib/src/store/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+    type Target = SysrootLock;
+
+    fn deref(&self) -> &Self::Target {
+        &self.sysroot
+    }
+}
+
+impl Storage {
+    pub fn new(sysroot: SysrootLock) -> Self {
+        let store = match env::var("BOOTC_STORAGE") {
```
jeckersb, line 52: oh I also added this bit to set the storage via env var (for now, at least), however this appears to not work as-is.  Seems the `from_str` below always fails so it's always defaulting.

cgwalters: Hmm, sanity check the `from_str` in a unit test?

cgwalters (approved):
Overall looks sane to me!

lib/src/store/mod.rs
jeckersb: Hm seems to be because it [defaults to kebab-case](https://docs.rs/clap/latest/clap/_derive/index.html#valueenum-attributes), but I tried adding `#[value(rename_all = "camelCase")]` and it looks like it's still using kebab-case.  I think maybe it's conflicting with `#[serde(rename_all = "camelCase")]`?  Really pushing the limits of my knowledge with attribute macro magic.

cgwalters: Ah yes confusing...we have both clap and serde in play here, and they operate in distinct worlds. Calling https://docs.rs/clap/latest/clap/trait.ValueEnum.html#method.from_str uses clap...so it's going to ignore whatever is configured for serde.

This has come up before, as some of the things we expose in the CLI are also serializable. 

This is a bit messy, as I personally tend to strongly prefer kebab-case for CLI things...but Kube-style APIs emphasize camelCase.

We could ensure no one wins by using `SCREAMING_SNAKE_CASE` (just kidding :smile: ))

Maybe we can add a clap alias for the camelCase names?

cgwalters: Another alternative is to avoid multi-word backend names, say e.g.

```
enum Backend {
 Ostree,
 Unified
}
``` 
that will come out to the same thing in both camelCase and kebab-case? as `ostree` and `unified`?

jeckersb: > Maybe we can add a clap alias for the camelCase names?

I think I like this one the best.

I must have been picking up the wrong image yesterday when I was testing this because [stripping it down in playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ffb4ff47e5f83d0cf8a40ad98664d11b), everything works fine.  That will match any case-insensive kebab or non-kebab option.

jeckersb:
Ok, rebased, added the alias, and added a test.

PR #698, opened 2024-07-17
 install/baseline: use sfdisk, not lsblk

cgwalters:
I think this may also help with the install flake we've been
seeing where we somehow get the partitions mixed up.

- Use `sfdisk` in the baseline install because it's what
  we're switching to using anyways later on
- Add some verification at least for the root and EFI partitions
  that we read back the expected type

### DIFF
--- a/lib/src/blockdev.rs
+++ b/lib/src/blockdev.rs
@@ -165,6 +165,21 @@ impl PartitionTable {
     pub(crate) fn path(&self) -> &Utf8Path {
         self.device.as_str().into()
     }
+
+    // Find the partition with the given offset (starting at 1)
+    pub(crate) fn find_partno(&self, partno: u32) -> Result<&Partition> {
+        let r = self
+            .partitions
+            .get(partno.checked_sub(1).expect("1 based partition offset") as usize)
+            .ok_or_else(|| anyhow::anyhow!("Missing partition for index {partno}"))?;
+        Ok(r)
+    }
+}
+
+impl Partition {
+    pub(crate) fn path(&self) -> &Utf8Path {
+        self.node.as_str().into()
+    }
 }
 
 #[context("Listing partitions of {dev}")]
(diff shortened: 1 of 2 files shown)


lib/src/blockdev.rs
```diff
@@ -165,6 +165,21 @@ impl PartitionTable {
     pub(crate) fn path(&self) -> &Utf8Path {
         self.device.as_str().into()
     }
+
+    // Find the partition with the given offset (starting at 1)
+    pub(crate) fn find_partno<'a>(&'a self, partno: u32) -> Result<&'a Partition> {
```
jeckersb, line 170: Explicit `'a` lifetime can be removed entirely here and elided instead

lib/src/install/baseline.rs
```diff
@@ -381,12 +379,15 @@ pub(crate) fn install_create_rootfs(
 
     // Initialize the /boot filesystem
     let bootdev = if let Some(bootpn) = boot_partno {
-        Some(findpart(bootpn)?)
+        Some(base_partitions.find_partno(bootpn)?)
     } else {
         None
     };
     let boot_uuid = if let Some(bootdev) = bootdev.as_deref() {
```
jeckersb, line 386: Remove `as_deref()`; switching from `findpart` to `find_partno` above changes `bootdev` from `Option<String>` to `Option<&Partition>` so it's redundant.

lib/src/install/baseline.rs
```diff
@@ -417,22 +418,22 @@ pub(crate) fn install_create_rootfs(
     // Create the underlying mount point directory, which should be labeled
     crate::lsm::ensure_dir_labeled(&target_rootfs, "boot", None, 0o755.into(), sepolicy)?;
     if let Some(bootdev) = bootdev.as_deref() {
```
jeckersb, line 420: Same as above here re: remove `as_deref()`

jeckersb (commented):
LGTM, just a few nits that clippy pointed out

jeckersb:
Hm not sure how i pushed overtop of you with `--force-with-lease` :cry:

jeckersb (approved):

PR #700, opened 2024-07-17
build: Use workspace global ints

cgwalters:
In relatively recent rust there's a nice way to globally configure clippy lints for the whole workspace. We can kill the `custom-lints` target because relatively
recently clippy has a lint for `todo!` and `dbg!` itself.


jeckersb (approved):
Nice!

PR #704, opened 2024-07-17
install: Fork `blockdev --rereadpt` instead of internal ioctl

cgwalters:
The `ioctl` to reread the partition table was the last thing pulling in `nix`. Honestly especially in the install path I have no problem forking external binaries, and since we're already depending on util-linux-core for other things like `sfdisk` and `lsblk`, it just makes sense to so here with the command whose entire role in life is to just issue that `ioctl`.

This drops out an older verison of `nix` (which is a fine crate, but again I think `rustix` is nicer). There's a different version pulled in via libsystemd, but we can get to that later.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -183,7 +183,6 @@ dependencies = [
  "libc",
  "liboverdrop",
  "libsystemd",
- "nix 0.29.0",
  "openssl",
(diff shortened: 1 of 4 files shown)


jeckersb (approved):
:+1:

PR #706, opened 2024-07-18
docs: Add experimental section for bound images

ckyrouac:


docs/src/experimental-logically-bound-images.md
```diff
@@ -0,0 +1,35 @@
...
+
+RUN ln -s /usr/share/containers/systemd/my-app.image /usr/lib/bootc-experimental/bound-images.d/my-app.image && \
+    ln -s /usr/share/containers/systemd/my-app.image /usr/lib/bootc-experimental/bound-images.d/my-app.image
+```
+
+## Limitations
+
+- Currently, only the Image field of a `.image` or `.container` file is used to pull the image. Any other field is ignored.
+- There is no cleanup during rollback.
+- Images can be deleted by manually running `podman rmi <image>`.
```
cgwalters, line 35: I'd generalize this slightly to say something like: 
Images are subject to default garbage collection semantics; e.g. a background job pruning images without a running container may prune them. They can also be manually removed via e.g. `podman rmi`.

Because in the case of kubelet, it *will* prune these images afaik by default.

cgwalters (approved):
Nice!

cgwalters (approved):

PR #707, opened 2024-07-18
install: Two prep changes for pulling bound images at install time

cgwalters:
Code similar to this was part of the original https://github.com/containers/bootc/pull/559 - this is prep for pulling bound images at install time

---

install: Split internal main phase into two

Split up:

- Initialize empty sysroot
- Install container to sysroot

This makes it easier to do more things that still
reference the sysroot.

(The git diff for this change is...really nice IMO,
 very straightforward)

---

install: Change helper to return deployment too

Prep for further changes.

---

### DIFF
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -542,15 +542,9 @@ pub(crate) fn print_configuration() -> Result<()> {
 }
 
 #[context("Creating ostree deployment")]
-async fn initialize_ostree_root_from_self(
-    state: &State,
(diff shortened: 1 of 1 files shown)


cgwalters:
```
mount: /run/bootc/mounts/rootfs: mount point is busy.
ERROR Installing to disk: Task Finalizing filesystem rootfs failed: ExitStatus(unix_wait_status(8192))
```

Huh, hadn't seen that one before...oh oh wait, I know.

PR #709, opened 2024-07-18
install: Warn if we're not installing to gpt

cgwalters:
I am not aware of any reason to use anything else; in theory we can leave this up to the OS, but...I think using `dos` format is really probably just a legacy mistake now.

And if we see anything else...well, let's at least warn.

### DIFF
--- a/lib/src/blockdev.rs
+++ b/lib/src/blockdev.rs
@@ -138,10 +138,18 @@ pub(crate) struct Partition {
     pub(crate) name: Option<String>,
 }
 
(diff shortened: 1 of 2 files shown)


cgwalters:
This correctly shows up in the [aarch64 integration test](https://github.com/containers/bootc/pull/709/checks?check_run_id=27623947656) because of https://github.com/osbuild/bootc-image-builder/issues/551

> Installing to `dos` format partitions is not recommended

cgwalters:
Rebased :surfer:

jeckersb (approved):
:+1:

PR #712, opened 2024-07-18
Bound install time

cgwalters:
utils: Drop unnecessary lint allow

This function doesn't use unsafe anymore.

---

boundimage: Use high level `deployment_fd` helper

This keeps things even simpler, it's the same thing we're doing
in `kargs.rs`.

---

boundimage: More struct definition to top

I think it's generally best to have type definitions come before
the code that references them.

---

boundimage: Drop const parameter

We were always passing this single constant value to the
function; let's just reference it inside the function
and simplify callers.

---

boundimage: Rename helper function

I think this name is a bit clearer.

---

boundimage: Make helpers `pub(crate)`

Prep for using them in the install path.

---

install: Pull bound images by default

Unless overridden by a CLI option, in which case they will
likely be pulled on firstboot.

---

### DIFF
--- a/lib/src/boundimage.rs
+++ b/lib/src/boundimage.rs
@@ -18,6 +18,17 @@ use ostree_ext::sysroot::SysrootLock;
 /// symbolic links to `.container` or `.image` files.
 const BOUND_IMAGE_DIR: &str = "usr/lib/bootc-experimental/bound-images.d";
 
+/// A subset of data parsed from a `.image` or `.container` file with
+/// the minimal information necessary to fetch the image.
+///
+/// In the future this may be extended to include e.g. certificates or
+/// other pull options.
+#[derive(PartialEq, Eq)]
+struct BoundImage {
+    image: String,
+    auth_file: Option<String>,
+}
+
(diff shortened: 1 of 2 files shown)


cgwalters:
OK, this works for me to pull bound images at install time by default.

However:

- I'm not happy with the `/var` situation (xref https://gitlab.com/fedora/bootc/base-images/-/issues/18 too); this also argues somewhat that bound images should be owned in the "bootc storage" as an additional store (also xref https://github.com/containers/bootc/issues/128#issuecomment-2229072527 )
- I think what we're very likely to need in the future is to support a model where we also look for bound images in `/var/lib/containers` in the installation environment, and pull from there if they exist. This would dramatically simplify both local development *and* support hermetic-style builds better
- Needs tests (but so does bound images in general)

ckyrouac (approved):
lgtm! Validated it works locally via podman-bootc.

cgwalters:
OK, rebased :surfer: on git main and lifting draft. I also added a `hide = true` to the install option as bound images are still classified as experimental.

PR #718, opened 2024-07-23
install: Force mkfs.xfs to overwrite

sacsant:
Reusing a disk (after manually deleting partitions on it) with xfs
as file system, install to-disk command fails complaning about
existing file system on root partition.

Creating root filesystem (xfs) on device /dev/sdb2 (size=107.4 GB)
> mkfs.xfs -m uuid=4826aa5b-0dd0-4382-a0c1-7c0b77cfd47d -L root /dev/sdb2
mkfs.xfs: /dev/sdb2 appears to contain an existing filesystem (xfs). 
mkfs.xfs: Use the -f option to force overwrite.

Use the -f option with mkfs.xfs to force overwrite.

### DIFF
--- a/lib/src/install/baseline.rs
+++ b/lib/src/install/baseline.rs
@@ -124,6 +124,7 @@ fn mkfs<'a>(
     dev: &str,
     fs: Filesystem,
     label: &str,
+    wipe: bool,
     opts: impl IntoIterator<Item = &'a str>,
 ) -> Result<uuid::Uuid> {
     let devinfo = crate::blockdev::list_dev(dev.into())?;
@@ -135,6 +136,9 @@ fn mkfs<'a>(
     );
     match fs {
         Filesystem::Xfs => {
+            if wipe {
+                t.cmd.arg("-f");
+            }
             t.cmd.arg("-m");
             t.cmd.arg(format!("uuid={u}"));
         }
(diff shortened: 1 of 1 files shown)


cgwalters:
Hm, you got this even when specifying `--wipe`? Or did you run wipefs manually?

cgwalters:
>  (after manually deleting partitions on it) 

Ah yes you did say that. Well...we have `to-disk --wipe` for this reason which note explicitly clears each child device too. (I don't know why `wipefs` doesn't have `--recursive` or so...probably wouldn't be hard to add)

How about: Only if `--wipe` is passed then we also pass `-f` to the mkfs invocations?

cgwalters:
There's a flip side to this in that this "existing partitions" check is the only thing right now that stops one from actively destroying e.g. the booted disk if `--wipe` is used: we definitely need to strengthen the `to-disk` path to bail hard if the device is mounted for example, even if `--wipe` is specified.

(xref https://lwn.net/Articles/941764/ )

cgwalters (approved):
Great, thanks!

cgwalters:
(This was missing a `cargo fmt`; in this repo remember to run `make validate` before pushing)

sacsant:
> (This was missing a `cargo fmt`; in this repo remember to run `make validate` before pushing)

Thanks. Will take care of this requirement for future PRs.

PR #703, opened 2024-07-17
Add JSON schema and guidance on API usage to docs

cgwalters:
- Add a hidden command to generate JSON schema
- Wire it up to the new `make update-generated`
- Describe how to use it in the docs

PR #722, opened 2024-07-24
build: move needless_borrow lint allows to be global

jeckersb:
A new instance of this snuck in under xtask, this will make sure it's
covered everywhere going forward.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,3 +58,7 @@ unused_must_use = "forbid"
 # These should only be in local code
 dbg_macro = "deny"
 todo = "deny"
+# These two are in my experience the lints which are most likely
+# to trigger, and among the least valuable to fix.
+needless_borrow = "allow"
+needless_borrows_for_generic_args = "allow"
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -8,10 +8,6 @@
 #![deny(missing_docs)]
 #![deny(missing_debug_implementations)]
 #![cfg_attr(feature = "dox", feature(doc_cfg))]
-// These two are in my experience the lints which are most likely
-// to trigger, and among the least valuable to fix.
-#![allow(clippy::needless_borrow)]
-#![allow(clippy::needless_borrows_for_generic_args)]
 
 mod boundimage;
 pub mod cli;
--- a/tests-integration/src/tests-integration.rs
+++ b/tests-integration/src/tests-integration.rs
@@ -1,6 +1,3 @@
-#![allow(clippy::needless_borrow)]
-#![allow(clippy::needless_borrows_for_generic_args)]
-
 use std::path::PathBuf;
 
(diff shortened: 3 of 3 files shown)


cgwalters (approved):
I'm OK with this but can you explain why you're adding this? 

Isn't this covered by the usage of global lints in `Cargo.toml` now?

Is this prep for another PR?

jeckersb:
> I'm OK with this but can you explain why you're adding this?
> 
> Isn't this covered by the usage of global lints in `Cargo.toml` now?
> 
> Is this prep for another PR?

Hm no it looks like the global lints aren't covering it.  I just noticed that clippy started complaining after https://github.com/containers/bootc/pull/703/files#diff-1b2f83403e0dae2d6172ad76c50e66039ec03402104f90900a29203790d21051R144

jeckersb:
Looks like I can fix it globally by just adding it to the root Cargo.toml so I'll change this to do that instead.

cgwalters:
I'm not getting any errors from `make validate` on current git main though. Oh...is this rustc version drift? My toolchain refresh cycle is a bit slow, was on 1.77. Just updated to 1.79, but I still don't see any clippy warnings on current git main.

jeckersb:
> I'm not getting any errors from `make validate` on current git main though. Oh...is this rustc version drift? My toolchain refresh cycle is a bit slow, was on 1.77. Just updated to 1.79, but I still don't see any clippy warnings on current git main.

With stable/1.79 I get:

```
$ cargo +stable clippy
   Compiling bootc-lib v0.1.13 (/var/home/jeckersb/git/bootc/lib)
    Checking tests-integration v0.1.0 (/var/home/jeckersb/git/bootc/tests-integration)
    Checking xtask v0.1.0 (/var/home/jeckersb/git/bootc/xtask)
warning: the borrowed expression implements the required traits
   --> xtask/src/xtask.rs:144:28
    |
144 |     std::fs::write(target, &schema)?;
    |                            ^^^^^^^ help: change this to: `schema`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrows_for_generic_args
    = note: `#[warn(clippy::needless_borrows_for_generic_args)]` on by default

warning: `xtask` (bin "xtask") generated 1 warning (run `cargo clippy --fix --bin "xtask"` to apply 1 suggestion)
    Checking bootc v0.1.9 (/var/home/jeckersb/git/bootc/cli)
    Finished `dev` profile [optimized + debuginfo] target(s) in 3.42s
```

cgwalters:
I'm a bit confused as I don't get that with the same rust version; but it's OK no need to dig in now...

jeckersb:
> I'm a bit confused as I don't get that with the same rust version; but it's OK no need to dig in now...

Was/is your local main branch behind upstream?  #703 only went in last night.

PR #724, opened 2024-07-24
Introduce bootc-owned container store, use for bound images

cgwalters:
Closes: https://github.com/containers/bootc/issues/721

- Initialize a containers-storage: instance at install time
  (that defaults to empty)
- Open it at the same time we open the ostree repo/sysroot
- Change bound images to use this

We are *NOT* yet changing the base bootc image pull to use this.
That's an obvious next step (xref
https://github.com/containers/bootc/pull/215 ) but will come later.

### DIFF
--- a/docs/src/experimental-logically-bound-images.md
+++ b/docs/src/experimental-logically-bound-images.md
@@ -14,7 +14,11 @@ This experimental feature enables an association of container "app" images to a 
 
 ## Using logically bound images
 
-Each image is defined in a [Podman Quadlet](https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html) `.image` or `.container` file. An image is selected to be bound by creating a symlink in the `/usr/lib/bootc/bound-images.d` directory pointing to a `.image` or `.container` file. With these defined, during a `bootc upgrade` or `bootc switch` the bound images defined in the new bootc image will be automatically pulled via podman.
+Each image is defined in a [Podman Quadlet](https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html) `.image` or `.container` file. An image is selected to be bound by creating a symlink in the `/usr/lib/bootc/bound-images.d` directory pointing to a `.image` or `.container` file. 
+
+With these defined, during a `bootc upgrade` or `bootc switch` the bound images defined in the new bootc image will be automatically pulled into the bootc image storage, and are available to container runtimes such as podman by explicitly configuring them to point to the bootc storage as an "additional image store", via e.g.:
+
+`podman --storage-opt=additionalimagestore=/usr/lib/bootc/storage run <image> ...`
 
 An example Containerfile
 
@@ -28,8 +32,17 @@ RUN ln -s /usr/share/containers/systemd/my-app.image /usr/lib/bootc/bound-images
     ln -s /usr/share/containers/systemd/my-app.image /usr/lib/bootc/bound-images.d/my-app.image
 ```
 
+In the `.container` definition, you should use:
+
+```
+GlobalArgs=--storage-opt=additionalimagestore=/usr/lib/bootc/storage
+```
+
+## Pull secret
+
+Images are fetched using the global bootc pull secret by default (`/etc/ostree/auth.json`). It is not yet supported to configure `PullSecret` in these image definitions.
+
 ## Limitations
 
-- Currently, only the Image field of a `.image` or `.container` file is used to pull the image. Any other field is ignored.
-- There is no cleanup during rollback.
-- Images are subject to default garbage collection semantics; e.g. a background job pruning images without a running container may prune them. They can also be manually removed via e.g. podman rmi.
+- Currently, only the Image field of a `.image` or `.container` file is used to pull the image; per above not even `PullSecret=` is supported.
+- Images are not yet garbage collected
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@ install:
 	install -D -m 0755 -t $(DESTDIR)$(prefix)/bin target/release/bootc
 	install -d -m 0755 $(DESTDIR)$(prefix)/lib/bootc/bound-images.d
 	install -d -m 0755 $(DESTDIR)$(prefix)/lib/bootc/kargs.d
+	ln -s /sysroot/ostree/bootc/storage $(DESTDIR)$(prefix)/lib/bootc/storage
 	install -d -m 0755 $(DESTDIR)$(prefix)/lib/systemd/system-generators/
 	ln -f $(DESTDIR)$(prefix)/bin/bootc $(DESTDIR)$(prefix)/lib/systemd/system-generators/bootc-systemd-generator
 	install -d $(DESTDIR)$(prefix)/lib/bootc/install
--- a/lib/src/boundimage.rs
+++ b/lib/src/boundimage.rs
@@ -5,7 +5,6 @@
 //! pre-pulled (and in the future, pinned) before a new image root
 //! is considered ready.
 
-use crate::task::Task;
 use anyhow::{Context, Result};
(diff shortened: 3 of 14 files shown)


cgwalters:
I've got this far:

```
 podman --storage-opt=additionalimagestore=/usr/lib/bootc/storage images
REPOSITORY             TAG         IMAGE ID      CREATED       SIZE        R/O
quay.io/fedora/fedora  40          d8a857bf322b  35 hours ago  233 MB      true
[root@localhost ~]# podman --storage-opt=additionalimagestore=/usr/lib/bootc/storage run --rm -ti quay.io/fedora/fedora:40 bash
Error: faccessat /var/lib/containers/storage/overlay/089e9cee22a3c6bf5eaa58e62df9569a9ac7c62ab2de25fec1d481fc27e2219c: no such file or directory
```

So...I need to dig into additionalimagestores here apparently.

rhatdan:
SELinux would require these directories to be labeled correctly.

vrothberg:
Does it work with `podman --root ... pull`?

rhatdan:
@cgwalters and I had a quick meeting yesterday where we say that `podman --root /sysboot... run` was failing similarly, so this looks like someing that sets up the addiitonal store is broken. Bottom line the additional store is broken, and Colin was going to look into that.

cgwalters:
It's actually using skopeo right now (there's some unused podman code) because podman doesn't support copying between two `containers-storage:`.

Anyways I figured it out...it was a bad interaction between `tempfile::TempDir` and mounts...we weren't unmounting cleanly which was obviously broken, but even with that I got tripped up by the fact that the storage stack creates a bind mount by default (I am curious why? need to dig) and so we need to unmount recursively.

Anyways with this latest things work for me!

```
$ podman --storage-opt=additionalimagestore=/usr/lib/bootc/storage run --rm -ti quay.io/fedora/fedora:40 echo hello
hello
$
```

So basically when we do this we'll need to tell people using bound images to specify `--storage-opt=additionalimagestore=/usr/lib/bootc/storage`. (TBD: do we actually want to do that globally by default? I'm uncertain)

vrothberg:
> podman doesn't support copying between two `containers-storage:`

`podman push $transport $transport` should work IIRC. Did you try that?



> So basically when we do this we'll need to tell people using bound images to specify `--storage-opt=additionalimagestore=/usr/lib/bootc/storage`. (TBD: do we actually want to do that globally by default? I'm uncertain)

I would add it.  That avoids duplicating the images in case non-bound users refer to it.

rhatdan:
I would suggest we do it by default.

I think `podman pull container-storage:/PATHTOSTORAGE` Should work. but their are a bunch of gotchas, specifically around configuration of the storage.conf file and the PATHTOSTORAGE is complicated to configure, but you might have already figured that out in skopeo.

cgwalters:
OK this one is getting farther. Still TODO:

- Update docs for bound images to describe the need to use `--storage-opt additionalimagestore=/usr/lib/bootc/storage`
- Introduce GC
- Tests

cgwalters:
OK, I think this is good to review/merge. I've fixed the TODOs above. Though there's clearly *a lot* more followup...for one thing I need to update the filesystem internals docs to talk about this, as it's suddenly a big new thing vs base ostree.

Bigger picture though this should all be very very low risk if one is not enabling any bound container images, which are also still classified as experimental.

But, we want people to try this out so getting it shipped in at least our continuous builds would help.

jmarrero:
We are getting close for 9.5 should we cut another release for this?

jmarrero (approved):
lgtm

docs/src/experimental-logically-bound-images.md
```diff
@@ -28,8 +32,17 @@ RUN ln -s /usr/share/containers/systemd/my-app.image /usr/lib/bootc/bound-images
     ln -s /usr/share/containers/systemd/my-app.image /usr/lib/bootc/bound-images.d/my-app.image
 ```
 
+In the `.container` definition, you should use:
+
+```
+GlobalArgs=--storage-opt=additionalimagestore=/usr/lib/bootc/storage
```
vrothberg, line 38: Couldn't we globally configure the additional image store via containers-storage.conf?  That would quite improve the UX.

cgwalters:
Fallout in https://github.com/containers/bootc/issues/747

docs/src/experimental-logically-bound-images.md
cgwalters: Yes, I think it's a possibility. That said, changing the storage configuration without having https://github.com/containers/storage/pull/1885 is...a bit ugly.

There are a few other things that have me slightly wary:

- All of this code is basically a no-op unless bound images are in use, but if we added an additional store it would mean every `podman` invocation would see it...it'd be empty, but it would be a global behavioral change
- There's been some lingering concerns I've seen in the past around what happens if...

Yeah ok, I think this a near showstopper for *globally* (i.e. in `storage.conf`) setting additionalimagestores (hereafter: AIS) as currently defined today. Basically, if an AIS has an image/layer, when pulling it will be reused. Makes sense!

But today, the bootc bound images logic will prune images that are no longer referenced.

This should work fine for bootc logically bound images (hereafter: LBI) because *all* layers in the stack live in the bootc store; podman sees every layer as read-only.

But yes, I just tested this, and things explode (as you'd expect) in the scenario where a lower read-only layer disappears for the case of an image/container that lives in `/var/lib/containers`.

cgwalters: I guess of course, bootc's GC logic could learn to take into account the `/var/lib/containers` store state...but we're starting to "cross the worlds" with stuff like that.

cgwalters: Now, one might be thinking...isn't the real fix for this to have *one storage* by default, i.e. on bootc systems podman and bootc default to having the same storage for images (underneath `/sysroot`, though containers would probably still make sense under `/var).

And yes maybe, but the thing that terrified me about going down that path is that typing "podman ..." has the potential to break the operating system. There's lots of docs talking about `podman system reset` and the like.

Could we make it reliable? Yes, clearly. But it'd require I think a lot of careful thought, code auditing, cooperation and integration.

rhatdan: It would be nice if podman had a way to recover from it. IE Pull down the missing image but I am not sure the missing image/layer would still exist or have a way to discover where it exist.

baude: we had agreed to create a pinning image mechanism for you that was going to avoid the reset problem no?  i thought that is the way we were going?

cgwalters: I am not aware of any commitment from anyone to work on image pinning on the podman side.

That said again, I should be very clear that my thoughts on this design changed as we got into the implementation. My original thought was basically "bootc is just running podman pull, how hard can it be Michael? 1 week?"

But yeah...having bootc touch `/var` (`/var/lib/containers`) just blows up our story around state, messes with transactional updates, partitioning, etc. So that led me rapidly down this path of having a bootc-owned storage.

rhatdan: Also don't we want the ability to roll back versions of images?

But bottom line we have the same issue if anyone sets up additional stores. If an image is used from an additional store and gets removed, then the container will no longer work. In RHIVOS we thought about this and came to the conclusion that we would know which images were being used so we would know what images to remove.  Also containers are ephemeral, they did not survive reboot.

cgwalters: > Also don't we want the ability to roll back versions of images?

bootc retains all LBIs that are referenced from any deployment, i.e. when you upgrade but then roll back, you are guaranteed that the LBIs are still there.

(However, there's ill-defined semantics currently around how LBIs work with floating tags, but basically "don't do that")

PR #727, opened 2024-07-25
Fix compiler warning on rust 1.80

jeckersb:
This dox/doc_cfg attr looks like it was originally included when bootc
was split off from ostree into its own repo but never actually used
here, so just remove it.


jmarrero (approved):
lgtm

PR #729, opened 2024-07-25
boundimage: Add tmt tests for bound images

ckyrouac:
These tests validate the basic use case of switching and upgrading an image with bound images that need to be pulled.


tests/booted/003-test-logically-bound.nu
```diff
@@ -0,0 +1,151 @@
...
+    }
+}
+
+def first_boot [] {
+    tap begin "bootc switch with bound images"
+
+    initial_setup
+
+    # build a bootc image that includes bound images
+    let images = echo [
```
cgwalters, line 83: The `echo` here shouldn't be necessary, in nushell I think it's basically a no-op. It notably *doesn't* convert the argument to a string.

cgwalters (approved):
Looks sane to me! Just one nonblocking nit

tests/booted/003-test-logically-bound.nu
cgwalters: (Here and a variety of other places)

cgwalters:
At the end of doing this, what did you think? I personally ended up somewhat ambivalent about the nushell usage for tests...my overall feeling was I could have written 70% of it faster in bash, except there's that 20% where you want to parse some JSON or operate on arrays and I need to go look up syntax and about 10% of horrible bash pitfalls.

tests/booted/003-test-logically-bound.nu
ckyrouac: ah - I'll clean these up before merging

ckyrouac:
I like it, I think these will be way easier to maintain in the long run. It was a little frustrating digging through the nu shell docs at times and chat gpt (and other llms) wasn't very helpful. Still, after the initial (relatively low) learning curve I think this a lot nicer than bash. I eventually got familiar enough with testing stuff directly in a nu shell prompt which seems more ergonomic than zsh or bash. I'll probably switch over to it as my main shell soon.

ckyrouac:
marking this as a draft because I started seeing some odd behavior related to handling reboots. I ran out of time today to figure out the cause and I'll be AFK for awhile. AFAICT the problem is that on each reboot all of the nu shell tmt tests run. So, trying to read the reboot count in each test gets tricky. I think for now we could just dump all the tests in a single file, but we probably want to isolate each test to a separate VM and run them in parallel or something.

cgwalters:
Oh wow...yes, oops. 

> but we probably want to isolate each test to a separate VM and run them in parallel or something.

Yes, this is quite doable, I'll take a look at figuring out the relevant tmt incantations.

cgwalters:
Let's track this issue here https://github.com/containers/bootc/issues/735

cgwalters:
Want to try rebasing on top of https://github.com/containers/bootc/pull/736 ? It will require something like:

`cp plans/test-20-local-upgrade.fmf plans/test-21-bound-images.fmf` and then replacing the test executed inside that fmf file

tests/booted/test-logically-bound.nu
```diff
@@ -0,0 +1,151 @@
...
+
+    verify_images $images $containers
+    tap ok
+}
+
+def main [] {
+    match $env.TMT_REBOOT_COUNT? {
+        null | "0" => first_boot,
+        null | "1" => second_boot,
+        null | "2" => third_boot,
```
cgwalters: The last two `null` here are unreachable (only the first can match), a bit surprising nu doesn't error on this. I've updated the code to drop them.

cgwalters (approved):
- rebased on main
- Added a fmf wrapper per new test SOP
- dropped the unnecessary `echo` (btw: doing so made the LSP plugin correctly do type inference! While nushell is somewhat obscure it's at least got a functioning language server and also functioning type inference; very unlikely to happen for bash...)

cgwalters:
```
$ tmt run plans -n test-21-logically-bound
/var/tmp/tmt/run-003

/plans/test-21-logically-bound
...
total: 1 test passed
```

:tada:

PR #730, opened 2024-07-25
install: Allocate a global tmpdir

cgwalters:
install: Allocate a global tmpdir

We allocate temporary things in a few places, and it's
handy to have a pre-created single directory for the
whole install process to use instead of creating
individual tempfiles.

---

install: Use tmpdir for target fetch verification

We create a transient ostree repo, to do so use the global
install tmpdir.

---


cgwalters:
Having a nice clean git history with a lot of incrementally justified and tested commits is nice, but some days the merge conflicts with one's own work gets tedious :smile:

jeckersb (approved):

PR #733, opened 2024-07-25
utils: Add a little CommandRunExt helper trait

cgwalters:
When I added the `Task` stuff it wasn't with the idea that we'd use it for *all* subprocess invocations.

I think in some cases we really just do want a `Command` instance without the extra wrappering.

Add an implementation (there are many of variants of this out there in lots of Rust codebases)

This makes sense to use instead of `Task` where we're using `.quiet()` so that we don't print the description, and especially where we don't expect the task to fail usually, so it's OK if the error message is odd.

### DIFF
--- a/lib/src/utils.rs
+++ b/lib/src/utils.rs
@@ -10,6 +10,25 @@ use ostree::glib;
 use ostree_ext::container::SignatureSource;
 use ostree_ext::ostree;
 
+/// Helpers intended for [`std::process::Command`].
+pub(crate) trait CommandRunExt {
+    fn run(&mut self) -> Result<()>;
+}
+
+impl CommandRunExt for Command {
+    /// Synchronously execute the child, and return an error if the child exited unsuccessfully.
+    fn run(&mut self) -> Result<()> {
+        let st = self.status()?;
+        if !st.success() {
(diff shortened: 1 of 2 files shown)


lib/src/utils.rs
```diff
@@ -10,6 +10,25 @@ use ostree::glib;
 use ostree_ext::container::SignatureSource;
 use ostree_ext::ostree;
 
+/// Helpers intended for [`std::process::Command`].
+pub(crate) trait CommandRunExt {
+    fn run(&mut self) -> Result<()>;
+}
+
+impl CommandRunExt for Command {
```
jmarrero, line 18: I see you are using Command::new and not CommandRunExt above. I am not sure I follow how this gets wired. I imagine that now everytime someone uses Command::new it will use this instead, but have no idea how that is happening.

jmarrero: mmm I guess it's using `trait` and then the impl that wires this.

jmarrero (approved):
lgtm

lib/src/utils.rs
cgwalters: Right, the trait adds a new method.

PR #734, opened 2024-07-25
tests: Run `tmt report` on failure

cgwalters:
It's annoying to have to go dig through the logs when something fails. We do have the tmt logs in CI, but again, more clicks.


cgwalters:
Rolled into https://github.com/containers/bootc/pull/736

PR #737, opened 2024-07-26
install: Work around bootc-image-builder using /run/osbuild/containers

cgwalters:
install: Factor out an osbuild module

Unfortunately the interactions/workarounds here are going to grow, so make
a module to encapsulate them.

---

install: Work around bootc-image-builder using /run/osbuild/containers

xref: https://github.com/osbuild/bootc-image-builder/issues/560

Basically osbuild/bib puts the host `/var/lib/containers` at
`/run/osbuild/containers`. If we detect this situation, bind
mount it to `/var/lib/containers` so that the container
stack we invoke at install time can find logically bound
images.

Closes: https://github.com/containers/bootc/issues/715

---

### DIFF
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -7,6 +7,7 @@
 // This sub-module is the "basic" installer that handles creating basic block device
 // and filesystem setup.
 pub(crate) mod baseline;
+mod osbuild;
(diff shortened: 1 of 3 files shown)


jmarrero (approved):
lgtm

PR #738, opened 2024-07-26
cli: Add a human readable format to status

jmarrero:
supersedes #602 and builds on top of #639

### DIFF
--- a/lib/src/status.rs
+++ b/lib/src/status.rs
@@ -1,4 +1,6 @@
 use std::collections::VecDeque;
+use std::io::IsTerminal;
+use std::io::Write;
 
 use anyhow::{Context, Result};
 use camino::Utf8Path;
@@ -305,6 +307,8 @@ pub(crate) async fn status(opts: super::cli::StatusOpts) -> Result<()> {
     let mut out = out.lock();
     let legacy_opt = if opts.json {
         OutputFormat::Json
+    } else if std::io::stdout().is_terminal() {
+        OutputFormat::HumanReadable
     } else {
         OutputFormat::Yaml
     };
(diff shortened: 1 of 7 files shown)


lib/src/status.rs
```diff
@@ -305,12 +306,24 @@ pub(crate) async fn status(opts: super::cli::StatusOpts) -> Result<()> {
...
+            OutputFormat::HumanReadable
+        } else {
+            OutputFormat::Yaml
+        }
     };
     let format = opts.format.unwrap_or(legacy_opt);
     match format {
         OutputFormat::Json => serde_json::to_writer(&mut out, &host).map_err(anyhow::Error::new),
         OutputFormat::Yaml => serde_yaml::to_writer(&mut out, &host).map_err(anyhow::Error::new),
+        OutputFormat::HumanReadable => {
```
cgwalters, line 319: Looks like a sane start to me, but we'll probably want to split this into a new function

lib/src/status.rs
```diff
@@ -305,12 +306,24 @@ pub(crate) async fn status(opts: super::cli::StatusOpts) -> Result<()> {
...
+            if let Some(booted) = host.status.booted {
+                println!("Current deployment image: {:?}", booted.image.unwrap().image.image);
```
cgwalters, line 321: I think we can avoid the unwrap() by referencing booted here right?

djach7:
Updated with most recent progress, still in draft status. 
Tasks still to do: 
- Format output to be more akin to `rpm-ostree status`
- Revise tests to function correctly with all fields
- Clean up code conventions and style to align with proper bootc/rust conventions

lib/Cargo.toml
```diff
@@ -44,6 +44,7 @@ toml = "0.8.12"
 xshell = { version = "0.2.6", optional = true }
 uuid = { version = "1.8.0", features = ["v4"] }
 tini = "1.3.0"
+struct_iterable = "0.1.1"
```
cgwalters, line 47: This crate is [not used very much](https://lib.rs/crates/struct_iterable/rev)...a bit of a red flag.

lib/src/status.rs
```diff
@@ -305,18 +308,106 @@ pub(crate) async fn status(opts: super::cli::StatusOpts) -> Result<()> {
...
+fn human_readable_output(mut out: impl Write, host: &Host) -> Result<()> {
+    for (status_string, status) in [
+        ("staged", &host.status.staged),
+        ("booted", &host.status.booted),
+        ("rollback", &host.status.rollback),
+    ] {
+        if let Some(host_status) = status {
+            let mut image_opts: ImageOpts = ImageOpts::default();
+            if let Some(image) = &host_status.image {
+                out.write_all(format!("Current {:?} image: {:?}\n", status_string, image.image.image).as_bytes())?;
```
cgwalters, line 344: Instead of `out.write_all(format!(...))` use `write!` which is both shorter and saves allocating a temporary string.

Also, we can avoid the literal `\n` with `writeln!`.

lib/src/status.rs
```diff
@@ -305,18 +308,106 @@ pub(crate) async fn status(opts: super::cli::StatusOpts) -> Result<()> {
...
+                    image_opts.version = ("No image version defined").to_owned();
+                }
+
+                image_opts.transport = image.image.transport.clone();
+                image_opts.digest = image.image_digest.clone();
+
+                for (field_name, field_value) in image_opts.iter() {
+                    if let Some(field_print) = field_value.downcast_ref::<String>() {
+                        out.write_all(format!("     Image {}: {:?}\n", field_name, field_print).as_bytes())?;
+                    }
```
cgwalters, line 358: I get why you wanted to do this...but I would say it's somewhat unusual in Rust to do things this way.

I don't think we're saving a *lot* here over something just like:

```
writeln!("    Image version: {}", image.version);
writeln!("    Image transport: {}", image.transport);
writeln!("    Image digest: {}", image.digest);
```
or so

lib/src/status.rs
```diff
@@ -305,18 +308,106 @@ pub(crate) async fn status(opts: super::cli::StatusOpts) -> Result<()> {
...
+#[test]
+fn test_human_readable() {
+    // Tests Staged and Booted, null Rollback
+    let mut SPEC_FIXTURE: &str = include_str!("fixtures/spec.yaml");
+    let mut host: Host = serde_yaml::from_str(SPEC_FIXTURE).unwrap();
+    let mut w = Vec::new();
+    human_readable_output(&mut w, &host).unwrap();
+    let w = String::from_utf8(w).unwrap();
+    dbg!(&w);
+    assert!(w.contains("quay.io/example/someimage:latest"));
```
cgwalters, line 380: I think it'd be really useful here to compare literal strings inline in the tests because that'd make it way easier for reviewers to see your proposed output without building your PR. Comparing the strings is nice with `similar_asserts::assert_eq!` which we use in a few tests already in this project.

lib/src/status.rs
```diff
@@ -305,18 +308,106 @@ pub(crate) async fn status(opts: super::cli::StatusOpts) -> Result<()> {
...
+
+    // Basic rhel for edge bootc install with nothing
```
cgwalters, line 382: I'd also consider splitting these into separate test functions.

lib/src/status.rs
```diff
@@ -305,19 +307,141 @@ pub(crate) async fn status(opts: super::cli::StatusOpts) -> Result<()> {
...
+    let mut w = Vec::new();
+    human_readable_output(&mut w, &host).unwrap();
+    let w = String::from_utf8(w).unwrap();
+    dbg!(&w);
+    let expected = indoc::indoc! { r"
+    Current staged image: quay.io/centos-bootc/centos-bootc:stream9
+        Image version: stream9.20240807.0 (No timestamp present)
+        Image transport: registry
+        Image digest: sha256:47e5ed613a970b6574bfa954ab25bb6e85656552899aa518b5961d9645102b38
+    No booted image present, checksum defined
```
cgwalters, line 419: We could probably phrase this in a way more understandable to the user. One may very well feel that they are in a "booted image", just not a *container* image right?

Hmm...how about: "Booted state is native ostree" ?

We could also actually go a step farther and detect the situation of rpm-ostree generated content, though that'd be a good bit more work.

lib/src/status.rs
```diff
@@ -305,19 +307,141 @@ pub(crate) async fn status(opts: super::cli::StatusOpts) -> Result<()> {
...
+    // Basic rhel for edge bootc install with nothing
+    let spec_fixture = include_str!("fixtures/spec-rfe-ostree-deployment.yaml");
+    let host = serde_yaml::from_str(spec_fixture).unwrap();
+    let mut w = Vec::new();
+    human_readable_output(&mut w, &host).unwrap();
+    let w = String::from_utf8(w).unwrap();
+    dbg!(&w);
+    let expected = indoc::indoc! { r"
+    No staged image present, checksum defined
+    No booted image present, checksum defined
```
cgwalters, line 399: See below on this

lib/src/status.rs
```diff
@@ -305,19 +307,141 @@ pub(crate) async fn status(opts: super::cli::StatusOpts) -> Result<()> {
...
+    human_readable_output(&mut w, &host).unwrap();
+    let w = String::from_utf8(w).unwrap();
+    dbg!(&w);
+    let expected = indoc::indoc! { r"
+    No staged image present
+    Current booted image: quay.io/centos-bootc/centos-bootc:stream9
+        Image version: stream9.20240807.0 (No timestamp present)
+        Image transport: registry
+        Image digest: sha256:47e5ed613a970b6574bfa954ab25bb6e85656552899aa518b5961d9645102b38
+    No rollback image present
```
cgwalters, line 440: One case I'd also like to see is where we have *both* staged and rollback...I think it'd be very useful to explicitly render the case of "upgrade" versus "switch" for example or so?

lib/src/status.rs
```diff
@@ -305,19 +307,141 @@ pub(crate) async fn status(opts: super::cli::StatusOpts) -> Result<()> {
...
+
+#[test]
+fn test_human_readable_base_spec() {
+    // Tests Staged and Booted, null Rollback
+    let spec_fixture: &str = include_str!("fixtures/spec-staged-booted.yaml");
+    let host: Host = serde_yaml::from_str(spec_fixture).unwrap();
+    let mut w = Vec::new();
+    human_readable_output(&mut w, &host).unwrap();
+    let w = String::from_utf8(w).unwrap();
+    dbg!(&w);
```
cgwalters, line 373: These are supposed to only live in local code, not committed to git.

lib/src/status.rs
```diff
@@ -305,19 +307,141 @@ pub(crate) async fn status(opts: super::cli::StatusOpts) -> Result<()> {
...
+            writeln!(out, "No {status_string} image present")?;
+        }
+    }
+    Ok(())
+}
+
+#[test]
+fn test_human_readable_base_spec() {
+    // Tests Staged and Booted, null Rollback
+    let spec_fixture: &str = include_str!("fixtures/spec-staged-booted.yaml");
```
cgwalters, line 368: One thing I think would help here is to factor out a helper function like

`fn human_status_from_spec_fixture(&str) -> Result<String>` or so

that gets used in each test.

cgwalters (commented):
Thanks for working on this!

I think this is getting close. I'd like to consider how to make the output even prettier (I know there's a lot of bikeshedding potential there). Some other projects use emojis (sometimes controversial), but maybe some ascii-art line drawing for associated data (like `systemctl status` does?)

lib/src/status.rs
djach7: Implemented a helper function, Following my interpretation of what you suggested. I used `.expect` to handled the result, should I instead be using the `.unwrap_or` that I used to handle the option types?

lib/src/status.rs
djach7: Added a test to do reflect this possible state

lib/src/status.rs
djach7: Updated the output line to match what you suggested, if you'd like me to dig in further let me know

lib/src/status.rs
```diff
@@ -305,19 +307,153 @@ pub(crate) async fn status(opts: super::cli::StatusOpts) -> Result<()> {
...
+
+#[test]
+fn test_human_readable_base_spec() {
+    // Tests Staged and Booted, null Rollback
+    let w = human_status_from_spec_fixture(include_str!("fixtures/spec-staged-booted.yaml"))
+        .expect("No spec found");
+    let expected = indoc::indoc! { r"
+    Current staged image: quay.io/example/someimage:latest
+        Image version: nightly (2023-10-14 19:22:15 UTC)
+        Image transport: registry
```
cgwalters, line 381: How about omitting this when it's `registry`? That's the 95% case I think.

lib/src/status.rs
```diff
@@ -305,19 +307,153 @@ pub(crate) async fn status(opts: super::cli::StatusOpts) -> Result<()> {
...
+                writeln!(out, "Current {status_string} state is native ostree")?;
+            }
+        } else {
+            writeln!(out, "No {status_string} image present")?;
+        }
+    }
+    Ok(())
+}
+
+fn human_status_from_spec_fixture(spec_fixture: &str) -> Result<String> {
```
cgwalters, line 365: Don't we get an unused function warning?

It may help to put everything here under a

```
#[cfg(test)]
mod tests {
...
}
```

as is done elsewhere in the code

cgwalters (approved):
Overall looks good! I would like to do some more changes at some point but it's better to do that as followups.

PR #741, opened 2024-07-29
A few prep commits for https://github.com/containers/bootc/pull/724

cgwalters:
utils: Allow non-static str for spinner

I have a use case for dynamic strings, this just requires
duplicating the string for the progress bar.

---

lib: Enable mount API for rustix

Will be used by future changes.

---

tests: Add generic post-install verification

Prep for more work.

---

utils: Capture stderr, add async

We want to capture stderr by default in these methods
so we provide useful errors.

Also add an async variant.

---

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -6,7 +6,8 @@ license = "MIT OR Apache-2.0"
 repository = "https://github.com/containers/bootc"
 readme = "README.md"
 publish = false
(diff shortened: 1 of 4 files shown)


cgwalters:
Blah, current rhel9 just has 1.75. I guess we should cap to that for now...

jmarrero (approved):
lgtm

PR #743, opened 2024-07-31
test: add centos stream 10 rpm build and e2e test

henrywang:
This PR includes:

1. add centos stream 10 all arch rpm build
2. add e2e test for centos stream 10

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -44,6 +44,10 @@ jobs:
       - centos-stream-9-aarch64
       - centos-stream-9-ppc64le
       - centos-stream-9-s390x
+      - centos-stream-10-x86_64
+      - centos-stream-10-aarch64
+      - centos-stream-10-ppc64le
+      - centos-stream-10-s390x
       - fedora-40-x86_64
       - fedora-40-aarch64
       - fedora-40-ppc64le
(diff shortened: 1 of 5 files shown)


henrywang:
Hi @lukewarmtemp, I found `systemd.unified_cgroup_hierarchy=0` was added in PR https://github.com/containers/bootc/pull/594.

I have to remove it in this PR because it caused boot failure on CS10 with error: `[!!!!!!] Refusing to run under cgroup v1, SY… specified on kernel command line.` CS10 deprecated boot from cgruop v1.

Does removing this karg impact test? Thanks.

lukewarmtemp:
> I there any test impact of remove this karg? Thanks.

You're free to remove it. Sorry, I just arbitrarily chose some kargs to test and didn't realize that it might conflict with others.

cgwalters:
Yes, we should definitely not set that.

henrywang:
> Looks sane, though one issue is the matrix here is getting pretty large, and fairly often we hit "artemis" provisioning failures e.g.

Yes. The most of time (99%) the provisioning issue happens on x86_64 runner machine. Looks the aarch64 runner does not hit the quota.

- In `to-disk` test, I removed `centos-stream-9-x86_64` and added `centos-stream-10-x86_64`. No additional x86_64 test added.
- I added `centos-stream-10-aarch64` in both `to-disk` and `to-existing-root` tests.

PR #744, opened 2024-07-31
Stabilize the current schema

cgwalters:
There's...a variety of warts in the current API schema. In particular I really dislike `.image.image.image`.

But...but...changing it now would in practice be somewhat difficult, and it's hard to say that just *that* tweak would be really useful.

Let's just stabilize the existing spec as `v1` and commit to maintaining it.

We've already added notes to ask callers to use `--format-version 1` to specify what they want so we can compatibly add new things in the future.

Speaking of, this also makes `--format-version=1` be the same thing as `--format-version=0` so that all the versions are just "1".

Closes: https://github.com/containers/bootc/issues/518


jmarrero (approved):
lgtm

PR #745, opened 2024-07-31
install/ppc64le: Also handle MBR partitions

cgwalters:
These may exist in the wild, even if we want to encourage GPT in the future.

Closes: https://github.com/containers/bootc/issues/742


cgwalters:
Only compile tested locally (by doing s/powerpc64/x86_64/ temporarily)

jmarrero (approved):
lgtm

PR #750, opened 2024-08-02
docs/logically-bound: A bit more docs

cgwalters:


docs/src/experimental-logically-bound-images.md
```diff
@@ -42,7 +53,12 @@ GlobalArgs=--storage-opt=additionalimagestore=/usr/lib/bootc/storage
...
+## Garbage collection
+
+The bootc image store is owned by bootc; images will be garbage collected when they are no longer referenced
+by a file in `/usr/lib/bootc/bound-images.d`.
+
 ## Limitations
 
-- Currently, only the Image field of a `.image` or `.container` file is used to pull the image; per above not even `PullSecret=` is supported.
-- Images are not yet garbage collected
+- The Image field of a `.image` or `.container` file is used to pull the image; this is the old field that is used.
```
jmarrero, line 63: generally lgtm, however something in the last sentence confuses me.  ```this is the old field that is used.``` after saying that the current way is .image or .container is how we pull the image makes me confused about what are you referring to as the old field here.

cgwalters: That sentence was just a corrupted mix of the old text and my attempt to update it, thanks for catching it!

Fixed.

jmarrero (approved):
lgtm

PR #751, opened 2024-08-02
store: Only print "Fetching bound image" if it's not there

cgwalters:
Also print how many bound images we have, since it's interesting and relevant information.


jmarrero (approved):
lgtm

PR #763, opened 2024-08-15
test: add logically bound images

henrywang:
This PR added logically bound images e2e test.


cgwalters (approved):
LGTM overall.

Going forward a bit, let's consider how we dedup/align the multiple test suites (I know this is my fault, it's tricky)

PR #764, opened 2024-08-15
Release 0.1.15

cgwalters:
The biggest feature here is the new support for "logically bound images", which is still classified as experimental, but we are very interested in feedback.


vrothberg (approved):
LGTM

PR #771, opened 2024-08-20
cli: Add `switch --apply`

cgwalters:
No reason not to have this.

Closes: https://github.com/containers/bootc/issues/770

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -59,6 +59,14 @@ pub(crate) struct SwitchOpts {
     #[clap(long)]
     pub(crate) quiet: bool,
 
+    /// Restart or reboot into the new target image.
+    ///
+    /// Currently, this option always reboots.  In the future this command
+    /// will detect the case where no kernel changes are queued, and perform
+    /// a userspace-only restart.
+    #[clap(long)]
+    pub(crate) apply: bool,
+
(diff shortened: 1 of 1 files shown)


lib/src/cli.rs
```diff
@@ -59,6 +59,14 @@ pub(crate) struct SwitchOpts {
     #[clap(long)]
     pub(crate) quiet: bool,
 
+    /// Restart or reboot into the new target image.
+    ///
+    /// Currently, this option always reboots.  In the future this command
+    /// will detect the case where no kernel changes are queued, and perform
+    /// a userspace-only restart.
+    #[clap(long, conflicts_with = "check")]
```
vrothberg, line 67: Looks like a copy-paste: `bootc switch` doesn't have a `--check` flag.

vrothberg (commented):
LGTM

lib/src/cli.rs
cgwalters: Thanks! Fixed.

henrywang:
The fedora 40 test failure is caused by https://pagure.io/fedora-infrastructure/issue/12110#comment-926243

djach7 (approved):
LGTM!

vrothberg:
> The fedora 40 test failure is caused by https://pagure.io/fedora-infrastructure/issue/12110#comment-926243

@henrywang do we need to use new images?

Issue #776, opened 2024-09-04
Add a hidden `bootc install wipe-ostree`


omertuc:
Right now the integration tests have some hacky code to clear out `ostree` stateroots  (`reset_root`), it would be better if they could rely on a bootc command to take care of that for them instead

_Originally posted by @cgwalters in https://github.com/containers/bootc/pull/622#discussion_r1742767547_

omertuc:
To that end @cgwalters suggested we use `Storage::write_deployments` with an empty slice, but unfortunately getting a `Storage` instance to operate on through `get_storage` requires us to run inside a bootc-booted system, which is [unsupported for running integration tests](https://github.com/containers/bootc/issues/404#issuecomment-2328853051), so we need to manually implement the ostree directory deletion

omertuc:
Even while avoiding the `get_storage` lock and instead running:

```rust
let sysroot = ostree::Sysroot::new_default();
sysroot.set_mount_namespace_in_use();
sysroot.load(gio::Cancellable::NONE)?;
sysroot
    .write_deployments(&[], gio::Cancellable::NONE)
    .context("removing deployments")?;
```

The `/ostree/deploy/foo` directory seems to stay intact

cgwalters:
Yeah, removing a deployment doesn't remove a stateroot today. But it should take care of the `/ostree/deploy/$stateroot/<deploy>` directories with their immutable bits etc.

omertuc:
> Yeah, removing a deployment doesn't remove a stateroot today. But it should take care of the `/ostree/deploy/$stateroot/<deploy>` directories with their immutable bits etc.

Ah of course, that makes sense. In that case, it seems to have done the job correctly. I'll try to plug it in the tests cleanup function in place of all the `chattr` stuff and see if it passes

omertuc:
Nah actually it didn't seem to work:

```
omer@localhost-live:~/bootc$ sudo podman run --rm --privileged -v /dev:/dev -v /var/lib/containers:/var/lib/containers --pid=host --security-opt label=disable -v /ostree:/ostree localhost/bootc bootc state wipe-ostree
omer@localhost-live:~/bootc$ sudo rm -rf /ostree/deploy/default/
rm: cannot remove '/ostree/deploy/default/deploy/a0003fa2e1d3d750a4d8ffc1f45c6eb5903049ebc533024255a4a1549d273d1f.0/bin': Operation not permitted
[...]

PR #777, opened 2024-09-05
cli: Add new `state wipe-ostree` subcommand

omertuc:
Solves #776

Right now the integration tests have some hacky code to clear out `ostree` stateroots  (see `reset_root` function), it would be better if they could rely on a bootc command to take care of that for them instead

This commit adds a new `state wipe-ostree` subcommand to the CLI that removes all `ostree` deployments from the system using native ostree bindings. This makes it easier to later rm -rf the stateroots and have a clean slate for the next test.

### DIFF
--- a/lib/src/deploy.rs
+++ b/lib/src/deploy.rs
@@ -16,8 +16,8 @@ use ostree_container::OstreeImageReference;
 use ostree_ext::container as ostree_container;
 use ostree_ext::container::store::{ImportProgress, PrepareResult};
 use ostree_ext::oci_spec::image::Descriptor;
-use ostree_ext::ostree;
 use ostree_ext::ostree::Deployment;
+use ostree_ext::ostree::{self, Sysroot};
 use ostree_ext::sysroot::SysrootLock;
 
 use crate::spec::ImageReference;
@@ -285,6 +285,14 @@ pub(crate) async fn prune_container_store(sysroot: &Storage) -> Result<()> {
         .prune_except_roots(&image_names)
         .await?;
     tracing::debug!("Pruned images: {}", pruned.len());
+    Ok(())
+}
+
+pub(crate) async fn wipe_ostree(sysroot: &Sysroot) -> Result<()> {
+    sysroot
+        .write_deployments(&[], gio::Cancellable::NONE)
+        .context("removing deployments")?;
+
     Ok(())
 }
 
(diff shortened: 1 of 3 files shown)


omertuc:
Tests failing because we need to check that /ostree even exists

vrothberg (approved):
LGTM

omertuc:
Is it normal for the tests to fail in this manner?

vrothberg:
> TASK [check secure boot status] ************************************************
fatal: [guest]: FAILED! => changed=true 
  cmd:
  - mokutil
  - --sb-state
  delta: '0:00:00.003638'
  end: '2024-09-06 08:17:47.889643'
  msg: non-zero return code
  rc: 255
  start: '2024-09-06 08:17:47.886005'
  stderr: This system doesn't support Secure Boot
  stderr_lines: <omitted>
  stdout: ''
  stdout_lines: <omitted>

@cgwalters WDYT?

lib/src/deploy.rs
```diff
@@ -288,6 +288,14 @@ pub(crate) async fn prune_container_store(sysroot: &Storage) -> Result<()> {
     Ok(())
 }
 
+pub(crate) async fn wipe_ostree(sysroot: &Sysroot) -> Result<()> {
```
cgwalters, line 291: We are absolutely not consistent about this, just an optional thing to maybe followup on later:

In theory an `async fn` shouldn't block the thread, but `write_deployments` is a totally blocking heavy filesystem operation. What we do in a few places is do heavy stuff like this inside a `tokio::task::spawn_blocking`. But again we're not *at all* consistent and it doesn't *really* matter right now because we aren't really doing many things concurrently at the moment.

But, the second we try to e.g. run as a systemd service that exposes an RPC API it would start to matter.

cgwalters (approved):
Nice thanks!

cgwalters:
The c10s tests are busted due to https://issues.redhat.com/browse/RHEL-56025 which should hopefully get sorted out soon

> stderr: This system doesn't support Secure Boot

Not sure on this one, @henrywang  ?

henrywang:
> > stderr: This system doesn't support Secure Boot
> 
> Not sure on this one, @henrywang ?

Hi @cgwalters , could you please link me this log? Thanks.

cgwalters:
I can't find it but @vrothberg where did you see that?

henrywang:
OH! I know this secure boot checking case. It's from https://github.com/containers/bootc/blob/1ba8d1e1b8863990b428fae6eb45ab754feb0e51/tests/e2e/playbooks/check-system.yaml#L23

Secure boot is one of system checkings, such as bios/uefi, tpm device, partition size, and partition table.

Issue #778, opened 2024-09-06
`async fn wipe_ostree` shouldn't block the thread


omertuc:
An  `async fn` shouldn't block the thread, but `write_deployments` is a totally blocking heavy filesystem operation. What we do in a few places is do heavy stuff like this inside a `tokio::task::spawn_blocking`. 

_Originally posted by @cgwalters in https://github.com/containers/bootc/pull/777#discussion_r1747218975_

PR #779, opened 2024-09-06
cli: wipe_ostree in spawn_blocking

omertuc:
Solves #778

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -21,7 +21,6 @@ use ostree_ext::keyfileext::KeyFileExt;
 use ostree_ext::ostree;
 use schemars::schema_for;
 
-use crate::deploy::wipe_ostree;
 use crate::deploy::RequiredHostSpec;
 use crate::lints;
 use crate::spec::Host;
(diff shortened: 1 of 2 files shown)


lib/src/cli.rs
```diff
@@ -929,7 +928,7 @@ async fn run_from_opt(opt: Opt) -> Result<()> {
             StateOpts::WipeOstree => {
                 let sysroot = ostree::Sysroot::new_default();
                 sysroot.load(gio::Cancellable::NONE)?;
-                wipe_ostree(&sysroot).await?;
+                tokio::task::spawn_blocking(move || crate::deploy::wipe_ostree(&sysroot)).await??;
```
cgwalters, line 931: This is a super minor nit, stylistic but I would have put the spawn_blocking inside `wipe_ostree` instead of outside.

Basically it's good to keep `cli.rs` a relatively thin wrapper around "parse options, dispatch to function"

omertuc: Yeah I considered that, but saw a few calls to `spawn_blocking` in `cli.rs` which convinced me to keep it outside, but it was mostly arbitrary. Will move it inside

cgwalters: Yep that's totally fine in this instance. However it's more common in some other code to *clone* the sysroot (it's effectively an `Arc<T>` internally (with gobject); we also promise that it's threadsafe, which is almost true).

cgwalters (approved):

Issue #781, opened 2024-09-10
Use the `os-ostree_sysroot_init_osname` API instead of the `os-init` libostree CLI


omertuc:
See discussion

https://github.com/containers/bootc/blob/d17e471eaa1da722bed005dfa820eac76aaf876e/lib/src/install.rs#L603-L606 

_Originally posted by @omertuc in https://github.com/containers/bootc/discussions/780_

PR #782, opened 2024-09-11
install: `sysroot.init_osname` API instead of CLI

omertuc:
Solves #781

This commit replaces the `ostree admin os-init` CLI invocation with the `sysroot.init_osname` API call. This is a more direct way to initialize the stateroot.


cgwalters (approved):
Nice, thank you!

PR #786, opened 2024-09-12
Release 0.1.16

cgwalters:
The major change here is we dropped our dependency on `gdisk` in favor of sfdisk from util-linux, which is more widely available and better maintained.


jeckersb (approved):
:ship:

PR #790, opened 2024-09-17
install: Drop SELinux-disabled warning

cgwalters:
We have e2e tests for this today. On a related topic, unfortunately the way bootc-image-builder sets up the container it runs bootc in it doesn't mount selinuxfs, so we see SELinux as disabled when it's not.

We should fix that, but it also avoids user confusion to drop the warning here.

While we're here, change things so we more consistently log the state computed.

Closes: https://github.com/containers/bootc/issues/419

### DIFF
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -894,6 +894,7 @@ impl RootSetup {
     }
 }
 
+#[derive(Debug)]
 pub(crate) enum SELinuxFinalState {
     /// Host and target both have SELinux, but user forced it off for target
     ForceTargetDisabled,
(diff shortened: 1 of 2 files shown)


mvo5 (approved):
This looks fine, thanks!

Orthogonal(ish) to this, shoudl we enable seinuxfs in our bib containers? It sounds like a good correctness fix?

cgwalters:
> Orthogonal(ish) to this, shoudl we enable seinuxfs in our bib containers? It sounds like a good correctness fix?

I think so yes

PR #791, opened 2024-09-17
TMT test tweaks

ckyrouac:
tmt: Add TMT_PLUGINS env to tmt run invocation

This is temporary while waiting for tmt to release the bootc provision
plugin.

----

tmt: Add TMT_VERBOSE env var

1 -> enabled. This adds -vvvvv to the tmt run invocation when using
`make test-tmt` to stream detailed test run output.

### DIFF
--- a/xtask/src/xtask.rs
+++ b/xtask/src/xtask.rs
@@ -181,10 +181,10 @@ fn test_tmt(sh: &Shell) -> Result<()> {
     println!("Discovered plans: {all_plan_files:?}");
 
     cmd!(sh, "cargo run -p tests-integration run-vm prepare-tmt").run()?;
-    // cc https://pagure.io/testcloud/pull-request/174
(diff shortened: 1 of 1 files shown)


xtask/src/xtask.rs
```diff
@@ -181,11 +181,25 @@ fn test_tmt(sh: &Shell) -> Result<()> {
...
+        let verbose_enabled = std::env::var("TMT_VERBOSE")
+            .ok()
+            .and_then(|s| s.parse::<u32>().ok())
+            .unwrap_or(0);
+
+        let verbose = if verbose_enabled == 1 {
+            Some("-vvvvv".to_string())
+        } else {
+            None
+        };
```
cgwalters, line 197: `let verbose = (verbose_enabled == 1).then_some("-vvvvv");` is a bit shorter, also avoids the unnecessary `to_string()` allocation here.

cgwalters (approved):

PR #792, opened 2024-09-18
tests: Add tmt test for bootc install with LBI

ckyrouac:
Along with the test, this will install the LBIs on the machine running the tests prior to running the tests. Also refactors the existing LBI test into a separate test plan to make room for the install tests.


xtask/src/xtask.rs
```diff
@@ -182,6 +182,13 @@ fn test_tmt(sh: &Shell) -> Result<()> {
 
     cmd!(sh, "cargo run -p tests-integration run-vm prepare-tmt").run()?;
 
+    // pull some small images that are used for LBI installation tests
+    cmd!(
+        sh,
+        "podman pull quay.io/curl/curl-base:latest quay.io/curl/curl:latest"
```
cgwalters, line 188: Minor: probably worth factoring out a `const TEST_IMAGES: &[&str] = ...` at the top of the file for increased visibility?

This one actually also relates to https://gitlab.com/fedora/bootc/examples/-/merge_requests/50#note_2041291332
which hmmm, I need to finish that PR

tests/containerfiles/lbi/usr/share/containers/systemd/jboss-webserver-5.image
```diff
@@ -0,0 +1,3 @@
+[Image]
+Image=registry.redhat.io/jboss-webserver-5/jws5-rhel8-operator:latest
+AuthFile=/root/auth.json
```
cgwalters, line 3: Is this actually used?

cgwalters (approved):
Looks nice to me! Thanks for working on this, I'm excited to set a better testing foundation for the project!

tests/containerfiles/lbi/usr/share/containers/systemd/jboss-webserver-5.image
ckyrouac: nope, added a comment to explain that it's meant to represent a quadlet image that is not logically bound.

cgwalters:
This just needs a `cargo fmt`

PR #794, opened 2024-09-18
utils: Add a `log_debug()` helper

cgwalters:
ref https://github.com/containers/bootc/discussions/793

### DIFF
--- a/utils/src/command.rs
+++ b/utils/src/command.rs
@@ -9,8 +9,12 @@ use anyhow::{Context, Result};
 
 /// Helpers intended for [`std::process::Command`].
 pub trait CommandRunExt {
+    /// Log (at debug level) the full child commandline.
+    fn log_debug(&mut self) -> &mut Self;
+
     /// Execute the child process.
     fn run(&mut self) -> Result<()>;
+
     /// Execute the child process, parsing its stdout as JSON.
     fn run_and_parse_json<T: serde::de::DeserializeOwned>(&mut self) -> Result<T>;
 }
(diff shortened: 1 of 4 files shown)


omertuc:
dco

cgwalters:
> dco

fixed, also did some other tweaks

utils/src/command.rs
```diff
@@ -68,9 +69,17 @@ impl CommandRunExt for Command {
     fn run(&mut self) -> Result<()> {
         let stderr = tempfile::tempfile()?;
         self.stderr(stderr.try_clone()?);
+        tracing::trace!("exec: {self:?}");
```
omertuc, line 72: I kinda hate that we'll have:

```
DEBUG exec: "lsblk" "--pairs" "--paths" "--inverse" "--output" "NAME,TYPE" "composefs"
TRACE exec: "lsblk" "--pairs" "--paths" "--inverse" "--output" "NAME,TYPE" "composefs"
```

But we can't append state to `Command` so not sure how to avoid it

cgwalters: Fixed that

utils/src/command.rs
```diff
@@ -68,9 +69,20 @@ impl CommandRunExt for Command {
...
         self.status()?.check_status(stderr)
     }
 
+    /// Output a debug-level log message with this command.
+    fn log_debug(&mut self) -> &mut Self {
+        // We unconditionally log at trace level, so avoid double logging
```
omertuc, line 78: To be pedantic, a user might able to avoid logging accidentally through running `output()`, `spawn()`, `status()`, etc instead of `run()`

cgwalters: Yeah, I have no good ideas at this point for that other than basically "don't do that"?

I think effectively our extension trait's role is to take over those "terminal invocations"...it'd seem weird to use the trait just for the debug logging which is pretty trivial to do externally.

This would be the kind of thing I'd worry about if we decided to go make an official crate from our helpers, but...

omertuc: Yeah that's fair

omertuc:
Just cherry-picked and it seems to work as expected, thanks

cgwalters:
Rebased :surfer:

jmarrero (approved):
lgtm

PR #795, opened 2024-09-18
Make lints stricter, apply crate wide

cgwalters:
Add `dead_code = "deny"` to our default lints; we had a compiler warning for this in main.

Fix the warning by moving the human readable test code into `#[cfg(test)]`.

While we're here, move the other lib.rs lints into the crate; enforcing docs for *everything* at first I thought might be heavy handed but actually is fine as it only applies to things that are `pub`, of which we don't actually have that much so it mainly forced me to add some stub docs for the modules, which is probably a good idea.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,14 +33,9 @@ jobs:
         uses: Swatinem/rust-cache@v2
         with:
           key: "tests"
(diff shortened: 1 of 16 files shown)


jmarrero (approved):
lgtm

PR #797, opened 2024-09-19
Update to ostree-ext 0.15

cgwalters:
Only one tiny change required thankfully for this one!

This drops out duplicate versions of oci-spec from our lockfile too which is nice.


jeckersb (approved):
:tada:

PR #798, opened 2024-09-19
Status tweaks

cgwalters:
status: Change transport display

Registry is the assumed default; let's be quieter (less output)
in this default/expected case.

If we're using something else (e.g. `oci`) then put it
inline with the image name as that's something one can
pass to e.g. `skopeo`.

---

status: Minor code refactorings

- Rename variable to `slot_name` as I think it's a bit clearer

---

status: Factor out helpers for rendering container image

Prep for more work; no functional changes.

---

### DIFF
--- a/lib/src/status.rs
+++ b/lib/src/status.rs
@@ -326,7 +326,7 @@ pub(crate) async fn status(opts: super::cli::StatusOpts) -> Result<()> {
 }
 
 fn human_readable_output(mut out: impl Write, host: &Host) -> Result<()> {
-    for (status_string, status) in [
+    for (slot_name, status) in [
         ("staged", &host.status.staged),
         ("booted", &host.status.booted),
         ("rollback", &host.status.rollback),
@@ -342,7 +342,7 @@ fn human_readable_output(mut out: impl Write, host: &Host) -> Result<()> {
                     // But for non-registry we include the transport
                     Cow::Owned(format!("{transport}:{imagename}"))
                 };
-                writeln!(out, "Current {status_string} image: {imageref}")?;
+                writeln!(out, "Current {slot_name} image: {imageref}")?;
 
                 let version = image
                     .version
@@ -358,10 +358,10 @@ fn human_readable_output(mut out: impl Write, host: &Host) -> Result<()> {
                 writeln!(out, "    Image version: {version} ({timestamp})")?;
                 writeln!(out, "    Image digest: {digest}")?;
             } else {
-                writeln!(out, "Current {status_string} state is native ostree")?;
+                writeln!(out, "Current {slot_name} state is native ostree")?;
             }
         } else {
-            writeln!(out, "No {status_string} image present")?;
+            writeln!(out, "No {slot_name} image present")?;
         }
     }
     Ok(())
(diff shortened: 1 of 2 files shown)


djach7:
Just tried to test this locally for a bit and kept running into this error: 

`
ERROR Status: Computing status: Booted deployment: Reading deployment metadata: Missing base image ref ostree/container/blob/d5f71a3300fe94bed5ec6cbf5a6e508a0da96d48057ba76d80a8426add6ef9ee
`

Any ideas what's causing that? I don't think I'm setting up the bootc environment wrong, as other commands (like `bootc help` and `bootc --version`) work fine

cgwalters:
see https://github.com/containers/bootc/issues/800

djach7:
Ah, I see. I'll review as much as I can then without being able to test the full thing. My apologies for not staying more on top of my github notifications

lib/src/status.rs
```diff
@@ -324,40 +325,62 @@ pub(crate) async fn status(opts: super::cli::StatusOpts) -> Result<()> {
...
+    image: &crate::spec::ImageStatus,
+) -> Result<()> {
+    let transport = &image.image.transport;
+    let imagename = &image.image.image;
+    // Registry is the default, so don't show that
+    let imageref = if transport == "registry" {
+        Cow::Borrowed(imagename)
+    } else {
+        // But for non-registry we include the transport
+        Cow::Owned(format!("{transport}:{imagename}"))
```
djach7, line 341: I need to read about how Cow works but I like this implementation. It does make a lot more sense.

cgwalters:
#800 only bites for the upgrade case, which is how it sailed through our primary CI. If you create a fresh image with current bootc git it will work. So e.g. `podman-bootc run <image>` etc.

lib/src/status.rs
```diff
@@ -324,40 +325,62 @@ pub(crate) async fn status(opts: super::cli::StatusOpts) -> Result<()> {
...
+        .map(|t| t.to_string())
+        .unwrap_or_else(|| "No timestamp present".to_owned());
+    let digest = &image.image_digest;
+
+    writeln!(out, "    Image version: {version} ({timestamp})")?;
+    writeln!(out, "    Image digest: {digest}")?;
+    Ok(())
+}
+
+fn human_render_ostree(mut out: impl Write, slot_name: &str, _ostree_commit: &str) -> Result<()> {
```
djach7, line 361: Is this where we'd implement the documentation for additional deployments like `rpm-ostree status`?

djach7 (approved):
LGTM, I need to read about Cow and I left a non-blocking question but it's a certainly a cleaner implementation overall. Agreed that `slot_name` is a better var name. 

Will double check again locally once #800 is resolved, but I'm ok with merging this before that.

lib/src/status.rs
cgwalters: Yep

lib/src/status.rs
cgwalters: Cow basically abstracts over borrowed/owned data here, it's honestly a pointless micro-optimization but it's not hard to do either.

PR #804, opened 2024-09-24
Bump ostree-ext to 0.15.1 and use proper digests where possible

jeckersb:

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -609,16 +609,17 @@ async fn upgrade(opts: UpgradeOpts) -> Result<()> {
         }
     } else {
         let fetched = crate::deploy::pull(repo, imgref, None, opts.quiet).await?;
-        let staged_digest = staged_image.as_ref().map(|s| s.image_digest.as_str());
-        let fetched_digest = fetched.manifest_digest.as_str();
+        let staged_digest = staged_image.map(|s| s.digest().expect("valid digest in status"));
+        let fetched_digest = &fetched.manifest_digest;
         tracing::debug!("staged: {staged_digest:?}");
         tracing::debug!("fetched: {fetched_digest}");
         let staged_unchanged = staged_digest
+            .as_ref()
             .map(|d| d == fetched_digest)
             .unwrap_or_default();
         let booted_unchanged = booted_image
             .as_ref()
-            .map(|img| img.manifest_digest.as_str() == fetched_digest)
+            .map(|img| &img.manifest_digest == fetched_digest)
             .unwrap_or_default();
         if staged_unchanged {
             println!("Staged update present, not changed.");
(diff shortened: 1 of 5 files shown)


jeckersb:
/me hopes to not break everything this time

lib/src/cli.rs
```diff
@@ -609,16 +609,16 @@ async fn upgrade(opts: UpgradeOpts) -> Result<()> {
         }
     } else {
         let fetched = crate::deploy::pull(repo, imgref, None, opts.quiet).await?;
-        let staged_digest = staged_image.as_ref().map(|s| s.image_digest.as_str());
-        let fetched_digest = fetched.manifest_digest.as_str();
+        let staged_digest = staged_image.and_then(|s| s.digest().ok());
```
cgwalters, line 612: I think we should either `?` or `expect("valid digest in status")` here but not `.ok()` which will silently discard invalid digests.

It really feels like it Shouldn't Happen and if it does we want a clear error.

lib/src/cli.rs
```diff
@@ -609,16 +609,16 @@ async fn upgrade(opts: UpgradeOpts) -> Result<()> {
...
+        let fetched_digest = &fetched.manifest_digest;
         tracing::debug!("staged: {staged_digest:?}");
         tracing::debug!("fetched: {fetched_digest}");
         let staged_unchanged = staged_digest
-            .map(|d| d == fetched_digest)
+            .map(|d| d == *fetched_digest)
```
cgwalters, line 617: Minor style nit I think coercing to a reference like `&d == fetched_digest` is perhaps more idomatic?

lib/src/cli.rs
cgwalters: (There's a hugely complex subthread one could spawn from this around "how robust are we to corrupted on disk state and how hard is it to recover" but I think the most likely fix for that is basically ensuring that we can always pull a new image even if for some reason we can't read the current one)

cgwalters: The rationale to `?` specifically is we already error out if e.g. the manifest JSON fails to parse because the disk state is corrupted, and "malformed digest" is basically that - the *writing* code shouldn't have let it happen, and we're treating "on disk state is malformed" as an error (but not a crash/panic).

lib/src/cli.rs
jeckersb: Yeah I suppose so.  That's more my historic C-like brain wanting to dereference two pointers to compare the values instead of comparing the actual pointers.  But rust already [does the dereference for you](https://doc.rust-lang.org/1.81.0/src/core/cmp.rs.html#1660) (I had to go look up how that actually works!).

lib/src/cli.rs
jeckersb: :+1: I went the `expect` route since we want to be explicit that errors there are... unexpected :sunglasses:

cgwalters (approved):

PR #805, opened 2024-09-25
install-to-disk: Check for mounts before install

### LINKED ISSUE #720: install to-disk: Check for mounts
Split from https://github.com/containers/bootc/pull/718#pullrequestreview-2194746026

> we definitely need to strengthen the to-disk path to bail hard if the device is mounted for example, even if --wipe is specified.

I looked at implementing this like the below, but it doesn't work because our mount namespace won't have the physical devices mounted.

Instead, I think we can implement this by using `findmnt -N 1 --source /path/to/dev` for each device...or just scraping the overall `findmnt -N 1` and matching up the devices on our own.

```
From c86a452b05a7b94c341269e147bc31d4c0a82feb Mon Sep 17 00:00:00 2001
Date: Tue, 23 Jul 2024 12:34:55 -0400
Subject: [PATCH] wip

---
 lib/src/blockdev.rs         | 17 +++++++++++++++++
 lib/src/install/baseline.rs |  4 ++++
 2 files changed, 21 insertions(+)

diff --git a/lib/src/blockdev.rs b/lib/src/blockdev.rs
index 3c69080..f18e187 100644
--- a/lib/src/blockdev.rs
+++ b/lib/src/blockdev.rs
@@ -36,6 +36,9 @@ pub(crate) struct Device {
     // Filesystem-related properties
+
+    // VFS properties
+    pub(crate) mountpoints: Option<Vec<String>>,
 }
 
 impl Device {
@@ -49,6 +52,20 @@ impl Device {
         self.children.as_ref().map_or(false, |v| !v.is_empty())
     }
 
+    /// Return the first mounted device (this device, or any of its children) along
+    /// with its mountpoints.
+    pub(crate) fn find_first_mounted(&self) -> Option<(&Device, &[String])> {
+        match self.mountpoints.as_deref() {
+            Some(mounts @ [_, ..]) => return Some((self, mounts)),
+            Some([]) | None => {}
+        }
+        self.children
+            .iter()
+            .flatten()
+            .filter_map(|c| c.find_first_mounted())
+            .next()
+    }
+
     // The "start" parameter was only added in a version of util-linux that's only
     // in Fedora 40 as of this writing.
     fn backfill_start(&mut self) -> Result<()> {
diff --git a/lib/src/install/baseline.rs b/lib/src/install/baseline.rs
index 180a3a1..fb95731 100644
--- a/lib/src/install/baseline.rs
+++ b/lib/src/install/baseline.rs
@@ -191,6 +191,10 @@ pub(crate) fn install_create_rootfs(
             opts.device
         );
     }
+    // Always disallow writing to a mounted device
+    if let Some((dev, mounts)) = device.find_first_mounted() {
+        anyhow::bail!("Device {} is mounted: {:?}", dev.path(), mounts)
+    }
 
     let run_bootc = Utf8Path::new(RUN_BOOTC);
     let mntdir = run_bootc.join("mounts");
-- 
2.44.0
```

djach7:
Works to improve `bootc install to-disk` flow by checking for mounts on the device to be installed to. If mounts are present then install will abort. 

Resolves #720

### DIFF
--- a/lib/src/mount.rs
+++ b/lib/src/mount.rs
@@ -12,6 +12,7 @@ use crate::task::Task;
 
 #[derive(Deserialize, Debug)]
 #[serde(rename_all = "kebab-case")]
+#[allow(dead_code)]
 pub(crate) struct Filesystem {
     // Note if you add an entry to this list, you need to change the --output invocation below too
     pub(crate) source: String,
@@ -21,6 +22,7 @@ pub(crate) struct Filesystem {
     pub(crate) fstype: String,
     pub(crate) options: String,
     pub(crate) uuid: Option<String>,
+    pub(crate) children: Option<Vec<Filesystem>>,
 }
 
 #[derive(Deserialize, Debug)]
@@ -28,7 +30,7 @@ pub(crate) struct Findmnt {
     pub(crate) filesystems: Vec<Filesystem>,
 }
 
-fn run_findmnt(args: &[&str], path: &str) -> Result<Filesystem> {
+fn run_findmnt(args: &[&str], path: &str) -> Result<Findmnt> {
     let o: Findmnt = Command::new("findmnt")
         .args([
             "-J",
@@ -40,6 +42,12 @@ fn run_findmnt(args: &[&str], path: &str) -> Result<Filesystem> {
         .arg(path)
         .log_debug()
         .run_and_parse_json()?;
+    Ok(o)
+}
+
+// Retrieve a mounted filesystem from a device given a matching path
+fn findmnt_filesystem(args: &[&str], path: &str) -> Result<Filesystem> {
+    let o = run_findmnt(args, path)?;
     o.filesystems
         .into_iter()
         .next()
@@ -50,13 +58,40 @@ fn run_findmnt(args: &[&str], path: &str) -> Result<Filesystem> {
 /// Inspect a target which must be a mountpoint root - it is an error
 /// if the target is not the mount root.
 pub(crate) fn inspect_filesystem(path: &Utf8Path) -> Result<Filesystem> {
-    run_findmnt(&["--mountpoint"], path.as_str())
+    findmnt_filesystem(&["--mountpoint"], path.as_str())
 }
 
 #[context("Inspecting filesystem by UUID {uuid}")]
 /// Inspect a filesystem by partition UUID
 pub(crate) fn inspect_filesystem_by_uuid(uuid: &str) -> Result<Filesystem> {
-    run_findmnt(&["--source"], &(format!("UUID={uuid}")))
+    findmnt_filesystem(&["--source"], &(format!("UUID={uuid}")))
+}
+
+// Check if a specified device contains an already mounted filesystem
+// in the root mount namespace
+pub(crate) fn is_mounted_in_pid1_mountns(path: &str) -> Result<bool> {
+    let o = run_findmnt(&["-N"], "1")?;
+
+    let mounted = o.filesystems.iter().any(|fs| is_source_mounted(path, fs));
+
+    Ok(mounted)
+}
+
+// Recursively check a given filesystem to see if it contains an already mounted source
+pub(crate) fn is_source_mounted(path: &str, mounted_fs: &Filesystem) -> bool {
+    if mounted_fs.source.contains(path) {
+        return true;
+    }
+
+    if let Some(ref children) = mounted_fs.children {
+        for child in children {
+            if is_source_mounted(path, child) {
+                return true;
+            }
+        }
+    }
+
+    false
 }
 
 /// Mount a device to the target path.
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -1802,6 +1802,7 @@ fn test_gather_root_args() {
         maj_min: "252:4".into(),
         options: "rw".into(),
         uuid: Some("965eb3c7-5a3f-470d-aaa2-1bcf04334bc6".into()),
+        children: None,
     };
     let r = find_root_args_to_inherit(&[], &inspect).unwrap();
     assert_eq!(r.mount_spec, "UUID=965eb3c7-5a3f-470d-aaa2-1bcf04334bc6");
--- a/lib/src/install/baseline.rs
+++ b/lib/src/install/baseline.rs
@@ -27,6 +27,7 @@ use super::State;
 use super::RUN_BOOTC;
 use super::RW_KARG;
 use crate::mount;
+use crate::mount::is_mounted_in_pid1_mountns;
 use crate::task::Task;
 
 // This ensures we end up under 512 to be small-sized.
@@ -161,6 +162,11 @@ pub(crate) fn install_create_rootfs(
     let device = crate::blockdev::list_dev(&opts.device)?;
     // Canonicalize devpath
     let devpath: Utf8PathBuf = device.path().into();
+
+    // Always disallow writing to mounted device
+    if is_mounted_in_pid1_mountns(&device.path())? {
+        anyhow::bail!("Device {} is mounted", device.path())
+    }
 
     // Handle wiping any existing data
     if opts.wipe {


.dockerignore
```diff
@@ -1,6 +0,0 @@
-.cosa
```
cgwalters, line 1: Shouldn't be deleted

lib/src/blockdev.rs
```diff
@@ -37,6 +38,9 @@ pub(crate) struct Device {
     // Filesystem-related properties
     pub(crate) label: Option<String>,
     pub(crate) fstype: Option<String>,
+
+    // VFS properties
+    pub(crate) mountpoints: Option<Vec<String>>,
```
cgwalters, line 43: See below

lib/src/blockdev.rs
```diff
@@ -50,6 +54,30 @@ impl Device {
         self.children.as_ref().map_or(false, |v| !v.is_empty())
     }
 
+    // Find first mounted device along with mountpoints
+    pub(crate) fn find_first_mounted(&self) -> Option<(&Device, String)> {
+        let output = Command::new("findmnt")
```
cgwalters, line 59: So we already have code in `mount.rs` that's wrapping `findmnt`. An overall challenge here is because the install code is in a mount namespace, we won't see the device as mounted to start when we scan the block device.

That's why the `mountpoints` property that we can get from lsblk will be empty.

I think what we want here is a new api in `mount.rs` like this:

```
pub(crate) fn is_mounted_in_pid_mountns(device: &Device, pid: rustix::process::Pid) -> Result<bool>` or so.

Then we traverse the the device and run `findmnt -N <pid> --source <device>` or so

**Review by cgwalters — COMMENTED** [review]
Thanks for picking this up!

File: .dockerignore (hunk 1 above)
Comment (djach7) [reply .dockerignore:1]: I'll update that, I was experimenting with some building so it's a bit hacky right now. Thanks for the spot

File: lib/src/blockdev.rs (hunk 3 above)
Comment (djach7) [reply lib/src/blockdev.rs:59]: So the `bootc install to-disk` code is in a mount namespace, which results in the device mountpoints being hidden from `findmnt`? 

For your suggestion, what process am I getting the pid from? Is that from the disk, since I'm feeding it into `findmnt` as the task id then?

Just trying to wrap my head around all this stuff, I appreciate the quick feedback a ton!

Comment (cgwalters) [reply lib/src/blockdev.rs:59]: > For your suggestion, what process am I getting the pid from?

Pid 1

Try playing with this from a root shell:

```
$ unshare -m
$ findmnt /tmp
$ mount --bind /var/tmp /tmp
$ findmnt /tmp
$ findmnt -N 1 /tmp
```

Note the second one (looking in the root mountns) didn't see the bind. Press ctrl-d to exit the `unshare` bash shell which will also tear down the mount.

Comment (djach7) [reply lib/src/blockdev.rs:59]: I think I have a better idea of it now, Joseph helped me walk through it yesterday too. My basic understanding is that `bootc install to-disk`  can't see the same pids (and hence the mountpoints) that the podman container can see, so I need to bridge the gap and allow install to-disk to see the mountpoints podman can. Then I can run `findmnt` with that specified pid and check for the mountpoints off of it

Comment (djach7) [reply lib/src/blockdev.rs:59]: I'll experiment with those commands you listed too, thank you

**Comment by djach7** [comment]
This should be working now, I made changes to parse the `findmnt` output for the device and bail if found. @cgwalters for the `findmnt` command I hardcoded `-N 1`,  is there any issue with that? I was attempting to pass it a pid (you can see that attempt in the wip commit if you're curious) but the pid being passed was unable to see the device I was attempting to mount to.

File: lib/src/blockdev.rs [hunk 4]
```diff
@@ -50,6 +50,24 @@ impl Device {
...
+    // Return true if device is mounted anywhere
+    pub(crate) fn is_mounted_in_pid_mounts(&self) -> Result<bool> {
+        let output = Command::new("findmnt")
+            .arg("-N")
+            .arg("1")
+            .arg("--output=SOURCE")
+            .output()
+            .expect("Failed to execute findmnt");
+
+        let mounts = String::from_utf8(output.stdout).unwrap();
```
cgwalters, line 62: While `findmnt` giving non-utf8 output is a "shouldn't happen" case, it's equally easy for us to use `?` here to propagate it as an error instead of aborting, so let's do that.

lib/src/blockdev.rs
```diff
@@ -50,6 +50,24 @@ impl Device {
...
     }
 
+    // Return true if device is mounted anywhere
+    pub(crate) fn is_mounted_in_pid_mounts(&self) -> Result<bool> {
+        let output = Command::new("findmnt")
+            .arg("-N")
+            .arg("1")
+            .arg("--output=SOURCE")
+            .output()
+            .expect("Failed to execute findmnt");
```
cgwalters, line 60: See below

lib/src/blockdev.rs
cgwalters: How about adding this code to `mount.rs` instead? We could aim to share code with `run_findmnt`...at least in terms of parsing JSON instead of text.

lib/src/blockdev.rs
```diff
@@ -50,6 +50,24 @@ impl Device {
         self.children.as_ref().map_or(false, |v| !v.is_empty())
     }
 
+    // Return true if device is mounted anywhere
+    pub(crate) fn is_mounted_in_pid_mounts(&self) -> Result<bool> {
```
cgwalters, line 54: How about "is_mounted_in_pid1_mountns"?

cgwalters (changes requested):

cgwalters:
Thanks for working on this BTW, I think it's getting close!

lib/src/blockdev.rs
djach7: Just pushed up a refactor into `mount.rs`, it now shares some code with the other findmnt functions in there. Is that more in line with what you were thinking?

lib/src/mount.rs
```diff
@@ -50,13 +58,42 @@ fn run_findmnt(args: &[&str], path: &str) -> Result<Filesystem> {
...
+        if is_source_mounted(&path, fs)? {
+            return Ok(true);
+        }
+    }
+
+    Ok(false)
+}
+
+// Recursively check a given filesystem to see if it contains an already mounted source
+pub(crate) fn is_source_mounted(path: &String, mounted_fs: &Filesystem) -> Result<bool> {
```
cgwalters, line 83: - This function looks infallible so I think it can just be `-> bool`
- Prefer using `&str` vs `&String` for parameters; recent blog: https://steveklabnik.com/writing/when-should-i-use-string-vs-str/

lib/src/mount.rs
```diff
@@ -50,13 +58,42 @@ fn run_findmnt(args: &[&str], path: &str) -> Result<Filesystem> {
...
 
 #[context("Inspecting filesystem by UUID {uuid}")]
 /// Inspect a filesystem by partition UUID
 pub(crate) fn inspect_filesystem_by_uuid(uuid: &str) -> Result<Filesystem> {
-    run_findmnt(&["--source"], &(format!("UUID={uuid}")))
+    findmnt_filesystem(&["--source"], &(format!("UUID={uuid}")))
+}
+
+// Check if a specified device contains an already mounted filesystem
+pub(crate) fn is_mounted_in_pid1_mountns(path: String) -> Result<bool> {
```
cgwalters, line 71: Let's make this take `&str`

lib/src/mount.rs
```diff
@@ -50,13 +58,42 @@ fn run_findmnt(args: &[&str], path: &str) -> Result<Filesystem> {
...
+    let o = run_findmnt(&["-N"], "1")?;
+    for fs in o.filesystems.iter() {
+        if is_source_mounted(&path, fs)? {
+            return Ok(true);
+        }
+    }
+
+    Ok(false)
```
cgwalters, line 79: This is all fine but with the below change could be condensed to something like
`Ok(o.filesystems.iter().any(|fs| is_source_mounted(path)))`

Or:
```
let mounted = o.filesystems.iter().any(|fs| is_source_mounted(path));
Ok(mounted)
```

lib/src/mount.rs
```diff
@@ -50,13 +58,42 @@ fn run_findmnt(args: &[&str], path: &str) -> Result<Filesystem> {
...
 }
 
 #[context("Inspecting filesystem by UUID {uuid}")]
 /// Inspect a filesystem by partition UUID
 pub(crate) fn inspect_filesystem_by_uuid(uuid: &str) -> Result<Filesystem> {
-    run_findmnt(&["--source"], &(format!("UUID={uuid}")))
+    findmnt_filesystem(&["--source"], &(format!("UUID={uuid}")))
+}
+
+// Check if a specified device contains an already mounted filesystem
```
cgwalters, line 70: "already mounted filesystem in the root mount namespace" or so.

lib/src/mount.rs
```diff
@@ -50,13 +58,42 @@ fn run_findmnt(args: &[&str], path: &str) -> Result<Filesystem> {
...
+            return Ok(true);
+        }
+    }
+
+    Ok(false)
+}
+
+// Recursively check a given filesystem to see if it contains an already mounted source
+pub(crate) fn is_source_mounted(path: &String, mounted_fs: &Filesystem) -> Result<bool> {
+    if mounted_fs.source.contains(path) {
```
cgwalters, line 84: A bit uncertain about `source.contains(path)` vs `mounted_fs.source.as_str() == path` here...I guess you're probably doing that as a way to handle e.g. `/dev/vda3` being mounted, but we're trying to match vs a whole device like `/dev/vda`?

Just thinking through this...what is probably cleaner actually is to do a set-intersection; on the [calling side](https://github.com/containers/bootc/pull/805/files#diff-e44cbc45f98e122e00481bbcf78595a5c93935f6ef5bd0e4dd32c6880f1f2cfaR167) loop through all children of the device.

djach7: I'm not sure I follow what you're saying, my understanding is that we want to see if `/dev/vda` (the device) is mounted or if any partition on it is mounted. That's what I wrote it for, so it will kickback the mounting error if any part of the device is mounted. 

Are you saying that this should only fail if the device alone is mounted? As in it shouldn't be checking the partitions of the device?

That's the reasoning behind the `source.contains(path)`, I can't be sure what number partition will be mounted but I can work off it having the same naming stem.

lib/src/mount.rs
djach7: Updated, will resolve when pushed

lib/src/mount.rs
djach7: Updated, will resolve when pushed

lib/src/mount.rs
djach7: Updated, will resolve when pushed

lib/src/mount.rs
cgwalters: Or maybe cleaner actually, make `is_mounted_in_pid1_mountns` take a `&Device` and have it iterate over children.

djach7: So instead of looping through the `findmnt` output and searching for a matching device, loop through the device's children and check each to see if it's mounted?

djach7: That was my original thought a couple of weeks ago when I first looked at this, I ended up going with parsing the `findmnt` output because it seemed more straightforward for reasons I can't exactly remember now.

 For the code block you noted above, should mount checking be within the `opts.wipe` handling? Or are you just using that as an example of what the loop would look like?

cgwalters: Eh nevermind, no need to bikeshed this one fully, I think it's fine as is. Let's just do the minor tweaks and get it in!

djach7: Copy that, will push as soon as I run through my testing

cgwalters (approved):

PR #813, opened 2024-10-02
Update ostree-ext

cgwalters:
In particular, this gets us much improved SELinux handling.


cgwalters:
OK well no one reviewed this after 3 full working days, which is a bit unfortunate. This ended up getting merged via https://github.com/containers/bootc/pull/816 anyways.

PR #819, opened 2024-10-11
status: Output targeted message in non-bootc case, update docs

cgwalters:
- Output a clear message when we're not booted
- Update the docs and help to reflect current reality (e.g.
  format is stable) and include some guidance on use cases


djach7 (approved):
LGTM :+1:

PR #822, opened 2024-10-11
kargs: Skip xattrs in tests

cgwalters:
This breaks in situations where `/tmp` is on e.g. overlayfs which sadly is the podman/docker default.


jmarrero (approved):
lgtm

PR #823, opened 2024-10-11
docs/filesystem: Mention toplevels and mountpoints

cgwalters:
We want to be clear that toplevel directories for mountpoints need to be created in the container build.

Also, this moves the transient root and stateoverlay to markdown level 2, where they should have been.

Closes: https://github.com/containers/bootc/issues/814


jmarrero (approved):
lgtm

PR #827, opened 2024-10-15
lbi: Add --remove-signatures to install time lbi copy

### LINKED ISSUE #812: Unable to bind a signed image at install time
When bootc tries to copy the signed image from the host to the install disk it fails with the following error:

```
ERROR Installing to disk: Pulling from host storage: registry.redhat.io/amq-streams/kafka-37-rhel9:2.7.0-13: Subprocess failed: ExitStatus(unix_wait_status(32000))
Checking if image destination supports signatures
Error: Copying this image would require changing layer representation, which we cannot do: "Would invalidate signatures"
```

Adding "--remove-signatures" to the `podman image push` command [here](https://github.com/containers/bootc/blob/27b58c1944596b26d3050a0dec884c6d1648dbd9/lib/src/imgstorage.rs#L299) fixes this error, however the result is an unsigned image. I did some digging through the [containers/image code](https://github.com/containers/image/blob/687363a95930190f415ee0d9360804408d508269/copy/single.go#L128), skopeo docs, and [containers-storage.conf](https://github.com/containers/storage/blob/main/docs/containers-storage.conf.5.md#storage-options-for-overlay-table) docs. It looks like the only way to copy and sign an image is to re-sign the image when copying it, I couldn't find a way to copy a signed image while preserving the signature. I might be missing something though.

ckyrouac:
We are unable to copy a signed image from c/storage -> c/storage while preserving the signature. See
https://github.com/containers/image/issues/2599

fixes #812

### DIFF
--- a/lib/src/imgstorage.rs
+++ b/lib/src/imgstorage.rs
@@ -296,7 +296,7 @@ impl Storage {
         let storage_dest = &format!(
             "containers-storage:[overlay@{STORAGE_ALIAS_DIR}+/proc/self/fd/{STORAGE_RUN_FD}]"
         );
-        cmd.args(["image", "push", image])
+        cmd.args(["image", "push", "--remove-signatures", image])
             .arg(format!("{storage_dest}{image}"));
         let mut cmd = AsyncCommand::from(cmd);
         cmd.run().await?;
--- a/tests/booted/test-logically-bound-install.nu
+++ b/tests/booted/test-logically-bound-install.nu
@@ -6,5 +6,6 @@ print "IMAGES:"
 podman --storage-opt=additionalimagestore=/usr/lib/bootc/storage images # for debugging
 assert ($images | any {|item| $item.column1 == "quay.io/curl/curl"})
 assert ($images | any {|item| $item.column1 == "quay.io/curl/curl-base"})
+assert ($images | any {|item| $item.column1 == "registry.redhat.io/ubi9/podman"}) # this image is signed
 
 tap ok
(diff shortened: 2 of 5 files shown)


cgwalters (approved):

cgwalters:
>   - Downloading successful, but checksum doesn't match. Calculated: ef16d195332d673efbe2fc9a316726df177a605e1caef3ca49f88a7181f973654347400cb50ee464dd7206fc9a59c34cf37d0baea158add0bde531cb7cab8039(sha512)  Expected: b2b0cdc3fdc6021ff6ddd33a4b508daa7050cc8cdfca7dd9aa6c854ed80dd39b4fd5bfa6e66db2c37e0ad25cee9f53e7b24b84d7f302f2ca07198439524d1c99(sha512)

CI is failing on rpm-md repository metadata desynchronization. It's *exactly* stuff like this that would be greatly mitigated if we [put rpms in registries instead](https://github.com/rpm-software-management/librepo/pull/323).

cgwalters:
@ckyrouac I know we'd had flakes in the integration tests lately but this one looks to have introduced this:

> Trying to pull registry.redhat.io/ubi9/podman:latest...
Error: initializing source docker://registry.redhat.io/ubi9/podman:latest: unable to retrieve auth token: invalid username/password: unauthorized: Please login to the Red Hat Registry using your Customer Portal credentials. Further instructions can be found here: https://access.redhat.com/RegistryAuthentication

So we'll need to set up a pull secret to make this work.

PR #830, opened 2024-10-15
lib: Bump version to 1.1.0

cgwalters:
We'll support all APIs and features we have now for the forseeable future.

Why not 1.0.0? Just to avoid making this version feel too "special".

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -165,7 +165,7 @@ dependencies = [
 
 [[package]]
 name = "bootc-lib"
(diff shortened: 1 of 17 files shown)


jmarrero:
looks like CI does not like 1.1.0 :package: 

```
error: failed to select a version for the requirement `bootc-lib = "^0.1"`
candidate versions found which didn't match: 1.1.0
```

jmarrero (approved):
lgtm

PR #832, opened 2024-10-16
Stabilize logically bound images

cgwalters:
The feature is working as is and provides a lot of value, so let's stabilize it.

Closes: https://github.com/containers/bootc/issues/128

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -19,6 +19,7 @@
 
 - [Upgrade and rollback](upgrades.md)
 - [Accessing registries and offline updates](registries-and-offline.md)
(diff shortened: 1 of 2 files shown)


vrothberg (approved):
LGTM

I still believe that one config file would be great for the UX.  When managing many images at once, I believe it's just easier to manage than individual .image files.

cgwalters:
> I still believe that one config file would be great for the UX. When managing many images at once, I believe it's just easier to manage than individual .image files.

I think if we were to do something here I'd probably say instead https://github.com/containers/bootc/issues/833 which would be a lot more invasive, but also solve more problems.

PR #840, opened 2024-10-19
docs: Various elaborations

cgwalters:
- Clarify c/storage instance and LBIs in multiple places
- Flesh out/clarify some differences between bootc and rpm-ostree


jmarrero (approved):
lgtm

PR #848, opened 2024-10-27
 deploy: Add a spinner

cgwalters:
deploy: Use bar.println

In some cases I wasn't seeing this output, and I think it's
because we hadn't `drop()`'d the bar. Since it's in scope
it's cleaner to use its `println` method.

---

deploy: Add a spinner

First, this operation was absolutely not asynchronous
and we need to spawn it in a helper thread on general
principle. Doing that is unfortunately VERY hacky
because we need to clone a lot of things and especially
there's an ergonomics hit here since `Deployment` is
incorrectly `!Send`.

But the main motivation is this operation can be slow sometimes,
so let's show progress output so the admin knows we're not blocked.

Closes: https://github.com/containers/bootc/issues/842

---

utils: Print and log time for async_task_with_spinner

Things we do via this mechanism can take some time; let's
log the elapsed time for greater visibility both to the
human output *and* log to the journal.

---

### DIFF
--- a/lib/src/deploy.rs
+++ b/lib/src/deploy.rs
@@ -19,11 +19,13 @@ use ostree_ext::oci_spec::image::{Descriptor, Digest};
 use ostree_ext::ostree::Deployment;
 use ostree_ext::ostree::{self, Sysroot};
 use ostree_ext::sysroot::SysrootLock;
+use ostree_ext::tokio_util::spawn_blocking_cancellable_flatten;
 
 use crate::spec::ImageReference;
 use crate::spec::{BootOrder, HostSpec};
 use crate::status::labels_of_config;
 use crate::store::Storage;
+use crate::utils::async_task_with_spinner;
 
 // TODO use https://github.com/ostreedev/ostree-rs-ext/pull/493/commits/afc1837ff383681b947de30c0cefc70080a4f87a
 const BASE_IMAGE_PREFIX: &str = "ostree/container/baseimage/bootc";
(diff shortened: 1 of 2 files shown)


vrothberg (commented):
LGTM

lib/src/deploy.rs
```diff
@@ -384,26 +386,50 @@ async fn deploy(
...
+                &opts,
+                Some(cancellable),
+            )?;
+            Ok(d.index())
+        }),
+    )
+    .await?;
+    // SAFETY: We must have a staged deployment
+    let staged = sysroot.staged_deployment().unwrap();
+    assert_eq!(staged.index(), r);
```
omertuc, line 431: Any reason to go for assertion/unwrap as opposed to `ensure!`/`?`

omertuc (commented):
looks good except for one question

lib/src/deploy.rs
cgwalters: I think the general guideline is:

- If the condition can ever be reached through external input, then use `?`
- If the condition represents an internal consistency check, `unwrap()` is OK

This is the latter: there's no way to change this externally. As the comments say we're just using an integer because the `Deployment` isn't `Send`, and we're verifying what we got back when re-parsing the state is what we expected. Nothing else could have changed that externally.

omertuc: Fair enough

omertuc:
Oh and FYI, IDK if you care but you added `tracing::debug!("Deploying commit {ostree_commit}");` under a seemingly irrelevant commit (adc37abfaac47dad094b2b902e904adda0d42946)

cgwalters:
> Oh and FYI, IDK if you care but you added tracing::debug!("Deploying commit {ostree_commit}"); under a seemingly irrelevant commit (https://github.com/containers/bootc/commit/adc37abfaac47dad094b2b902e904adda0d42946)

I dropped that

cgwalters:
> /lgtm

This repository doesn't use Prow, so to merge it needs a Github approval state

omertuc (approved):

PR #849, opened 2024-10-27
cli: Reset executable name

cgwalters:
In some cases we re-exec ourselves; seeing the temporary file in there is super ugly. In general ideally in the future we support reliably querying journal logs from bootc; at least this improves things so that one can do
`journalctl _COMM=bootc`.


ckyrouac (approved):

cgwalters:
> error[E0658]: `c".."` literals are experimental

Ah I c...too old rust in c9s. Will fall back

PR #858, opened 2024-10-30
tmt: Update bootc-install.py plugin

ckyrouac:
This contains the latest plugin code from
https://github.com/teemtee/tmt/pull/3161. It includes automatically spinning up a podman-machine when podman is not rootful. It also fixes a bug introduced in the latest tmt version related to starting a web server via a systemd service.

### DIFF
--- a/hack/provision-derived.sh
+++ b/hack/provision-derived.sh
@@ -22,8 +22,10 @@ case "$variant" in
       # tmt wants rsync
       dnf -y install cloud-init rsync
       ln -s ../cloud-init.target /usr/lib/systemd/system/default.target.wants
-      # And tmt wants to write to /usr/local/bin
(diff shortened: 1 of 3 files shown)


cgwalters:
Looks like we get farther now but

```
                    Error: nu::shell::external_command
                    
                      × External command failed
                        ╭─[/var/tmp/tmt/run-002/plans/test-20-local-upgrade/tree/tests/booted/test-image-pushpull-upgrade.nu:52:5]
                     51 │     # And reboot into it
                     52 │     tmt-reboot
                        ·     ─────┬────
                        ·          ╰── executable was not found
                     53 │ }
                        ╰────
```

ckyrouac:
should be fixed now. (my local setup was picking up an old tmt version installed via pip)

cgwalters:
Every CI check is green! :heavy_check_mark:

PR #866, opened 2024-11-04
Cargo.lock: Bump tar and cap-std-ext

cgwalters:
We need the updated tar for
https://github.com/ostreedev/ostree-rs-ext/pull/679

And the updated cap-std-ext fixes an important bug.


jeckersb (approved):
:+1:

PR #871, opened 2024-11-05
List logically bound images

omertuc:
Solves the second part of https://github.com/containers/bootc/issues/846

The (hidden) image list command now has a `--type` flag to allow users to list only logical images, only host images, or all images.

Also the command has been adjusted so that it can run even when not booted off of a bootc system. In that case, it will only list logical images. If a user tries to list host images without a booted system, an error will be thrown.

The command also has a `--format` flag to allow users to choose between a human-readable table format and a JSON format.

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -46,6 +46,7 @@ toml = "0.8.12"
 xshell = { version = "0.2.6", optional = true }
 uuid = { version = "1.8.0", features = ["v4"] }
 tini = "1.3.0"
+comfy-table = "7.1.1"
 
 [dev-dependencies]
 indoc = { workspace = true }
--- a/lib/src/image.rs
+++ b/lib/src/image.rs
@@ -2,33 +2,122 @@
 //!
 //! APIs for operating on container images in the bootc storage.
 
-use anyhow::{Context, Result};
+use anyhow::{bail, Context, Result};
 use bootc_utils::CommandRunExt;
+use cap_std_ext::cap_std::{self, fs::Dir};
+use clap::ValueEnum;
+use comfy_table::{presets::NOTHING, Table};
 use fn_error_context::context;
 use ostree_ext::container::{ImageReference, Transport};
+use serde::Serialize;
 
-use crate::imgstorage::Storage;
+use crate::{
+    boundimage::query_bound_images,
+    cli::{ImageListFormat, ImageListType},
+};
 
 /// The name of the image we push to containers-storage if nothing is specified.
 const IMAGE_DEFAULT: &str = "localhost/bootc";
 
-#[context("Listing images")]
-pub(crate) async fn list_entrypoint() -> Result<()> {
-    let sysroot = crate::cli::get_storage().await?;
-    let repo = &sysroot.repo();
+#[derive(Clone, Serialize, ValueEnum)]
+enum ImageListTypeColumn {
+    Host,
+    Logical,
+}
 
-    let images = ostree_ext::container::store::list_images(repo).context("Querying images")?;
-
-    println!("# Host images");
-    for image in images {
-        println!("{image}");
+impl std::fmt::Display for ImageListTypeColumn {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.to_possible_value().unwrap().get_name().fmt(f)
     }
-    println!();
+}
 
-    println!("# Logically bound images");
-    let mut listcmd = sysroot.get_ensure_imgstore()?.new_image_cmd()?;
-    listcmd.arg("list");
-    listcmd.run()?;
+#[derive(Serialize)]
+struct ImageOutput {
+    image_type: ImageListTypeColumn,
+    image: String,
+    // TODO: Add hash, size, etc? Difficult because [`ostree_ext::container::store::list_images`]
+    // only gives us the pullspec.
+}
+
+#[context("Listing host images")]
+fn list_host_images(sysroot: &crate::store::Storage) -> Result<Vec<ImageOutput>> {
+    let repo = sysroot.repo();
+    let images = ostree_ext::container::store::list_images(&repo).context("Querying images")?;
+
+    Ok(images
+        .into_iter()
+        .map(|image| ImageOutput {
+            image,
+            image_type: ImageListTypeColumn::Host,
+        })
+        .collect())
+}
+
+#[context("Listing logical images")]
+fn list_logical_images(root: &Dir) -> Result<Vec<ImageOutput>> {
+    let bound = query_bound_images(root)?;
+
+    Ok(bound
+        .into_iter()
+        .map(|image| ImageOutput {
+            image: image.image,
+            image_type: ImageListTypeColumn::Logical,
+        })
+        .collect())
+}
+
+async fn list_images(list_type: ImageListType) -> Result<Vec<ImageOutput>> {
+    let rootfs = cap_std::fs::Dir::open_ambient_dir("/", cap_std::ambient_authority())
+        .context("Opening /")?;
+
+    let sysroot: Option<crate::store::Storage> =
+        if ostree_ext::container_utils::running_in_container() {
+            None
+        } else {
+            Some(crate::cli::get_storage().await?)
+        };
+
+    Ok(match (list_type, sysroot) {
+        // TODO: Should we list just logical images silently here, or error?
+        (ImageListType::All, None) => list_logical_images(&rootfs)?,
+        (ImageListType::All, Some(sysroot)) => list_host_images(&sysroot)?
+            .into_iter()
+            .chain(list_logical_images(&rootfs)?)
+            .collect(),
+        (ImageListType::Logical, _) => list_logical_images(&rootfs)?,
+        (ImageListType::Host, None) => {
+            bail!("Listing host images requires a booted bootc system")
+        }
+        (ImageListType::Host, Some(sysroot)) => list_host_images(&sysroot)?,
+    })
+}
+
+#[context("Listing images")]
+pub(crate) async fn list_entrypoint(
+    list_type: ImageListType,
+    list_format: ImageListFormat,
+) -> Result<()> {
+    let images = list_images(list_type).await?;
+
+    match list_format {
+        ImageListFormat::Table => {
+            let mut table = Table::new();
+
+            table
+                .load_preset(NOTHING)
+                .set_header(vec!["REPOSITORY", "TYPE"]);
+
+            for image in images {
+                table.add_row(vec![image.image, image.image_type.to_string()]);
+            }
+
+            println!("{table}");
+        }
+        ImageListFormat::Json => {
+            let mut stdout = std::io::stdout();
+            serde_json::to_writer_pretty(&mut stdout, &images)?;
+        }
+    }
 
     Ok(())
 }
@@ -79,7 +168,7 @@ pub(crate) async fn push_entrypoint(source: Option<&str>, target: Option<&str>) 
 /// Thin wrapper for invoking `podman image <X>` but set up for our internal
 /// image store (as distinct from /var/lib/containers default).
 pub(crate) async fn imgcmd_entrypoint(
-    storage: &Storage,
+    storage: &crate::imgstorage::Storage,
     arg: &str,
     args: &[std::ffi::OsString],
 ) -> std::result::Result<(), anyhow::Error> {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -176,6 +176,7 @@ dependencies = [
  "chrono",
  "clap",
  "clap_mangen",
+ "comfy-table",
  "fn-error-context",
  "hex",
  "indicatif",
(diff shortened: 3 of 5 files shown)


cgwalters (commented):
This looks sane to me to start!

I suspect though to really meet the goal we need to think about machine-readable output, which raises its own set of questions.

Also should be straightforward to change `test-logically-bound-install.nu` to cover this right?

omertuc:
>I suspect though to really meet the goal we need to think about machine-readable output, which raises its own set of questions.

I've played around with formats:

```
[omer@hal9000 ~]$ sudo bootc image list
 IMAGE TYPE   IMAGE
 host         docker://quay.io/otuchfel/bootc:listss
 logical      quay.io/curl/curl-base:latest
```

```
[omer@hal9000 ~]$ sudo bootc image list --help
List fetched images stored in the bootc storage.

Note that these are distinct from images stored via e.g. `podman`.

Usage: bootc image list [OPTIONS]

Options:
      --list-type <LIST_TYPE>
          Type of image to list

          [default: all]

          Possible values:
          - all:     List all images
          - logical: List only logically bound images
          - host:    List only host images

      --list-format <LIST_FORMAT>
          [default: table]

          Possible values:
          - table: Human readable table format
          - json:  JSON format
          - csv:   CSV format

  -h, --help
          Print help (see a summary with '-h')
```

```
[omer@hal9000 ~]$ sudo bootc image list --list-format json
[
  {
    "image_type": "Host",
    "image": "docker://quay.io/otuchfel/bootc:listss"
  },
  {
    "image_type": "Logical",
    "image": "quay.io/curl/curl-base:latest"
  }
]
```

```
[omer@hal9000 ~]$ sudo bootc image list --list-format csv
image_type,image
Host,docker://quay.io/otuchfel/bootc:listss
Logical,quay.io/curl/curl-base:latest
```

```
[omer@hal9000 ~]$ sudo bootc image list --list-format json --list-type logical
[
  {
    "image_type": "Logical",
    "image": "quay.io/curl/curl-base:latest"
  }
]
```

```
[omer@hal9000 ~]$ sudo bootc image list --list-format json --list-type logical | jq '.[].image' -r
quay.io/curl/curl-base:latest
```

omertuc:
>Also should be straightforward to change test-logically-bound-install.nu to cover this right?

Seems so, I'm trying to get at-least one successful run of the `tmt` tests on a clean `main` branch, then I'll see how I can extend the `.nu` script

cgwalters:
I don't see the need to have both csv and json, let's pick json.

This overlaps a bit with https://github.com/containers/bootc/issues/522 in that in theory we could expose this via varlink too? But eh, json seems like a sane baseline.

omertuc:
> I don't see the need to have both csv and json, let's pick json.

Removed CSV (it was trivial to add and to remove so no harm done), it was mostly just an example of how we can add formats easily anyway

omertuc:
>Seems so, I'm trying to get at-least one successful run of the tmt tests on a clean main branch, then I'll see how I can extend the .nu script

Extending a bootc image with the following commands seems to be enough to prepare it , once booted on a VM, for cloning the bootc repo and running `make test-tmt` successfully:

```Containerfile
FROM localhost/bootc

RUN dnf install -y --setopt=install_weak_deps=False cargo git genisoimage libvirt python-pip qemu-kvm rsync && \
    dnf clean all && \
    rm -rf /var/cache/dnf /var/log/dnf* /var/log/yum.*

RUN python3 -m pip install testcloud tmt && \
    rm -rf ~/.cache/pip

RUN systemctl enable virtqemud.socket
```

cgwalters:
> Is anything like this documented anywhere? Am I even doing it right or am I missing some obvious simpler way?

I run most things from a toolbox container personally where I have this stuff already installed. What you're doing there is "recursive" which is cool, but I am unsure I would have that be a primary target.

lib/src/image.rs
```diff
@@ -2,33 +2,112 @@
...
+fn list_logical_images(sysroot: &crate::store::Storage) -> Result<Vec<ImageOutput>> {
+    let output = sysroot
+        .get_ensure_imgstore()?
+        .new_image_cmd()?
+        .arg("list")
+        .arg("--format={{.Repository}}:{{.Tag}}")
+        .log_debug()
+        .output()?;
+
+    ensure!(
```
cgwalters, line 65: Can you rebase on https://github.com/containers/bootc/pull/880 ? There's quite a lot of subtleties in subprocesses and that helper code gets more things right (including handling truncating stderr at a reasonable size, etc.)

lib/src/image.rs
```diff
@@ -2,33 +2,112 @@
...
+        .collect())
+}
+
+#[context("Listing logical images")]
+fn list_logical_images(sysroot: &crate::store::Storage) -> Result<Vec<ImageOutput>> {
+    let output = sysroot
+        .get_ensure_imgstore()?
+        .new_image_cmd()?
+        .arg("list")
+        .arg("--format={{.Repository}}:{{.Tag}}")
```
cgwalters, line 61: Alternatively we could use `--format=json`

lib/src/image.rs
```diff
@@ -2,33 +2,112 @@
...
-    println!("# Host images");
-    for image in images {
-        println!("{image}");
+    match list_format {
+        ImageListFormat::Table => {
+            println!("{}", tabled::Table::new(images).with(Style::blank()));
+        }
+        ImageListFormat::Json => {
+            let json = serde_json::to_string_pretty(&images)?;
+            println!("{}", json);
```
cgwalters, line 108: It's a lot more efficient to directly write to stdout, we do this elsewhere:

```
                let mut stdout = std::io::stdout().lock();
                serde_json::to_writer_pretty(&mut stdout, &images)?;
```

lib/src/image.rs
```diff
@@ -2,33 +2,112 @@
...
+) -> Result<()> {
+    // TODO: Get the storage from the container image, not the booted storage
+    let sysroot: crate::store::Storage = crate::cli::get_storage().await?;
 
-    let images = ostree_ext::container::store::list_images(repo).context("Querying images")?;
+    let images = match list_type {
+        ImageListType::All => {
+            let images = list_host_images(&sysroot)?;
+            let images = images.into_iter().chain(list_logical_images(&sysroot)?);
+            images.collect()
```
cgwalters, line 96: This is fine as is but I would probably write this as
```
let mut images = list_host_images(&sysroot);
images.extend(list_logical_images(&sysroot)?);
images
```

cgwalters (commented):
A few nits but looking sane to me!

lib/Cargo.toml
```diff
@@ -46,6 +46,7 @@ toml = "0.8.12"
 xshell = { version = "0.2.6", optional = true }
 uuid = { version = "1.8.0", features = ["v4"] }
 tini = "1.3.0"
+tabled = "0.16.0"
```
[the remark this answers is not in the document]
omertuc: I did zero analysis, I will have a look if there's any alternative that's lighter / less dependent / more popular / better maintained / etc

lib/src/image.rs
omertuc: That was my initial attempt, but that doesn't contain any trivially accessible equivalent of `{{.Repository}}:{{.Tag}}` that I could find. In the Go Template [handling code](https://github.com/containers/podman/blob/72f1617facbb83fbc1e74583f62394dfc6142109/cmd/podman/images/list.go#L190-L193), `Repository` and `Tag` are coming from the [imageReporter](https://github.com/containers/podman/blob/72f1617facbb83fbc1e74583f62394dfc6142109/cmd/podman/images/list.go#L336-L340) struct, but the JSON [handling code dumps a different struct called `image`](https://github.com/containers/podman/blob/72f1617facbb83fbc1e74583f62394dfc6142109/cmd/podman/images/list.go#L164-L168) which, when [populated from `imageReporter`](https://github.com/containers/podman/blob/72f1617facbb83fbc1e74583f62394dfc6142109/cmd/podman/images/list.go#L173-L176), doesn't copy the `Repository` and `Tag` fields

lib/src/image.rs
omertuc: Went for

```rust
        ImageListType::All => list_host_images(&sysroot)?
            .into_iter()
            .chain(list_logical_images(&sysroot)?)
            .collect(),
```

lib/src/image.rs
omertuc: Done

lib/Cargo.toml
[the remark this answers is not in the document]
omertuc: Switched to `comfy-table` as it seems just a bit more popular, also felt a bit simpler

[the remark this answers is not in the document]
cgwalters: I'm comfytable with that.

lib/src/image.rs
```diff
@@ -2,33 +2,122 @@
...
+}
+
+#[context("Listing logical images")]
+fn list_logical_images(root: &Dir) -> Result<Vec<ImageOutput>> {
+    let bound = query_bound_images(root)?;
+
+    Ok(bound
+        .iter()
+        .map(|x| ImageOutput {
+            image: x.image.clone(),
```
cgwalters, line 63: minor: I think if we used `into_iter` above, then we could drop the `clone` here

lib/src/image.rs
```diff
@@ -2,33 +2,122 @@
...
+async fn list_images(list_type: ImageListType) -> Result<Vec<ImageOutput>> {
+    let rootfs = cap_std::fs::Dir::open_ambient_dir("/", cap_std::ambient_authority())
+        .context("Opening /")?;
+
+    let sysroot: Option<crate::store::Storage> =
+        if ostree_ext::container_utils::running_in_container() {
+            None
+        } else {
+            Some(crate::cli::get_storage().await?)
+        };
```
cgwalters, line 78: Hum, interesting. This one might be a candidate for a helper in `imgstorage.rs`?

tests-integration/src/container.rs
```diff
@@ -23,7 +23,7 @@ pub(crate) fn test_bootc_upgrade() -> Result<()> {
         assert!(!st.success());
         let stderr = String::from_utf8(o.stderr)?;
         assert!(
-            stderr.contains("this command requires a booted host system"),
+            stderr.contains("This command requires a booted host system"),
```
cgwalters, line 26: This change to this "this" broke the integration test; doesn't seem related.

lib/src/image.rs
```diff
@@ -2,33 +2,122 @@
...
+    // only gives us the pullspec.
+}
+
+#[context("Listing host images")]
+fn list_host_images(sysroot: &crate::store::Storage) -> Result<Vec<ImageOutput>> {
+    let repo = sysroot.repo();
+    let images = ostree_ext::container::store::list_images(&repo).context("Querying images")?;
+
+    Ok(images
+        .iter()
```
cgwalters, line 48: Ditto re `into_iter`

lib/src/image.rs
```diff
@@ -2,33 +2,122 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.to_possible_value().unwrap().get_name().fmt(f)
+    }
+}
+
+#[derive(Serialize)]
+struct ImageOutput {
+    image_type: ImageListTypeColumn,
+    image: String,
+    // TODO: Add hash, size, etc? Difficult because [`ostree_ext::container::store::list_images`]
```
cgwalters, line 38: We could query the hash; size...is not directly exposed today, but we could.

cgwalters (changes requested):

tests-integration/src/container.rs
omertuc: yeah whoops, removed the `t` arbitrarily to make sure I'm running what I'm editing, but then put it back the wrong way...

lib/src/image.rs
omertuc: But it would require changes in the ostree ext project right?

lib/src/image.rs
omertuc: I personally can't imagine how this would be useful anywhere else, the `None` was mostly to control flow in this function

lib/src/image.rs
cgwalters: Yeah (though that project is now folded in here) but to be clear my comment was just an aside - we should definitely not block on it!

lib/src/image.rs
```diff
@@ -2,33 +2,122 @@
...
+        .into_iter()
+        .map(|image| ImageOutput {
+            image: image.image,
+            image_type: ImageListTypeColumn::Logical,
+        })
+        .collect())
+}
+
+async fn list_images(list_type: ImageListType) -> Result<Vec<ImageOutput>> {
+    let rootfs = cap_std::fs::Dir::open_ambient_dir("/", cap_std::ambient_authority())
```
cgwalters, line 70: This is OK as is but generally so far I've been trying to do things like this closer to `cli.rs`, keeping functions in other places expecting ambient state.

cgwalters (approved):

PR #874, opened 2024-11-05
Merge bootupd

cgwalters:
Initial draft for https://github.com/coreos/bootupd/issues/432


Conan-Kudo:
> bootupd hasn't really seen usage outside of the ostree sphere, and maintaining it as its own independent thing has huge overhead (e.g. constant dependabot things, CI maintenance, releases)

Doing this will preclude any consideration for adoption outside of Atomic systems. I had been talking to @travier about doing this for regular Fedora sometime in the next year or so, but if it's merged inside of bootc, I will abandon interest in that.

cgwalters:
Closing for now as there is an offer to help maintain bootupd separately.

PR #878, opened 2024-11-06
tests: Drop python tests

cgwalters:
It's not useful to have two frameworks here, let's just use nushell. ref https://github.com/containers/bootc/issues/868


mvo5 (approved):
Thank you, it makes sense to focus on one!

PR #881, opened 2024-11-07
install: Check if running in container earlier

ckyrouac:
Previously, the check to verify install to-filesystem is being run
within a container happened fairly late in prepare_install. This moves
the check up before some of the other container specific checks. Now,
the error should occur when trying a basic install to-filesystem
invocation, e.g. `bootc install to-filesystem /var/mnt`.

This also adds a test to verify the error occurs with minimal arguments
and adds host_is_container to the global state.


lib/src/install.rs
```diff
@@ -1321,6 +1324,12 @@ async fn install_with_sysroot(
 }
 
 async fn install_to_filesystem_impl(state: &State, rootfs: &mut RootSetup) -> Result<()> {
+    // this is to prevent the compiler from complaining about unused variables
+    // currently, host_is_container will always be true at this point
+    if !state.host_is_container {
+        tracing::debug!("not in a container");
```
cgwalters, line 1330: I think we should actually error out, right? We basically silently do the wrong thing today by default. i.e. something like:

```
anyhow::bail!("installation must be run in a container")
```

(Though when we do that we also need to ignore this if `--source-imageref` is defined)

lib/src/containerenv.rs
```diff
@@ -20,6 +20,10 @@ pub(crate) struct ContainerExecutionInfo {
     pub(crate) rootless: Option<String>,
 }
 
+pub(crate) fn is_container(rootfs: &Dir) -> bool {
+    rootfs.exists(PATH)
```
cgwalters, line 24: Hmm but we have `get_container_execution_info()` (defined just below) that we should be calling but somehow aren't fatally erroring out on right?

I'm confused since it sure looks like we *should* be calling that function and erroring out when run directly...

cgwalters:
What we also should do here I think is add a test case...mmm in the nushell bits should be easy that verifies we *do* currently get an error.

lib/src/install.rs
ckyrouac: yea I can add that to be double sure. I think [this](https://github.com/containers/bootc/blob/ad95463ee2f5877bd21b4c971b0c798846d48b58/lib/src/install.rs#L1166) should cause it to bail out before getting to this point. I added this call here to avoid the compiler error for now while I work on the other “real” checks.

lib/src/containerenv.rs
ckyrouac: yea, I can add a test case to make sure we bail out. I tested it manually and it was working. I split this into a separate function just so it can be added to the global state and eventually used to toggle selinux/kargs/etc. checks.

ckyrouac:
reworked this to focus on more clearly validating if install to-filesystem is running in a container.

cgwalters (approved):
Nice, just some minor nonblocking nits for the test

PR #882, opened 2024-11-07
install: Stop reading kargs from container root, use ostree

cgwalters:
Part of https://github.com/containers/bootc/issues/879

Basically we want to be able to `bootc install` outside of a container. For the same reasons actually that we already support parsing kargs from an ostree commit (without materializing it as a filesystem), just expose that API via `pub(crate)` and use it in between the "pull" and "deploy" phases.

We basically do the same thing on `bootc upgrade`.


cgwalters:
(Existing tests cover this)

cgwalters:
```
---- replace=alongside with ssh keys and a karg, and SELinux disabled ----
command exited with non-zero code `sudo /bin/sh -c "grep testing-kargsd=3 /boot/loader/entries/*.conf"`: 1
```

huh...digging

ckyrouac (approved):

PR #887, opened 2024-11-08
store: Support importing images without `/ostree`

cgwalters:
A sticking point keeping ostree in the picture here for containers was SELinux handling. When we started this effort I'd feared rewriting.

But recently we changed things such that we label derived images using the policy from the final root.

This is a relatively small change in code size and complexity, that allows us to import images that don't have "ostree stuff" in them at all, i.e. there's no `/ostree/repo/objects`.

The advantage here is that this significantly simplifies constructing base images.

The main disadvantage today for people who build images this way is that we end up re-labeling and re-checksumming all objects.

But, the real fix for that in the future will be for us to rework things such that we support `security.selinux` for example as native xattrs in the tar stream.

### DIFF
--- a/ostree-ext/src/container/store.rs
+++ b/ostree-ext/src/container/store.rs
@@ -227,7 +227,7 @@ pub struct PreparedImport {
     /// The layers containing split objects
     pub ostree_layers: Vec<ManifestLayerState>,
     /// The layer for the ostree commit.
-    pub ostree_commit_layer: ManifestLayerState,
+    pub ostree_commit_layer: Option<ManifestLayerState>,
     /// Any further non-ostree (derived) layers.
     pub layers: Vec<ManifestLayerState>,
 }
@@ -235,7 +235,8 @@ pub struct PreparedImport {
 impl PreparedImport {
     /// Iterate over all layers; the commit layer, the ostree split object layers, and any non-ostree layers.
     pub fn all_layers(&self) -> impl Iterator<Item = &ManifestLayerState> {
-        std::iter::once(&self.ostree_commit_layer)
+        self.ostree_commit_layer
+            .iter()
             .chain(self.ostree_layers.iter())
             .chain(self.layers.iter())
     }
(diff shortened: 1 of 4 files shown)


cgwalters:
This was migrated from https://github.com/ostreedev/ostree-rs-ext/pull/674 btw.

jmarrero (approved):
lgtm

if I understand correctly, there is still a requirement then for the container image to be composefs not just any image can be used? Or any image like ubi could end up being used here?

cgwalters:
> if I understand correctly, there is still a requirement then for the container image to be composefs not just any image can be used? Or any image like ubi could end up being used here?

composefs is something we generate client side; it's not directly related *except* in that today one needs to explicitly turn on ostree-composefs in the container image's content.

>  Or any image like ubi could end up being used here?

See https://gitlab.com/fedora/bootc/tracker/-/issues/32#note_2201993683 - so yes, one could derive from ubi too.

cgwalters:
I was asked about this PR recently, and just to clarify: I'd consider this PR "supported" in the sense that we will absolutely fix any bugs that arise, and we do have some tests, but it's just less battle tested than the previous path.

At some point we will probably try to change fedora-bootc to do this by default, but we have to be really cautious because we need to support people upgrading from old systems where the bootc won't support this.

So the overall goal is basically:

- Make it easier to onboard fresh new base images without `/ostree` stuff that don't have backcompat concerns
- Let this support trickle back into existing systems using bootc and test it out more, planning to rework to a new image format when we're confident most systems-being-upgraded-from have the support

(This issue also touches on https://github.com/containers/bootc/issues/640 in that in theory it could help us get out of these bootstrap requirements; the bootc-in-target could handle any mixups)

cgwalters:
To extend the previous comment, we now test the "no ostree in container" flow pretty well too in various places, it's the default output of the https://docs.fedoraproject.org/en-US/bootc/building-custom-base/ flow.

PR #890, opened 2024-11-08
tmt: Add TMT_TEST env variable to specify single test to run

ckyrouac:
e.g. `TMT_TEST=test-21-logically-bound-switch make test-tmt` will run all the tmt setup code then run test-21-logically-bound-switch. This is useful when developing a test.

### DIFF
--- a/xtask/src/xtask.rs
+++ b/xtask/src/xtask.rs
@@ -155,7 +155,7 @@ fn update_generated(sh: &Shell) -> Result<()> {
 }
 
 #[context("test-integration")]
-fn test_tmt(sh: &Shell) -> Result<()> {
+fn all_plan_files(sh: &Shell) -> Result<Vec<(u32, String)>> {
     // We need to split most of our tests into separate plans because tmt doesn't
     // support automatic isolation. (xref)
     let mut all_plan_files =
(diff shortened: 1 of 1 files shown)


cgwalters:
Yes! I was literally looking for this 10 minutes ago :smile:

xtask/src/xtask.rs
```diff
@@ -186,14 +186,26 @@ fn test_tmt(sh: &Shell) -> Result<()> {
...
+}
+
+#[context("test-integration")]
+fn test_tmt(sh: &Shell) -> Result<()> {
+    let tests = match std::env::var("TMT_TEST") {
+        Ok(test) => {
+            vec![(0, test)]
+        }
+        Err(_) => all_plan_files(sh)?,
+    };
```
cgwalters, line 199: I think this can be `let tests = std::env::var("TMT_TEST").unwrap_or_else(|| vec![(0, test]));`

cgwalters (approved):
OK as is, just an optional suggestion

xtask/src/xtask.rs
```diff
@@ -186,14 +186,26 @@ fn test_tmt(sh: &Shell) -> Result<()> {
...
+
+    println!("Discovered plans: {tests:?}");
```
cgwalters, line 201: Now when `TMT_TEST` is set we'll now only say that we discovered what they asked for right?

How about doing this like:
```
let mut tests = all_plan_files()?;
if let Ok(name) std::env::var("TMT_TEST") {
    tests.retain(|x| x.1.as_str() == name);
    if tests.is_empty() {
        anyhow::bail!("Failed to match test: {name}");
    }
}
```

or so

ckyrouac: good idea, updated

PR #891, opened 2024-11-08
kargs: Handle when kargs.d doesn't exist

cgwalters:
I'm experimenting with custom base images and hit on this not existing.


jmarrero (approved):
lgtm

PR #905, opened 2024-11-18
install: Some cleanups around root_path

cgwalters:
The extra cloning we were doing here looked odd to me. I think this ends up being cleaner make the toplevel variable mutable and replace it only in the circumstance we detect an ostree deployment.

### DIFF
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -1636,40 +1636,42 @@ pub(crate) async fn install_to_filesystem(
     opts: InstallToFilesystemOpts,
     targeting_host_root: bool,
 ) -> Result<()> {
-    let fsopts = opts.filesystem_opts;
-    let root_path = &fsopts.root_path;
+    let mut fsopts = opts.filesystem_opts;
 
-    let st = root_path
-        .symlink_metadata()
-        .with_context(|| format!("Querying target filesystem {root_path}"))?;
-    if !st.is_dir() {
-        anyhow::bail!("Not a directory: {root_path}");
+    // Check that the target is a directory
+    {
+        let root_path = &fsopts.root_path;
+        let st = root_path
+            .symlink_metadata()
+            .with_context(|| format!("Querying target filesystem {root_path}"))?;
+        if !st.is_dir() {
+            anyhow::bail!("Not a directory: {root_path}");
+        }
     }
 
+    // If we're installing to an ostree root, then find the physical root from
+    // the deployment root.
     let possible_physical_root = fsopts.root_path.join("sysroot");
     let possible_ostree_dir = possible_physical_root.join("ostree");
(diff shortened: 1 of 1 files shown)


lib/src/install.rs
```diff
@@ -1636,40 +1636,42 @@ pub(crate) async fn install_to_filesystem(
...
 
+    // If we're installing to an ostree root, then find the physical root from
+    // the deployment root.
     let possible_physical_root = fsopts.root_path.join("sysroot");
     let possible_ostree_dir = possible_physical_root.join("ostree");
-    let root_path = if possible_ostree_dir.exists() {
+    if possible_ostree_dir.exists() {
         tracing::debug!(
-            "ostree detected in {possible_ostree_dir}, assuming / is a deployment root and using {possible_physical_root} instead of {root_path} as target root"
+            "ostree detected in {possible_ostree_dir}, assuming / is a deployment root and using {possible_physical_root} as target root"
```
omertuc, line 1658: Maybe even

```suggestion
            "ostree detected in {possible_ostree_dir}, assuming {root_path} is a deployment root and using {possible_physical_root} as target root"
```

omertuc (changes requested):
Looks good

omertuc:
>The extra cloning we were doing here looked odd to me. I think this ends up being cleaner make the toplevel variable mutable and replace it only in the circumstance we detect an ostree deployment.

I personally find mutations harder to follow as a code reader/debugger, even if they're sometimes "simpler" (mostly in terms of LoC) and more performant, but it's all very subjective

omertuc (approved):

lib/src/install.rs
```diff
@@ -1636,40 +1636,42 @@ pub(crate) async fn install_to_filesystem(
...
 
+    // If we're installing to an ostree root, then find the physical root from
+    // the deployment root.
     let possible_physical_root = fsopts.root_path.join("sysroot");
     let possible_ostree_dir = possible_physical_root.join("ostree");
-    let root_path = if possible_ostree_dir.exists() {
+    if possible_ostree_dir.exists() {
         tracing::debug!(
-            "ostree detected in {possible_ostree_dir}, assuming / is a deployment root and using {possible_physical_root} instead of {root_path} as target root"
+            "ostree detected in {possible_ostree_dir}, assuming {root_path} is a deployment root and using {possible_physical_root} as target root"
```
omertuc, line 1658: Oh sorry, there's no `root_path`

omertuc (changes requested):

cgwalters:
> I personally find mutations harder to follow as a code reader/debugger, even if they're sometimes "simpler" (mostly in terms of LoC) and more performant, but it's all very subjective

Agreed that the mutation can be confusing, but I think in this case it is clearer than creating a shadowed copy of the same value.

However I have some other followups here

PR #909, opened 2024-11-19
install: Two rootfs cleanups

cgwalters:
install: Reduce usage of absolute path for rootfs

In the install flow we have both `rootfs` and `rootfs_fd` which
hold the physical root. Using fd-relative accesses where we
can provides a lot of advantages, so switch most uses over
to the file descriptor.

---

install: Rename rootfs -> physical_root

In the install flow we juggle *three* file systems in general:

- The container/host root
- The physical root
- The deployment root

"rootfs" in theory could be any of those three. In the install code
it's the physical (target) root, so rename the variable
to clarify.

---

### DIFF
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -12,7 +12,7 @@ mod osbuild;
 pub(crate) mod osconfig;
 
 use std::io::Write;
-use std::os::fd::AsFd;
+use std::os::fd::{AsFd, AsRawFd};
 use std::os::unix::process::CommandExt;
 use std::path::Path;
 use std::process::Command;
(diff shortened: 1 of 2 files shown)


cgwalters:
```
                    
                      × Assignment operations require a variable.
                        ╭─[/var/tmp/tmt/run-002/plans/test-20-local-upgrade/tree/tests/booted/test-image-pushpull-upgrade.nu:86:5]
                     85 │     let booted_digest = $booted.imageDigest
                     86 │     print booted_digest = $booted_digest
                        ·     ─────────┬─────────
                        ·              ╰── needs to be a variable
                     87 │     # We should already be fetching updates from container storage
                        ╰────
                      help: try assigning to a variable or a cell path of a variable
                    
```

Hmm, I think the nushell update here in EPEL broke our test scripts. Looking at this

cgwalters:
https://github.com/nushell/nushell/issues/14333 is also related to this I think

jeckersb (approved):
lgtm

PR #910, opened 2024-11-19
install: Fix broken warn_on_host_root check

omertuc:
The `warn_on_host_root` check was broken when we added support for installing on already-ostree systems.

See https://github.com/containers/bootc/issues/907

The solution is to use the original user provided root_path for the fd passed to warn_on_host_root, rather than the modified one, as that will always match /proc/0/root's fsid (in ostree systems systemd is running with the deployment root as its root, and this is what we have mounted as /:/target)

### DIFF
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -1716,6 +1716,14 @@ pub(crate) async fn install_to_filesystem(
         }
     }
 
+    // Check to see if this happens to be the real host root
+    if !fsopts.acknowledge_destructive {
+        let root_path = &fsopts.root_path;
+        let rootfs_fd = Dir::open_ambient_dir(root_path, cap_std::ambient_authority())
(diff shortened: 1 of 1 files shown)


cgwalters (commented):
Thanks for spotting this regression...clearly we should have some CI coverage that verifies we print the warning.

This code is quite complex and entangled now. I am wondering if we couldn't fix this by just moving the warning up *above* where we re-compute and `fsopts.root_path`.

Actually also staring at this code, I did https://github.com/containers/bootc/pull/912

cgwalters:
> Yeah I wanted to do that, but that would mean the check would run before prepare_install which would change the behavior and I wasn't confident this is desired. I'll gladly change the implementation if you think this is not a problem

https://github.com/containers/bootc/pull/912 would address that right?

cgwalters (approved):

PR #911, opened 2024-11-19
tests: Fix syntax error from newer nushell

cgwalters:
Updated versions of nu make this a syntax error apparently.


jeckersb (approved):
I will totally take your word for it on this one, having never touched any of the nushell stuff before :smile:

PR #912, opened 2024-11-19
install: Move re-exec earlier

cgwalters:
As the comment says everything we do before
`prepare_install` is something we might do *twice* so let's keep it as early as possible.

Looking at the history of this code (and its logical inputs) I believe there's no reason for its placement in the middle of this function other than just "stuff appeared
before it organically".

Yes, it means some argument validation will happen after the re-exec but that's fine.


jeckersb (approved):
:+1:

PR #914, opened 2024-11-19
clippy: Deny `str::len`

cgwalters:
When I was working on some column printing code
with Unicode I got bit by using `str::len`...and
digging in I found that clippy actually just
merged a lint to go the *other* way; more in
the link in the code.

Turning on a lint showed one place that should
have been using `chars().count()` and one that
should have been validating ASCII. Fix those.


jeckersb (approved):
Strings are hard.  At least rust makes you confront that reality upfront!

PR #915, opened 2024-11-20
install: Add `ensure-completion` verb, wire up ostree-deploy → bootc

cgwalters:
When bootc was created, it started to become a superset of ostree;
in particular things like `/usr/lib/bootc/kargs.d` and logically
bound images.

However...Anaconda today is still invoking `ostree container image deploy`.

Main fix
--------

When bootc takes over the `/usr/libexec/ostree/ext/ostree-container`
entrypoint, make the existing `ostree container image deploy` CLI actually
just call back into bootc to fix things up. No additional work required other
than getting an updated bootc in the Anaconda ISO.

Old Anaconda ISOs
-----------------

But, a further problem here is that Anaconda is only updated once
per OS major+minor - e.g. there won't be an update to it for the lifetime
of RHEL 9.5 or Fedora 41. We want the ability to ship new
features and bugfixes in those OSes (especially RHEL9.5).

So given that we have a newer bootc in the target container, we can
do this:

```
%post --erroronfail
bootc install ensure-completion
%end
```

And will fix things up. Of course there's fun $details here...the
way Anaconda implements `%post` is via a hand-augmented `chroot`
i.e. a degenerate container, and we need to escape that and
fix some things up (such as a missing cgroupfs mount).

Summmary
--------

- With a newer bootc in the ISO, everything just works
- For older ISOs, one can add the `%post` above as a workaround.

Implementation details: Cross-linking bootc and ostree-rs-ext
-------------------------------------------------------------

This whole thing is very confusing because now, the linkage
between bootc and ostree-rs-ext is bidirectional. In the case
of `bootc install to-filesystem`, we end up calling into ostree-rs-ext,
and we *must not* recurse back into bootc, because at least for
kernel arguments we might end up applying them *twice*. We do
this by passing a CLI argument.

The second problem is the crate-level dependency; right now they're
independent crates so we can't have ostree-rs-ext actually
call into bootc directly, as convenient as that would be. So we
end up forking ourselves as a subprocess. But that's not too bad
because we need to carry a subprocess-based entrypoint *anyways*
for the Anaconda `%post` case.

Implementation details: /etc/resolv.conf
----------------------------------------

There's some surprising stuff going on in how Anaconda handles
`/etc/resolv.conf` in the target root that I got burned by. In
Fedora it's trying to query if systemd-resolved is enabled in
the target or something?

I ended up writing some code to just try to paper over this
to ensure we have networking in the `%post` where we need
it to fetch LBIs.

---

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -182,6 +182,15 @@ pub(crate) enum InstallOpts {
     /// will be wiped, but the content of the existing root will otherwise be retained, and will
     /// need to be cleaned up if desired when rebooted into the new root.
     ToExistingRoot(crate::install::InstallToExistingRootOpts),
+    /// Intended for use in environments that are performing an ostree-based installation, not bootc.
+    ///
+    /// In this scenario the installation may be missing bootc specific features such as
+    /// kernel arguments, logically bound images and more. This command can be used to attempt
+    /// to reconcile. At the current time, the only tested environment is Anaconda using `ostreecontainer`
+    /// and it is recommended to avoid usage outside of that environment. Instead, ensure your
+    /// code is using `bootc install to-filesystem` from the start.
+    #[clap(hide = true)]
+    EnsureCompletion {},
     /// Output JSON to stdout that contains the merged installation configuration
     /// as it may be relevant to calling processes using `install to-filesystem`
     /// that in particular want to discover the desired root filesystem type from the container image.
@@ -345,6 +354,15 @@ pub(crate) enum InternalsOpts {
     OstreeContainer {
         #[clap(allow_hyphen_values = true)]
         args: Vec<OsString>,
+    },
+    #[cfg(feature = "install")]
+    /// Invoked from ostree-ext to complete an installation.
+    BootcInstallCompletion {
+        /// Path to the sysroot
+        sysroot: Utf8PathBuf,
+
+        // The stateroot
+        stateroot: String,
     },
 }
 
@@ -989,6 +1007,10 @@ async fn run_from_opt(opt: Opt) -> Result<()> {
                 crate::install::install_to_existing_root(opts).await
             }
             InstallOpts::PrintConfiguration => crate::install::print_configuration(),
+            InstallOpts::EnsureCompletion {} => {
+                let rootfs = &Dir::open_ambient_dir("/", cap_std::ambient_authority())?;
+                crate::install::completion::run_from_anaconda(rootfs).await
+            }
         },
         #[cfg(feature = "install")]
         Opt::ExecInHostMountNamespace { args } => {
@@ -1025,6 +1047,11 @@ async fn run_from_opt(opt: Opt) -> Result<()> {
             InternalsOpts::Cleanup => {
                 let sysroot = get_storage().await?;
                 crate::deploy::cleanup(&sysroot).await
+            }
+            #[cfg(feature = "install")]
+            InternalsOpts::BootcInstallCompletion { sysroot, stateroot } => {
+                let rootfs = &Dir::open_ambient_dir("/", cap_std::ambient_authority())?;
+                crate::install::completion::run_from_ostree(rootfs, &sysroot, &stateroot).await
             }
         },
         #[cfg(feature = "docgen")]
--- a/lib/src/install/completion.rs
+++ b/lib/src/install/completion.rs
@@ -1,0 +1,301 @@
+//! This module handles finishing/completion after an ostree-based
+//! install from e.g. Anaconda.
+
+use std::io;
+use std::os::fd::AsFd;
+use std::process::Command;
+
+use anyhow::{Context, Result};
+use bootc_utils::CommandRunExt;
+use camino::Utf8Path;
+use cap_std_ext::{cap_std::fs::Dir, dirext::CapStdExtDirExt};
+use fn_error_context::context;
+use ostree_ext::{gio, ostree};
+use rustix::fs::Mode;
+use rustix::fs::OFlags;
+
+use super::config;
+
+/// An environment variable set by anaconda that hints
+/// we are running as part of that environment.
+const ANACONDA_ENV_HINT: &str = "ANA_INSTALL_PATH";
+/// The path where Anaconda sets up the target.
+/// <https://anaconda-installer.readthedocs.io/en/latest/mount-points.html#mnt-sysroot>
+const ANACONDA_SYSROOT: &str = "mnt/sysroot";
+/// Global flag to signal we're in a booted ostree system
+const OSTREE_BOOTED: &str = "run/ostree-booted";
+/// The very well-known DNS resolution file
+const RESOLVCONF: &str = "etc/resolv.conf";
+/// A renamed file
+const RESOLVCONF_ORIG: &str = "etc/resolv.conf.bootc-original";
+/// The root filesystem for pid 1
+const PROC1_ROOT: &str = "proc/1/root";
+/// The cgroupfs mount point, which we may propagate from the host if needed
+const CGROUPFS: &str = "sys/fs/cgroup";
+/// The path to the temporary global ostree pull secret
+const RUN_OSTREE_AUTH: &str = "run/ostree/auth.json";
+/// A sub path of /run which is used to ensure idempotency
+pub(crate) const RUN_BOOTC_INSTALL_RECONCILED: &str = "run/bootc-install-reconciled";
+
+/// Assuming that the current root is an ostree deployment, pull kargs
+/// from it and inject them.
+fn reconcile_kargs(sysroot: &ostree::Sysroot, deployment: &ostree::Deployment) -> Result<()> {
+    let deployment_root = &crate::utils::deployment_fd(sysroot, deployment)?;
+    let cancellable = gio::Cancellable::NONE;
+
+    let current_kargs = deployment
+        .bootconfig()
+        .expect("bootconfig for deployment")
+        .get("options");
+    let current_kargs = current_kargs
+        .as_ref()
+        .map(|s| s.as_str())
+        .unwrap_or_default();
+    tracing::debug!("current_kargs={current_kargs}");
+    let current_kargs = ostree::KernelArgs::from_string(&current_kargs);
+
+    // Keep this in sync with install_container
+    let install_config = config::load_config()?;
+    let install_config_kargs = install_config
+        .as_ref()
+        .and_then(|c| c.kargs.as_ref())
+        .into_iter()
+        .flatten()
+        .map(|s| s.as_str())
+        .collect::<Vec<_>>();
+    let kargsd = crate::kargs::get_kargs_in_root(deployment_root, std::env::consts::ARCH)?;
+    let kargsd = kargsd.iter().map(|s| s.as_str()).collect::<Vec<_>>();
+
+    current_kargs.append_argv(&install_config_kargs);
+    current_kargs.append_argv(&kargsd);
+    let new_kargs = current_kargs.to_string();
+    tracing::debug!("new_kargs={new_kargs}");
+
+    sysroot.deployment_set_kargs_in_place(deployment, Some(&new_kargs), cancellable)?;
+    Ok(())
+}
+
+/// A little helper struct which on drop renames a file. Used for putting back /etc/resolv.conf.
+#[must_use]
+struct Renamer<'d> {
+    dir: &'d Dir,
+    from: &'static Utf8Path,
+    to: &'static Utf8Path,
+}
+
+impl<'d> Renamer<'d> {
+    fn _impl_drop(&mut self) -> Result<()> {
+        self.dir
+            .rename(self.from, self.dir, self.to)
+            .map_err(Into::into)
+    }
+
+    fn consume(mut self) -> Result<()> {
+        self._impl_drop()
+    }
+}
+
+impl<'d> Drop for Renamer<'d> {
+    fn drop(&mut self) {
+        let _ = self._impl_drop();
+    }
+}
+/// Work around https://github.com/containers/buildah/issues/4242#issuecomment-2492480586
+/// among other things. We unconditionally replace the contents of `/etc/resolv.conf`
+/// in the target root with whatever the host uses (in Fedora 41+, that's systemd-resolved for Anaconda).
+#[context("Copying host resolv.conf")]
+fn ensure_resolvconf<'d>(rootfs: &'d Dir, proc1_root: &Dir) -> Result<Option<Renamer<'d>>> {
+    // Now check the state of etc/resolv.conf in the target root
+    let meta = rootfs
+        .symlink_metadata_optional(RESOLVCONF)
+        .context("stat")?;
+    let renamer = if meta.is_some() {
+        rootfs
+            .rename(RESOLVCONF, &rootfs, RESOLVCONF_ORIG)
+            .context("Renaming")?;
+        Some(Renamer {
+            dir: &rootfs,
+            from: RESOLVCONF_ORIG.into(),
+            to: RESOLVCONF.into(),
+        })
+    } else {
+        None
+    };
+    // If we got here, /etc/resolv.conf either didn't exist or we removed it.
+    // Copy the host data into it (note this will follow symlinks; e.g.
+    // Anaconda in Fedora 41+ defaults to systemd-resolved)
+    proc1_root
+        .copy(RESOLVCONF, rootfs, RESOLVCONF)
+        .context("Copying new resolv.conf")?;
+    Ok(renamer)
+}
+
+/// Bind a mount point from the host namespace into our root
+fn bind_from_host(
+    rootfs: &Dir,
+    src: impl AsRef<Utf8Path>,
+    target: impl AsRef<Utf8Path>,
+) -> Result<()> {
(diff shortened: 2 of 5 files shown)


cgwalters:
This seems to be working well in my hand-rolled testing.
(BTW I am testing via https://github.com/virt-manager/virt-manager/issues/739#issuecomment-2489529672 )

Still TODO:

[ ] Add [LBI pulling](https://github.com/containers/bootc/pull/860)
[ ] Add tests
[ ] Also try to re-do the ostree deployment to handle fixups like [this one](https://github.com/ostreedev/ostree-rs-ext/pull/679) or [this one](https://github.com/ostreedev/ostree-rs-ext/pull/669)

cgwalters:
OK I've rebased this on top of https://github.com/containers/bootc/pull/860 and we successfully pull LBIs at Anaconda install time too now.

cgwalters:
There were a surprising number of things I hit. One of them for example is that anaconda's hand-rolled chroot/container doesn't mount `cgroupfs` which makes `podman` quite unhappy so we do so manually in https://github.com/containers/bootc/pull/915/files#diff-66bc72c28514e2546fbe456aee74a321866d5a9147136ef99251eec1e08be8ddR107

cgwalters:
The plus side of this PR as is is that it has near-zero risk unless explicitly turned on in the two places right now. The anaconda `%post` path is obviously opt-in. And the extensions to the `ostree container image deploy` path also only turn on when bootc takes those over from e.g. rpm-ostree today.

I've given this a fair bit of manual testing, but I think what will help here is to get this into e.g. Fedora rawhide and that'll get things running through the daily integration testing.

cgwalters:
> Broken link

Hmm, does it work for anyone else?

omertuc:
I think the repo is just private

cgwalters:
Duh sorry, here's the raw scripts. But yes ideally these bits are generalized into https://github.com/virt-manager/virt-manager/issues/739

```
#!/bin/bash
set -xeuo pipefail

#cd=/var/srv/walters/machine-images/centos-stream9/CentOS-Stream-9-20241118.0-x86_64-boot.iso
cd=/var/srv/walters/machine-images/fedora/Fedora-Everything-netinst-x86_64-Rawhide-20241117.n.0.iso
virt-install --connect qemu:///session \
--name anaconda-bootc-test \
--vcpus 4 \
--ram 8192 \
--disk size=40 \
--location ${cd} \
--os-variant rhel9.4 \
--initrd-inject=./inst.ks \
--memorybacking=source.type=memfd,access.mode=shared \
--filesystem=/var/srv/,host-var-srv,driver.type=virtiofs \
--noautoconsole \
--extra-args "inst.ks=file:/inst.ks inst.updates=https://fedorapeople.org/~walters/updates.img"
```

```
%pre --erroronfail
mkdir -p /mnt/host-var-srv
mount -t virtiofs -o ro host-var-srv /mnt/host-var-srv
# This code actually takes over things
cp -p /mnt/host-var-srv/walters/bootc.fedora /usr/bin/bootc
ln -sfr /usr/bin/bootc /usr/libexec/libostree/ext/ostree-container
%end

# Basic setup
text
network --bootproto=dhcp --device=link --activate
# Basic partitioning
clearpart --all --initlabel --disklabel=gpt
reqpart --add-boot
part / --grow --fstype xfs

#ostreecontainer --transport oci --url /mnt/host-var-srv/walters/oci:bootc-lbi
ostreecontainer --transport oci --url /mnt/host-var-srv/walters/oci:bootc-lb

rootpw <password>
sshkey --username root "<key>"
#reboot

%post
curl -L --head https://quay.io
%end

%post --erroronfail
bootc install ensure-completion
%end
```

cgwalters:
Rebased :surfer:

cgwalters:
@omertuc mind reviewing?

cgwalters:
OK no traction, @vrothberg can you do it?

allisonkarlitskaya (commented):
I'm sorry that I didn't say it explicitly earlier: I don't have enough knowledge to review this. :/

lib/src/cli.rs
```diff
@@ -183,6 +183,24 @@ pub(crate) enum InstallOpts {
...
+    /// In this scenario the installation may be missing bootc specific features such as
+    /// kernel arguments, logically bound images and more. This command can be used to attempt
+    /// to reconcile. At the current time, the only tested environment is Anaconda using `ostreecontainer`
+    /// and it is recommended to avoid usage outside of that environment. Instead, ensure your
+    /// code is using `bootc install to-filesystem` from the start.
+    #[clap(hide = true)]
+    EnsureCompletion {
+        /// When provided, we assume that we're being invoked from our own
+        /// ostree-ext codebase.
+        #[clap(long)]
```
omertuc, line 197:
```suggestion
        #[clap(long, requires = "stateroot")]
```

lib/src/cli.rs
```diff
@@ -971,6 +989,17 @@ async fn run_from_opt(opt: Opt) -> Result<()> {
                 crate::install::install_to_existing_root(opts).await
             }
             InstallOpts::PrintConfiguration => crate::install::print_configuration(),
+            InstallOpts::EnsureCompletion { sysroot, stateroot } => {
+                let rootfs = &Dir::open_ambient_dir("/", cap_std::ambient_authority())?;
+                if let Some(sysroot) = sysroot {
+                    let stateroot = stateroot.as_deref().ok_or_else(|| {
+                        anyhow::anyhow!("Expected stateroot when --sysroot is set")
+                    })?;
```
omertuc, line 997:
```suggestion
                    let stateroot = stateroot.as_deref().unwrap();
```

ostree-ext/src/globals.rs
```diff
@@ -87,7 +87,10 @@ pub fn get_global_authfile(root: &Dir) -> Result<Option<(Utf8PathBuf, File)>> {
 }
 
 /// Return the path to the global container authentication file, if it exists.
-fn get_global_authfile_impl(root: &RootDir, am_uid0: bool) -> Result<Option<(Utf8PathBuf, File)>> {
+pub fn get_global_authfile_impl(
```
omertuc, line 90: Is this `pub` relevant to this PR?

omertuc (changes requested):
Initial superficial review, will post another for `completion.rs`

lib/src/cli.rs
cgwalters: FWIW it only worked to make them mutually required

```
diff --git a/lib/src/cli.rs b/lib/src/cli.rs
index acd986da..572fd0d2 100644
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -194,11 +194,11 @@ pub(crate) enum InstallOpts {
     EnsureCompletion {
         /// When provided, we assume that we're being invoked from our own
         /// ostree-ext codebase.
-        #[clap(long)]
+        #[clap(long, requires = "stateroot")]
         sysroot: Option<Utf8PathBuf>,
 
         /// Must be set if sysroot is set
-        #[clap(long)]
+        #[clap(long, requires = "sysroot")]
         stateroot: Option<String>,
     },
     /// Output JSON to stdout that contains the merged installation configuration
```

lib/src/cli.rs
cgwalters: This isn't true with just `requires`...the only way to make this work would be to split up the argument. Which I'll do instead.

ostree-ext/src/container/deploy.rs
```diff
@@ -122,6 +136,23 @@ pub async fn deploy(
...
         )?;
+
+        // We end up re-executing ourselves as a subprocess because
+        // otherwise right now we end up with a circular dependency between
+        // crates. We need an option to skip though so when the *main*
+        // bootc install code calls this API, we don't do this as it
+        // will have already been handled.
+        if !options.skip_completion {
+            let st = Command::new("bootc")
+                .args(["install", "ensure-completion", "--sysroot=."])
```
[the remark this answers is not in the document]
cgwalters: Yes, `cwd_dir` changes the directory. I added a comment

lib/src/install/completion.rs
```diff
@@ -0,0 +1,298 @@
...
+    )?;
+    Dir::reopen_dir(&proc1_root.as_fd()).map_err(Into::into)
+}
+
+/// Core entrypoint invoked when we are likely being invoked from inside Anaconda as a `%post`.
+pub(crate) async fn run_from_anaconda(rootfs: &Dir) -> Result<()> {
+    // unshare our mount namespace, so any *further* mounts aren't leaked.
+    // Note that because this does a re-exec, anything *before* this point
+    // should be idempotent.
+    crate::cli::ensure_self_unshared_mount_namespace()?;
```
omertuc, line 198: Requires root, and will clumsily fail. Since this is what people will see if they run `bootc install ensure-completion` (e.g. out of curiosity while digging into our anaconda procedures), maybe we should error out more gracefully?

lib/src/install/completion.rs
```diff
@@ -0,0 +1,298 @@
...
+
+    if std::env::var_os(ANACONDA_ENV_HINT).is_none() {
+        // Be loud if a user is invoking this outside of the expected setup.
+        medium_visibility_warning(&format!("Missing environment variable {ANACONDA_ENV_HINT}"));
```
omertuc, line 202: Consider hard fail for safety?

lib/src/install/completion.rs
```diff
@@ -0,0 +1,298 @@
...
+    rootfs: &Dir,
+    src: impl AsRef<Utf8Path>,
+    target: impl AsRef<Utf8Path>,
+) -> Result<()> {
+    fn bind_from_host_impl(rootfs: &Dir, src: &Utf8Path, target: &Utf8Path) -> Result<()> {
+        rootfs.create_dir_all(target)?;
+        if rootfs.is_mountpoint(target)?.unwrap_or_default() {
+            return Ok(());
+        }
+        let target = format!("/mnt/sysroot/{target}");
```
omertuc, line 142: Maybe it's supposed to be obvious but what is `/mnt/sysroot`?

omertuc: EDIT: https://anaconda-installer.readthedocs.io/en/latest/mount-points.html#mnt-sysroot

omertuc: (Or better move it to a constant with a comment above it)

lib/src/install/completion.rs
cgwalters: I was thinking we didn't want this to just blow up if Anaconda changed this environment variable, I don't think it's a stable API. But...yeah they're probably unlikely to remove it.

lib/src/install/completion.rs
```diff
@@ -0,0 +1,298 @@
...
+    // Note that because this does a re-exec, anything *before* this point
+    // should be idempotent.
+    crate::cli::ensure_self_unshared_mount_namespace()?;
+
+    if std::env::var_os(ANACONDA_ENV_HINT).is_none() {
+        // Be loud if a user is invoking this outside of the expected setup.
+        medium_visibility_warning(&format!("Missing environment variable {ANACONDA_ENV_HINT}"));
+    } else {
+        // In the way Anaconda sets up the bind mounts today, this doesn't exist. Later
+        // code expects it to exist, so do so.
```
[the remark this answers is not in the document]
cgwalters: Yes, we need a shared constant for this for sure, will look at that as a followup

cgwalters:
Ah, CI fell over due to missing `#[cfg(feature = "install")]`...it's tempting to drop that as I doubt anyone cares. Or maybe only have the `to-disk` path be under `#[cfg(feature = "install-to-disk")]` since it's *that* code that's more the "demo"/"toy".

cgwalters:
Anyways, this one is updated.

ostree-ext/src/container/deploy.rs
```diff
@@ -122,6 +136,24 @@ pub async fn deploy(
...
+
+        // We end up re-executing ourselves as a subprocess because
+        // otherwise right now we end up with a circular dependency between
+        // crates. We need an option to skip though so when the *main*
+        // bootc install code calls this API, we don't do this as it
+        // will have already been handled.
+        if !options.skip_completion {
+            // Note that the sysroot is provided as `.`  but we use cwd_dir to
+            // make the process current working directory the sysroot.
+            let st = Command::new("bootc")
```
omertuc, line 148: Maybe instead

```suggestion
            let st = Command::new(<argv0>)
```

So that people running a custom non-PATH binary of bootc (during dev / debug) don't get surprised when the re exec is a different binary altogether

omertuc (changes requested):
Review still incomplete

cgwalters:
> Are we happy with these names? Can you update the commit messages to mention both options / the difference between them?

The internals one is totally internal and not something a human should ever invoke, that's why I didn't add it even to the commit message.

omertuc:
Testing my understanding:

The reason

```
%post --erroronfail
bootc install ensure-completion
%end
```

uses the image's bootc and not the bootc that ships with the OS that anaconda runs on, is because %post's "degenerate container" is actually a chroot into the filesystem that we just installed the image too, thus ensuring the new bootc binary is available there?

ostree-ext/src/container/deploy.rs
```diff
@@ -122,6 +136,24 @@ pub async fn deploy(
             flags,
             cancellable,
         )?;
+
+        // We end up re-executing ourselves as a subprocess because
+        // otherwise right now we end up with a circular dependency between
+        // crates. We need an option to skip though so when the *main*
+        // bootc install code calls this API, we don't do this as it
+        // will have already been handled.
+        if !options.skip_completion {
```
omertuc, line 145: Does this mean that all future users of `ostree container image deploy`  (where options.skip_completion` is `false`) will have bootc completion running for them? 

What about users doing it on non-bootc images? For missing InstallConfiguration (kargs) I see we use Option::into_iter() which will make it a no-op

Similarly for bound images, it will just be empty

So it should work in theory, but did you test this in practice? that the command is happy with completely non-bootc images?

omertuc (changes requested):

ostree-ext/src/container/deploy.rs
omertuc: I'm basically afraid that this PR will introduce a regression / unexpected errors in users of `ostree container image deploy` in non-bootc images

omertuc:
Other than the two open reviews, lgtm fwiw

cgwalters:
> uses the image's bootc and not the bootc that ships with the OS that anaconda runs on, is because %post's "degenerate container" is actually a chroot into the filesystem that we just installed the image to (using ostree container image deploy), thus ensuring the new bootc binary is available there?

Exactly!

ostree-ext/src/container/deploy.rs
cgwalters: I didn't explicitly test that yet, and we should. However note that none of this code  will run *in practice* until the moment when we take over from the version in rpm-ostree.

cgwalters: But yes basically the heart of this right now is looking for files (kargs, bound image symlinks) and acting on them if present, and if they don't it's a no-op.

omertuc:
OK then just https://github.com/containers/bootc/pull/915#discussion_r1871616405 needs to be addressed

ostree-ext/src/container/deploy.rs
cgwalters: Good catch! Changed to use `/proc/self/exe`

omertuc (approved):

cgwalters:
For reference, this does work for LBIs but doesn't for kargs with anaconda; see https://gitlab.com/fedora/bootc/tests/bootc-workflow-test/-/merge_requests/555#note_2284993995

It does work for *non* Anaconda cases that directly invoke `ostree container image deploy` - the problem is that Anaconda also then later invokes `ostree admin instutil set-kargs` (it doesn't need to do this, it could compute the kargs up front and pass them to the deploy), but when it does that it does it without `--merge` which just blows away the other kargs, which is what the patch linked there fixes.

In theory, we could change `ensure-completion` to also fix this, but it would be even messier implementation wise.

The LBI support is most important to wedge in to older Anacondas I think, and otherwise we should just proceed with fixing newer Anacondas.

kfox1111:
Thats.... unfortunate. Its the exact use case that broke for me. :/

Though I have a workaround...

Renlor:
> And will fix things up. Of course there's fun $details here...the
> way Anaconda implements %post is via a hand-augmented chroot
> i.e. a degenerate container, and we need to escape that and
> fix some things up (such as a missing cgroupfs mount).

Not super important, mostly for my own understanding: Why not run the post script with `--nochroot` so you don't first have to escape chroot to fix up the install environment?  Something to do with using the installed images bootc instead of the anaconda one?

cgwalters:
> Why not run the post script with --nochroot

Because for Fedora and RHEL we don't respin the Anaconda ISO post GA, so the `/usr/bin/bootc` on the ISO gets older and older over time and the whole idea of this is to be able to have a potentially newer `bootc` binary from the *container being installed* to be able to fix things up.

Renlor:
That makes sense, thanks for taking the time to help me understand the implications.

One more question if you don't mind my curiosity:  What are the downsides of calling the newly installed bootc (/mnt/sysroot/usr/bin/bootc) from outside of the chroot?

PR #916, opened 2024-11-22
 cli: Add interception for ostree extension verbs

cgwalters:
cli: Factor out helper to parse "callname"

No functional changes; prep for further work.

---

cli: Prepare for other argv0 dispatching

Clean up the interception to prepare for other cases.

---

cli: Add interception for ostree extension verbs

This allows us to fully own the symlinks in `/usr/libexec/libostree/ext`.

---

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -2,8 +2,7 @@
 //!
 //! Command line tool to manage bootable ostree-based containers.
 
-use std::ffi::CString;
-use std::ffi::OsString;
+use std::ffi::{CString, OsStr, OsString};
 use std::io::Seek;
 use std::os::unix::process::CommandExt;
 use std::process::Command;
@@ -879,6 +878,18 @@ where
     run_from_opt(Opt::parse_including_static(args)).await
 }
 
+/// Find the base binary name from argv0 (without a full path). The empty string
+/// is never returned; instead a fallback string is used. If the input is not valid
+/// UTF-8, a default is used.
+fn callname_from_argv0(argv0: &OsStr) -> &str {
+    let default = "bootc";
+    std::path::Path::new(argv0)
+        .file_name()
+        .and_then(|s| s.to_str())
+        .filter(|s| !s.is_empty())
+        .unwrap_or(default)
+}
+
 impl Opt {
     /// In some cases (e.g. systemd generator) we dispatch specifically on argv0.  This
     /// requires some special handling in clap.
@@ -890,9 +901,9 @@ impl Opt {
         let mut args = args.into_iter();
         let first = if let Some(first) = args.next() {
             let first: OsString = first.into();
-            let argv0 = first.to_str().and_then(|s| s.rsplit_once('/')).map(|s| s.1);
+            let argv0 = callname_from_argv0(&first);
             tracing::debug!("argv0={argv0:?}");
-            if matches!(argv0, Some(InternalsOpts::GENERATOR_BIN)) {
+            if argv0 == InternalsOpts::GENERATOR_BIN {
                 let base_args = ["bootc", "internals", "systemd-generator"]
                     .into_iter()
                     .map(OsString::from);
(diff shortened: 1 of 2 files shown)


lib/src/cli.rs
```diff
@@ -879,6 +878,18 @@ where
     run_from_opt(Opt::parse_including_static(args)).await
 }
 
+/// Find the base binary name from argv0 (without a full path). The empty string
+/// is never returned; instead a fallback string is used.
+fn callname_from_argv0(argv0: &OsStr) -> &str {
+    let default = "bootc";
+    let argv0 = argv0.to_str().filter(|s| !s.is_empty()).unwrap_or(default);
```
allisonkarlitskaya, line 885: This is pretty semantically dense...

allisonkarlitskaya: My understanding of the logic here: if argv0 fails to convert to `str`, or in case that string is empty, use a default value, and then try to split it based on `/`, taking the right side, but if the split fails or the right side is empty, use the original (or default) string instead.

It gets difficult to read because of some non-obvious stuff like `.to_str()` returning an `Option<>` and `.filter()` working on `Option<>` as if it were a 0-or-1-item list.  To me it looks like you're trying to filter non-empty ... characters?  No wait, that doesn't make sense...

I feel like you could use a bit more early returning and/or open-coding of stuff.  Even a bit of `match` could help here... and isn't there some `basename` type feature somewhere we can use?

This is obviously entirely stylistic...

lib/src/cli.rs
```diff
@@ -1022,6 +1033,31 @@ async fn run_from_opt(opt: Opt) -> Result<()> {
     }
 }
 
+#[test]
+fn test_callname() {
+    let mapped_cases = [
+        ("", "bootc"),
+        ("/foo/bar", "bar"),
+        ("../foo/bar", "bar"),
+        ("usr/bin/ostree-container", "ostree-container"),
```
allisonkarlitskaya, line 1049: Not tested: osstr that fails to `.to_str()`, string ending with `/`.

lib/src/cli.rs
allisonkarlitskaya: This function will also return `/` for the input `/`, but I think maybe you'd prefer `bootc` in that case?  ie: if the string successfully splits but the right-hand result of the splitting is empty maybe you should prefer a fallback to the default instead of taking the entire string?

lib/src/cli.rs
```diff
@@ -903,10 +903,14 @@ impl Opt {
             let first: OsString = first.into();
             let argv0 = callname_from_argv0(&first);
             tracing::debug!("argv0={argv0:?}");
-            if argv0 == InternalsOpts::GENERATOR_BIN {
-                let base_args = ["bootc", "internals", "systemd-generator"]
-                    .into_iter()
-                    .map(OsString::from);
+            let mapped = match argv0 {
+                InternalsOpts::GENERATOR_BIN => {
+                    Some(["bootc", "internals", "systemd-generator"].as_slice())
```
allisonkarlitskaya, line 908: What is `.as_slice()` doing here?  I'd have assumed that it already is a slice...  Copying?  Could you rather do that with a ref or something?

lib/src/cli.rs
```diff
@@ -907,6 +907,9 @@ impl Opt {
                 InternalsOpts::GENERATOR_BIN => {
                     Some(["bootc", "internals", "systemd-generator"].as_slice())
                 }
+                "ostree-container" | "ostree-ima-sign" | "ostree-provisional-repair" => {
+                    Some(["bootc", "internals", "ostree-ext"].as_slice())
```
allisonkarlitskaya, line 911: (thinking out loud)

My brain is saying "...you dropped the verb here" but actually it gets in via the `.chain()` below taking the entire argument list again. not just argv[1..].  But that means that we also include the `::GENERATOR_BIN` part (without path mangling) after the `systemd-generator` argument.   Which is presumably correct because that's how it already works today....

allisonkarlitskaya (commented):
Mostly style fly-bys from me on this one.  I'm afraid I don't understand the larger picture.  Let's talk in standup?

lib/src/cli.rs
cgwalters: This coerces the arrays to slices so both values from the match have the same type.

lib/src/cli.rs
cgwalters: > This is pretty semantically dense...

Heh for sure.

lib/src/cli.rs
cgwalters: > My brain is saying "...you dropped the verb here" but actually it gets in via the .chain() below taking the entire argument list again.

Not quite, the `next()` from the iterator popped off the first arg.

I pushed another commit "cli: Strengthen generator parsing test" which shows the existing code was right.

lib/src/cli.rs
cgwalters: I reworked this one to use `Path::file_name()` and yeah it got way cleaner.

cgwalters:
@allisonkarlitskaya mind stamping this?

allisonkarlitskaya (approved):
:+1:

PR #919, opened 2024-11-25
install: Automatically set up bind mounts if not provided

cgwalters:
install: Automatically set up /dev and /var/lib/containers

We're looking again at the ergonomics of `bootc install to-existing-root`.
This uses the "mounting into mount namespaces" from the new
mount API to automatically set up `/dev` and `/var/lib/containers`
if they weren't provided to `podman run`, which shrinks what's
needed a bit.

Closes: https://github.com/containers/bootc/issues/826

---

install: Drop skopeo-in-host fallback code

Now that we unconditionally mount /var/lib/containers, drop
the hacky fork-skopeo-in-host-mountns code.

---

Drop test references to /dev and /var/lib/containers mounts

Keep the bind mounts in the docs though for now because many
people will be using the current docs with older bootc.

---

install: Drop need for -v /:/target

Build on our new logic for bind mounting from the host mountns
to also drop the need for the `-v /:/target` in the alongside
install code.

---

install: Centralize PID1 definition

Hooray for `const`.

---

tests: Drop more bind mount instances

These should be unnecessary.

---


cgwalters:
One thing I will say I'm pretty proud of is this code *almost* worked the first try, including the whole fd passing over the socketpair. The famous "in Rust if it compiles, it works". The only thing I missed was that for `/var/lib/containers` because it's not present in our base image, `podman run ... -v /var/lib/containers:/var/lib/containers` will automatically create the directory. So we just needed a `std::fs::create_dir_all()` in our mount reconciliation path.

lib/src/mount.rs
```diff
@@ -124,3 +137,134 @@ pub(crate) fn is_same_as_host(path: &Utf8Path) -> Result<bool> {
...
+    // Allocate a socket pair to use for sending file descriptors.
+    let (sock_parent, sock_child) = rustix::net::socketpair(
+        AddressFamily::UNIX,
+        SocketType::STREAM,
+        SocketFlags::CLOEXEC,
+        None,
+    )
+    .context("socketpair")?;
+    const DUMMY_DATA: &[u8] = &[b'!'];
+    match unsafe { libc::fork() } {
```
cgwalters, line 159: rustix doesn't bind this, probably for good reason. Probably the best cleanup here would be to fork + exec a helper early on (before any threads are possibly created) and talk to it via IPC. But I think this is fine as is.

lib/src/install.rs
```diff
@@ -1269,6 +1205,8 @@ async fn prepare_install(
     tracing::debug!("Target image reference: {target_imgref}");
 
     // A bit of basic global state setup
+    crate::mount::ensure_mirrored_host_mount("/dev")?;
+    crate::mount::ensure_mirrored_host_mount("/var/lib/containers")?;
```
vrothberg, line 1209: What if a user configures storage.conf to point to another path?

cgwalters: Yeah, we already hit one instance of that. But fixing that is a distinct thing, orthogonal to dropping the requirements for the bind mounts being set up externally, right?

I will file a separate issue.

tests/e2e/playbooks/install.yaml
```diff
@@ -54,9 +54,7 @@
          --privileged \
          --tls-verify=false \
          --pid=host \
-         -v /dev:/dev \
          -v /:/target \
```
omertuc, line 57:
```suggestion
```

lib/src/mount.rs
```diff
@@ -23,6 +23,14 @@ use serde::Deserialize;
 
 use crate::task::Task;
 
+/// Well known identifier for pid 1
+pub(crate) const PID1: Pid = const {
+    match Pid::from_raw(1) {
+        Some(v) => v,
+        None => panic!("Expected to parse pid1"),
+    }
```
cgwalters, line 31: Tangential but this one is set to just be `const { Pid::from_raw(1).unwrap() }` in the next Rust which will be a big big improvement for `const`.

tests/e2e/playbooks/install.yaml
cgwalters: Thanks, applied!

omertuc:
Title should be edited

cgwalters:
The PR title? Updated.

cgwalters:
Hmm, so [one job](https://github.com/containers/bootc/pull/919/checks?check_run_id=33554884644) failed like this:

> 
Initializing ostree layout
ERROR Installing to disk: Creating ostree deployment: Pulling: Creating importer: failed to invoke method OpenImage: failed to invoke method OpenImage: reference "[overlay@/var/lib/containers/storage+/run/containers/storage:overlay.mountopt=nodev,metacopy=on]192.168.100.1:5000/bootc-workflow-test:95ad@sha256:71116de023138a0960d152377f1c2a353a8c0139ab6adfbd8bf45399364c5fa6" does not resolve to an image ID: identifier is not an image

Which looks like it could be caused by this PR, but I don't have any idea why it would just silently not work there...what I was thinking *could* fail is if we were running from an older OS without the new mount API but I think all operating system versions we care about right now have it. I have no idea how to rerun testing farm jobs to see if this was a flake...

omertuc:
Amend force to rerun all?

omertuc (approved):

PR #921, opened 2024-11-25
feat(upgrade): Add total progress bar and JSON output

antheas:
Adds a total progress bar to the progress output of switch and upgrade that also shows time remaining.

Then, uses the total progress bar to allow sending a jsonl output to stderr that can be read by userspace software. Has to be stderr, as bootc has some spurious prints on stdout. Progressbar is hidden as it writes to stderr.

Progress bar styling needs some cleanup. Jsonl output needs a type arg to identify the type of output, so that there can be other jsonl jsons for e.g., deployment.

### DIFF
--- a/lib/src/deploy.rs
+++ b/lib/src/deploy.rs
@@ -21,6 +21,7 @@ use ostree_ext::ostree::{self, Sysroot};
 use ostree_ext::sysroot::SysrootLock;
 use ostree_ext::tokio_util::spawn_blocking_cancellable_flatten;
 
+use crate::progress_jsonl::{Event, ProgressWriter, SubTaskBytes, SubTaskStep, API_VERSION};
 use crate::spec::ImageReference;
 use crate::spec::{BootOrder, HostSpec};
 use crate::status::labels_of_config;
@@ -141,11 +142,20 @@ fn prefix_of_progress(p: &ImportProgress) -> &'static str {
 async fn handle_layer_progress_print(
     mut layers: tokio::sync::mpsc::Receiver<ostree_container::store::ImportProgress>,
     mut layer_bytes: tokio::sync::watch::Receiver<Option<ostree_container::store::LayerProgress>>,
+    digest: Box<str>,
     n_layers_to_fetch: usize,
+    layers_total: usize,
+    bytes_to_download: u64,
+    bytes_total: u64,
+    prog: ProgressWriter,
+    quiet: bool,
 ) {
     let start = std::time::Instant::now();
     let mut total_read = 0u64;
     let bar = indicatif::MultiProgress::new();
+    if quiet {
+        bar.set_draw_target(indicatif::ProgressDrawTarget::hidden());
+    }
     let layers_bar = bar.add(indicatif::ProgressBar::new(
         n_layers_to_fetch.try_into().unwrap(),
     ));
@@ -157,7 +167,8 @@ async fn handle_layer_progress_print(
             .template("{prefix} {bar} {pos}/{len} {wide_msg}")
             .unwrap(),
     );
-    layers_bar.set_prefix("Fetching layers");
+    let taskname = "Fetching layers";
+    layers_bar.set_prefix(taskname);
     layers_bar.set_message("");
     byte_bar.set_prefix("Fetching");
     byte_bar.set_style(
@@ -167,6 +178,9 @@ async fn handle_layer_progress_print(
                 )
                 .unwrap()
         );
+
+    let mut subtasks = vec![];
+    let mut subtask: SubTaskBytes = Default::default();
     loop {
         tokio::select! {
             // Always handle layer changes first.
@@ -174,18 +188,44 @@ async fn handle_layer_progress_print(
             layer = layers.recv() => {
                 if let Some(l) = layer {
                     let layer = descriptor_of_progress(&l);
+                    let layer_type = prefix_of_progress(&l);
+                    let short_digest = &layer.digest().digest()[0..21];
                     let layer_size = layer.size();
                     if l.is_starting() {
+                        // Reset the progress bar
                         byte_bar.reset_elapsed();
                         byte_bar.reset_eta();
                         byte_bar.set_length(layer_size);
-                        let layer_type = prefix_of_progress(&l);
-                        let short_digest = &layer.digest().digest()[0..21];
                         byte_bar.set_message(format!("{layer_type} {short_digest}"));
+
+                        subtask = SubTaskBytes {
+                            subtask: layer_type.into(),
+                            description: format!("{layer_type}: {short_digest}").clone().into(),
+                            id: format!("{short_digest}").clone().into(),
+                            bytes_cached: 0,
+                            bytes: 0,
+                            bytes_total: layer_size,
+                        };
                     } else {
                         byte_bar.set_position(layer_size);
                         layers_bar.inc(1);
                         total_read = total_read.saturating_add(layer_size);
+                        // Emit an event where bytes == total to signal completion.
+                        subtask.bytes = layer_size;
+                        subtasks.push(subtask.clone());
+                        prog.send(Event::ProgressBytes {
+                            api_version: API_VERSION.into(),
+                            task: "pulling".into(),
+                            description: format!("Pulling Image: {digest}").into(),
+                            id: (*digest).into(),
+                            bytes_cached: bytes_total - bytes_to_download,
+                            bytes: total_read,
+                            bytes_total: bytes_to_download,
+                            steps_cached: (layers_total - n_layers_to_fetch) as u64,
+                            steps: layers_bar.position(),
+                            steps_total: n_layers_to_fetch as u64,
+                            subtasks: subtasks.clone(),
+                        }).await;
                     }
                 } else {
                     // If the receiver is disconnected, then we're done
@@ -197,9 +237,26 @@ async fn handle_layer_progress_print(
                     // If the receiver is disconnected, then we're done
                     break
                 }
-                let bytes = layer_bytes.borrow();
-                if let Some(bytes) = &*bytes {
+                let bytes = {
+                    let bytes = layer_bytes.borrow_and_update();
+                    bytes.as_ref().cloned()
+                };
+                if let Some(bytes) = bytes {
                     byte_bar.set_position(bytes.fetched);
+                    subtask.bytes = byte_bar.position();
+                    prog.send_lossy(Event::ProgressBytes {
+                        api_version: API_VERSION.into(),
+                        task: "pulling".into(),
+                        description: format!("Pulling Image: {digest}").into(),
+                        id: (*digest).into(),
+                        bytes_cached: bytes_total - bytes_to_download,
+                        bytes: total_read + byte_bar.position(),
+                        bytes_total: bytes_to_download,
+                        steps_cached: (layers_total - n_layers_to_fetch) as u64,
+                        steps: layers_bar.position(),
+                        steps_total: n_layers_to_fetch as u64,
+                        subtasks: subtasks.clone().into_iter().chain([subtask.clone()]).collect(),
+                    }).await;
                 }
             }
         }
@@ -221,6 +278,27 @@ async fn handle_layer_progress_print(
     )) {
         tracing::warn!("writing to stdout: {e}");
     }
+
+    // Since the progress notifier closed, we know import has started
+    // use as a heuristic to begin import progress
+    // Cannot be lossy or it is dropped
+    prog.send(Event::ProgressSteps {
+        api_version: API_VERSION.into(),
+        task: "importing".into(),
+        description: "Importing Image".into(),
+        id: (*digest).into(),
+        steps_cached: 0,
+        steps: 0,
+        steps_total: 1,
+        subtasks: [SubTaskStep {
+            subtask: "importing".into(),
+            description: "Importing Image".into(),
+            id: "importing".into(),
+            completed: false,
+        }]
+        .into(),
+    })
+    .await;
 }
 
 /// Wrapper for pulling a container image, wiring up status output.
@@ -230,6 +308,7 @@ pub(crate) async fn pull(
     imgref: &ImageReference,
     target_imgref: Option<&OstreeImageReference>,
     quiet: bool,
+    prog: ProgressWriter,
 ) -> Result<Box<ImageState>> {
     let ostree_imgref = &OstreeImageReference::from(imgref.clone());
     let mut imp = new_importer(repo, ostree_imgref).await?;
@@ -250,20 +329,52 @@ pub(crate) async fn pull(
     ostree_ext::cli::print_layer_status(&prep);
     let layers_to_fetch = prep.layers_to_fetch().collect::<Result<Vec<_>>>()?;
     let n_layers_to_fetch = layers_to_fetch.len();
-    let printer = (!quiet).then(|| {
-        let layer_progress = imp.request_progress();
-        let layer_byte_progress = imp.request_layer_progress();
-        tokio::task::spawn(async move {
-            handle_layer_progress_print(layer_progress, layer_byte_progress, n_layers_to_fetch)
-                .await
-        })
+    let layers_total = prep.all_layers().count();
+    let bytes_to_fetch: u64 = layers_to_fetch.iter().map(|(l, _)| l.layer.size()).sum();
+    let bytes_total: u64 = prep.all_layers().map(|l| l.layer.size()).sum();
+
+    let prog_print = prog.clone();
+    let digest = prep.manifest_digest.clone();
+    let digest_imp = prep.manifest_digest.clone();
+    let layer_progress = imp.request_progress();
+    let layer_byte_progress = imp.request_layer_progress();
+    let printer = tokio::task::spawn(async move {
+        handle_layer_progress_print(
+            layer_progress,
+            layer_byte_progress,
+            digest.as_ref().into(),
+            n_layers_to_fetch,
+            layers_total,
+            bytes_to_fetch,
+            bytes_total,
+            prog_print,
+            quiet,
+        )
+        .await
     });
     let import = imp.import(prep).await;
-    if let Some(printer) = printer {
-        let _ = printer.await;
-    }
+    let _ = printer.await;
+    // Both the progress and the import are done, so import is done as well
+    prog.send(Event::ProgressSteps {
+        api_version: API_VERSION.into(),
+        task: "importing".into(),
+        description: "Importing Image".into(),
+        id: digest_imp.clone().as_ref().into(),
+        steps_cached: 0,
+        steps: 1,
+        steps_total: 1,
+        subtasks: [SubTaskStep {
+            subtask: "importing".into(),
+            description: "Importing Image".into(),
+            id: "importing".into(),
+            completed: true,
+        }]
+        .into(),
+    })
+    .await;
     let import = import?;
     let wrote_imgref = target_imgref.as_ref().unwrap_or(&ostree_imgref);
+
     if let Some(msg) =
         ostree_container::store::image_filtered_content_warning(repo, &wrote_imgref.imgref)
             .context("Image content warning")?
@@ -450,8 +561,53 @@ pub(crate) async fn stage(
     stateroot: &str,
     image: &ImageState,
     spec: &RequiredHostSpec<'_>,
+    prog: ProgressWriter,
 ) -> Result<()> {
+    let mut subtask = SubTaskStep {
+        subtask: "merging".into(),
+        description: "Merging Image".into(),
+        id: "fetching".into(),
+        completed: false,
+    };
+    let mut subtasks = vec![];
+    prog.send(Event::ProgressSteps {
+        api_version: API_VERSION.into(),
+        task: "staging".into(),
+        description: "Deploying Image".into(),
+        id: image.manifest_digest.clone().as_ref().into(),
+        steps_cached: 0,
+        steps: 0,
+        steps_total: 3,
+        subtasks: subtasks
+            .clone()
+            .into_iter()
+            .chain([subtask.clone()])
+            .collect(),
+    })
+    .await;
     let merge_deployment = sysroot.merge_deployment(Some(stateroot));
+
+    subtask.completed = true;
+    subtasks.push(subtask.clone());
+    subtask.subtask = "deploying".into();
+    subtask.id = "deploying".into();
+    subtask.description = "Deploying Image".into();
+    subtask.completed = false;
+    prog.send(Event::ProgressSteps {
+        api_version: API_VERSION.into(),
+        task: "staging".into(),
+        description: "Deploying Image".into(),
+        id: image.manifest_digest.clone().as_ref().into(),
+        steps_cached: 0,
+        steps: 1,
+        steps_total: 3,
+        subtasks: subtasks
+            .clone()
+            .into_iter()
+            .chain([subtask.clone()])
+            .collect(),
+    })
+    .await;
     let origin = origin_from_imageref(spec.image)?;
     let deployment = crate::deploy::deploy(
         sysroot,
@@ -462,14 +618,74 @@ pub(crate) async fn stage(
     )
     .await?;
 
+    subtask.completed = true;
+    subtasks.push(subtask.clone());
+    subtask.subtask = "bound_images".into();
+    subtask.id = "bound_images".into();
+    subtask.description = "Pulling Bound Images".into();
+    subtask.completed = false;
+    prog.send(Event::ProgressSteps {
+        api_version: API_VERSION.into(),
+        task: "staging".into(),
+        description: "Deploying Image".into(),
+        id: image.manifest_digest.clone().as_ref().into(),
+        steps_cached: 0,
+        steps: 1,
+        steps_total: 3,
+        subtasks: subtasks
+            .clone()
+            .into_iter()
+            .chain([subtask.clone()])
+            .collect(),
+    })
+    .await;
     crate::boundimage::pull_bound_images(sysroot, &deployment).await?;
 
+    subtask.completed = true;
+    subtasks.push(subtask.clone());
+    subtask.subtask = "cleanup".into();
+    subtask.id = "cleanup".into();
+    subtask.description = "Removing old images".into();
+    subtask.completed = false;
+    prog.send(Event::ProgressSteps {
+        api_version: API_VERSION.into(),
+        task: "staging".into(),
+        description: "Deploying Image".into(),
+        id: image.manifest_digest.clone().as_ref().into(),
+        steps_cached: 0,
+        steps: 2,
+        steps_total: 3,
+        subtasks: subtasks
+            .clone()
+            .into_iter()
+            .chain([subtask.clone()])
+            .collect(),
+    })
+    .await;
     crate::deploy::cleanup(sysroot).await?;
     println!("Queued for next boot: {:#}", spec.image);
     if let Some(version) = image.version.as_deref() {
         println!("  Version: {version}");
     }
     println!("  Digest: {}", image.manifest_digest);
+
+    subtask.completed = true;
+    subtasks.push(subtask.clone());
+    prog.send(Event::ProgressSteps {
+        api_version: API_VERSION.into(),
+        task: "staging".into(),
+        description: "Deploying Image".into(),
+        id: image.manifest_digest.clone().as_ref().into(),
+        steps_cached: 0,
+        steps: 3,
+        steps_total: 3,
+        subtasks: subtasks
+            .clone()
+            .into_iter()
+            .chain([subtask.clone()])
+            .collect(),
+    })
+    .await;
 
     Ok(())
 }
--- a/lib/src/progress_jsonl.rs
+++ b/lib/src/progress_jsonl.rs
@@ -1,0 +1,290 @@
+//! Output progress data using the json-lines format. For more information
+//! see <https://jsonlines.org/>.
+
+use anyhow::Result;
+use fn_error_context::context;
+use serde::Serialize;
+use std::borrow::Cow;
+use std::os::fd::{FromRawFd, OwnedFd, RawFd};
+use std::sync::Arc;
+use std::time::Instant;
+use tokio::io::{AsyncWriteExt, BufWriter};
+use tokio::net::unix::pipe::Sender;
+use tokio::sync::Mutex;
+
+// Maximum number of times per second that an event will be written.
+const REFRESH_HZ: u16 = 5;
+
+pub const API_VERSION: &str = "org.containers.bootc.progress/v1";
+
+/// An incremental update to e.g. a container image layer download.
+/// The first time a given "subtask" name is seen, a new progress bar should be created.
+/// If bytes == bytes_total, then the subtask is considered complete.
+#[derive(Debug, serde::Serialize, serde::Deserialize, Default, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct SubTaskBytes<'t> {
+    /// A machine readable type for the task (used for i18n).
+    /// (e.g., "ostree_chunk", "ostree_derived")
+    #[serde(borrow)]
+    pub subtask: Cow<'t, str>,
+    /// A human readable description of the task if i18n is not available.
+    /// (e.g., "OSTree Chunk:", "Derived Layer:")
+    #[serde(borrow)]
+    pub description: Cow<'t, str>,
+    /// A human and machine readable identifier for the task
+    /// (e.g., ostree chunk/layer hash).
+    #[serde(borrow)]
+    pub id: Cow<'t, str>,
+    /// The number of bytes fetched by a previous run (e.g., zstd_chunked).
+    pub bytes_cached: u64,
+    /// Updated byte level progress
+    pub bytes: u64,
+    /// Total number of bytes
+    pub bytes_total: u64,
+}
+
+/// Marks the beginning and end of a dictrete step
+#[derive(Debug, serde::Serialize, serde::Deserialize, Default, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct SubTaskStep<'t> {
+    /// A machine readable type for the task (used for i18n).
+    /// (e.g., "ostree_chunk", "ostree_derived")
+    #[serde(borrow)]
+    pub subtask: Cow<'t, str>,
+    /// A human readable description of the task if i18n is not available.
+    /// (e.g., "OSTree Chunk:", "Derived Layer:")
+    #[serde(borrow)]
+    pub description: Cow<'t, str>,
+    /// A human and machine readable identifier for the task
+    /// (e.g., ostree chunk/layer hash).
+    #[serde(borrow)]
+    pub id: Cow<'t, str>,
+    /// Starts as false when beginning to execute and turns true when completed.
+    pub completed: bool,
+}
+
+/// An event emitted as JSON.
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+#[serde(
+    tag = "type",
+    rename_all = "PascalCase",
+    rename_all_fields = "camelCase"
+)]
+pub enum Event<'t> {
+    /// An incremental update to a container image layer download
+    ProgressBytes {
+        /// The version of the progress event format.
+        #[serde(borrow)]
+        api_version: Cow<'t, str>,
+        /// A machine readable type (e.g., pulling) for the task (used for i18n
+        /// and UI customization).
+        #[serde(borrow)]
+        task: Cow<'t, str>,
+        /// A human readable description of the task if i18n is not available.
+        #[serde(borrow)]
+        description: Cow<'t, str>,
+        /// A human and machine readable unique identifier for the task
+        /// (e.g., the image name). For tasks that only happen once,
+        /// it can be set to the same value as task.
+        #[serde(borrow)]
+        id: Cow<'t, str>,
+        /// The number of bytes fetched by a previous run.
+        bytes_cached: u64,
+        /// The number of bytes already fetched.
+        bytes: u64,
+        /// Total number of bytes. If zero, then this should be considered "unspecified".
+        bytes_total: u64,
+        /// The number of steps fetched by a previous run.
+        steps_cached: u64,
+        /// The initial position of progress.
+        steps: u64,
+        /// The total number of steps (e.g. container image layers, RPMs)
+        steps_total: u64,
+        /// The currently running subtasks.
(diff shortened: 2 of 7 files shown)


cgwalters:
Thanks for working on this! I've been oscillating a lot on API/RPC frameworks in https://github.com/containers/bootc/issues/522

"jsonl" is yet another option there that is more targeted just for progress reporting (which is indeed *mainly* what we need now).

But - but - if we do that I'd really like to survey the field and see whether there are any other similar progress APIs.

For example, I think this API right now would constrain us to fetching a single layer at a time ; we could add a layer identifier to the metadata for example? Basically what we include in jsonl should be sufficient to render a "multi-progress" bar.

antheas:
This PR has some issues right now with spurious prints. I will try to fix them today and add different stages for deploy steps.

Incl. Fixing the failed test 

As far as multilayer reporting goes, the way I did this PR is for overall progress so it does not make a difference how many layers are being downloaded at a time. The syntax could be extended with a layers dictionary in the future though.

antheas:
Ok, tried to fix spurious prints. Seems to be impossible.

Since println! and eprintln! are used liberally, the functions cannot be dual purposed to json output.

Other than piping stdout to stderr with `set_print()` which is not in the rust stable API.

cgwalters:
Right, we should have e.g. `--json-progress-fd=<fdnum>` which would be the write half of a `pipe()` e.g.

cgwalters (changes requested):
Marking requested changes per discussion

(Also other minor bits, you need to run `cargo fmt`)

lib/src/deploy.rs
```diff
@@ -150,23 +161,30 @@ async fn handle_layer_progress_print(
...
-    byte_bar.set_prefix("Fetching");
     byte_bar.set_style(
         indicatif::ProgressStyle::default_bar()
                 .template(
-                    " └ {prefix} {bar} {binary_bytes}/{binary_total_bytes} ({binary_bytes_per_sec}) {wide_msg}",
+                    " └ {bar:20} {msg} ({binary_bytes}/{binary_total_bytes})",
                 )
                 .unwrap()
         );
+    total_byte_bar.set_prefix("Total");
```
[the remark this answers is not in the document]
antheas: I can chain PR them. The code for the total progress bar is used for the JSON output.

cgwalters:
Before we go too much farther on this can you give a read and comment on https://github.com/containers/bootc/issues/522 too please? Specifically WRT "jsonl" vs varlink vs gRPC and your use case.

lib/src/deploy.rs
```diff
@@ -45,6 +46,17 @@ pub(crate) struct ImageState {
     pub(crate) ostree_commit: String,
 }
 
+/// Download information
+#[derive(Debug, serde::Serialize)]
+pub struct JsonProgress {
```
cgwalters, line 51: In combination with my suggestion of having a `mod jsonl` I think this struct would go there. Just to future proof things a bit more I'd like to try to make this a little bit more abstract so that it could handle slightly more generic tasks.

For example, a `fsck` like operation wouldn't have "done_bytes" and "download_bytes" at least in that same sense, but it'd be good to have support for a similar progress bar for it.

In rpm-ostree we have such a generic progress API (over DBus), it's a bit clunky but see https://github.com/coreos/rpm-ostree/blob/f4fa2c5fe78774775ef3ce4f27d4199fac48f7a9/rust/src/console_progress.rs#L20

Basically I think we can structure this so clients can view progress data in a progressively more complex fashion; if they don't care about the details for an operation then can just take a simple "n/total" numbers and that could be pretty easily reused for an operation like fsck.

cgwalters (commented):
Some more comments.

Do you mind if I force push some updates to this PR and we can co-author it?

antheas:
Go ahead and push

lib/src/deploy.rs
antheas: I am usually using interpreted languages, so I can distinguish between different JSON objects based on "stage". However, if that is not the case for something like rust, then this struct would need to go into another struct as a union on the bottom.

cgwalters: In Rust serde has good support for most JSON enum conventions: https://serde.rs/enum-representations.html

cgwalters:
OK, I pushed a minor cleanup - stepping out for the day and I am in theory out the rest of this week, so releasing my mutex on this one.

Overall...I think my disposition is to just get this jsonl stuff in and we can sit on the larger IPC question for later.

One thing I'd still really love to know though: is anyone else out there doing something like this (specifically jsonl for *progress bars*?).

Oh hum, there's already a crate for this and it has [a fair number of reverse dependencies](https://crates.io/crates/serde-jsonlines/reverse_dependencies). That's encouraging. Only briefly skimming the source to one of them, [this code](https://github.com/jwodder/confab/blob/6bb9df8ac31675d1a24c0c3e2ea040f01bdf5123/tests/cli.rs#L195) looks quite related.

Like *ideally* we share a "schema" with an existing project here or at least something similar. (Maybe a good way to think of it is: could the schema we invent here be sufficiently generic that one could use it for e.g. `dnf` or `apt` too?)

antheas:
Neither the format or the concept behind this are particularly novel. The combination perhaps is.

Jsonlines is widely used in machine learning because it is dead simple, to the point it is not worth using a library for (unless performance is required), and supports streaming and compression due to it having separators very often.

Binding JSON to a UI is pretty standard data-binding too (e.g., look at React, and especially backends such as Redux).

I think there is a bit of a rough edge due to the fact that cli commands are used together with json, so those two words are mixed. 

If you want some ideas for IPC, when making hhd, I took the json idea a bit to the extreme, and did server side rendering over json, using a set of pre-defined components (the spec is [here](https://github.com/hhd-dev/hhd/blob/master/docs/http.md)).

Essentially, the `api/settings` endpoint sends a JSON of the currently available settings, one of which may be a progress bar, and then there is an `api/state` endpoint that contains all of the current values and updates similarly to jsonl. It also does internalization too, so all UIs share the translations. The UI just shows the settings that it was given and updates their values, with hhd acting as the model and controller in MVC.

That document was last updated in Dec. I think one month into making hhd and it is almost a year later now. There have been no breaking changes to that API since then, so I'd say the idea works pretty well. There are around 20x more settings now, and the project is 5x the size. 

The downside of that API is that it decides how the UI should be structured, and it makes it hard to interact with it over cli. But since the UI works so well now nobody asks for that. As for the UI, the only updates it has are to add stylized components ([here](https://github.com/hhd-dev/hhd-ui/releases)). It is actually quite nice to be able to add features without having to update the UI(s).

I will try to work through this PR this week and I think we will try to deploy this PR soon too. Just the progress bar itself is a nice QOL.

lib/src/progress_jsonl.rs
```diff
@@ -8,12 +8,36 @@ use std::os::fd::{FromRawFd, RawFd};
...
 use serde::Serialize;
 
-pub(crate) struct JsonlWriter {
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+pub struct LayerState {
+    pub digest: String,
+    pub downloaded: u64,
+    pub size: u64,
+    /// "cached", "waiting", "downloading", "done"
+    pub status: String,
```
antheas, line 17: This should be an enum

antheas:
Turns out the mutex was needed after all. Progress reader spins a thread so it consumed the writer

Added deployment stages as well:
```
{"stage":"fetch","bytes_done":199213658,"bytes_download":205161290,"bytes_total":4816175316,"layers_done":2,"layers_download":3,"layers_total":73,"layers":null}
{"stage":"fetch","bytes_done":201337882,"bytes_download":205161290,"bytes_total":4816175316,"layers_done":2,"layers_download":3,"layers_total":73,"layers":null}
{"stage":"fetch","bytes_done":202688154,"bytes_download":205161290,"bytes_total":4816175316,"layers_done":2,"layers_download":3,"layers_total":73,"layers":null}
{"stage":"fetch","bytes_done":204013466,"bytes_download":205161290,"bytes_total":4816175316,"layers_done":2,"layers_download":3,"layers_total":73,"layers":null}
Fetched layers: 195.66 MiB in 63 seconds (3.12 MiB/s)
{"stage":"deploy","n_steps":3,"step":0,"name":"deploying"}
Deploying...done (13 seconds)
{"stage":"deploy","n_steps":3,"step":1,"name":"pulling_bound_images"}
{"stage":"deploy","n_steps":3,"step":2,"name":"cleaning_up"}
Pruned images: 1 (layers: 6, objsize: 597.9 MB)
Queued for next boot: ghcr.io/ublue-os/bazzite:testing
  Version: testing-41.20241125.1
  Digest: sha256:97e85bdef72416134bde5afc7083a0222baa6fdc12a5d50a14cace1afd431d81
```

Inbetween these two lines there is a very suspicious 10-40s gap.  So something needs to be done there progress wise.
```
{"stage":"fetch","bytes_done":204013466,"bytes_download":205161290,"bytes_total":4816175316,"layers_done":2,"layers_download":3,"layers_total":73,"layers":null}
Fetched layers: 195.66 MiB in 63 seconds (3.12 MiB/s)
```

antheas:
```
$ sudo /tmp/bootc switch ghcr.io/ublue-os/bazzite-deck-gnome:testing --json-fd=0 --quiet
{"stage":"fetch","bytes_done":196820234,"bytes_download":198878861,"bytes_total":5766354157,"layers_done":4,"layers_download":5,"layers_total":73,"layers":null}
{"stage":"fetch","bytes_done":198753546,"bytes_download":198878861,"bytes_total":5766354157,"layers_done":4,"layers_download":5,"layers_total":73,"layers":null}
{"stage":"fetch","bytes_done":198878861,"bytes_download":198878861,"bytes_total":5766354157,"layers_done":5,"layers_download":5,"layers_total":73,"layers":null}
Fetched layers: 189.67 MiB in 7 seconds (26.19 MiB/s)
Committing image to OSTree
{"stage":"deploy","n_steps":5,"step":0,"name":"committing"}
{"stage":"deploy","n_steps":5,"step":1,"name":"sanity_check"}
{"stage":"deploy","n_steps":5,"step":2,"name":"deploying"}
⠙ Deploying                                                                                                                                                           
(bootc:13285): GLib-CRITICAL **: 22:09:10.120: g_atomic_ref_count_dec: assertion 'old_value > 0' failed
⠒ Deploying                                                                                                                                                           
(bootc:13285): GLib-CRITICAL **: 22:09:10.506: g_atomic_ref_count_dec: assertion 'old_value > 0' failed

(bootc:13285): GLib-CRITICAL **: 22:09:10.507: g_atomic_ref_count_dec: assertion 'old_value > 0' failed

(bootc:13285): GLib-CRITICAL **: 22:09:10.507: g_atomic_ref_count_dec: assertion 'old_value > 0' failed

(bootc:13285): GLib-CRITICAL **: 22:09:10.534: g_atomic_ref_count_dec: assertion 'old_value > 0' failed

(bootc:13285): GLib-CRITICAL **: 22:09:10.536: g_atomic_ref_count_dec: assertion 'old_value > 0' failed

(bootc:13285): GLib-CRITICAL **: 22:09:10.536: g_atomic_ref_count_dec: assertion 'old_value > 0' failed

(bootc:13285): GLib-CRITICAL **: 22:09:10.541: g_atomic_ref_count_dec: assertion 'old_value > 0' failed
⠂ Deploying                                                                                                                                                           
(bootc:13285): GLib-CRITICAL **: 22:09:10.557: g_atomic_ref_count_dec: assertion 'old_value > 0' failed
  Deploying: done (12 seconds)                                                                                                                                        {"stage":"deploy","n_steps":5,"step":3,"name":"pulling_bound_images"}
{"stage":"deploy","n_steps":5,"step":4,"name":"cleaning_up"}
Pruned images: 1 (layers: 6, objsize: 996.9 MB)
Queued for next boot: ghcr.io/ublue-os/bazzite-deck-gnome:testing
  Version: testing-41.20241125.1
  Digest: sha256:8265108a96bf3402c377b146a4071215e337d8ef84026a6e886981684bf57381
```

Looks a lot more granular and nicer now. Corruption bug still there.

cgwalters:
Hmm, I think "add update --check json" is scope creeping this; it relates to https://github.com/containers/bootc/pull/472#issuecomment-2061219087

Let's keep this one to what *isn't* in `status --json` already; the biggest thing there is just the progress notifications.

antheas:
> Hmm, I think "add update --check json" is scope creeping this; it relates to [#472 (comment)](https://github.com/containers/bootc/pull/472#issuecomment-2061219087)
> 
> Let's keep this one to what _isn't_ in `status --json` already; the biggest thing there is just the progress notifications.

I guess I was not aware of the following:
https://github.com/containers/bootc/pull/472#issuecomment-2061219087

Sure, we can strip that commit

cgwalters:
Let me take this one back for a bit

cgwalters:
> (bootc:13285): GLib-CRITICAL **: 22:09:10.120: g_atomic_ref_count_dec: assertion 'old_value > 0' failed

I'm not getting that one here...that's quite worrying. Can you try to get a core dump + stack via e.g. `env G_DEBUG=fatal-warnings bootc ...` ?

antheas:
```
       Message: Process 23271 (bootc) of user 0 terminated abnormally without generating a coredump.
Coredump entry has no core attached (neither internally in the journal nor externally on disk).
bazzite@antheas-ally-x:~$ sudo G_DEBUG=fatal-warnings ./bootc update
```
Unfortunate, gdp complains about debug info

antheas:
image

It's going somewhere. Rollbacks with skopeo and the loading bar are still missing

image

cgwalters:
OK I pushed an update to this, here's what the json looks like now:

```
"type":"begin","task":"Fetching layers","bytes":1052727512,"bytes_total":2610463862,"steps":31,"steps_total":70}
{"type":"sub-task-bytes","subtask":"ostree chunk d901b0107f5b7eff234e7","bytes":4120577,"total":41266953}
{"type":"sub-task-bytes","subtask":"ostree chunk d901b0107f5b7eff234e7","bytes":24042599,"total":41266953}
{"type":"sub-task-bytes","subtask":"ostree chunk d901b0107f5b7eff234e7","bytes":40377447,"total":41266953}
{"type":"sub-task-bytes","subtask":"ostree chunk 9dc61caa4fc9a69b614b8","bytes":8193,"total":13272558}
{"type":"sub-task-bytes","subtask":"ostree chunk 9dc61caa4fc9a69b614b8","bytes":13272558,"total":13272558}
{"type":"sub-task-bytes","subtask":"ostree chunk fe8a149a3f8a314d64409","bytes":5692519,"total":32825173}
{"type":"sub-task-bytes","subtask":"ostree chunk fe8a149a3f8a314d64409","bytes":22887527,"total":32825173}
{"type":"sub-task-bytes","subtask":"ostree chunk 294da4439e4cb61cd9b2c","bytes":8193,"total":11481423}
{"type":"sub-task-bytes","subtask":"ostree chunk 294da4439e4cb61cd9b2c","bytes":11481423,"total":11481423}
...
```

Rough changes:

- Dropped the new progress bar; I'm not inherently opposed but I'd like to take it as a separate change
- In general worked to minimize the diff vs current code for ease of review
- Reworked the schema to be more stateful, with distinct events, which requires a bit more work on the consumer side, but I think is clearer to understand
- The new schema is also more prepared for multiple concurrent operations
- Dropped the changes to deploy - let's be sure we *can* do that but I'd like to laser focus on the download phase to start
- Changed the lower level ostree-ext code to close the progress channels (fixing the FIXME)
- Made the progress writer async
- Taught the progress writer about `send` vs `send_lossy` (so the `REFRESH_HZ` lives in the writer, not in the producing code)

cgwalters:
> It's going somewhere. Rollbacks with skopeo and the loading bar are still missing

Cool. What do you mean "with skopeo" there?

antheas:
> OK I pushed an update to this, here's what the json looks like now:
> 
> ```
> {"type":"begin","task":"Fetching layers","bytes":1052727512,"bytes_total":2610463862,"steps":31,"steps_total":70}
> {"type":"sub-task-bytes","subtask":"ostree chunk d901b0107f5b7eff234e7","bytes":4120577,"total":41266953}
> {"type":"sub-task-bytes","subtask":"ostree chunk d901b0107f5b7eff234e7","bytes":24042599,"total":41266953}
> {"type":"sub-task-bytes","subtask":"ostree chunk d901b0107f5b7eff234e7","bytes":40377447,"total":41266953}
> {"type":"sub-task-bytes","subtask":"ostree chunk 9dc61caa4fc9a69b614b8","bytes":8193,"total":13272558}
> {"type":"sub-task-bytes","subtask":"ostree chunk 9dc61caa4fc9a69b614b8","bytes":13272558,"total":13272558}
> {"type":"sub-task-bytes","subtask":"ostree chunk fe8a149a3f8a314d64409","bytes":5692519,"total":32825173}
> {"type":"sub-task-bytes","subtask":"ostree chunk fe8a149a3f8a314d64409","bytes":22887527,"total":32825173}
> {"type":"sub-task-bytes","subtask":"ostree chunk 294da4439e4cb61cd9b2c","bytes":8193,"total":11481423}
> {"type":"sub-task-bytes","subtask":"ostree chunk 294da4439e4cb61cd9b2c","bytes":11481423,"total":11481423}
> ...
> {"type":"complete"}
> ```
> 
> Rough changes:
> 
> * Dropped the new progress bar; I'm not inherently opposed but I'd like to take it as a separate change
> * In general worked to minimize the diff vs current code for ease of review
> * Reworked the schema to be more stateful, with distinct events, which requires a bit more work on the consumer side, but I think is clearer to understand

I am in favor of denormalization here. A couple of bytes do not hurt and give flexibility to the receiver to handle the progress bar as they want. As it currently stands, e.g., a bash script cannot consume this and the receiver needs to have decoding logic. The previous standard could be bound to a UI directly.

Begin should also contain some versioning, which is something I omitted as well.

> * The new schema is also more prepared for multiple concurrent operations

New schema drops total image size, which would be required to show a dual stage progress bar that considers cached layers. Subtasks must contain bytes, where there are e.g., deploy steps in the pipeline that do not contain a download, so the new standard cannot be used for them.

> * Dropped the changes to deploy - let's be sure we _can_ do that but I'd like to laser focus on the download phase to start

Since deployment takes 30-40% of the update time, it is required for a proper progress bar.

The diff is smaller which is indeed true.

> > It's going somewhere. Rollbacks with skopeo and the loading bar are still missing
> 
> Cool. What do you mean "with skopeo" there?

Some users might want to rebase to a slightly older version.  So the idea would be to pull the manifest with skopeo, which also contains the tags, and allow them to choose one from the 5 previous ones.

cgwalters:
> New schema drops total image size, 

It's there in the `Begin` task

`{ "type":"begin","task":"Fetching layers","bytes":1052727512,"bytes_total":2610463862,"steps":31,"steps_total":70}`

> I am in favor of denormalization here. A couple of bytes do not hurt and give flexibility to the receiver to handle the progress bar as they want. As it currently stands, e.g., a bash script cannot consume this and the receiver needs to have decoding logic. The previous standard could be bound to a UI directly.

OK I think it's a fair point that we should develop this with a reference bash script (or other consumer code) to validate ergonomics. Now, "cannot consume" seems inaccurate but if it makes it easier to consume by emitting some redundant data, sure.

> Subtasks must contain bytes, where there are e.g., deploy steps in the pipeline that do not contain a download, so the new standard cannot be used for them.

Yeah, I was thinking of adding a `SubTaskStep` that just displays a message and increments the steps.

> Since deployment takes 30-40% of the update time, it is required for a proper progress bar.

Do you have the [latest ostree](https://github.com/ostreedev/ostree/releases/tag/v2024.9) ?

> Some users might want to rebase to a slightly older version. So the idea would be to pull the manifest with skopeo, which also contains the tags, and allow them to choose one from the 5 previous ones.

OK, makes sense! Hmm, I wonder if we should add this to bootc too just to streamline the minor bits of referencing the right image and authfiles, but obviously something we can do later if need be.

antheas:
> > New schema drops total image size,
> 
> It's there in the `Begin` task

Yeah, I looked at it again today. It was a bit late yesterday, oups.

> `{ "type":"begin","task":"Fetching layers","bytes":1052727512,"bytes_total":2610463862,"steps":31,"steps_total":70}`
> 
> > I am in favor of denormalization here. A couple of bytes do not hurt and give flexibility to the receiver to handle the progress bar as they want. As it currently stands, e.g., a bash script cannot consume this and the receiver needs to have decoding logic. The previous standard could be bound to a UI directly.
> 
> OK I think it's a fair point that we should develop this with a reference bash script (or other consumer code) to validate ergonomics. Now, "cannot consume" seems inaccurate but if it makes it easier to consume by emitting some redundant data, sure.

Its probably around 100 lines of bash v4+, since it'd need associative arrays.

> > Subtasks must contain bytes, where there are e.g., deploy steps in the pipeline that do not contain a download, so the new standard cannot be used for them.
> 
> Yeah, I was thinking of adding a `SubTaskStep` that just displays a message and increments the steps.

I will add it today so I have something to work with

> > Since deployment takes 30-40% of the update time, it is required for a proper progress bar.
> 
> Do you have the [latest ostree](https://github.com/ostreedev/ostree/releases/tag/v2024.9) ?

Version: '2024.9'

Seems so. In any case, we are not talking about extreme lengths here. Total update is 4-8m depending on cache invalidation with around 2m being deployment. Committing is around 40s, staging another 30s, and I think around 30s is used for cleanup typically.

> > Some users might want to rebase to a slightly older version. So the idea would be to pull the manifest with skopeo, which also contains the tags, and allow them to choose one from the 5 previous ones.
> 
> OK, makes sense! Hmm, I wonder if we should add this to bootc too just to streamline the minor bits of referencing the right image and authfiles, but obviously something we can do later if need be.

The logic for getting relative images is a bit distro specific, but the RepoTags skopeo arg is all that is needed.

The new REFRESH_HZ will not work for parallel pulls as it might be unfairly discriminate between layers and skew the progress bar. But it is good enough for now.

cgwalters:
> The new REFRESH_HZ will not work for parallel pulls as it might be unfairly discriminate between layers and skew the progress bar. But it is good enough for now.

I spent a few minutes thinking probably too much about micro-optimizing this but one thing I bet would work well would be to cache `ioctl(FIONREAD)` (ref [pipe docs](https://man7.org/linux/man-pages/man7/pipe.7.html)) after a write and skip emitting the next lossy event if the consumer hasn't managed to read from the pipe. This would get us rough "update as fast as consumer can process" semantics.

cgwalters:
> Seems so. In any case, we are not talking about extreme lengths here. Total update is 4-8m depending on cache invalidation with around 2m being deployment. Committing is around 40s, staging another 30s, and I think around 30s is used for cleanup typically.

Perhaps part of the problem here is I spend 90% of my interactive time with a really fast workstation (i9-13900k with multiple NVMe) - the slower comuters I tend to have automated and don't look at directly interactively much.

The deployment phase specifically for me is ~6s now for example, but yeah...two minutes is clearly something we should have progress on. I am guessing that even with the ostree fix the core problem is we're still doing work of O(metadata).

antheas:
> > The new REFRESH_HZ will not work for parallel pulls as it might be unfairly discriminate between layers and skew the progress bar. But it is good enough for now.
> 
> I spent a few minutes thinking probably too much about micro-optimizing this but one thing I bet would work well would be to cache `ioctl(FIONREAD)` (ref [pipe docs](https://man7.org/linux/man-pages/man7/pipe.7.html)) after a write and skip emitting the next lossy event if the consumer hasn't managed to read from the pipe. This would get us rough "update as fast as consumer can process" semantics, without filling the pipe with redundant data.

I was thinking the same as I'd have to spin up a thread. Even then, the way it is denormalized now does not allow for missing messages and still showing correct progress or not stopping the download. When parallel pulls happen, there is potential for skew too.

> > Seems so. In any case, we are not talking about extreme lengths here. Total update is 4-8m depending on cache invalidation with around 2m being deployment. Committing is around 40s, staging another 30s, and I think around 30s is used for cleanup typically.
> 
> Perhaps part of the problem here is I spend 90% of my interactive time with a really fast workstation (i9-13900k with multiple NVMe) - the slower comuters I tend to have automated and don't look at directly interactively much.
> 
> The deployment phase specifically for me is ~6s now for example, but yeah...two minutes is clearly something we should have progress on. I am guessing that even with the ostree fix the core problem is we're still doing work of O(metadata).

The ROG Ally X has the Z1 Extreme, which is a variant of the 7840U that is the fastest ultrathin CPU AMD makes, and a PCIe Gen4 x4 SSD rated at 5GB/s. Single core is at 80% of the i9, multi at ~50%.

https://www.cpubenchmark.net/compare/5322vs5022/AMD-Ryzen-7-7840U-vs-Intel-i9-13900K

I think its the image size. We also have not moved to composefs yet.

cgwalters:
Different topic: One thing I was thinking that'd be very helpful for this is to try to change things so the indicatif code *also* consumes these events - that would make sure it's constantly tested and maintained. I can maybe attempt that next time you want to hand this PR back!

antheas:
> > I was thinking the same as I'd have to spin up a thread.
> 
> You mentioned wanting to support bash which makes sense, but since you mention threads, which language are you targeting? Python?

Yes I am targeting python. I am almost done actually. The only thing left is the rebase dialog loading the previous 5 versions and the progress bar.

https://github.com/hhd-dev/hhd/blob/master/src/hhd/plugins/bootc/__init__.py

> > Even then, the way it is denormalized now does not allow for missing messages and still showing correct progress or not stopping the download.
> 
> The default with `send` is that messages cannot be lost, we only use `send_lossy` for byte-level progress where it's fine if it gets dropped, it just means the progress bar updates less frequently.

Yeah, but if the `lossy` messages consume the buffer and a non-lossy message comes, you are forced to stop. For example, currently, I have an event loop that runs every 2 seconds. If I used that, it would block the update a couple of times so I'd have to spin up a thread.

> > When parallel pulls happen, there is potential for skew too.
> 
> Can you elaborate on this? What would be skewed specifically? We could still at most only drop updates for byte-level progress even with multiple layers at once; a client will still get the start/end notifications for each layer reliably right?

Yeah, they would get the end notifications for the layers, but if e.g., for some reason only the byte updates of a single layer get through while 5 of them are downloading and they go from 20%-80% that is a skew of around 3 layer sizes.

antheas:
> Different topic: One thing I was thinking that'd be very helpful for this is to try to change things so the indicatif code _also_ consumes these events - that would make sure it's constantly tested and maintained. I can maybe attempt that next time you want to hand this PR back!

I plan to get to this PR later today after I finish the full rebase dialog. Could also do tomorrow.

I see four issues right now: deployment progress missing, layer progress skew, potential blocking of the updates, and i18n. For the latter, the way the API is right now does not support internalization. The previous version did.

So before we do the indicatif integration it would be good for all of those to resolve.

antheas:
Ok, so I did partially normalize the schema.

Then I tried to test it and I can't anymore because tokio errors out with `ERROR Switching: Creating progress writer: not a pipe` when using the fd 0 because its not uniquely owned

antheas:
Python 3.13 is dropping pipes, so unless this uses fd 0 it is going to get tricky

cgwalters:
In this stateless model, how would this look when we transition from e.g. pull to commit to deploy type operations? I guess a renderer would detect that `task` has changed? And when task changed, we assume all previous operations were completed?
For a transition to final "complete" we'd emit an event for zero bytes/steps with no subtasks e.g.?

Now we have an enum with one element; I guess that gives us some future extensibility but not obvious to me what we'd add there.

> Python 3.13 is dropping pipes, so unless this uses fd 0 it is going to get tricky

Reference? The python subprocess `pass_fds` should be usable for this, basically call `os.pipe` and provide the write side to `pass_fds` and give that fd number to `--json-fd` right? I can't find anything about them removing either of those APIs. It would seem pretty dramatic if they did try given the [11k references to os.pipe](https://github.com/search?q=lang%3Apython+%22os.pipe%28%29%22+-is%3Afork&type=code) and [1.5k references to pass_fds](https://github.com/search?q=lang%3Apython+%22pass_fds%3D%22+-is%3Afork&type=code). 

> Then I tried to test it and I can't anymore because tokio errors out with ERROR Switching: Creating progress writer: not a pipe when using the fd 0 because its not uniquely owned

Hmm. OK, so we should definitely have a small demo python parser for this executed as part of the test suite.

antheas:
> In this stateless model, how would this look when we transition from e.g. pull to commit to deploy type operations? I guess a renderer would detect that `task` has changed? And when task changed, we assume all previous operations were completed? For a transition to final "complete" we'd emit an event for zero bytes/steps with no subtasks e.g.?

Task and id would change. Yes, the caller would assume they completed. Then, the caller could hardcode a ratio between the pull task and the deploy task to show a single progress bar. Or the caller could show two different progress bars. 

> Now we have an enum with one element; I guess that gives us some future extensibility but not obvious to me what we'd add there.

As for deploy, we can add a `ProgressStep` struct to the enum that contains no bytes. The subtasks there can have a boolean `completed`, start as false and become true. I ran out of time to do that today. 

But I did finish the rest of the updater. It includes checking for updates, updating, canceling updates, rollbacks, rebase to different branches and for each branch pinning one of the last 5 versions, undoing the pin, and detecting when rpm-ostree incompatibilities exist and offering to do rpm-ostree reset

> > Python 3.13 is dropping pipes, so unless this uses fd 0 it is going to get tricky
> 
> Reference? The python subprocess `pass_fds` should be usable for this, basically call `os.pipe` and provide the write side to `pass_fds` and give that fd number to `--json-fd` right? I can't find anything about them removing either of those APIs. It would seem pretty dramatic if they did try given the [11k references to os.pipe](https://github.com/search?q=lang%3Apython+%22os.pipe%28%29%22+-is%3Afork&type=code) and [1.5k references to pass_fds](https://github.com/search?q=lang%3Apython+%22pass_fds%3D%22+-is%3Afork&type=code) on GitHub (just a small sample of code in the wild).

Pass_fds is optional anyway.

https://docs.python.org/3/library/pipes.html

Yeah, `os.pipe()` should work. I missed that one. I will try that tomorrow.

> 
> > Then I tried to test it and I can't anymore because tokio errors out with ERROR Switching: Creating progress writer: not a pipe when using the fd 0 because its not uniquely owned
> 
> Hmm. OK, so we should definitely have a small demo python parser for this executed as part of the test suite.

I did not use python. I was trying to use bash with it just to see the output by eye. The previous implementation with `--json-fd` worked. Perhaps you can share your script.

cgwalters:
> Task and id would change. Yes, the caller would assume they completed. Then, the caller could hardcode a ratio between the pull task and the deploy task to show a single progress bar. Or the caller could show two different progress bars.

So it's possible now that an entire "task" could be skipped in theory if it happened fast enough. I guess it's a bit of a corner case: do we care about showing an 100% complete bar if the task happened fast enough? I feel like this would matter a little bit more if we changed the default TTY renderer to consume this stream? But...eh I can't claim this is a serious problem either, I just still find myself leaning towards the previous schema.

antheas:
> > Task and id would change. Yes, the caller would assume they completed. Then, the caller could hardcode a ratio between the pull task and the deploy task to show a single progress bar. Or the caller could show two different progress bars.
> 
> So it's possible now that an entire "task" could be skipped in theory if it happened fast enough. I guess it's a bit of a corner case: do we care about showing an 100% complete bar if the task happened fast enough? I feel like this would matter a little bit more if we changed the default TTY renderer to consume this stream? But...eh I can't claim this is a serious problem either, I just still find myself leaning towards the previous schema.

I guess so, well, i can partially undo the change and fire the last message as required. This seems like a sane choice.

cgwalters:
> I guess so, well, i can partially undo the change and fire the last message as required. This seems like a sane choice.

Yeah that'd help. Actually I started to write tests for this and that's where the lossiness could definitely show up as a flake. Always emitting a message when changing tasks seems good enough probably.

antheas:
I still cannot get it to work still. I thought it was inheritability, indeed `os.pipe` is not inheritable. However, even even with `os.set_inheritable(w, True)` (or pipe2 which is the same) it still does not work.

`fatal runtime error: IO Safety violation: owned file descriptor already closed`
```python
import subprocess
import os

proc = None
try:
    r, w = os.pipe2(0)
    proc = subprocess.Popen(
        [
            "sudo",
            "/tmp/bootc",
            "switch",
            "ghcr.io/ublue-os/bazzite:unstable",
            f"--json-fd={w}",
        ],
    )

    with os.fdopen(r) as f:
        while True:
            line = f.readline()
            if not line:
                break
            print(line, end="")

    proc.wait()

except KeyboardInterrupt:
    if proc:
        proc.terminate()
```

cgwalters:
You need to provide the `pass_fds` to `Popen`

antheas:
I tried both individually. You are right, it needed both inheritability and pass_fds together

antheas:
I readded staging and import as separate stages.

Then fixed up the script and made it close correctly after the image is merged.

```python
import subprocess
import os

proc = None
try:
    r, w = os.pipe2(0)

    os.set_inheritable(r, True)

    proc = subprocess.Popen(
        [
            "/tmp/bootc",
            "switch",
            "ghcr.io/ublue-os/bazzite:testing",
            f"--json-fd",
            str(w),
        ],
        pass_fds=[w],
        stderr=subprocess.DEVNULL,
    )

    # Close fd in parent process
    os.close(w)

    with os.fdopen(r) as f:
        while proc.poll() is None:
            line = f.readline()
            if not line:
                break
            print(line, end="")

    proc.wait()

except KeyboardInterrupt:
    if proc:
        proc.terminate()
```

antheas:
I fixed the style to match `bootc status --json`.

And I think its ready to fork and use for my usecase/it's getting close to be mergeable.

antheas:
And here it is
https://github.com/hhd-dev/hhd/commit/de34916c28748c2b023109748f89dbade4433067

[bootc_update.webm](https://github.com/user-attachments/assets/fde5bcef-6be7-4ad0-b587-9b709545d025)

cgwalters:
Juggling a whole lot here but I should be able to circle back to this one by end of the week. One thing that would be helpful (that I was planning to do, to be clear) is add integration testing for this in our tmt-based tests (the current unit test is too synthetic). If you or someone else wants to jump on that it'd help, but again since the current test suite is a bit bespoke and a bit of an investment to learn I was planning to do it myself (also to help get familiar with the code).

The other thing we probably should do is have a basic doc for this in `docs/src/bootc-via-api.md` or so.

antheas:
No rush from my side, i pinned the current hash on a different branch so that we can deploy that.

I can try to work a bit on the docs later in the week but I am also busy this week so I do not think I can jump on the tests.

mvo5:
Just a quick drive-by/sidenote - in osbuild we had a similar discussion and have a similar json based progress/status format, we went with jsonseq though mostly because there is https://datatracker.ietf.org/doc/html/rfc7464 which defines it. But it seems there is no clear "standard" in the json streaming world so jsonn is a fine choice too.

antheas:
@mvo5 so essentially it is the same here but we prepend with 0x1E?

It is something to consider. I will leave the choice up to Colin.

mvo5:
> @mvo5 so essentially it is the same here but we prepend with 0x1E?

Yes, that is my understanding too.

> It is something to consider. I will leave the choice up to Colin.

Yeah, I have no strong opinion either way, they are all so similar, we were on the fence flipped a coin (well, picked it because ofthe rfc)

cgwalters:
The low level protocol aspect ties into the higher level question of whether there's any way we could share a schema for the actual JSON, I'd love to be able to do that. I mean if we land this as is in bootc, surely at some point we'd end up bridging the bootc json progress into an osbuild progress?

Is the osbuild progress API stable or could we change it?

On the question of jsonl vs jsonseq, well...you know what's kind of maddening honestly is that varlink chose to terminate with a NUL byte instead. Honestly none of these are really difficult to handle...

jsonl seems maybe more widely used because it's just so simple. But I would say IMO varlink's NUL terminator is actually the most obvious here and even though it's not a RFC, it is at least used.

So umm...I dunno, I guess if I had to pick between all 3 of these things I'd say let's go with the varlink style NUL.

But ultimately again it's a less interesting question than trying to share a progress schema right?

antheas:
I'd rather stay with jsonseq or jsonl. Both are separated with \n so they can be readlined

Then jsonseq has the extra delimiter at the start that can be skipped

NUL would make a lot of file handlers close

cgwalters:
> NUL would make a lot of file handlers close

I don't think that's true, do you have a reference? Go and Rust both have nice high level APIs for this. I am sure there is something in Python but if it exists in the stdlib I couldn't find it in a quick search (or query with a LLM)

- https://godocs.io/bufio#Reader.ReadSlice
- https://doc.rust-lang.org/std/io/trait.BufRead.html#method.read_until

Anyways though, yeah I don't think this matters hugely, I am also totally fine with jsonl or jsonseq. Though AFAICS nothing in the serde_json docs actually guarantees it won't emit a literal newline so maybe that's an argument for jsonseq (or trying to add that guarantee to serde_json).

antheas:
> > NUL would make a lot of file handlers close
> 
> I don't think that's true, do you have a reference? Go and Rust both have nice high level APIs for this. I am sure there is something in Python but if it exists in the stdlib I couldn't find it in a quick search (or query with a LLM)
> 
> * https://godocs.io/bufio#Reader.ReadSlice
> * https://doc.rust-lang.org/std/io/trait.BufRead.html#method.read_until
> 
> Anyways though, yeah I don't think this matters hugely, I am also totally fine with jsonl or jsonseq. Though AFAICS nothing in the serde_json docs actually guarantees it won't emit a literal newline so maybe that's an argument for jsonseq (or trying to add that guarantee to serde_json).

I got confused, EOL is different. But in any case, sticking `\0` in the middle of the output is bound to cause (unnecessary) problems.

jsonseq still appends `\n`, its the same standard. But it has 0x13 as a backup. In that way it is better because you can e.g., pretty print the output optionally. So I'd support using that instead. The wart with it is that you have to "eat" 0x13 if you dont stop on it but its not a big deal and I can do that when this merges.

cgwalters:
@mvo5 I'm looking at https://github.com/osbuild/osbuild/commit/a1eaf3da3a805d09b2372a3135bf5d25aae9545a#diff-0d30cb1d123e195a69ca1c376d6493f22171312db57f78f468d1481b3bfbaa06R93 and thinking about this.

 It looks like in that schema there's no explicit separation like we have now between "byte downloads" vs "steps"...that seems desirable here. Something like an optional "disposition": "bytes" maybe?

It also looks like there's no "total number of bytes/steps"...one could hack that by I guess just creating progress entries for them too that stay at zero, and encourage UI renderers to not show progress bars for steps that are still at zero?

cgwalters (approved):
There's some things I'd like to tweak and get tested more but we just cut a release, and I'd like this to get some max soak time in git main and our continuous releases so let's get this in.

cgwalters:
Just rebased 🏄 (to get CI fixes from git main) and squashed ⏬ (because I tend to prefer a relatively lean "git log" even though renovate ruins it).

PR #925, opened 2024-12-01
feat: add thin build profile for embedded applications

antheas:
Add a profile that allows squeezing every single bit out of the output executable.

Embedded applications will inevitably require squeezing every single bit out of bootc. Current release executable stands at 140mb due to including debuginfo. This is stripped by the RPM but consumers/forks that do not do that are left with the symbols.

Runs with  `cargo build --profile=thin`. 

Produces an executable with size 5.8mb, current fedora executable of ver 1.1.2 is 7.1mb.

The idea here would be to use for an embedded web installer, that is used by e.g., double clicking an executable or installing an apk file which adds a light EFI executable/bootloader to the device's boot/EFI partition.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,19 @@ panic = "abort"
 # We assume we're being delivered via e.g. RPM which supports split debuginfo
 debug = true
 
+[profile.thin]
+# drop bootc size when split debuginfo is not available and go a step
+# further in size optimization (when tested from 140mb, to 12mb without 
+# symbols/debuginfo, to 5.8mb with extra optimizations)
+# https://github.com/johnthagen/min-sized-rust
+# cargo build --profile=thin
+inherits = "release"
+debug = false       # Re-strip debug symbols
+strip = true        # Strip symbols from binary
+lto = true          # Use full lto to remove dead code
+opt-level = 's'     # Optimize for size with vector vectorization
+codegen-units = 1   # Reduce number of codegen units to increase optimizations
+
 [profile.releaselto]
 codegen-units = 1
 inherits = "release"


cgwalters:
I rebased this to pick up https://github.com/containers/bootc/pull/927

cgwalters:
> places a signed shim, grub, kernel, and initramfs

(Tangential: the first 3 are well known, the latter is the tricky part as it ties kernel to userspace and has larger implications. I think if the goal isn't to create "sealed" systems it's not necessary to "sign" the initramfs so much as ensure that there's a chain from kernel commandline to at least find the initramfs and do some basic validation of it if desired)

> The initramfs will contain an embedded installer with bootc, some wifi firmware, and https://lvgl.io/, so it will never switchroot and stay within the initramfs.

Another probably viable path is to download the bootstrap content onto the main Windows partition and then mount that from the bootstrap EFI path. I was thinking Bitlocker might be a problem for this but it looks like [cryptsetup supports unlocking it](https://man.archlinux.org/man/core/cryptsetup/cryptsetup.8.en#BITLK_(WINDOWS_BITLOCKER_COMPATIBLE)_EXTENSION).

That'd lift all concerns about the EFI partition size; what lands there could just be a small amount of code to bootstrap.

This all said...there's a lot of moving parts to this versus just asking people to insert a USB stick - there's already maintained Windows apps to flash USB sticks. What we could clearly optimize though is having a two-phase thing where the USB stick OS is just a launcher that can run `bootc install` or equivalent as a live OS (and can act as its own "additional image store" and have cached layers that we don't redownload)?

Anyways this PR clearly isn't the right place to discuss this, let me know if you end up creating one (would it be part of ublue?).

PR #933, opened 2024-12-04
test: Remove e2e test

henrywang:
This PR removed e2e test here because QE already has daily e2e test for bootc latest main copr build.

xref: https://github.com/containers/bootc/issues/868


cgwalters (approved):

cgwalters:
BTW I asked and the recent constant e2e failures here are because it's unable to provision virt-capable metal machines in the target cloud.

I would like to investigate again using nested virt in GCP at least instead as we've had pretty decent luck with using that for at least sanity testing.

henrywang:
> I would like to investigate again using nested virt in GCP at least instead as we've had pretty decent luck with using that for at least sanity testing.

I use GCP VM as runner to run RHEL for Edge test (nested virt required) before. It works.
We can drop testing farm and use tmt to run test on GCP VM directly.

PR #935, opened 2024-12-04
Install with fsverity enabled + required

cgwalters:
The goal of this PR is to align with [this PR to ostree](https://github.com/ostreedev/ostree/pull/3354) so that to to hard require fsverity for `bootc install`, and chain it to the ostree config, you can do e.g.:

```
RUN echo -e '[composefs]\nenabled=verity' > /usr/lib/ostree/prepare-root.conf
```


cgwalters:
Needs a rebase, conflicts should be relatively straightforward to fix

allisonkarlitskaya:
Can you explain some scenarios where we might want to have fs-verity disabled?  Is it just on filesystems that can't support it?  Because even without checking the verity data or anything, I find the "this inode is now immutable" thing to be extremely compelling, particularly in the presence of multiple hardlinks....

cgwalters:
TODO: Also check for https://github.com/ostreedev/ostree/pull/3354 - if that's set then we should key off it too.

In fact, maybe that should be the sole interface for now.

cgwalters:
So this is getting closer, but I'm seeing an issue where some zero-sized objects (different ones due to different selinux labels) don't have fsverity enabled. It must be an ostree bug, but I wasn't able to reproduce in a quick test.

cgwalters:
Split out of this PR so far as prep:

- https://github.com/containers/bootc/pull/1175
- https://github.com/containers/bootc/pull/1178

cgwalters:
TODO:

- Docs for this
- Support for upgrades (we should check post fetching, before reboot, check if new image enables fsverity, if so do it *before* rebooting)

jmarrero (approved):
lgtm

PR #937, opened 2024-12-05
tests: Fix lbi test

cgwalters:
tests: Adapt for newer nushell

It looks like the default `column1` became `column0` which
broke this. But really we don't need a table, we just need a list;
and piping to `from csv` was a weird way to go about splitting
lines into a list.

Change the assertion to just expect a list which simplifies
the code.

---

Drop copy of tmt bootc plugin

It's now merged and shipped, and us having a copy breaks things.

---

Closes: https://github.com/containers/bootc/issues/936


cgwalters:
One thing I hadn't realized is that nushell is still [making breaking changes](https://github.com/nushell/nushell/issues?q=label%3Apr%3Abreaking-change+is%3Aclosed) - this is the second one of these that bit us, and I don't think we have really complex/unusual code here.

This makes me less sanguine about using it for this but...for now let's just keep this and hope it settles down.

cgwalters:
Huh, the integration test there hit
> ERROR Switching: Pulling: Importing: Parsing layer blob sha256:73a9da8ed490b51b03fe0109a75cea245a5fda85410ebd76820aeb110854a7f5: failed to invoke method FinishPipe: failed to invoke method FinishPipe: write |1: broken pipe

Again...I thought we'd had that one fixed!

jeckersb:
> Huh, the integration test there hit
> 
> > ERROR Switching: Pulling: Importing: Parsing layer blob sha256:73a9da8ed490b51b03fe0109a75cea245a5fda85410ebd76820aeb110854a7f5: failed to invoke method FinishPipe: failed to invoke method FinishPipe: write |1: broken pipe
> 
> Again...I thought we'd had that one fixed!

Do not like :-1:

cgwalters:
Rebased just to see if that helps

cgwalters:
YES! Integration tests passed again.

henrywang (approved):

PR #938, opened 2024-12-05
test: Remove tmt bootc plugin code and use released bootc plugin

henrywang:
Remove tests-intetration folder because TMT bootc plugin replaces tests-integration


cgwalters:
Thanks for looking at this! I had started on a similar patch but it started to get big and so I did just https://github.com/containers/bootc/pull/937

These don't logically conflict but let's land 937 first?

cgwalters:
Hmm but now we entirely lost the tmt tests running on PRs here?

cgwalters:
Test jobs seem dead now with
> fatal: not a git repository (or any parent up to mount point /)

henrywang:
Yes! I got everything works now.

cgwalters (approved):
Nice!

PR #939, opened 2024-12-05
Use a shared const and helpers for run/ostree-booted

cgwalters:
Just a code cleanup.

Closes: https://github.com/containers/bootc/issues/934

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -17,6 +17,7 @@ use fn_error_context::context;
 use ostree::gio;
 use ostree_container::store::PrepareResult;
 use ostree_ext::container as ostree_container;
+use ostree_ext::container_utils::ostree_booted;
 use ostree_ext::keyfileext::KeyFileExt;
 use ostree_ext::ostree;
 use schemars::schema_for;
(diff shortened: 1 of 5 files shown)


lib/src/cli.rs
```diff
@@ -611,7 +612,7 @@ fn prepare_for_write() -> Result<()> {
     if ostree_ext::container_utils::running_in_container() {
         anyhow::bail!("Detected container; this command requires a booted host system.");
     }
-    if !std::path::Path::new("/run/ostree-booted").exists() {
+    if !is_ostree_booted()? {
         anyhow::bail!("This command requires an ostree-booted host system");
     }
```
omertuc, line 617: "if condition then bail" can be ensure

```suggestion
    anyhow::ensure!(is_ostree_booted()?, "This command requires an ostree-booted host system");
```

(If we're already changing this code)

ostree-ext/src/container_utils.rs
```diff
@@ -64,6 +71,17 @@ pub fn is_bare_split_xattrs() -> Result<bool> {
     }
 }
 
+/// Returns true if the system appears to have been booted via ostree.
+/// This accesses global state in /run.
+pub fn is_ostree_booted() -> io::Result<bool> {
```
omertuc, line 76:
```suggestion
pub fn ostree_booted() -> io::Result<bool> {
```

Just a suggestion, up to you, I feel it reads better

omertuc (changes requested):

cgwalters:
Good suggestions thanks, updated!

omertuc (approved):

omertuc (approved):

PR #940, opened 2024-12-06
ext: Replace terminal_size with comfy-table

cgwalters:
I was looking at our vendoring set and while it's not actually relevant I found myself wondering why we had *three* versions of `windows-sys`. Having that many crate versions is often a signal that there's an unmaintained dependency.

And indeed, `terminal_size` is no longer cool. The "in" crowd has moved on to newer, hipper things. Life moves fast, we need to keep up.

(OK but yes also this drops some manual column printing code
 we had which is also a win)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -120,7 +120,7 @@ dependencies = [
  "miniz_oxide",
  "object",
  "rustc-demangle",
- "windows-targets 0.52.6",
+ "windows-targets",
 ]
 
 [[package]]
@@ -351,7 +351,7 @@ dependencies = [
  "num-traits",
  "serde",
  "wasm-bindgen",
- "windows-targets 0.52.6",
+ "windows-targets",
 ]
 
 [[package]]
@@ -1422,6 +1422,7 @@ dependencies = [
  "chrono",
  "clap",
  "clap_mangen",
+ "comfy-table",
  "containers-image-proxy",
  "flate2",
  "fn-error-context",
@@ -1449,7 +1450,6 @@ dependencies = [
  "similar-asserts",
  "tar",
  "tempfile",
- "terminal_size",
  "tokio",
  "tokio-stream",
  "tokio-util",
@@ -1497,7 +1497,7 @@ dependencies = [
  "libc",
  "redox_syscall",
  "smallvec",
- "windows-targets 0.52.6",
+ "windows-targets",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


omertuc:
Trying to test this PR, I built it into a bootc image, `bootc switch`'d into it, then I was wondering if I was getting the ostree-rs-ext version from this repo or some other one:

```
[omer@hal9000 ~]$ ls -lah /usr/libexec/libostree/ext/ostree-container
lrwxrwxrwx. 1 root root 23 Jan  1  1970 /usr/libexec/libostree/ext/ostree-container -> ../../../bin/rpm-ostree
```

Why is this pointing at ../../../bin/rpm-ostree? I would expected it to point at a binary related to ostree-ext?

omertuc:
image
Not sure which is worse :laughing:

cgwalters:
The `| head` thing is an old, old problem... https://github.com/rust-lang/rust/issues/46016

cgwalters:
I'm not getting the extra newlines here...I did sanity check this too before pushing

omertuc:
https://github.com/user-attachments/assets/e644399c-e911-4c21-bef9-a754d689a519

omertuc:
> Ahh wait I _do_ get the newlines if one of the output lines is long enough to wrap past the terminal size. I think this may be fixable with [docs.rs/comfy-table/latest/comfy_table/enum.ContentArrangement.html](https://docs.rs/comfy-table/latest/comfy_table/enum.ContentArrangement.html) ?

Yeah

```rust
use comfy_table::{presets::NOTHING, Table};

fn main() {
    let mut table = Table::new();

    table
        .load_preset(NOTHING)
        .set_header(vec!["REPOSITORY", "TYPE"])
        .set_content_arrangement(comfy_table::ContentArrangement::Dynamic);

    table.add_row(vec!["foo", "bar"]);
    table.add_row(vec![
        "fooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooo",
        "baaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaar",
    ]);

    println!("{table}");
}
```

Seems well behaved

omertuc:
```rust
use std::env::args;

use comfy_table::{presets::NOTHING, Table};

fn main() {
    let mut table = Table::new();

    table
        .load_preset(NOTHING)
        .set_header(vec!["REPOSITORY", "TYPE"])
        .set_content_arrangement(match args().nth(1).as_deref() {
            Some("d") => comfy_table::ContentArrangement::Disabled,
            Some("dyn") => comfy_table::ContentArrangement::Dynamic,
            _ => comfy_table::ContentArrangement::DynamicFullWidth,
        });

    table.add_row(vec!["foo", "bar"]);
    table.add_row(vec![
        "fooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooo",
        "baaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaar",
    ]);

    println!("{table}");
}
```

Not sure I understand the difference between `Dynamic` and `DynamicFull`, didn't make a diff in my toy demo


image

omertuc (approved):

omertuc (approved):

PR #943, opened 2024-12-06
Cmd pdeathsig

cgwalters:
utils: Add a lifecycle_bind helper for Command

In almost all children we fork, we want the child to reliably
exit if we do (e.g. especially if we panic). The Linux
PR_SET_PDEATHSIG is just great for this.

---

utils: Use `run` helper for editor

I just happened to glance at this code, this gives us stderr
in the error, etc.

---

### DIFF
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -840,7 +840,8 @@ async fn install_container(
 /// Run a command in the host mount namespace
 pub(crate) fn run_in_host_mountns(cmd: &str) -> Command {
     let mut c = Command::new("/proc/self/exe");
-    c.args(["exec-in-host-mount-namespace", cmd]);
+    c.lifecycle_bind()
+        .args(["exec-in-host-mount-namespace", cmd]);
     c
 }
 
--- a/lib/src/utils.rs
+++ b/lib/src/utils.rs
@@ -117,16 +117,12 @@ pub(crate) fn spawn_editor(tmpf: &tempfile::NamedTempFile) -> Result<()> {
     let argv0 = editor_args
         .next()
         .ok_or_else(|| anyhow::anyhow!("Invalid editor: {editor}"))?;
-    let status = Command::new(argv0)
+    Command::new(argv0)
         .args(editor_args)
         .arg(tmpf.path())
         .lifecycle_bind()
-        .status()
-        .context("Spawning editor")?;
-    if !status.success() {
-        anyhow::bail!("Invoking editor: {editor} failed: {status:?}");
-    }
-    Ok(())
+        .run()
+        .with_context(|| format!("Invoking editor {editor} failed"))
 }
 
 /// Convert a combination of values (likely from CLI parsing) into a signature source
--- a/lib/src/utils.rs
+++ b/lib/src/utils.rs
@@ -5,6 +5,7 @@ use std::process::Command;
 use std::time::Duration;
 
 use anyhow::{Context, Result};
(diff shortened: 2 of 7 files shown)


utils/src/command.rs
```diff
@@ -84,6 +88,19 @@ impl CommandRunExt for Command {
         self.status()?.check_status(stderr)
     }
 
+    #[allow(unsafe_code)]
+    fn lifecycle_bind(&mut self) -> &mut Self {
+        // SAFETY: This API is safe to call in a forked child.
+        unsafe {
+            self.pre_exec(|| {
+                rustix::process::set_parent_process_death_signal(Some(
```
allisonkarlitskaya, line 96: Oh, that's real nice.  If you wanna do that from Python you're gonna be ctyping your way into `prctl()` with some hardcoded constants...

utils/src/command.rs
```diff
@@ -84,6 +88,19 @@ impl CommandRunExt for Command {
...
+                    rustix::process::Signal::Term,
+                ))
+                .map_err(Into::into)
```
allisonkarlitskaya, line 99: Would it have worked to use `?` here to get the coercion?

utils/src/command.rs
```diff
@@ -84,6 +88,19 @@ impl CommandRunExt for Command {
         self.status()?.check_status(stderr)
     }
 
+    #[allow(unsafe_code)]
+    fn lifecycle_bind(&mut self) -> &mut Self {
+        // SAFETY: This API is safe to call in a forked child.
```
allisonkarlitskaya, line 93: A docs reference would be good (even if it's just to say that this is implemented with prctl and point to the list in signal-safety manpage)...

lib/src/utils.rs
```diff
@@ -117,16 +117,12 @@ pub(crate) fn spawn_editor(tmpf: &tempfile::NamedTempFile) -> Result<()> {
...
         .args(editor_args)
         .arg(tmpf.path())
         .lifecycle_bind()
-        .status()
-        .context("Spawning editor")?;
-    if !status.success() {
-        anyhow::bail!("Invoking editor: {editor} failed: {status:?}");
-    }
-    Ok(())
+        .run()
```
allisonkarlitskaya, line 124: Thanks for putting this is a separate commit with a good message.  I was really confused when I first pulled this up in files view. :)

allisonkarlitskaya (approved):
This could go in as-is, but see comments/questions below (mostly superficial, or for the sake of my own curiosity)...

lib/src/install.rs
```diff
@@ -840,7 +840,8 @@ async fn install_container(
 /// Run a command in the host mount namespace
 pub(crate) fn run_in_host_mountns(cmd: &str) -> Command {
     let mut c = Command::new("/proc/self/exe");
-    c.args(["exec-in-host-mount-namespace", cmd]);
+    c.lifecycle_bind()
```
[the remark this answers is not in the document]
cgwalters: This function is a bit unusual in that it ends up returning `Command` so it needs to be this way.

utils/src/command.rs
cgwalters: Hmm, the rustix docs for this function do link directly to the prctl man page right?

cgwalters: I think while we *could* have more safety explanations in cases like this, what we have here is actually better than some of the other (few) `unsafe` cases and if we were to try to raise the bar we'd need to a bit more

PR #944, opened 2024-12-06
lints: Add check for /usr/etc

cgwalters:
Ref https://github.com/containers/bootc/issues/942

We don't want to support this, lint against it.


cgwalters:
> Shall we document it?

Done in https://github.com/containers/bootc/pull/951

PR #953, opened 2024-12-09
install: Guide user towards the correct podman flags

omertuc:
See commit message

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -7,7 +7,7 @@ use std::io::Seek;
 use std::os::unix::process::CommandExt;
 use std::process::Command;
 
-use anyhow::{Context, Result};
+use anyhow::{ensure, Context, Result};
 use camino::Utf8PathBuf;
 use cap_std_ext::cap_std;
 use cap_std_ext::cap_std::fs::Dir;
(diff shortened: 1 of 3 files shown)


cgwalters:
Needs a `#[cfg(feature = "install")]`

lib/src/cli.rs
```diff
@@ -576,15 +578,46 @@ pub(crate) async fn get_storage() -> Result<crate::store::Storage> {
...
-        anyhow::bail!("This command requires full root privileges (CAP_SYS_ADMIN)");
-    }
+pub(crate) fn require_root(is_container: bool) -> Result<()> {
+    ensure!(
+        rustix::process::getuid().is_root(),
+        match is_container {
+            true =>
+                "The user inside the container from which you are running this command must be root",
+            false => "This command must be executed as the root user",
+        }
```
cgwalters, line 588: I'd write this 
`if is_container { "The user inside the container from which you are running this command must be root" } else {  "This command must be executed as the root user" }`

lib/src/cli.rs
```diff
@@ -576,15 +578,46 @@ pub(crate) async fn get_storage() -> Result<crate::store::Storage> {
...
+    );
+
+    ensure!(
+        rustix::thread::capability_is_in_bounding_set(rustix::thread::Capability::SystemAdmin)?,
+        match is_container {
+            true => "The container must be executed with the podman --privileged flag",
```
cgwalters, line 594: In the end podman is not the only container runtime, it's also valid to run some of this stuff as a kubernetes pod in some cases for example too. How about "The container must have full privileges (e.g. --privileged)" ?

It's fine as is honestly, but let's just keep things like this in mind as we go forward.

lib/src/cli.rs
```diff
@@ -576,15 +578,46 @@ pub(crate) async fn get_storage() -> Result<crate::store::Storage> {
...
+
+#[context("Querying root podman")]
+pub(crate) fn require_root_podman() -> Result<()> {
+    let proc_self = Dir::open_ambient_dir("/proc/self", cap_std::ambient_authority())?;
+    let uid_map = proc_self.open("uid_map")?;
+    let uid_map = std::io::BufReader::new(uid_map);
+
+    let map_lines = uid_map.lines().collect::<std::io::Result<Vec<_>>>()?;
+
+    // TODO: I'm unaware of more official channels to check for rootless podman, we should
```
cgwalters, line 612: We have an existing related check in https://github.com/containers/bootc/blob/af4d43f93c4867dd73a491f85025f90bf5bf9406/lib/src/install.rs#L1014

cgwalters (changes requested):
Marking requested changes per above

lib/src/install/completion.rs
```diff
@@ -245,7 +245,7 @@ pub(crate) async fn run_from_anaconda(rootfs: &Dir) -> Result<()> {
 
 /// From ostree-rs-ext, run through the rest of bootc install functionality
 pub async fn run_from_ostree(rootfs: &Dir, sysroot: &Utf8Path, stateroot: &str) -> Result<()> {
-    crate::cli::require_root()?;
+    crate::cli::require_root(false)?;
```
cgwalters, line 248: Non-blocking: One larger cleanup I think we could do is have an `Environment` struct that gathers global state like this (uid, whether we have cap_sys_admin, whether `/run/ostree-booted` exists, whether we appear to be in a container), and then query requirements on it.

e.g. here we're passing `false` for `is_container` but we don't actually know that's false or not.

Not a blocker, just an optional followup.

cgwalters (approved):
The improvement from the commit message looks great! 

We could have some tests, but I wouldn't block on that.

PR #956, opened 2024-12-10
timer: Import defaults from systemd-sysupdate.timer

cgwalters:
Inspired by https://github.com/coreos/rpm-ostree/pull/5183/commits/db108a30fe94929bcab7313e0224faccde5edc0d

### DIFF
--- a/systemd/bootc-fetch-apply-updates.timer
+++ b/systemd/bootc-fetch-apply-updates.timer
@@ -1,14 +1,23 @@
 [Unit]
 Description=Apply bootc updates
 Documentation=man:bootc(8)
+# TODO: Tweak this to more strongly conditionalize on using
+# bootc for host updates.
 ConditionPathExists=/run/ostree-booted
 
 [Timer]
+# This is copied from systemd-sysupdate.timer; it's just an arbitrary
+# starting point.
+#
+# Trigger the update 1 hour after boot, and then – on average – every 6h, but
+# randomly distributed in a 2h…6h interval. In addition trigger things
+# persistently once on each Saturday, to ensure that even on systems that are
+# never booted up for long we have a chance to do the update.
 OnBootSec=1h
-# This time is relatively arbitrary and obviously expected to be overridden/changed
-OnUnitInactiveSec=8h
-# When deploying a large number of systems, it may be beneficial to increase this value to help with load on the registry.
-RandomizedDelaySec=2h
+OnUnitActiveSec=2h
+OnCalendar=Sat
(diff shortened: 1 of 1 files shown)


systemd/bootc-fetch-apply-updates.timer
```diff
@@ -1,14 +1,23 @@
...
 Documentation=man:bootc(8)
+# TODO: Tweak this to more strongly conditionalize on using
+# bootc for host updates.
 ConditionPathExists=/run/ostree-booted
 
 [Timer]
+# This is copied from systemd-sysupdate.timer; it's just an arbitrary
+# starting point.
+#
+# Trigger the update 1 hour after boot, and then – on average – every 6h, but
```
jeckersb, line 12: `every 6h` should be `every 4h` here?  If it's randomly picking between 2h and 6h.  Maybe I am misunderstanding something though.

jeckersb (approved):
:+1:, one minor correction if you want to fix (assuming I'm not misunderstanding things).

systemd/bootc-fetch-apply-updates.timer
cgwalters: The comment here is mostly copied verbatim, so any errors are in the original. But it is good to double check. In looking at this...I think offhand you're right that the comment here is wrong or misleading.

It does seem offhand to me that the timer is going to be actually on average 2h (after initial bootup and completion of the timer, ignoring the time it takes to actually execute the check).

jeckersb: I am going to just jot down my flimsy understanding here of how this would play out.  Corrections very much welcomed :smile: 

- T-0: system boots
- T+1h: `OnBootSec=1h` is "triggered", but doesn't fire immediately due to `RandomizedDelaySec=4h` being applied to all "subtimers" within this timer?
- T+~3h: (assume `RandomizedDelaySec=4h` always just produces 2h) The timer actually fires for the first time, activating the associated service unit
- T+~5h: `OnUnitActiveSec=2h` is "triggered" but as above with `OnBootSec` doesn't fire immediately due to delay
- T+~7h: The timer fires the service for the second time
- (loop here)
- T+11h: service fires
- T+15h: service fires
- (and so on looping)
- I guess Saturday at 00:00 the `OnCalendar` bit fires, but is still subjected to the ~2h delay
- Saturday 2am: service fires
- (subtimer for `OnUnitActive` resets?)
- Saturday 4am: `OnUnitActive` triggered but delays
- Saturday 6am: service fires

Something like that?

PR #958, opened 2024-12-10
donotmerge: IBX changes

omertuc:
Some (trash, temporary) changes I had to perform on bootc to get IBX to work

- `./Containerfile` builds a bootable image from an existing normal IBX seed
- `./build_seed.sh` is just a helper script

To install on recipient

```
sudo podman run --privileged --env RUST_LOG=trace -v /var/lib/containers/storage:/var/lib/containers/storage --pid=host -it quay.io/your/seed:image bootc install to-existing-root --acknowledge-destructive --stateroot foo
```

This PR only exists to start discussions around concrete issues, it's not meant to be merged.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,0 +1,26 @@
+FROM quay.io/centos/centos:stream9 as build
+COPY hack/build.sh /build.sh
+RUN /build.sh && rm -v /build.sh
+COPY . /build
+WORKDIR /build
+RUN mkdir -p /build/target/dev-rootfs  # This can hold arbitrary extra content
+# See https://www.reddit.com/r/rust/comments/126xeyx/exploring_the_problem_of_faster_cargo_docker/
+# We aren't using the full recommendations there, just the simple bits.
+RUN --mount=type=cache,target=/build/target --mount=type=cache,target=/var/roothome make test-bin-archive && mkdir -p /out && cp target/bootc.tar /out
+RUN mkdir -p /build/target/dev-rootfs  # This can hold arbitrary extra content
+
+FROM quay.io/otuchfel/ostbackup:serv1 as seed
+
+# ____________________________________________________________________________
+
+FROM quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:5b1124faf4b73753b4679085604dd8cb810c4a7a2e659978f5c80183bb165f94
+
+LABEL com.openshift.lifecycle-agent.seed_format_version=3
+
+RUN mkdir -p /usr/lib/bootc/install
+RUN echo -e '[install.filesystem.root]\ntype = "xfs"' > /usr/lib/bootc/install/00-bootc.toml
+
+COPY --from=seed --exclude=ostree.tgz / /var/tmp/seed
+
+COPY --from=build /out/bootc.tar /tmp
+RUN tar -C / -xvf /tmp/bootc.tar && rm -vrf /tmp/*
--- a/Makefile
+++ b/Makefile
@@ -39,10 +39,10 @@ install-all: install install-ostree-hooks
 	install -D -m 0755 target/release/tests-integration $(DESTDIR)$(prefix)/bin/bootc-integration-tests 
 
 bin-archive: all
-	$(MAKE) install DESTDIR=tmp-install && $(TAR_REPRODUCIBLE) --zstd -C tmp-install -cf target/bootc.tar.zst . && rm tmp-install -rf
+	$(MAKE) install DESTDIR=tmp-install && $(TAR_REPRODUCIBLE) -C tmp-install -cf target/bootc.tar . && rm tmp-install -rf
 
 test-bin-archive: all
-	$(MAKE) install-all DESTDIR=tmp-install && $(TAR_REPRODUCIBLE) --zstd -C tmp-install -cf target/bootc.tar.zst . && rm tmp-install -rf
+	$(MAKE) install-all DESTDIR=tmp-install && $(TAR_REPRODUCIBLE) -C tmp-install -cf target/bootc.tar . && rm tmp-install -rf
 
 test-tmt:
 	cargo xtask test-tmt
--- a/build_seed.sh
+++ b/build_seed.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+SCRIPT_DIR=$(dirname $0)
+
+cd $SCRIPT_DIR
+
+podman build -t bootcseed -f Containerfile .
+podman tag bootcseed:latest quay.io/otuchfel/bootc:seed
+podman push quay.io/otuchfel/bootc:seed
--- a/hack/Containerfile
+++ b/hack/Containerfile
@@ -14,7 +14,7 @@ WORKDIR /build
 RUN mkdir -p /build/target/dev-rootfs  # This can hold arbitrary extra content
 # See https://www.reddit.com/r/rust/comments/126xeyx/exploring_the_problem_of_faster_cargo_docker/
 # We aren't using the full recommendations there, just the simple bits.
-RUN --mount=type=cache,target=/build/target --mount=type=cache,target=/var/roothome make test-bin-archive && mkdir -p /out && cp target/bootc.tar.zst /out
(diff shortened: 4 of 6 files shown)


Containerfile
```diff
@@ -0,0 +1,26 @@
+FROM quay.io/centos/centos:stream9 as build
```
omertuc, line 1: We must build bootc because it doesn't ship with RHCOS yet

Containerfile
```diff
@@ -0,0 +1,26 @@
...
+RUN /build.sh && rm -v /build.sh
+COPY . /build
+WORKDIR /build
+RUN mkdir -p /build/target/dev-rootfs  # This can hold arbitrary extra content
+# See https://www.reddit.com/r/rust/comments/126xeyx/exploring_the_problem_of_faster_cargo_docker/
+# We aren't using the full recommendations there, just the simple bits.
+RUN --mount=type=cache,target=/build/target --mount=type=cache,target=/var/roothome make test-bin-archive && mkdir -p /out && cp target/bootc.tar /out
+RUN mkdir -p /build/target/dev-rootfs  # This can hold arbitrary extra content
+
+FROM quay.io/otuchfel/ostbackup:serv1 as seed
```
omertuc, line 12: Your normal seed image goes here.

Containerfile
omertuc, line 1: In an ideal future all seeds will simply be "bootc seeds", but for now we have this Containerfile to make a "bootc seed" out of a normal seed

Makefile
```diff
@@ -39,10 +39,10 @@ install-all: install install-ostree-hooks
 	install -D -m 0755 target/release/tests-integration $(DESTDIR)$(prefix)/bin/bootc-integration-tests 
 
 bin-archive: all
-	$(MAKE) install DESTDIR=tmp-install && $(TAR_REPRODUCIBLE) --zstd -C tmp-install -cf target/bootc.tar.zst . && rm tmp-install -rf
```
omertuc, line 42: -zstd unsupported on `quay.io/centos/centos:stream9`

build_seed.sh
omertuc, line 1: Some helper script, ignore

hack/Containerfile
omertuc, line 1: This file isn't really used in this PoC, ignore those changes

lib/src/install.rs
```diff
@@ -946,7 +946,7 @@ pub(crate) fn reexecute_self_for_selinux_if_needed(
     override_disable_selinux: bool,
 ) -> Result<SELinuxFinalState> {
     // If the target state has SELinux enabled, we need to check the host state.
-    if srcdata.selinux {
```
omertuc, line 949: The re-exec is failing for some reason, I skipped it and things seems to work alright (maybe because I `sudo setenforce permissive`)

lib/src/install.rs
```diff
@@ -1222,7 +1162,7 @@ async fn prepare_install(
     crate::mount::ensure_mirrored_host_mount("/dev")?;
     crate::mount::ensure_mirrored_host_mount("/var/lib/containers")?;
     ensure_var()?;
-    setup_tmp_mounts()?;
```
omertuc, line 1225: This was failing for some reason so I had to take it out, I'm not quite sure what purpose it serves anyway, but need to dig into why it was failing

lib/src/install.rs
omertuc: update: it works even without setenforce

omertuc:
WHile the install completes "successfully", SSHing after rebooting doesn't work

```
ssh -o IdentityFile=./bip-orchestrate-vm/ssh-key/key -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no core@192.168.126.10
ssh: connect to host 192.168.126.10 port 22: Connection refused
make: *** [Makefile:425: ssh] Error 255
```

omertuc:
image

omertuc:
Host SSH keys were not created... Suspecting selinux?

Containerfile
```diff
@@ -0,0 +1,26 @@
...
+
+# ____________________________________________________________________________
+
+FROM quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:5b1124faf4b73753b4679085604dd8cb810c4a7a2e659978f5c80183bb165f94
+
+LABEL com.openshift.lifecycle-agent.seed_format_version=3
+
+RUN mkdir -p /usr/lib/bootc/install
+RUN echo -e '[install.filesystem.root]\ntype = "xfs"' > /usr/lib/bootc/install/00-bootc.toml
```
cgwalters, line 21: I think this is unnecessary after FCOS starts inheriting from fedora-bootc, you just haven't picked up that change in the base image here right?

Makefile
cgwalters: Wait that's definitely not true. Do you have a really old image? Builds don't pull by default so unfortunately you need to explicitly e.g. `podman pull quay.io/centos-bootc/centos-bootc:stream9` or so.

lib/src/install.rs
cgwalters: Failing with what error message?

lib/src/install.rs
cgwalters: Again what error message?

Containerfile
omertuc: I've added that because I was just starting off from a very old PoC by @!javipolo, it might be completely unnecessary. I'll try to do without it

Makefile
omertuc: I don't believe I have an old image

lib/src/install.rs
omertuc:
```
ERROR Installing to filesystem: Ensuring selinux install_t type: Ensuring selinux install_t type: fsetxattr(security.selinux): Operation not supported (os error 95)
```

But I'm now starting to think this just might be due to my removal of `setup_tmp_mounts`

lib/src/install.rs
omertuc: `ERROR Installing to filesystem: Linking tmp mounts to host: Exchanging /var/tmp <=> /var/tmp.tmp: Invalid cross-device link (os error 18)`

Looking into it

lib/src/install.rs
omertuc: Temporarily working around the `setup_tmp_mounts` error mentioned in the other comment below using `-v /tmp:/tmp -v /var/tmp:/var/tmp` 

But now the re-exec fails with a new error

```
ERROR Installing to filesystem: Ensuring selinux install_t type: Failed to enter install_t (running as unconfined_u:unconfined_r:spc_t:s0) - use BOOTC_SETENFORCE0_FALLBACK=1 to override
```

omertuc: Weirdly the error seems to disappear when `-v /tmp:/tmp` is removed. Confirming that this is true

omertuc: Yep that is indeed the case.

To summarize:

- No tmp mounts at all leads to `ERROR Installing to filesystem: Linking tmp mounts to host: Exchanging /var/tmp <=> /var/tmp.tmp: Invalid cross-device link (os error 18)`

- Both `/tmp/` and `/var/tmp` mounts leads to `ERROR Installing to filesystem: Ensuring selinux install_t type: Failed to enter install_t (running as unconfined_u:unconfined_r:spc_t:s0) - use BOOTC_SETENFORCE0_FALLBACK=1 to override`

- Only `/var/tmp` mount works

Containerfile
omertuc: Yeah it works fine without it

Containerfile
romfreiman: hmm not sure. we can generate new seeds I guess. no need to deal with backwards compatibility at this point. but whatever is easier for the POC of course

omertuc:
Force push because of rebase

omertuc:
> Host SSH keys were not created... Suspecting selinux?
> 
> EDIT: neither did the key I added with `--root-ssh-authorized-keys`

OK figured out it was because of `/etc/ssh/sshd_config.d/40-rhcos-defaults.conf` disabling root login completely, not just with password

cgwalters:
> OK figured out it was because of /etc/ssh/sshd_config.d/40-rhcos-defaults.conf disabling root login completely, not just with password

Ouch...maybe we should revisit that.

PR #961, opened 2024-12-11
Various updates for `--progress-fd`

antheas:
deploy: Avoid cloning progress

Have the printer take ownership and return it. This is generally
prep for making the writer more stateful which will make it easier
to work with.

---

cli: Rename --json-fd to --progress-fd

The format is not as important as what the option *does*. There's
multiple things that can be JSON, but this is specifically
about dynamic progress information.

---

cli: Make --progress-fd hide=true for now

We're still debating some of the interface/semantics here
so let's mark this as hidden.

---

docs: Describe `--progress-fd`, add rendered JSON schema

Signed-off-by: Antheas Kapenekakis <git@antheas.dev>

---

### DIFF
--- a/docs/src/experimental-progress-fd.md
+++ b/docs/src/experimental-progress-fd.md
@@ -12,9 +12,8 @@ The format of data output over `--progress-fd` is [JSON Lines](https://jsonlines
 which is a series of JSON objects separated by newlines (the intermediate
 JSON content is guaranteed not to contain a literal newline).
 
-The current API version is `org.containers.bootc/progress/v1`. You can find
-the JSON schema describing this version here:
-[progress-v1.schema.json](progress-v1.schema.json).
+You can find the JSON schema describing this version here:
+[progress-v0.schema.json](progress-v0.schema.json).
 
 Deploying a new image with either switch or upgrade consists
 of three stages: `pulling`, `importing`, and `staging`. The `pulling` step
--- a/docs/src/experimental-progress-fd.md
+++ b/docs/src/experimental-progress-fd.md
@@ -1,0 +1,33 @@
+
+# Interactive progress with `--progress-fd`
+
+This is an experimental feature; tracking issue: <https://github.com/containers/bootc/issues/1016>
+
+While the `bootc status` tooling allows a client to discover the state
+of the system, during interactive changes such as `bootc upgrade`
+or `bootc switch` it is possible to monitor the status of downloads
+or other operations at a fine-grained level with `-progress-fd`.
+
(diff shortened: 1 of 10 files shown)


cgwalters:
I did some tweaks here, though we still need to generate a schema for the progress and probably clarify even more strongly the guidelines for how we evolve it.

antheas:
Would also be good to decide if you prefer `org.containers.bootc/progress/v1` or `org.containers.bootc.progress/v1` and match them. The code uses the latter, the docs the former.

cgwalters:
> Would also be good to decide if you prefer org.containers.bootc/progress/v1 or org.containers.bootc.progress/v1 and match them. The code uses the latter, the docs the former.

I think we should use a semantic version here instead (without any namespace prefixing because it's obvious the data here is coming from bootc), and a semantic version would allow us to better track protocol evolutions (e.g. a client can know to expect new fields). That said of course in practice JSON often used precisely because it's "loose" and new fields can be added at any time.

Really also instead of having a version number in each event, I think it'd make sense to emit that version once as a special event type.

docs/src/experimental-progress-fd.md
```diff
@@ -0,0 +1,33 @@
+
+# Interactive progress with `--progress-fd`
+
+This is an experimental feature; tracking issue: <https://github.com/containers/bootc/issues/1016>
+
+While the `bootc status` tooling allows a client to discover the state
+of the system, during interactive changes such as `bootc upgrade`
+or `bootc switch` it is possible to monitor the status of downloads
+or other operations at a fine-grained level with `-progress-fd`.
```
jeckersb, line 9: missing a `-` in `--progress-fd`

lib/src/progress_jsonl.rs
```diff
@@ -198,8 +217,17 @@ impl ProgressWriter {
...
+            inner.sent_start = true;
+            let start = Event::Start {
+                version: API_VERSION.into(),
+            };
+            Self::send_impl_inner(inner, &start).await?;
+        }
+
         // For messages that can be dropped, if we already sent an update within this cycle, discard this one.
-        // TODO: Also consider querying the pipe buffer and also dropping if we can't do this write.
+        // TODO: Also consider querying the pipe buffer and also dropping if wqe can't do this write.
```
jeckersb, line 230: typo snuck in here

jeckersb (approved):
LGTM, just two typo nits

jeckersb:
missed that automerge was on, will file typo fix followup :)

PR #964, opened 2024-12-12
install: Lift `hide=true` from `install ensure-completion`

cgwalters:
While this has obscure use cases right now, we will need to support it for the forseeable future, so just lift its `hide=true` state so it's clear that it exists.

Update the generated man pages.


cgwalters:
Rolling into https://github.com/containers/bootc/pull/965

PR #972, opened 2024-12-13
docs/bootc-install: Give more guidance on `root=` and `rootflags=`

cgwalters:
We need to change Anaconda and bootc-image-builder to stop generating `/etc/fstab` for this reason.

### DIFF
--- a/docs/src/bootc-install.md
+++ b/docs/src/bootc-install.md
@@ -148,9 +148,12 @@ can be configured off at build time via Cargo features.
 ### Using `bootc install to-filesystem`
 
 The usual expected way for an external storage system to work
-is to provide `root=<UUID>` type kernel arguments. Note that
-if a separate `/boot` is needed (e.g. for LUKS) you will also need to
-provide `--boot-mount-spec UUID=...`.
+is to provide `root=<UUID>` and `rootflags` kernel arguments
+to describe to the inital RAM disk how to find and mount the
+root partition. For more on this, see the below section
+discussing mounting the root filesystem.
+
+Note that if a separate `/boot` is needed (e.g. for LUKS) you will also need to provide `--boot-mount-spec UUID=...`.
 
 The `bootc install to-filesystem` command allows an operating
 system or distribution to ship a separate installer that creates more complex block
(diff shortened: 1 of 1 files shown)


docs/src/bootc-install.md
```diff
@@ -222,6 +225,39 @@ a different sandboxing tool (e.g. [bubblewrap](https://github.com/containers/bub
...
+
+Systems using systemd will often default to using
+[systemd-fstab-generator](https://www.freedesktop.org/software/systemd/man/latest/systemd-fstab-generator.html)
+and/or [systemd-gpt-auto-generator](https://www.freedesktop.org/software/systemd/man/latest/systemd-gpt-auto-generator.html#).
+Support for the latter though for the root filesystem is conditional on EFI and a bootloader implementing the bootloader interface.
+
+Outside of the discoverable partition model, a common baseline default for installers is to set `root=UUID=`
+(and optionally `rootflags=`) kernel arguments as machine specific state.
+When using `install to-filesystem`, you should provide these as explicit
+kernel arguments.
```
vrothberg, line 244: Can you give an example?  This stuff is outside my comfort zone, so having some example would increase my understanding.

vrothberg (commented):
Is there a workaround to avoid the remount service from failing?

docs/src/bootc-install.md
cgwalters: This section of the document is targeted at people doing generic installers, disk image image generation tools and custom installs in general.

There's multiple ways to set kernel arguments that are described in the docs already right?

cgwalters:
> Is there a workaround to avoid the remount service from failing?

No, unfortunately.

I commented in https://bugzilla.redhat.com/show_bug.cgi?id=2332319#c1 which should cover the Anaconda case you hit. We can link that as a known bug until Anaconda is updated.

The change that would be by far the easiest/biggest impact would be to change the systemd fstab generator to handle this but I've hesitated to raise it there as I think conceptually it's more that the fstab is wrong.

vrothberg:
> I commented in https://bugzilla.redhat.com/show_bug.cgi?id=2332319#c1 which should cover the Anaconda case you hit. We can link that as a known bug until Anaconda is updated.

Love that!

vrothberg (approved):

PR #973, opened 2024-12-13
docs/images: Document composefs here too

cgwalters:
This needs to be called out way more strongly than it is now. For example a lot of our assumptions about being able to put content anywhere (i.e. outside of `/usr` which happens especially with containers) which rely on it.


jmarrero (approved):
lgtm

PR #976, opened 2024-12-16
lib: tidy up module declarations

jeckersb:
This was bothering me as I was doing something else

Organize these in some sane manner:

- Non-feature-gated modules, sorted
- Then feature-gated modules, grouped and sorted by feature name


jeckersb:
I kinda wish we didn't have to tack the `#[cfg(feature = "install")]` bit over and over again, but afaict you'd have to pull in something like [cfg_if](https://docs.rs/cfg-if/latest/cfg_if/) and that's overkill for such a minor annoyance.

PR #978, opened 2024-12-17
build-sys: Drop custom version glue code

cgwalters:
It was annoying me that rust-analyzer was showing an error for our `version.rs`. Something we're doing there is wrong, maybe related to the output directory?

But when I dig in a bit I don't understand how we ended up with this custom code when the clap docs show use of the `clap::crate_version` macro which looks just fine to me.

Not having a `build.rs` just simplifies things.


jeckersb (approved):
:scissors:

PR #981, opened 2024-12-18
cli: Centralize progress parsing a bit

cgwalters:
Add a shared structure (so we only document the option once) and add a `TryFrom` chain which cleans up the conversion from the options into a `ProgressWriter`.


jmarrero (approved):
lgtm

PR #982, opened 2024-12-18
WIP: Add progress for install

cgwalters:
It turns out to be *really* hard to make this work with my default dev setup of podman-machine...FIFOs don't work over virtiofs for unsurprising reasons, and podman doesn't do fd passing with podman-remote.

I think the best fix is to add pipe fd passing with podman remote and have it act as a copying proxy.

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -22,6 +22,7 @@ use ostree_ext::keyfileext::KeyFileExt;
 use ostree_ext::ostree;
 use schemars::schema_for;
 use serde::{Deserialize, Serialize};
+use tokio::net::unix::pipe::Sender;
 
 use crate::deploy::RequiredHostSpec;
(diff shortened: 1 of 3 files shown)


cgwalters:
> It turns out to be really hard to make this work with my default dev setup of podman-machine...FIFOs don't work over virtiofs for unsurprising reasons, and podman doesn't do fd passing with podman-remote.

A giant amount of related discussion in https://github.com/containers/podman-bootc/pull/95
Note that PR was using vsock. It might make sense for us to directly support vsock, but eh I think we can just make it the responsibility of the caller to bridge between local and hypervisor if needed, I shouldn't have tried to do it in this PR. We can just accept a fd and that's it.

cgwalters:
CLosing in favor of https://github.com/bootc-dev/bootc/pull/1431

Issue #975, opened 2024-12-16
non-UTF8 filename support


cgwalters:
Since its creation ostree (current bootc backend) intentionally didn't support non-UTF8 filenames. In theory, we could lift that restriction as we look towards a direct composefs backend.

I personally think it's almost always going to be a mistake to ship non-UTF8 filenames in containers, but just filing this so we have a tracker.

Also, it'd probably be good to add this to `container lint`.

cgwalters:
Yeah, the UTF-8 restriction was an intentional choice many years ago for ostree. Another two example opinionated choices:

- Not having modification times at all
- No support for device files

The mtime thing is still a huge overall issue...and devices are basically always a bad idea (even the overlayfs whiteout nesting one is [as discussed](https://github.com/containers/bootc/issues/644) better done another way)

But of course since OCI didn't add these restrictions we need to be compatible and yes, composefs will help with that.

It's actually really nice that with composefs specifically the non-UTF8 filenames don't leak onto the *physical* filesystem and only impact the container using them.

PR #983, opened 2024-12-19
lints: add check for non-utf-8 filesystem content

allisonkarlitskaya:
As mentioned in #975, we should add a lint to ensure that all filenames are UTF-8, giving nice errors (with full pathnames) in the event that we encounter invalid filenames.

We also check symlink targets.

Add a unit test that tries various valid and invalid scenarios.

### DIFF
--- a/lib/src/lints.rs
+++ b/lib/src/lints.rs
@@ -4,18 +4,26 @@
 
 use std::env::consts::ARCH;
 
-use anyhow::Result;
+use anyhow::{bail, ensure, Result};
 use cap_std::fs::Dir;
 use cap_std_ext::cap_std;
 use cap_std_ext::dirext::CapStdExtDirExt as _;
 use fn_error_context::context;
+
+use crate::utils::openat2_with_retry;
 
 /// check for the existence of the /var/run directory
 /// if it exists we need to check that it links to /run if not error
 /// if it does not exist error.
 #[context("Linting")]
 pub(crate) fn lint(root: &Dir) -> Result<()> {
-    let lints = [check_var_run, check_kernel, check_parse_kargs, check_usretc];
+    let lints = [
+        check_var_run,
+        check_kernel,
+        check_parse_kargs,
+        check_usretc,
+        check_utf8,
+    ];
     for lint in lints {
         lint(&root)?;
     }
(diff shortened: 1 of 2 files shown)


allisonkarlitskaya:
From what I can see of how we get our root `Dir` reference in `cli.rs`, and the recommended instructions for running this in a `RUN` command in the `Containerfile`, I think this might end up walking into `/dev`, `/sys`, `/proc` and so on as well...

On the other hand, any bind mounts that are present might prevent us from scanning the underlying content.  Imagine some scenario where `/etc/resolv.conf` in the image was a symlink that contained non-utf8 character, but we wouldn't be able to access that because it gets bind-mounted-over for the duration of the `RUN` command.

Maybe we should do a temporary bind mount of (only) the rootfs for purposes of linting?

Thoughts?

cgwalters:
> From what I can see of how we get our root Dir reference in cli.rs, and the recommended instructions for running this in a RUN command in the Containerfile, I think this might end up walking into /dev, /sys, /proc and so on as well...

Yeah we should avoid that for sure. There's relevant similar code in e.g. `remove_all_on_mount_recurse`

allisonkarlitskaya:
I don't really trust checking dev vs. rootdev anymore: we've had some reports of that that show that overlayfs with composefs can produce strange effects in this case... like in https://github.com/containers/composefs-rs/issues/41#issuecomment-2514883271

The overlayfs docs themselves warn about this:
> While directories will report an st_dev from the overlay-filesystem, non-directory objects may report an st_dev from the lower filesystem or upper filesystem that is providing the object. Similarly st_ino will only be unique when combined with st_dev, and both of these can change over the lifetime of a non-directory object. Many applications and tools ignore these values and will not be affected.

https://docs.kernel.org/filesystems/overlayfs.html

The openat2 `NO_XDEV` flag approach seems kinda okay-ish I guess... does cap-std have support for this?

I think the best thing we could do here is some trickery with cloning the root mount to an fd using the new mount API and then iterating through that fd...

cgwalters:
> I don't really trust checking dev vs. rootdev anymore: we've had some reports of that that show that overlayfs with composefs can produce strange effects in this case... like in https://github.com/containers/composefs-rs/issues/41#issuecomment-2514883271

Yeah, fair.

OK so another approach which is also used in the referenced code is this: https://github.com/containers/bootc/pull/984


> The openat2 NO_XDEV flag approach seems kinda okay-ish I guess... does cap-std have support for this?

Not directly but we can also just use rustix for stuff like this too.

cgwalters:
> Okay, "some trickery" isn't very tricky at all. It mostly looks like just using open_tree() without the AT_RECURSIVE flag. rustix binds this, so I'll take a look.

That is awesome! I didn't know that was usable unprivileged. Indeed it's by far the best approach. I noticed this wasn't doc'd so :arrow_right: https://github.com/brauner/man-pages-md/pull/15

lib/src/cli.rs
```diff
@@ -991,7 +992,9 @@ impl Opt {
 
 /// Internal (non-generic/monomorphized) primary CLI entrypoint
 async fn run_from_opt(opt: Opt) -> Result<()> {
-    let root = &Dir::open_ambient_dir("/", cap_std::ambient_authority())?;
+    // We want to open *only* the root filesystem of the container image
+    // ie: without /sys, /proc, /etc/resolve.conf bind mounts, etc.
```
cgwalters, line 996: This is a big change. I think in most cases it's right...OK actually this patch is just changing one of several copies of this we have and it's *definitely* safe and correct for the linter. But...

lib/src/cli.rs
```diff
@@ -1071,7 +1074,7 @@ async fn run_from_opt(opt: Opt) -> Result<()> {
                 late_dir: _,
             } => {
                 let unit_dir = &Dir::open_ambient_dir(normal_dir, cap_std::ambient_authority())?;
-                crate::generator::generator(root, unit_dir)
+                crate::generator::generator(&root, unit_dir)
```
cgwalters, line 1077: I think this will actually break things on the booted system because `etc` is a bind mount right?

allisonkarlitskaya: Ya.  Good point.  I wasn't thinking about the booted system.

I might de-factor this variable then.  It only has 2 users (if you count the two "internals" commands as one), and they need it for different things...

cgwalters (approved):
Nice thanks!

cgwalters:
Ah CI needs some `#[cfg(feature = "install")]` massaging...I am quite tempted again to drop that feature or just scope it to `to-disk`

lib/src/mount.rs
```diff
@@ -146,6 +146,23 @@ pub(crate) fn is_same_as_host(path: &Utf8Path) -> Result<bool> {
     Ok(devstat.f_fsid == hostdevstat.f_fsid)
 }
 
+/// Open the named path as a new mount attached to an fd.
+#[context("Opening mount tree from pid")]
```
cgwalters, line 150: This isn't "from pid" anymore

lib/src/mount.rs
```diff
@@ -146,6 +146,23 @@ pub(crate) fn is_same_as_host(path: &Utf8Path) -> Result<bool> {
...
+    // Open the target mount path as a file descriptor.
+    let recursive = if recursive {
+        OpenTreeFlags::AT_RECURSIVE
+    } else {
+        OpenTreeFlags::empty()
+    };
+    rustix::mount::open_tree(
+        rustix::fs::CWD,
+        path.as_std_path(),
+        OpenTreeFlags::OPEN_TREE_CLOEXEC | OpenTreeFlags::OPEN_TREE_CLONE | recursive,
```
cgwalters, line 161: Also per https://github.com/brauner/man-pages-md/pull/15 we can't actually use `OPEN_TREE_CLONE` for the linting case because the default for docker/podman is not to be in a user namespace with `CAP_SYS_ADMIN`.

(Eventually of course it'd make huge amounts of sense to try to drive a userns to be enabled by default for builds and runs but it has a lot of potential fallout)

allisonkarlitskaya: I think we could do this if we created our own user and mount namespace.  That works for `bwrap`, even as a non-root user, and even running inside of a container itself.  It's starting to get a bit "heavy", but as a general feature to have inside of bootc, it would be quite useful.  I already imagine some uses for it in composefs-rs as well, for the "unified" example...

Of course, we already have the code in `mount.rs` for entering another mount namespace and sending a mount fd out of it back to ourselves... this would probably end up looking quite a lot like that, but instead of entering a new namespace, we create our own...

cgwalters:
After a live chat we decided to just rely on `openat2(.. NO_XDEV)`, updated for that.

cgwalters:
Also noting here just so it doesn't get lost:

We also discussed having `container lint --root /path/to/root` which would be usable in a multi-stage build like this:

```
FROM someimage as rootfs
...


FROM otherimage as linter
RUN --mount=from=rootfs,target=/rootfs bootc container lint --rootfs=/rootfs
```

An advantage of this approach is that we'd be able to scan the filesystem tree without having container runtime state (`/proc`, `/dev`) overmounted at all...which would actually allow us to e.g. also lint [against device nodes](https://github.com/ostreedev/ostree/issues/2568).

But...the ergonomics would be obviously less good.

And ultimately the most powerful way to do this is to have a container scanner that runs outside of the running image, which would allow us to lint the actual image structure itself (e.g. tiny layers that should be coalesced). Or of course, a container -> container optimizer.

PR #984, opened 2024-12-19
tree-wide: Use cap-std-ext is_mountpoint() API

cgwalters:
I moved it there a while ago, now we can drop the copy of it
here.

---


cgwalters:
I changed this one, it's now way simpler.

PR #985, opened 2024-12-19
docs: remove duplicate manpage for bootc-fetch-apply-updates.service

jeckersb:


jeckersb:
Hm, looks like the labeler didn't add `documentation` here.  It's running the full suite and actually skipped the docs job :(

jeckersb:
> Hm, looks like the labeler didn't add `documentation` here. It's running the full suite and actually skipped the docs job :(

I think #986 fixes this

cgwalters (approved):

PR #986, opened 2024-12-19
ci: labeler: use globstar where applicable

jeckersb:
It appears the single glob '*' doesn't match recursively, so use the
globstar '**' where we want to match anything within a subtree


jeckersb:
[relevant labeler docs](https://github.com/actions/labeler?tab=readme-ov-file#basic-examples)

PR #987, opened 2024-12-19
`install` feature is always on, add `install-to-disk`

cgwalters:
I consider `bootc install to-filesystem` support a key feature of bootc. In theory today one can still set up a system directly with `ostree` and we will continue to support that.

But things like logically bound images we do want to be initialized from the start and that's with `bootc install to-filesystem`.

Maintaining the feature just has a logistical annoyance any time one touches the install code as we often end up needing to carefully `#[cfg(feature = "install")]` in many places in an infectious way.

Also as we head towards enabling factory reset
https://github.com/containers/bootc/issues/404
we really do want some of the install code enabled there.

However, `to-disk` is much more of a "demo". I don't want bootc to grow too much knowledge around block devices. Complex setups (LVM, LUKS) etc. are the domain of external installers and provisioning tools.

So the feature gate is now on that (which is still on by default).

We ended up with more `#[cfg(feature = "install-to-disk")]` than I'd have liked, but some of that can be fixed subsequently.

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -53,9 +53,11 @@ similar-asserts = { workspace = true }
 static_assertions = { workspace = true }
 
 [features]
-default = ["install"]
-# This feature enables `bootc install`.  Disable if you always want to use an external installer.
-install = []
+default = ["install-to-disk"]
+# This feature enables `bootc install to-disk`, which is considered just a "demo"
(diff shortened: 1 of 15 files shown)


jmarrero (approved):
lgtm

cgwalters:
Rebased :surfer: for trivial conflict with https://github.com/containers/bootc/pull/990

PR #988, opened 2024-12-20
install: Add /usr/share/doc/bootc/baseimage

cgwalters:
This directory will contain expected files in the base image.

That said, I may change the container import path to auto-create
at least the sysroot dir and symlink at some point and these
can just be dropped.

And for that matter after
https://github.com/ostreedev/ostree/commit/9a0acd7249bb0c7f55c2bf56e5073902cd60038b
"libostree/deploy: enable composefs by default"
we can likely just drop the prepare-root bit too.

But for now this is needed.

Motivated by improving base image generation from
https://gitlab.com/fedora/bootc/tracker/-/issues/32

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,9 @@ install:
 	  fi; \
 	  done
 	install -D -m 0644 -t $(DESTDIR)/$(prefix)/lib/systemd/system systemd/*.service systemd/*.timer systemd/*.path systemd/*.target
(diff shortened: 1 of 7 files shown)


cgwalters:
Oh duh of course, can't ship empty directories in git...annoying.

jeckersb (approved):
:+1:

PR #1001, opened 2025-01-02
 install: Never traverse mount points with `--wipe`

cgwalters:
utils: Move open_dir_noxdev here

Prep for using it elsewhere.

---

install: Never traverse mount points with `--wipe`

If we encounter a mount point when attempting to
wipe a filesystem, then something has definitely gone
wrong. At the install phase we should only be operating
on a single physical filesystem.

Motivated by https://github.com/containers/bootc/issues/997

---

### DIFF
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -28,6 +28,7 @@ use camino::Utf8Path;
 use camino::Utf8PathBuf;
 use cap_std::fs::{Dir, MetadataExt};
 use cap_std_ext::cap_std;
+use cap_std_ext::cap_std::fs::FileType;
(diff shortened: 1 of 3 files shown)


cgwalters:
Ah yep, CI fails in the `--replace=alongside` case now:

> ERROR Installing to filesystem: Removing boot directory content: Removing entries (noxdev): Found unexpected mount point "efi"

We do need to specially handle this case...arguably we should support not wiping the ESP at all, but anyways we do want to traverse this mount for now. Or perhaps, special case finding the ESP, wiping it, and then not traversing the mount.

jmarrero (approved):
lgtm

PR #1002, opened 2025-01-03
install: Add `dracut` subdirectory to baseimage reference

cgwalters:
ref https://gitlab.com/fedora/bootc/tracker/-/issues/32#note_2270189835

This will keep track of the generic bootc+dracut integration glue here instead of replicating it in different base images.

(Perhaps bootc actually injects code into the initramfs in the
 future instead of ostree, for example)


jmarrero (approved):
lgtm

PR #1003, opened 2025-01-03
tests: Handle not having skopeo present

cgwalters:
Fallout from the ostree-ext merge. We were missing a `BuildRequires`
in the spec, which caused our unit tests to fail in Fedora.
I since added that BuildRequires, but let's change our tests
to skip if it's not present.


cgwalters:
Also tangential but https://github.com/rpm-software-management/rpm/discussions/3504

jmarrero (approved):
lgtm

PR #1009, opened 2025-01-06
Two changes for `bootc-status-updated.path`

cgwalters:
build-sys: Enable `bootc-status-updated.path` by default

Followup to https://github.com/containers/bootc/pull/977

In chatting with John we had some agreement to statically
enable the `.path` unit by default upstream here since
it's a feature we want to be available by default
so that tooling can react dynamically to changes.

---

bootc-status-updated: Add `ConditionPathExists=/run/ostree-booted`

This is a general best practice for our units to ensure
they don't apply outside of bootc systems.

---


jmarrero (approved):
lgtm

jeckersb (approved):
:partying_face:

PR #1012, opened 2025-01-07
docs/relationships: add section about ostree

jlebon:
I think this info can probably be pieced together from the rest of the docs, but let's add a dedicated section for it. Came up in an internal chat.


cgwalters (approved):

cgwalters:
Ah, the docs are failing, but not because of this PR, will fix independently.

PR #1018, opened 2025-01-09
baseimage: Add kernel-install integration

jmarrero:
This is part of the kernel-integration work:
- https://github.com/coreos/rpm-ostree/pull/5135
- https://gitlab.com/fedora/bootc/base-images/-/merge_requests/62
- https://github.com/coreos/rpm-ostree/pull/5209

WIP while we release rpm-ostree.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,10 @@ install:
 	cp -PfT baseimage/base/ostree $(DESTDIR)/$(prefix)/share/doc/bootc/baseimage/base/ostree 
 	# Ensure we've cleaned out any possibly older files
 	rm -vrf $(DESTDIR)$(prefix)/share/doc/bootc/baseimage/dracut
+	rm -vrf $(DESTDIR)$(prefix)/share/doc/bootc/baseimage/systemd
+	# Copy dracut and systemd config files
 	cp -Prf baseimage/dracut $(DESTDIR)$(prefix)/share/doc/bootc/baseimage/dracut
+	cp -Prf baseimage/systemd $(DESTDIR)$(prefix)/share/doc/bootc/baseimage/systemd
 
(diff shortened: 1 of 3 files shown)


jmarrero:
I will add the journal one in a separate PR with its corresponding MR for the base image removing it there. Just to keep things separate.

cgwalters:
Looks sane to me, I think we just need an update to the `README.md`

jmarrero:
ok, I added the changes to the Makefile too following what we do for Dracut, but the question is, if that should be behind some flag since it's technically optional? Or maybe I misunderstood about needing to add this to the Makefile.

cgwalters:
Note that everything in the baseimage directory is installed into `/usr/share/doc/bootc` - so it's just *informative*. It's up to base image builders to copy it in explicitly (and we aren't yet doing that in fedora-bootc at all yet).

So this can't break anything today. The changes LGTM as is!

jmarrero:
Ok, cool, I re-read the Makefile as you were typing :D and noticed the doc, but that makes sense now I will leave this PR alone then.

cgwalters (approved):

jmarrero:
This breaks kernel-install somehow... I would have expected that the dropin would not get triggered. 

```
dracut[I]: *** Creating image file '/tmp/.tmpU2x9Mg/initramfs.img' ***
dracut[I]: *** Creating initramfs image file '/tmp/.tmpU2x9Mg/initramfs.img' done ***
+ test -f /usr/lib/modules/6.11.4-301.fc41.x86_64/initramfs.img
Error: Process completed with exit code 1.
```
I need to dig.

jmarrero:
Since this is informational at the moment, we should be good to merge this here.

PR #1025, opened 2025-01-14
ostree-ext/tar/export: Fix reexport of var/tmp

cgwalters:
A lot of technical debt here. A long time ago I added this hacky bit to inject var/tmp is the container stream even if it wasn't in the ostree commit.

Today things shipped by `rpm-ostree compose image` like FCOS don't have `var/tmp` in the commit.

But then more recently we started shipping `/var/tmp` in base images directly.

Now I'm working on https://github.com/coreos/rpm-ostree/issues/5221 where we're rechunking from a rootfs that does have var/tmp and that ends up in the ostree commit.

The path comparison here was wrong because the tar stream we generate has the paths start with `./` and a literal comparison doesn't match `./var/tmp` != `var/tmp`.

Add a canonicalization helper and use it for this.

### DIFF
--- a/ostree-ext/src/tar/export.rs
+++ b/ostree-ext/src/tar/export.rs
@@ -62,6 +62,22 @@ fn map_path_v1(p: &Utf8Path) -> &Utf8Path {
     }
 }
 
+/// Given two paths, which may be absolute (starting with /) or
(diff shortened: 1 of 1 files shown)


jmarrero (approved):
lgtm

PR #1026, opened 2025-01-14
contrib/packaging/bootc.spec: sync with current Fedora rawhide

jmarrero:


jmarrero:
mmm I think the rpm-build still can't deal with auto release let's see how the build goes.

jmarrero:
seems stream 9/rhel9 needs us to be conservative here.

jmarrero:
The rpms builds CI need more masa

cgwalters:
In c10s:

```
cargo_build: invalid option -- '-'
error: /builddir/build/originals/bootc.spec: line 71: Unknown option - in cargo_build(naf:)
  1<    (%)
  0<  (%cargo_build)
```

Offhand...I think the rust-toolset in rhel10 is too old? I think it supports `-f` but not `--features`?

cgwalters:
Ok and c9s died on

```
thread 'lints::tests::test_baseimage_root' panicked at lib/src/lints.rs:300:64:
called `Option::unwrap()` on a `None` value
```

Hmm...ummm, I guess because something in the RPM build path is overriding the cargo stuff and we're missing `CARGO_MANIFEST_PATH` ?

What a mess

cgwalters:
Can you merge in 
```
From 8388662ce6086e40c63e8774d2568402a3055660 Mon Sep 17 00:00:00 2001
Date: Tue, 14 Jan 2025 16:58:53 -0500
Subject: [PATCH] tests: Ignore baseimage root test if CARGO_MANIFEST_PATH is
 not set

This is only in relatively new Rust (not in RHEL 9.5).

---
 lib/src/lints.rs | 5 ++++-
 1 file changed, 4 insertions(+), 1 deletion(-)

diff --git a/lib/src/lints.rs b/lib/src/lints.rs
index 9c0942cb..d382e8e4 100644
--- a/lib/src/lints.rs
+++ b/lib/src/lints.rs
@@ -297,7 +297,10 @@ mod tests {
         assert!(check_baseimage_root(&td).is_err());
 
         // Copy our reference base image content from the source dir
-        let manifest = std::env::var_os("CARGO_MANIFEST_PATH").unwrap();
+        let Some(manifest) = std::env::var_os("CARGO_MANIFEST_PATH") else {
+            // This was only added in relatively recent cargo
+            return Ok(());
+        };
         let srcdir = Path::new(&manifest)
             .parent()
             .unwrap()
-- 
2.47.0

```

I think we *don't* need to backport that downstream because we don't need to care about the unit tests on older rust (in rhel).

PR #1027, opened 2025-01-14
container/lint: Add --rootfs

cgwalters:
Intended to pair with things like https://github.com/coreos/rpm-ostree/pull/5211 where we operate on a non-booted root.


jmarrero (approved):
lgtm

PR #1028, opened 2025-01-14
ostree-ext: surface libostree signature verification text

jlebon:
Right now when using OSTree remote verification, there's no indication
that the OSTree commit was successfully verified.

Capture the return string from `OstreeRepo.signature_verify_commit_data`
and return it as part of the public `LayeredImageState` object.

We could return a more structured signature object here, but this is
sufficient for our purposes for now and saves clients from having to
regenerate similar looking text.

In the `ostree container` CLI, print this verification text.

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -28,7 +28,7 @@ use crate::lints;
 use crate::progress_jsonl::{ProgressWriter, RawProgressFd};
 use crate::spec::Host;
 use crate::spec::ImageReference;
-use crate::utils::sigpolicy_from_opts;
+use crate::utils::sigpolicy_from_opt;
 
 /// Shared progress options
 #[derive(Debug, Parser, PartialEq, Eq)]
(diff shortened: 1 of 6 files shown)


jlebon:
Haven't added any tests to this. Didn't really look at the testing setup in this repo yet.

cgwalters (approved):
Also of note at some point IMO we should do the same for skopeo re surfacing verification status, although I think the maintainer felt that was an anti-pattern I don't agree.

lib/src/cli.rs
```diff
@@ -111,10 +111,6 @@ pub(crate) struct SwitchOpts {
     #[clap(long)]
     pub(crate) enforce_container_sigpolicy: bool,
 
-    /// Enable verification via an ostree remote
-    #[clap(long)]
-    pub(crate) ostree_remote: Option<String>,
```
cgwalters, line 116: Technically we claimed public CLI/API stability but I can't find any users of this on public source code, and I can't imagine there are any private ones.

To be clear: Everyone should use container signing, not ostree commit signatures.

PR #1031, opened 2025-01-15
tests: Quiet skopeo --help

cgwalters:
We were spewing the help text out in tests, quiet that. Regression from previous change.


jeckersb (approved):
LGTM

PR #1032, opened 2025-01-15
ostree-ext: Serialize xattrs into tar stream as well 

cgwalters:
We really want this for https://github.com/coreos/rpm-ostree/pull/5222
to be able to rebuild images from their container-synthesized rootfs.

Really, the only xattr we don't want to emit in to the tar stream
is security.selinux for now.

Eventually we should try to switch to putting that into the tar
stream too, but it needs more validation.


jmarrero (approved):
lgtm

cgwalters:
Interesting. The unit tests passed in GHA, but failed in copr. Need to dig into that...

cgwalters:
No idea what's up with the unit tests in COPR but I added some code to dump the filesystem tree if the checksum doesn't match

cgwalters:
Ah hah! It's just c{9,10}s failing because those have ostree-2024.10 which has https://github.com/ostreedev/ostree/pull/3346 which sorts xattrs. I need to fix our test suite to do that too.

PR #995, opened 2024-12-27
ci: Land a new upstream release in Fedora with Packit

henrywang:
Trigger the build in Fedora Koji build system as a reaction to a new dist-git commit. Create a new update in Fedora Bodhi for successful Koji build.

I can't send PR in https://src.fedoraproject.org/rpms/bootc. According to https://fedoraproject.org/wiki/Infrastructure/fedorapeople.org#Accessing_your_fedorapeople.org_space, my fedora account should be in one group. Anyone can help to add my fedora account (@xiaofwan) into group? Thanks.

PR #1034, opened 2025-01-16
Packit: Add official Fedora downstream jobs

lsm5:
Fixes: #995


lsm5:
AFAIK, the only way to verify this change is to cut a new release with this diff included. Should be safe to merge though.

PR #1035, opened 2025-01-16
deny: Sync with rpm-ostree

cgwalters:
We should really have a more centrally-maintained `cargo-deny` configuration. I'd argue to maintain it here in bootc to start, but this will be a common thing for other projects in github.com/containers and elsewhere.

Anyways, this needed updating for the new Unicode-3.0 license in some updated unicode crates that I saw in rpm-ostree.

While we're here, quiet the duplicate crate warning in the CI job, as it's just noise. Keep it when running locally so we have some visibility if we care about it.


jmarrero (approved):
lgtm

PR #1037, opened 2025-01-16
spec: fix build with rhsm

yselkowitz:
The Go macros for Fedora, RHEL 10, and ELN (the future RHEL 11) are all in sync; only RHEL 9 and earlier need the old syntax.

Downstream PR: https://src.fedoraproject.org/rpms/bootc/pull-request/7

### DIFF
--- a/contrib/packaging/bootc.spec
+++ b/contrib/packaging/bootc.spec
@@ -67,7 +67,7 @@ Provides: ostree-cli(ostree-container)
 %cargo_prep -v vendor
 
 %build
-%if 0%{?rhel} == 10
+%if 0%{?fedora} || 0%{?rhel} >= 10
     %cargo_build %{?with_rhsm:-f rhsm}
 %else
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
Ah...and you're doing this because it broke in ELN right?

cgwalters:
Missing DCO https://github.com/containers/bootc/pull/1037/checks?check_run_id=35728011800

yselkowitz:
> Ah...and you're doing this because it broke in ELN right?

Indeed, but a local build `--with rhsm` would also break on Fedora.

cgwalters:
> Indeed, but a local build --with rhsm would also break on Fedora.

Right, fair but no one is asking for that :smile: 

I think it's important to include even brief info on the real-world use case in commit messages. In this case, I honestly just forgot about ELN and it's helpful as a reminder for everyone else.

PR #1039, opened 2025-01-17
lints: Various improvements

cgwalters:
Internal only change, but prep for future work; I want to track the version lints were introduced in, support clearly differentiating between warnings and fatal errors, etc.

Specifically motivated by custom base image work
https://github.com/coreos/rpm-ostree/issues/5221
but also obviously useful in general.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -196,6 +196,7 @@ dependencies = [
  "similar-asserts",
  "static_assertions",
  "tempfile",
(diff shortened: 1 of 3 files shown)


vrothberg (approved):
LGTM

Smart idea!  Independent of this PR:  Could we improve the docs of the lint command?  It would be great to know which linters are being run and what we're checking for.  Ecosystem engineering is looking into it for best practices etc.

PR #1040, opened 2025-01-17
build.sh: Use our embedded spec

cgwalters:
One CI run just got a server error fetching the spec from Fedora. We have a spec here (which I don't like but we do) so use it to lower CI flakes.


cgwalters:
Looks like generic copr infra flakes on this one

jmarrero (approved):
lgtm

PR #1041, opened 2025-01-17
systemd: enable bootc-status-updated.target on startup

jeckersb:
Any services that react to status changes should be idempotent, and
should run on boot.  It is likely (but not guaranteed) that during
boot we are in the process of switching from a previous deployment to
a new one, and thus what was previously "staged" is now "booted" and
what was previously "booted" is now "rollback", so anything that cares
about status is going to want to handle that.

### DIFF
--- a/systemd/bootc-status-updated.target
+++ b/systemd/bootc-status-updated.target
@@ -2,3 +2,7 @@
 Description=Target for bootc status changes
 Documentation=man:bootc-status-updated.target(8)
 StopWhenUnneeded=true
+ConditionPathExists=/run/ostree-booted
+
+[Install]
+WantedBy=multi-user.target
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,7 @@ install:
 	install -D -m 0644 -t $(DESTDIR)/$(prefix)/lib/systemd/system systemd/*.service systemd/*.timer systemd/*.path systemd/*.target
 	install -d -m 0755 $(DESTDIR)/$(prefix)/lib/systemd/system/multi-user.target.wants
 	ln -s ../bootc-status-updated.path $(DESTDIR)/$(prefix)/lib/systemd/system/multi-user.target.wants/bootc-status-updated.path
+	ln -s ../bootc-status-updated.target $(DESTDIR)/$(prefix)/lib/systemd/system/multi-user.target.wants/bootc-status-updated.target
 	install -D -m 0644 -t $(DESTDIR)/$(prefix)/share/doc/bootc/baseimage/base/usr/lib/ostree/ baseimage/base/usr/lib/ostree/prepare-root.conf
 	install -d -m 755 $(DESTDIR)/$(prefix)/share/doc/bootc/baseimage/base/sysroot
 	cp -PfT baseimage/base/ostree $(DESTDIR)/$(prefix)/share/doc/bootc/baseimage/base/ostree 


cgwalters:
In the general case, don't we actually want to trigger `bootc-status-updated.target` as part of `multi-user.target` by default?

jeckersb:
> In the general case, don't we actually want to trigger `bootc-status-updated.target` as part of `multi-user.target` by default?

Hmm yeah probably.  In a perfect world we don't want to trigger it if nothing has changed (you rebooted the system without any sort of update/switch/rollback queued), but I think that's better than it *not* triggering when the status *did* change.

cgwalters:
This heavily relates to a discussion elsewhere for a metadata flag of "is this the first boot of a deployment"? If we had that, then we could key off it in our systemd generator and only enqueue the change notification if we found that.

ref https://github.com/ostreedev/ostree/issues/3032

But for now I think it's actually OK to run on boot by default - most tools that want to hook into this should be idempotent anyways, i.e. they need to handle the case where nothing changed. Not that anything will run by default except for the subman case today anyways.

jeckersb:
> Marking requested-changes per [#1041 (comment)](https://github.com/containers/bootc/pull/1041#issuecomment-2598757130)

Updated to static enable the `.target` alongside the `.path` so both will be enabled and activated at boot.  Then [the centos-stream MR for rhsm](https://gitlab.com/redhat/centos-stream/containers/bootc/-/merge_requests/361) should just work, since the target will trigger the service once it's enabled by the preset.  Unfortunately this means we'll need another bootc release to land in c10s/c9s before this all works (unless I hack around it temporarily in the base image build, which I would rather not do).

systemd/bootc-status-updated.target
```diff
@@ -2,3 +2,6 @@
 Description=Target for bootc status changes
 Documentation=man:bootc-status-updated.target(8)
 StopWhenUnneeded=true
```
cgwalters, line 4: Since we're now enabling this default; I think we should probably also add here `ConditionPathExists=/run/ostree-booted`? In practice it's probably effectively the same, but the difference is that everyone who isn't using bootc will I think see "skipped as condition not met" vs "reached target bootc-status-updated" and I think the former makes more sense.

Alternatively: We could just dynamically enable this from our generator in that condition.

cgwalters (approved):

systemd/bootc-status-updated.target
jeckersb: :+1: good catch, updated

cgwalters (approved):

cgwalters:
@jeckersb can you comment here? It sounds like this flips the semantics, so that we run on firstboot, but we stop rerunning when the path changed?

jeckersb:
> "If this option is set, a unit will be automatically cleaned up if no other active unit requires it. Defaults to false."
> 
> And `multi-user.target` stays active. I suspect we could fix this by having an intermediate stub `bootc-first-boot.service` or so that pulls this in, and making _that_ be `WantedBy=multi-user.target`?

Yeah something like that.  I will play around with it and see if I can conjure up a solution that works without making things worse yet again.  Maybe involving `BindsTo` or `PartOf`?  They're all related but subtly different.

jeckersb:
https://github.com/containers/bootc/pull/1052

PR #1045, opened 2025-01-17
ostree-ext/encapsulate: Add some error contexts

cgwalters:
I was getting EPERM on creating the dir, and it was a pain to track down.


jmarrero (approved):
lgtm

PR #1048, opened 2025-01-20
install/blockdev: Break cyclic build dependency

cgwalters:
The install code and the blockdev code call each other. Clean up blockdev to only use `bootc-utils`.
Prep for splitting out the blockdev stuff into at least a separate internal crate; ref
https://github.com/coreos/bootupd/pull/820

### DIFF
--- a/lib/src/blockdev.rs
+++ b/lib/src/blockdev.rs
@@ -14,9 +14,6 @@ use fn_error_context::context;
 use regex::Regex;
 use serde::Deserialize;
 
-#[cfg(feature = "install-to-disk")]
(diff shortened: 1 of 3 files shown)


HuijingHei (approved):
Overall LGTM, just one minor question, is it necessary to remove all `#[cfg(feature = "install-to-disk")]` in `blockdev.rs` ?

cgwalters:
> Overall LGTM, just one minor question, is it necessary to remove all #[cfg(feature = "install-to-disk")] in blockdev.rs ?

It will be in the future when splitting this into a separate crate yes. At the moment those suppress some unused variable warnings.

PR #1050, opened 2025-01-22
Split off a blockdev internal crate

cgwalters:
This code has been forked between bootupd and coreos-installer and here. This is prep for having bootupd pull it from bootc's git so we can deduplicate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -151,6 +151,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "blockdev"
+version = "0.0.0"
+dependencies = [
+ "anyhow",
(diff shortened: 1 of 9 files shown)


HuijingHei (approved):
LGTM

HuijingHei:
One small question, should we use `bootc-blockdev` like https://github.com/coreos/bootupd/pull/821 ?

cgwalters:
> One small question, should we use bootc-blockdev like https://github.com/coreos/bootupd/pull/821 ?

We do when importing it elsewhere, but yeah it would likely make sense to rename all of the internal crates to have a `bootc-` prefix.

PR #1051, opened 2025-01-22
lints: Add warning level, add a check for /var/log

cgwalters:
This will help catch e.g. `/var/log/dnf.log` and such; no one really wants to leak those into containers by default.

Add a `Warning` level lint - this one shouldn't be fatal by default. But also add the ability to opt in to making warnings fatal.


jmarrero (approved):
lgtm

PR #1052, opened 2025-01-22
Add bootc-status-updated-onboot.target 

jeckersb:
The bootc-status-updated.target unit cannot directly be
WantedBy=multi-user.target, because we rely on StopWhenUnneeded=true
so the target will be activated each time bootc-status-updated.path
triggers it.  Because multi-user.target remains active, this would
hold bootc-status-updated.target in an active state as well, and the
path unit activating would not have the desired effect of retriggering
the target unit.

The new bootc-status-updated-onboot.target is intended to serve the
same purpose as bootc-status-updated.target but will activate (and
remain active) on boot.  Any interested units should be WantedBy both
targets.  This updates bootc-publish-rhsm-facts.service to do just
that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ install:
 	install -D -m 0644 -t $(DESTDIR)/$(prefix)/lib/systemd/system systemd/*.service systemd/*.timer systemd/*.path systemd/*.target
 	install -d -m 0755 $(DESTDIR)/$(prefix)/lib/systemd/system/multi-user.target.wants
 	ln -s ../bootc-status-updated.path $(DESTDIR)/$(prefix)/lib/systemd/system/multi-user.target.wants/bootc-status-updated.path
-	ln -s ../bootc-status-updated.target $(DESTDIR)/$(prefix)/lib/systemd/system/multi-user.target.wants/bootc-status-updated.target
+	ln -s ../bootc-status-updated-onboot.target $(DESTDIR)/$(prefix)/lib/systemd/system/multi-user.target.wants/bootc-status-updated-onboot.target
 	install -D -m 0644 -t $(DESTDIR)/$(prefix)/share/doc/bootc/baseimage/base/usr/lib/ostree/ baseimage/base/usr/lib/ostree/prepare-root.conf
 	install -d -m 755 $(DESTDIR)/$(prefix)/share/doc/bootc/baseimage/base/sysroot
 	cp -PfT baseimage/base/ostree $(DESTDIR)/$(prefix)/share/doc/bootc/baseimage/base/ostree 
--- a/systemd/bootc-publish-rhsm-facts.service
+++ b/systemd/bootc-publish-rhsm-facts.service
@@ -9,3 +9,4 @@ ExecStart=/usr/bin/bootc internals publish-rhsm-facts
 
 [Install]
 WantedBy=bootc-status-updated.target
+WantedBy=bootc-status-updated-onboot.target
--- a/systemd/bootc-status-updated-onboot.target
+++ b/systemd/bootc-status-updated-onboot.target
@@ -1,5 +1,7 @@
 [Unit]
-Description=Target for bootc status changes
+Description=Target for bootc status changes on boot
 Documentation=man:bootc-status-updated.target(8)
-StopWhenUnneeded=true
 ConditionPathExists=/run/ostree-booted
+
+[Install]
+WantedBy=multi-user.target


Makefile
```diff
@@ -26,8 +26,10 @@ install:
 	  done
 	install -D -m 0644 -t $(DESTDIR)/$(prefix)/lib/systemd/system systemd/*.service systemd/*.timer systemd/*.path systemd/*.target
 	install -d -m 0755 $(DESTDIR)/$(prefix)/lib/systemd/system/multi-user.target.wants
+	install -d -m 0755 $(DESTDIR)/$(prefix)/lib/systemd/system/bootc-multi-user.service.wants
 	ln -s ../bootc-status-updated.path $(DESTDIR)/$(prefix)/lib/systemd/system/multi-user.target.wants/bootc-status-updated.path
-	ln -s ../bootc-status-updated.target $(DESTDIR)/$(prefix)/lib/systemd/system/multi-user.target.wants/bootc-status-updated.target
+	ln -s ../bootc-status-updated.target $(DESTDIR)/$(prefix)/lib/systemd/system/bootc-multi-user.service.wants/bootc-status-updated.target
```
cgwalters, line 31: This can just be a `Wants=bootc-status-updated.target` directly in the `.service` file right?

systemd/bootc-multi-user.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Dummy bootc service that activates during system boot
+Documentation=man:bootc(8)
+
+[Service]
+Type=oneshot
+ExecStart=/bin/true
```
cgwalters, line 7: Not to make you retest all of this again, but I suspect this could be actually be a `.target` unit and not a `.service` right? The role of `.target` units is precisely for  synthetic ordering.

If we *were* to make this a service (or even a target)...I would bikeshed that we shouldn't call it `-multi-user` since that has an implication we have something to do with user logins.

How about `bootc-status-updated-onboot.target`?

cgwalters: (If you have tested this PR as is and it all works, I am ok to just ship this and do further cleanup as a followup, to be clear)

jeckersb: Yeah I like this better.  I was trying to avoid making the "dependent" services (rhsm and potential future others) be WantedBy multiple things, but the hacky dummy service it resulted in is not great.

Makefile
jeckersb: re: below, I'll just drop the `.service` entirely, and then we'll do the static enable of `bootc-status-updated-onboot.target` -> `multi-user.target`.

cgwalters:
Only tangentially related, this is sure a bit complex but...it's I think overall better than say `/etc/NetworkManager/dispatcher.d` type things that end up being messier; `NetworkManager-dispatcher.service` ends up needing high privileges because we have no idea what the code will do. Also from that service is this "eww" bit:

```
# We want to allow scripts to spawn long-running daemons, so tell
# systemd to not clean up when nm-dispatcher exits
KillMode=process
```

This approach is cleaner than that apparently straightforward "just run these binaries in a directory in order", albeit at the cost of more systemd gymnastics.

jeckersb:
> Only tangentially related, this is sure a bit complex but...it's I think overall better than say `/etc/NetworkManager/dispatcher.d` type things that end up being messier; `NetworkManager-dispatcher.service` ends up needing high privileges because we have no idea what the code will do. Also from that service is this "eww" bit:
> 
> ```
> # We want to allow scripts to spawn long-running daemons, so tell
> # systemd to not clean up when nm-dispatcher exits
> KillMode=process
> ```
> 
> This approach is cleaner than that apparently straightforward "just run these binaries in a directory in order", albeit at the cost of more systemd gymnastics.

Yeah the latest version I just pushed I think is actually pretty clean.  Took a lot of painful iterating to get there... but I am happy with this finally and it seems to work.  I went through and tested by cherry-picking to c10s bootc, mockbuilding, generating a repo, overriding the base image build to include said repo, and then doing a `podman-bootc run`.  Then testing within that doing a `bootc switch` and rebooting.  At each step of the way, checked `subscription-manager facts | grep bootc` and everything looked sane:

- At initial boot
- After `bootc switch` but before rebooting
- After rebooting

So hopefully no more surprises and I can finally be done with this.

cgwalters (approved):

systemd/bootc-publish-rhsm-facts.service
```diff
@@ -9,3 +9,4 @@ ExecStart=/usr/bin/bootc internals publish-rhsm-facts
 
 [Install]
 WantedBy=bootc-status-updated.target
+WantedBy=bootc-status-updated-onboot.target
```
cgwalters, line 12: (Sorry again nonblocking but...I thought we'd make `bootc-status-onboot.target` have `Wants=bootc-status-updated.target`, then each individual unit is again just triggered solely by `bootc-status-updated.target)

jeckersb: I think it has to be this way, because otherwise we'd have the same problem as before, except transitive.  `multi-user.target` would be active, so `bootc-status-updated-onboot.target` would also remain active, and in turn `bootc-status-updated.target` would remain active, so the path unit wouldn't retrigger things.

cgwalters: (Queuing this PR to merge as is, so let's not update it further; this discussion is about followups)

What I was thinking was:

- `multi-user.target` -> `bootc-status-updated-onboot.target` -> `bootc-status-updated.target`
- `bootc-status-updated.path` -> `bootc-status-updated.target`

Now again the problem as is is:

> "If this option is set, a unit will be automatically cleaned up if no other active unit requires it. Defaults to false."

But isn't the fix there to also add `StopWhenUnneeded=true` to `bootc-status-updated-onboot.target`?

jeckersb: No, because it's always "needed" by multi-user.target

jeckersb: Really I think the RFE is probably for systemd to distinguish between "needed" for things that are soft (`Wants`) vs hard (`Requires`).  The docs specifically say:

```
If this option is set, a unit will be automatically cleaned up if no other active unit requires it.
```

But "requires" there is not `Requires`.

PR #1057, opened 2025-01-24
build-sys: Two minor patches

cgwalters:
build-sys: Split workspace members to multiple lines

To lessen the chance of future conflicts when changing things here.

---

Rename internal blockdev crate to bootc-blockdev

To make a bit clearer this is an internal-to-bootc thing; but also
because other projects like bootupd may start referencing it.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -151,7 +151,20 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


jmarrero (approved):
lgtm

PR #1058, opened 2025-01-24
lints: Also test main execution

cgwalters:
We unit test each lint, but not the main function. As it grows in complexity it'll be useful to unit test.

- Change the function to write to a provided output, not always stdout/stderr
- Add a `passing_fixture` helper that passes all lints
- Fix the baseimage test not to use `??` because that turns the *inner* error into an outer error which is not what we want.


jeckersb (approved):
LGTM

PR #1059, opened 2025-01-24
ostree-ext: Add bootc feature

cgwalters:
This is aiming to fix https://github.com/coreos/rpm-ostree/issues/5251 Basically we can't re-invoke the calling process if we aren't actually bootc.

(There are solutions that make this work, but what we'll hopefully
 do shortly is propagate the logic to older streams to have
 bootc take over the handling anyways)


jmarrero (approved):
lgtm

PR #1061, opened 2025-01-26
misc: Don't use inline const to preserve compat with rust 1.78

jmarrero:
An attempt to fix build failure on: https://github.com/coreos/rpm-ostree/pull/5253


cgwalters:
Thanks for looking at this. I'm not opposed but we *should* have newer rust; 1.78 is more than 8 months old now.

I pushed a different attempted fix for this in https://github.com/coreos/rpm-ostree/pull/5253/commits/8ba6c2d67afd7715ce3c00a29348f4f89aa48259

PR #1063, opened 2025-01-27
cli: add `system-reinstall-bootc` binary

omertuc:
See commit messages

### DIFF
--- a/system-reinstall-bootc/src/podman.rs
+++ b/system-reinstall-bootc/src/podman.rs
@@ -1,0 +1,60 @@
+use std::process::Command;
+
+use super::ROOT_KEY_MOUNT_POINT;
+use crate::users::UserKeys;
+
+pub(crate) fn command(image: &str, root_key: &Option<UserKeys>) -> Command {
+    let mut podman_command_and_args = [
+        // We use podman to run the bootc container. This might change in the future to remove the
+        // podman dependency.
+        "podman",
+        "run",
+        // The container needs to be privileged, as it heavily modifies the host
+        "--privileged",
+        // The container needs to access the host's PID namespace to mount host directories
+        "--pid=host",
+        // Since https://github.com/containers/bootc/pull/919 this mount should not be needed, but
+        // some reason with e.g. quay.io/fedora/fedora-bootc:41 it is still needed.
+        "-v",
+        "/var/lib/containers:/var/lib/containers",
+    ]
+    .map(String::from)
+    .to_vec();
+
+    let mut bootc_command_and_args = [
+        "bootc",
+        "install",
+        // We're replacing the current root
+        "to-existing-root",
+        // The user already knows they're reinstalling their machine, that's the entire purpose of
+        // this binary. Since this is no longer an "arcane" bootc command, we can safely avoid this
+        // timed warning prompt. TODO: Discuss in https://github.com/containers/bootc/discussions/1060
+        "--acknowledge-destructive",
+    ]
+    .map(String::from)
+    .to_vec();
+
+    if let Some(root_key) = root_key.as_ref() {
+        let root_authorized_keys_path = root_key.authorized_keys_path.clone();
+
+        podman_command_and_args.push("-v".to_string());
+        podman_command_and_args.push(format!(
+            "{root_authorized_keys_path}:{ROOT_KEY_MOUNT_POINT}"
+        ));
+
+        bootc_command_and_args.push("--root-ssh-authorized-keys".to_string());
+        bootc_command_and_args.push(ROOT_KEY_MOUNT_POINT.to_string());
+    }
+
+    let all_args = [
+        podman_command_and_args,
+        vec![image.to_string()],
+        bootc_command_and_args,
+    ]
+    .concat();
+
+    let mut command = Command::new(&all_args[0]);
+    command.args(&all_args[1..]);
+
+    command
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -638,6 +638,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "339544cc9e2c4dc3fc7149fd630c5f22263a4fdf18a98afd0075784968b5cf00"
 
 [[package]]
+name = "dialoguer"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "658bce805d770f407bc62102fca7c2c64ceef2fbcb2b8bd19d2765ce093980de"
+dependencies = [
+ "console",
+ "shell-words",
+ "tempfile",
+ "thiserror 1.0.69",
+ "zeroize",
+]
+
+[[package]]
 name = "digest"
 version = "0.10.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1952,6 +1965,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "shell-words"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "24188a676b6ae68c3b2cb3a01be17fbf7240ce009799bb56d5b1409051e78fde"
+
+[[package]]
 name = "shlex"
 version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2081,6 +2100,23 @@ dependencies = [
  "pkg-config",
  "toml",
  "version-compare",
+]
+
+[[package]]
+name = "system-reinstall-bootc"
+version = "0.1.9"
+dependencies = [
+ "anyhow",
+ "bootc-utils",
+ "clap",
+ "dialoguer",
+ "log",
+ "rustix",
+ "serde",
+ "serde_json",
+ "serde_yaml",
+ "tracing",
+ "uzers",
 ]
 
 [[package]]
@@ -2419,6 +2455,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "uzers"
+version = "0.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4df81ff504e7d82ad53e95ed1ad5b72103c11253f39238bcc0235b90768a97dd"
+dependencies = [
+ "libc",
+ "log",
+]
+
+[[package]]
 name = "valuable"
 version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2726,6 +2772,12 @@ dependencies = [
  "quote",
  "syn 2.0.87",
 ]
+
+[[package]]
+name = "zeroize"
+version = "1.8.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ced3678a2879b30306d323f4542626697a464a97c0a07c9aebf7ebca65cd4dde"
 
 [[package]]
 name = "zstd"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -170,7 +170,6 @@ dependencies = [
  "anyhow",
  "bootc-lib",
  "bootc-utils",
- "clap",
  "log",
  "tokio",
  "tracing",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -169,11 +169,11 @@ version = "0.1.9"
 dependencies = [
  "anyhow",
  "bootc-lib",
+ "bootc-utils",
  "clap",
  "log",
  "tokio",
  "tracing",
- "tracing-subscriber",
 ]
 
 [[package]]
@@ -249,6 +249,7 @@ dependencies = [
  "tempfile",
  "tokio",
  "tracing",
+ "tracing-subscriber",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,7 @@
 [workspace]
 members = [
     "cli",
+    "system-reinstall-bootc",
     "lib",
     "ostree-ext",
     "utils",
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ all:
     
 install:
 	install -D -m 0755 -t $(DESTDIR)$(prefix)/bin target/release/bootc
+	install -D -m 0755 -t $(DESTDIR)$(prefix)/bin target/release/system-reinstall-bootc
 	install -d -m 0755 $(DESTDIR)$(prefix)/lib/bootc/bound-images.d
 	install -d -m 0755 $(DESTDIR)$(prefix)/lib/bootc/kargs.d
 	ln -s /sysroot/ostree/bootc/storage $(DESTDIR)$(prefix)/lib/bootc/storage
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -10,10 +10,8 @@ default-run = "bootc"
 
 # See https://github.com/coreos/cargo-vendor-filterer
 [package.metadata.vendor-filter]
-# This list of platforms is not intended to be exclusive, feel free
-# to extend it.  But missing a platform will only matter for the case where
-# a dependent crate is *only* built on that platform.
-platforms = ["x86_64-unknown-linux-gnu", "aarch64-unknown-linux-gnu", "powerpc64le-unknown-linux-gnu", "s390x-unknown-linux-gnu", "riscv64gc-unknown-linux-gnu"]
+# For now we only care about tier 1+2 Linux.  (In practice, it's unlikely there is a tier3-only Linux dependency)
+platforms = ["*-unknown-linux-gnu"]
 
 [dependencies]
 anyhow = { workspace = true }
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -19,10 +19,10 @@ platforms = ["x86_64-unknown-linux-gnu", "aarch64-unknown-linux-gnu", "powerpc64
 anyhow = { workspace = true }
 bootc-lib = { version = "1.0", path = "../lib" }
 clap = { workspace = true }
+bootc-utils = { path = "../utils" }
 tokio = { workspace = true, features = ["macros"] }
 log = "0.4.21"
 tracing = { workspace = true }
-tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 
 [lints]
 workspace = true
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -18,7 +18,6 @@ platforms = ["x86_64-unknown-linux-gnu", "aarch64-unknown-linux-gnu", "powerpc64
 [dependencies]
 anyhow = { workspace = true }
 bootc-lib = { version = "1.0", path = "../lib" }
-clap = { workspace = true }
 bootc-utils = { path = "../utils" }
 tokio = { workspace = true, features = ["macros"] }
 log = "0.4.21"
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -1,25 +1,14 @@
 //! The main entrypoint for bootc, which just performs global initialization, and then
 //! calls out into the library.
-
 use anyhow::Result;
 
 /// The code called after we've done process global init and created
 /// an async runtime.
 async fn async_main() -> Result<()> {
-    // Don't include timestamps and such because they're not really useful and
-    // too verbose, and plus several log targets such as journald will already
-    // include timestamps.
-    let format = tracing_subscriber::fmt::format()
-        .without_time()
-        .with_target(false)
-        .compact();
-    // Log to stderr by default
-    tracing_subscriber::fmt()
-        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
-        .event_format(format)
-        .with_writer(std::io::stderr)
-        .init();
-    tracing::trace!("starting");
+    bootc_utils::initialize_tracing();
+
+    tracing::trace!("starting bootc");
+
     // As you can see, the role of this file is mostly to just be a shim
     // to call into the code that lives in the internal shared library.
     bootc_lib::cli::run_from_iter(std::env::args()).await
--- a/contrib/packaging/bootc.spec
+++ b/contrib/packaging/bootc.spec
@@ -62,6 +62,14 @@ Provides: ostree-cli(ostree-container)
 %description
 %{summary}
 
+%package reinstall
+Summary: Utility to reinstall the current system using bootc
+Requires: podman
+# The reinstall subpackage intentionally does not require bootc, as it pulls in many unnecessary dependencies
+
+%description reinstall
+This package provides a utility to simplify reinstalling the current system to a given bootc image.
+
 %prep
 %autosetup -p1 -a1
 # Default -v vendor config doesn't support non-crates.io deps (i.e. git)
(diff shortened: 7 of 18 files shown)


reinstall-cli/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+edition = "2021"
+license = "MIT OR Apache-2.0"
+repository = "https://github.com/containers/bootc"
+readme = "README.md"
+publish = false
+# For now don't bump this above what is currently shipped in RHEL9.
+rust-version = "1.75.0"
+
+# See https://github.com/coreos/cargo-vendor-filterer
+[package.metadata.vendor-filter]
```
cgwalters, line 13: We can use the workspace version of this in the main cargo.toml

reinstall-cli/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+# This list of platforms is not intended to be exclusive, feel free
+# to extend it.  But missing a platform will only matter for the case where
+# a dependent crate is *only* built on that platform.
+platforms = ["x86_64-unknown-linux-gnu", "aarch64-unknown-linux-gnu", "powerpc64le-unknown-linux-gnu", "s390x-unknown-linux-gnu", "riscv64gc-unknown-linux-gnu"]
+
+[dependencies]
+anyhow = { workspace = true }
+bootc-lib = { version = "1.0", path = "../lib" }
```
cgwalters, line 21: I'm fine with this short term but in the medium term...it seems like it'd be cleaner/clearer to have this use only at most our `utils` crate and is otherwise just a consumer of existing public APIs which I think is pretty much the case, it's just you refactored the tracing bits to be shared.

But I think we could actually put the tracing bits into `utils` maybe? Dunno.

reinstall-cli/src/main.rs
```diff
@@ -0,0 +1,84 @@
...
+fn run() -> Result<()> {
+    bootc_lib::cli::tracing_util::initialize_tracing();
+
+    tracing::trace!("starting bootc-reinstall");
+
+    prompt()?;
+
+    let command_and_args = [
+        // Rootless is not supported
+        "sudo",
```
cgwalters, line 16: I personally get a bit weirded out when tools invoke `sudo` automatically. It's not wrong per se, but it can lead to confusion. Do you think we really need it?

reinstall-cli/src/main.rs
```diff
@@ -0,0 +1,84 @@
...
+        "to-existing-root",
+        // The user already knows they're reinstalling their machine, that's the entire purpose of
+        // this binary. Since this is no longer an "arcane" bootc command, we can safely avoid this
+        // timed warning prompt. TODO: Discuss in https://github.com/containers/bootc/discussions/1060
+        "--acknowledge-destructive",
+    ];
+
+    Command::new(command_and_args[0])
+        .args(&command_and_args[1..])
+        .status()
```
cgwalters, line 43: At this point we could consider just `exec` too?

cgwalters (commented):
Nice, thanks for getting the ball rolling here!

reinstall-cli/Cargo.toml
omertuc: Yeah I wanted to have it in a separate crate but didn't want to create one

Didn't know we had a utils crate

Anyway it seems the resulting RPM is still very tiny. The linker must discard all the entirety of the unused bootc library

reinstall-cli/src/main.rs
omertuc: 100% definitely, just copy pasted the command without giving it much thought. The user should do the sudo themselves

reinstall-cli/src/main.rs
omertuc: I kinda want to print the command at the end, in case it fails. But otherwise yeah exec makes perfect sense

reinstall-cli/Cargo.toml
omertuc: image

Do you mean I should also take it out of `cli/Cargo.toml`?

cgwalters: Oh yes, it's not needed there.

reinstall-cli/src/main.rs
cgwalters: We can still print an error if the exec fails right? There's several other instances of `exec()` in this project.

reinstall-cli/Cargo.toml
cgwalters: > Didn't know we had a utils crate

It may not even be needed here, it doesn't have that much.

> Anyway it seems the resulting RPM is still very tiny. The linker must be discarding the entirety of the bootc library save for the used bit

LTO can work really well, yeah.

I'm not worried about size really, it's just more that ideally we show that one can build *external* installer tools on top of our publicly exposed APIs.

reinstall-cli/src/main.rs
omertuc: >We can still print an error if the exec fails right? 

Isn't the point of `exec` that the `podman` process replaces the `bootc-installer` process? Nothing can run after an `exec`. Am I misunderstanding?

reinstall-cli/Cargo.toml
omertuc: Can't say I fully understand your intentions and remarks with regards to exposed APIs though. The new binary is not really a consumer of bootc functionality, it's just a command runner. The only reason I made it use the library in the first place was just to share a bit of tracing boilerplate

reinstall-cli/src/main.rs
cgwalters: As far as I can tell right now, nothing is running after the `Command`? Hence we could use exec. But it's a really minor point honestly.

reinstall-cli/Cargo.toml
cgwalters: Yep we're agreed!

reinstall-cli/src/main.rs
omertuc: The `?` operator is running after the command. Thus:

> kinda want to print the command at the end, in case it fails.

cgwalters: Yes, handling the error from exec requires some special handling because it always returns an error (it never returns on success). Here's an example of that in this project:
https://github.com/containers/bootc/blob/5f0dd2361a8ba33ff339359676eb006f3cc8c28d/lib/src/install.rs#L881

omertuc: Oh I didn't realize it returns on error. I thought it never returns period. Will do that

omertuc: Oh and it only errors if there was a problem running the command itself. If the command that was ran fails (non-zero exit code), it will not return an error

cgwalters: I'd say we should probably print the command unconditionally before we start, regardless of error or not; WDYT?

omertuc: Definitely, but by that point it will be buried under 50 layer podman pull logs. So I thought it would be nice to have it after

reinstall-cli/src/podman.rs
```diff
@@ -0,0 +1,66 @@
...
+        "--acknowledge-destructive",
+    ]
+    .map(String::from)
+    .to_vec();
+
+    if let Some(root_key) = root_key.as_ref() {
+        let root_authorized_keys_path = root_key.authorized_keys_path.clone();
+
+        podman_command_and_args.push("-v".to_string());
+        podman_command_and_args.push(format!(
```
cgwalters, line 41: For later followup: I think we could support `--root-ssh-authorized-keys-fd` or so, and pass it via file descriptor. That'd avoid the bind dance here.

Or maybe we should learn `--ssh-home-dir`, and directly scrape all keys from there?

reinstall-cli/src/prompt.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    // TODO: Handle https://github.com/console-rs/dialoguer/issues/77
+    let selected_user_indices: Vec<usize> = dialoguer::MultiSelect::new()
+        .with_prompt("Select the users you want to install in the system (along with their authorized SSH keys)")
+        .items(&keys)
+        .interact()?;
+
+    Ok(selected_user_indices
+        .iter()
+        // Safe unwrap because we know the index is valid
```
cgwalters, line 27: I dug because I was curious what happens if `MultiSelect` was passed an empty set, looks like it's this: https://github.com/console-rs/dialoguer/blob/6244c77656e8f3e934a2981c5c7e9dacf938a492/src/prompts/multi_select.rs#L211

OK.

reinstall-cli/src/users.rs
```diff
@@ -0,0 +1,106 @@
...
+            tracing::debug!(
+                "Skipping user {} with shell {} because it can't login",
+                user_info.name().to_string_lossy(),
+                user_info.shell().to_string_lossy()
+            );
+            continue;
+        }
+
+        let home_dir = user_info.home_dir();
+        let user_authorized_keys_path = home_dir.join(".ssh/authorized_keys");
```
cgwalters, line 62: It's a bit harder than this in the general case; in CoreOS derivatives we've pushed for https://github.com/coreos/ssh-key-dir for example.

reinstall-cli/src/users.rs
```diff
@@ -0,0 +1,106 @@
...
+/// See safety information in [`uzers::all_users`]. We are not impacted because we're not
+/// multi-threaded.
+pub(crate) fn get_all_users_keys() -> Result<Vec<UserKeys>> {
+    #[allow(unsafe_code)]
+    let iter = unsafe { uzers::all_users() };
+
+    let mut result = Vec::new();
+
+    for user_info in iter {
+        if user_info.uid() < 1000 && user_info.uid() != 0 {
```
cgwalters, line 40: Note though 1000 is recommend, it's not *required* to be the barrier. See https://systemd.io/UIDS-GIDS/

But see above

reinstall-cli/src/users.rs
```diff
@@ -0,0 +1,106 @@
...
+        )
+    }
+}
+
+/// # Safety
+/// See safety information in [`uzers::all_users`]. We are not impacted because we're not
+/// multi-threaded.
+pub(crate) fn get_all_users_keys() -> Result<Vec<UserKeys>> {
+    #[allow(unsafe_code)]
+    let iter = unsafe { uzers::all_users() };
```
cgwalters, line 35: It may be simpler and more reliable here to parse the output of `loginctl -j`. That will only give us logged in users, which in the common case of non-root user + sudo will give us both, and in the case of just logging in as root won't offer other users, which also seems right.

reinstall-cli/src/users.rs
```diff
@@ -0,0 +1,106 @@
...
+        {
+            tracing::debug!(
+                "Skipping user {} with shell {} because it can't login",
+                user_info.name().to_string_lossy(),
+                user_info.shell().to_string_lossy()
+            );
+            continue;
+        }
+
+        let home_dir = user_info.home_dir();
```
cgwalters, line 61: Not that it *really* matters here, but just commenting here because in the general case reading/writing a user's home directory from a privileged process requires some care, as it's otherwise easy for a user to trick root into following a symlink and reading/writing outside of it. (Really more generally having a higher privileged process read/write to a directory which is writable by a lower privileged process)

The strongest and most portable way is to fork a helper process that's `setuid` to the user, and in this case have it do the operations.

In our case since we're just reading (not writing here) we can address the symlink problem using cap-std, which we use pervasively in this project anyways.

reinstall-cli/src/podman.rs
```diff
@@ -0,0 +1,66 @@
...
+
+    [
+        podman_command_and_args,
+        vec![image.to_string()],
+        bootc_command_and_args,
+    ]
+    .concat()
+}
+
+pub(crate) fn run_podman(all_args: Vec<String>) -> Result<()> {
```
cgwalters, line 57: This isn't specific to running podman? See below

reinstall-cli/src/podman.rs
```diff
@@ -0,0 +1,66 @@
...
+    Command::new(all_args[0].clone())
+        .args(all_args[1..].iter())
+        .status()
+        .context(format!(
+            "Failed to run the command \"{}\"",
+            all_args.join(" ")
```
cgwalters, line 63: I think the debug output of `Command` correctly handles shell quoting, which is pretty useful because it means you can copy-paste.

It looks like the main goal of this `run_podman` is really to add the error context from the command so we could just call it `run_cmd` or something?

But that said, I notice an immediate large problem here in that we're not checking the return code of the child...this type of thing is why I added the `CommandRunExt` helpers in the bootc-utils crate.

I think we could extend it with something like `run_with_cmd_errcontext()` or so?

reinstall-cli/src/main.rs
```diff
@@ -0,0 +1,36 @@
...
+
+    let config = config::ReinstallConfig::load().context("loading config")?;
+
+    tracing::trace!("starting bootc-reinstall");
+    // Rootless podman is not supported by bootc
+    ensure!(getuid().is_root(), "Must run as the root user");
+    let all_args = podman::command(&config.bootc_image, prompt::get_root_key()?);
+    println!("Going to run command: {}", all_args.join(" "));
+    prompt::temporary_developer_protection_prompt()?;
+    podman::run_podman(all_args)?;
```
cgwalters, line 24: One thing that'd be cool I think is on failure, save all the provided interactive answers in e.g. `/run/bootc/reinstall.yaml` or so, and tell the user:

```
Installation answers saved in /run/bootc/reinstall.yaml
error: Out of disk space
```

Then when we re-run we detect that and offer to resume from saved answers.

reinstall-cli/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "bootc-reinstall"
```
cgwalters, line 2: By far the biggest thing we need to figure out before trying to land this is one of the hardest problems: naming the thing.

I am not opposed to `bootc-reinstall`...but I'm not in love with it either. One minor thing I mentioned elsewhere is that if we have both `bootc` and `bootc-reinstall` then shells will stop tab completion without adding a space.

I guess in *most* cases they'll be mutually exclusive? i.e. you only have `bootc-reinstall` on non-bootc hosts? But I'm sure people will ask us why we don't ship this on bootc images too...

Another option which I like a little better is `system-reinstall-bootc` - it's longer to type which is a good thing here, doesn't clash with tab completion for `bootc`, and it feels more "tightly scoped".

Having it *not* have the `bootc` prefix also makes clearer that logically it's a higher level tool that in theory *could* live separately or be maintained elsewhere.

IOW I think people would get legitimately confused if we have `bootc install` and `bootc-reinstall`. It seems easier to message `system-reinstall-bootc` is a standalone tool.

Open to other names too of course...

reinstall-cli/src/config/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+            Some(value) => value
+                .as_str()
+                .context("bootc_image must be a string")?
+                .to_string(),
+            None => bail!("bootc_image is required"),
+        };
+
+        ensure!(
+            value.is_empty(),
+            "unknown keys {:?} in config file",
```
cgwalters, line 38: Confused why we aren't using `#[derive(Deserialize)]` on the struct and https://serde.rs/container-attrs.html#deny_unknown_fields

cgwalters (commented):
Awesome progress!! Various comments follow

reinstall-cli/src/config/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+pub(crate) struct ReinstallConfig {
+    pub(crate) bootc_image: String,
+}
+
+impl ReinstallConfig {
+    pub fn parse_from_cli(cli: cli::Cli) -> Result<Self> {
+        Ok(Self {
+            bootc_image: cli.bootc_image,
+        })
+    }
```
germag, line 18: why returning `Result<Self>`  instead of just `Self` ?

reinstall-cli/src/config/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+                .context("bootc_image must be a string")?
+                .to_string(),
+            None => bail!("bootc_image is required"),
+        };
+
+        ensure!(
+            value.is_empty(),
+            "unknown keys {:?} in config file",
+            value.keys().map(|key| key.to_string()).join(", ")
+        );
```
germag, line 40: you can replace the closure with `ToString::to_string`
```
value.keys().map(ToString::to_string).join(", ")
```

reinstall-cli/src/config/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+    pub fn load() -> Result<Self> {
+        Ok(match std::env::var("BOOTC_REINSTALL_CONFIG") {
+            Ok(var) => {
+                ensure_no_cli_args()?;
+
+                ReinstallConfig::parse_from_config_file(
+                    &std::fs::read(&var)
+                        .context(format!("reading BOOTC_REINSTALL_CONFIG file {}", var))?,
+                )
+                .context(format!("parsing BOOTC_REINSTALL_CONFIG file {}", var))?
```
germag, line 56: inline `var`:
```
                        .context(format!("... {var}"))?,
```

reinstall-cli/src/podman.rs
```diff
@@ -0,0 +1,66 @@
+use super::ROOT_KEY_MOUNT_POINT;
+use crate::users::UserKeys;
+use anyhow::{Context, Result};
+use std::process::Command;
+
+pub(crate) fn command(image: &str, root_key: Option<UserKeys>) -> Vec<String> {
+    let mut podman_command_and_args = [
```
germag, line 7: nit: `root_key: &Option<UserKeys>`

reinstall-cli/src/podman.rs
germag, line 57: No need to use a `Vec`: `run_podman(all_args: &[String]) -> Result<()> `

germag (commented):
Just a general comment, even if this simple comment, I think It will better to split some part in different commits, like adding tracing, etc.

reinstall-cli/src/podman.rs
omertuc: How would we pass fd's across podman run? Do you mean using something like `--preserve-fds`?

cgwalters: > Do you mean using something like --preserve-fds?

Yes, exactly.

reinstall-cli/src/users.rs
omertuc: But what do we do on the target machine? Try to emulate the same structure as the original machine? Just copy all the keys and dump them naively in the target machine? 

How do we make sure the binary behind `AuthorizedKeysCommand` even behaves the same on the source and target machines?

Since `AuthorizedKeysCommand` is an opaque box, it's actually impossible to know what we would have to do on the target machine such that it still outputs the same keys

This all seems like a bad idea the more I think about it

I think we might have to special case the coreos scenario, as solving this in the general seems literally impossible

cgwalters: That seems reasonably straightforward. @germag and I had a live chat about this too and my take is:

- For this initial version, we can aim to tightly constrain our starting points. If we even just handle the stock ssh configuration for e.g. RHEL9.5 and RHEL CoreOS (already enough diversity there since `AuthorizedKeysCommand` is in play in one but not the other) 
- If we even need to add distribution/OS detection and bail out on untested setups (say ubuntu or whatever) I think that's fine to start, with maybe a `--allow-untested-ssh` CLI escape hatch) and obviously ideally it's reasonably easy for us or other folks to add more cases later

reinstall-cli/src/users.rs
omertuc: I assume you're referring to `loginctl --output=json`?

That doesn't seem to give me any information about root, or the user's home directories, even with `--all`

reinstall-cli/src/users.rs
omertuc: Long comment only 3 minutes after my second comment, just to make sure - please make sure you've seen my second comment

reinstall-cli/src/config/mod.rs
omertuc: Oops, just leftover from copy pasting from [my previous project where this could fail due to JSON serialization of env::args()](https://github.com/rh-ecosystem-edge/recert/blob/e9428cc982b13bce78f366a628680188086a1797/src/config.rs#L453)

Which come to think of would be nice feature to have here as well, if we wanna leave any trace of the reinstallation configuration, it would be nice to have the raw CLI flags included in that trace

reinstall-cli/src/users.rs
cgwalters: > That doesn't seem to give me any information about root, or the user's home directories, even with --all

I mean that `loginctl` replaces `uzers::all_users()` - we don't actually want to iterate over *all* users (necessarily) just logged in ones. (bonus: we kill an `unsafe`). We'd still use `uzers::get_user_by_name` on the names we find from `loginctl` to gather other data.

As far as `root`, if it doesn't show up from logged in users we could just add it to the list automatically.

omertuc: Alright

reinstall-cli/Cargo.toml
cgwalters: > but I feel it's gonna get pushback as too long. 

Two things:
- It's not something we want people to type casually though :smile: 
- At least in current Fedora Workstation, `system-` has only one other possible completion in `system-config-abrt`, so you just need to type `system-r<TAB>` at most which is not bad.

The `system-` nomenclature has a history in Red Hat Linux derivatives at least where there was a push for a while to have standardized CLI and GUI tools like `system-config-network` (goes back to [RHEL5](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/5/html/5.11_technical_notes/system-config-network)) but those have long since gone away overall, freeing up the `system-` namespace (apart from system*d* of course but the chance for confusion with that suite is fixed by the `-bootc` suffix I think)

reinstall-cli/src/podman.rs
omertuc: Followup under #1080

reinstall-cli/src/users.rs
omertuc: Followup in #1081

reinstall-cli/src/main.rs
omertuc: Followup in #1082

system-reinstall-bootc/src/podman.rs
```diff
@@ -0,0 +1,60 @@
...
+        // We use podman to run the bootc container. This might change in the future to remove the
+        // podman dependency.
+        "podman",
+        "run",
+        // The container needs to be privileged, as it heavily modifies the host
+        "--privileged",
+        // The container needs to access the host's PID namespace to mount host directories
+        "--pid=host",
+        // Since https://github.com/containers/bootc/pull/919 this mount should not be needed, but
+        // some reason with e.g. quay.io/fedora/fedora-bootc:41 it is still needed.
```
cgwalters, line 17: We should try to root cause some of this. Let's pile onto https://github.com/containers/bootc/issues/1053

cgwalters (approved):
Looks sane to me! Missing DCO is the main thing

reinstall-cli/Cargo.toml
omertuc: Let's continue discussion over at #1083

(issue and not discussion because we need to reach a final decision)

omertuc:
Branched off some discussions / issues / further work into:

#1080 #1081 #1082 #1083

omertuc:
Still need to fix Cargo.lock

Couldn't build because need to upgrade to Fedora 41 since the ostree dependency became 2025

cgwalters:
> Branched off some discussions / issues / further work into:

I also added [a new label](https://github.com/containers/bootc/issues?q=is%3Aissue%20state%3Aopen%20label%3Aarea%2Fsystem-reinstall-bootc) too!

reinstall-cli/src/users.rs
omertuc: Seems like the exact flag changed in recent systemd versions  

Related:
https://github.com/kolide/launcher/issues/1758
https://github.com/systemd/systemd/issues/15275
https://github.com/systemd/systemd/pull/16176

PR #1068, opened 2025-02-01
Install config verity prep

cgwalters:
Prep for https://github.com/containers/bootc/pull/935


cgwalters:
Man, this has been a crazy chain of things I hit. The vendoring thing was surprisingly complicated. But OK, now the latest thing is that composefs-rs's MSRV is newer than what's in RHEL9.5 (not C9S, so we know we're good for 9.6+).

jmarrero (approved):
lgtm

PR #1069, opened 2025-02-01
Update to ostree-rs 0.20

cgwalters:
Notably this drops out several duplicate crates.

Depends on https://github.com/ostreedev/ostree/pull/3378 and (and publishing the update to crates.io)


jeckersb (approved):
lgtm

PR #1071, opened 2025-02-03
ostree-ext: Add `image clear-cached-update`

cgwalters:
This is effectively an internals command for us to aid debugging.

At some point it may make sense to expose this
at the bootc level too.

### DIFF
--- a/ostree-ext/src/container/store.rs
+++ b/ostree-ext/src/container/store.rs
@@ -23,9 +23,10 @@ use futures_util::TryFutureExt;
 use oci_spec::image::{
     self as oci_image, Arch, Descriptor, Digest, History, ImageConfiguration, ImageManifest,
 };
+use ostree::glib::FromVariant;
 use ostree::prelude::{Cast, FileEnumeratorExt, FileExt, ToVariant};
 use ostree::{gio, glib};
-use std::collections::{BTreeSet, HashMap};
+use std::collections::{BTreeMap, BTreeSet, HashMap};
 use std::iter::FromIterator;
 use tokio::sync::mpsc::{Receiver, Sender};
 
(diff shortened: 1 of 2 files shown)


jmarrero (approved):
lgtm

ostree-ext/src/container/store.rs
```diff
@@ -1196,6 +1197,37 @@ fn parse_cached_update(meta: &glib::VariantDict) -> Result<Option<CachedImageUpd
...
+#[context("Clearing cached update {imgref}")]
+pub fn clear_cached_update(repo: &ostree::Repo, imgref: &ImageReference) -> Result<()> {
+    let cancellable = gio::Cancellable::NONE;
+    let ostree_ref = ref_for_image(imgref)?;
+    let rev = repo.require_rev(&ostree_ref)?;
+    let Some(commitmeta) = repo.read_commit_detached_metadata(&rev, cancellable)? else {
+        return Ok(());
+    };
+
+    // SAFETY: We know this is an a{sv}
```
jeckersb, line 1210: I have no idea what `a{sv}` means here, not a blocker but please enlighten me :smiley:

jeckersb (approved):

ostree-ext/src/container/store.rs
cgwalters: gvariant is a bit like a binary JSON, although also with a richer type system (e.g. 32 and 64 bit signed/unsigned integers, not just "number" etc.).

See https://docs.gtk.org/glib/gvariant-format-strings.html - specifically `a{sv}` is a lot like a default json dictionary (string keys, heterogenous values).

jeckersb: thanks for the lesson :mortar_board:

PR #1073, opened 2025-02-04
docs/filesystem: Flesh out `etc` a bit more

cgwalters:
- Mention `ostree admin config-diff`
- Note that metadata also counts as a modification


vrothberg (approved):
LGTM
Thank you !

Issue #76, opened 2023-03-15
Add `bootc apply-live`


cgwalters:
This is probably mainly draining the logic from rpm-ostree into ostree-ext, then re-using it here.

But...a whole lot of suddenly OS-specific issues come to the fore.  For example, should we try to distinguish between "new content" and "changes"?

cgwalters:
This one also relates to https://github.com/containers/bootc/issues/165

I know at least some people have also been asking for an "apply live by default" mode.  Note that there's a super tricky detail of that in that if there's kernel changes in the new root, we must in general do something like keep `/usr/lib/modules/$kver` mounted/copied into the running root.

alexlarsson:
> This one also relates to #165
> 
> I know at least some people have also been asking for an "apply live by default" mode. Note that there's a super tricky detail of that in that if there's kernel changes in the new root, we must in general do something like keep `/usr/lib/modules/$kver` mounted/copied into the running root.

Could we bind mount /usr/lib/modules somewhere else like /run/booted-kernel-modules during the initial boot, and then have the kernel also look for modules there? Then we don't have to mess with trying to have union the new and old modules dirs?

fabiendupont:
Could we use the same logic as the [DNF needs-restarting plugin](https://dnf-plugins-core.readthedocs.io/en/latest/needs_restarting.html)? This gives hints whether processes, services or the whole system needs a restart to account for the changes. And this would keep us close to DNF.

It might also be nice to have a `--dry-run` option to know before applying a change whether it will require a process / service / system restart. It would also be interesting to have the ability to use the same logic to compare two images out of band, e.g. in a CI/CD pipeline.

SeanMooney:
> Could we use the same logic as the [DNF needs-restarting plugin](https://dnf-plugins-core.readthedocs.io/en/latest/needs_restarting.html)? This gives hints whether processes, services or the whole system needs a restart to account for the changes. And this would keep us close to DNF.
> 
> It might also be nice to have a `--dry-run` option to know before applying a change whether it will require a process / service / system restart. It would also be interesting to have the ability to use the same logic to compare two images out of band, e.g. in a CI/CD pipeline.

the intent behind apply-live is to avoid reboots, where a kernel update is actually required a more natural fit would be to have a kexec flow to switch directly to the new kernel or defer that until the next actual reboot is requested.

have bootc apply-live ever triver a whole system restart would make it un usabel for some usecases.
[...]

PR #1076, opened 2025-02-04
Add kexec support

Mstrodl:
When --kexec is passed, an upgrade or switch will load
the new deployment into the kernel's kexec buffer

I don't know how to get a bootc dev environment quickly,
so it's very possible this doesn't actually work.
Please feel free to take this as a base and make changes :)

### DIFF
--- a/lib/src/deploy.rs
+++ b/lib/src/deploy.rs
@@ -478,6 +478,27 @@ pub(crate) fn get_base_commit(repo: &ostree::Repo, commit: &str) -> Result<Optio
     Ok(r)
 }
 
+#[context("Loading deployment into kexec")]
+async fn kexec_load(sysroot: &Storage, deployment: &Deployment) -> Result<()> {
+    // Clone all the things to move to worker thread
+    let sysroot = sysroot.sysroot.clone();
+    // ostree::Deployment is incorrently !Send 😢 so convert it to an integer
+    let deployment_index = deployment.index() as usize;
+
+    async_task_with_spinner(
+        "Deploying",
+        spawn_blocking_cancellable_flatten(move |cancellable| -> Result<()> {
+            let deployments = sysroot.deployments();
+            let deployment = &deployments[deployment_index];
+
+            sysroot.deployment_kexec_load(&deployment, Some(cancellable))?;
+            Ok(())
+        }),
+    )
+    .await?;
+    Ok(())
+}
+
 #[context("Writing deployment")]
 async fn deploy(
     sysroot: &Storage,
@@ -552,6 +573,12 @@ fn origin_from_imageref(imgref: &ImageReference) -> Result<glib::KeyFile> {
     Ok(origin)
 }
 
+#[derive(Debug, Clone, Default)]
+#[non_exhaustive]
+pub(crate) struct StageOptions {
+    pub(crate) deploy_kexec: bool,
+}
+
 /// Stage (queue deployment of) a fetched container image.
 #[context("Staging")]
 pub(crate) async fn stage(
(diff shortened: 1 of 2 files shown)


Mstrodl:
Related to, but doesn't fix #76

lgarbarini:
I literally just sat down to make this change 😂. Just tested with `bootc upgrade --kexec --apply` on my dev environment and it just works!

lib/src/deploy.rs
```diff
@@ -478,6 +478,27 @@ pub(crate) fn get_base_commit(repo: &ostree::Repo, commit: &str) -> Result<Optio
     Ok(r)
 }
 
+#[context("Loading deployment into kexec")]
+async fn kexec_load(sysroot: &Storage, deployment: &Deployment) -> Result<()> {
```
lgarbarini, line 482: I personally might find it nice to have positive confirmation on stdout that the kernel has been loaded via `kexec` to reduce the risk of triggering a time consuming firmware/hard reboot. Seems like this is done elsewhere here with println.

```
[tester@fedora ~]$ sudo ./bootc upgrade --kexec 
layers already present: 79; layers needed: 2 (674 bytes)
Fetched layers: 674 B in 1 second (846 B/s)                                                                                                                                                                                                     
  Deploying: done (7 seconds)                                                                                                                                                                                                                   Pruned images: 0 (layers: 0, objsize: 33.7 kB)
Queued for next boot: [...]
  Version: 41.20250123.0
  Digest: sha256:[...]
  Deploying: done (0 seconds)                                                                                                                                                                                                                   Total new layers: 81    Size: 3.0 GB
Removed layers:   2     Size: 675 bytes
Added layers:     2     Size: 674 bytes
```

Mstrodl: Changed `stage` to have a print saying it kexec'd successfully.

lgarbarini:
@cgwalters this looks good to me! What else needs to be done to merge?

lib/src/deploy.rs
```diff
@@ -552,6 +573,12 @@ fn origin_from_imageref(imgref: &ImageReference) -> Result<glib::KeyFile> {
     Ok(origin)
 }
 
+#[derive(Debug, Clone, Default)]
+#[non_exhaustive]
+pub(crate) struct StageOptions {
```
cgwalters, line 578: Not opposed but this feels slightly overkill at this point - we don't have *that* many parameters to the function.

(And if we did we may as well move the progress writer in?)

I think a simple `kexec: bool` would be OK no?

lib/src/deploy.rs
```diff
@@ -560,7 +587,11 @@ pub(crate) async fn stage(
     image: &ImageState,
     spec: &RequiredHostSpec<'_>,
     prog: ProgressWriter,
+    StageOptions { deploy_kexec }: StageOptions,
 ) -> Result<()> {
+    let steps_total = 4 + (deploy_kexec as u64);
+    let mut steps = 0;
```
cgwalters, line 593: How about:
```
let mut steps = 4;
if deploy_kexec {
  steps += 1;
}
```
?

Feels like it will "scale" a bit more cleanly if we have more conditionals later.

(This all said, I'd really like to have higher level wrappers for the progress API because the numbers here are SO easy to mess up)

cgwalters (commented):
Sorry for the delay and thanks so much for starting this!

Maybe it could work out more generally to have `bootc prepare-kexec` that can be applied as a secondary step after any operation?

Hmm also ideally this would be something visible in `bootc status`...digging in very slightly behind what `kexec --status` does it looks like there's `/sys/kernel/kexec_loaded` which just gives a `1` or `0`...pretty primitive, but eh.

I guess we could store some state in e.g. `/run/bootc`...or hmm, arguably the ostree API should store something in the per-deployment state.

(I should have thought about this when we merged the ostree PR, but at the bootc level I really try hard to think about spec/status and machine readable APIs)

lib/src/deploy.rs
Mstrodl: Yeah, that's pretty much the same idea, but I see how that would be cleaner.

kfox1111:
Very interesting. Is this still being worked on?

PR #1078, opened 2025-02-05
 tar/export: Fix error with zero-sized hardlinked files

cgwalters:
tar/export: Add error context

---

tar/export: Fix error with zero-sized hardlinked files

In the case where we already emitted an object into the tar stream,
we wouldn't previously set the file type and size into the tar
header structure that we passed into the hardlink emission function.
This would cause an error when accessing the (logically) uninitialized `size()`.

Fix the writer function to always set the file type and size unconditionally,
and also rework the consumer side to make the logic even clearer.

---


jeckersb:
Just ran into this myself while testing https://github.com/coreos/rpm-ostree/pull/5275, gonna pull this in and sanity check it

PR #1086, opened 2025-02-06
blockdev: Extend lsblk output with parttype and partuuid

cgwalters:
We want parttype in general, for the same reason we wanted it from `sfdisk`. Prep for maybe using this in bootupd for https://github.com/coreos/bootupd/issues/837


jeckersb (approved):
lgtm

PR #1092, opened 2025-02-07
reinstall: warn if no users with authorized ssh keys are found

omertuc:


system-reinstall-bootc/src/main.rs
```diff
@@ -20,8 +20,13 @@ fn run() -> Result<()> {
 
     let config = config::ReinstallConfig::load().context("loading config")?;
 
-    let mut reinstall_podman_command =
-        podman::command(&config.bootc_image, &prompt::get_root_key()?);
+    let root_key = &prompt::get_root_key()?;
+
+    if root_key.is_none() {
+        tracing::warn!("NO USERS WITH AUTHORIZED SSH KEYS FOUND! You might not be able to log in after reinstalling the system.");
```
cgwalters, line 26: Well, but this is where we could also detect if e.g. cloud-init or other tools are in the target image right?

Also, something strongly related to this is I think we probably are going to need to scope creep this tool slightly into supporting non-cloud flows; I've seen people doing that already with `to-existing-root` - basically the scenario here is:

- Install Fedora Server (or whatever), have a physical console
- Want to replace it w/bootc image

In this flow we'd need to support a root password, not ssh keys. One general purpose option here would be to support running `systemd-firstboot`. Or of course, we could just try to automatically scrape a root password in addition too.

I kind of like that latter one as an expedient option - in clouds we typically won't have a root password so it'd be a no-op.

---

We can track all of the above as an issue, but how about just slightly softening the tone here, and also making this a yes/no continue question instead of just a warning?

PR #1094, opened 2025-02-07
reinstall: avoid duplicate users in `loginctl_users`

### LINKED ISSUE #1093: reinstall: duplicate usernames
Image

omertuc:
Fixes #1093

### DIFF
--- a/system-reinstall-bootc/src/users.rs
+++ b/system-reinstall-bootc/src/users.rs
@@ -5,14 +5,20 @@ use rustix::process::geteuid;
 use rustix::process::getuid;
 use rustix::thread::set_thread_res_uid;
 use serde_json::Value;
+use std::collections::BTreeSet;
 use std::fmt::Display;
 use std::fmt::Formatter;
 use std::process::Command;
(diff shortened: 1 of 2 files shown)


omertuc:
Draft cause still testing

system-reinstall-bootc/src/users.rs
```diff
@@ -5,12 +5,13 @@ use rustix::process::geteuid;
...
 use rustix::thread::set_thread_res_uid;
 use serde_json::Value;
+use std::collections::HashSet;
 use std::fmt::Display;
 use std::fmt::Formatter;
 use std::process::Command;
 use uzers::os::unix::UserExt;
 
-fn loginctl_users() -> Result<Vec<String>> {
+fn loginctl_users() -> Result<HashSet<String>> {
```
cgwalters, line 14: Probably better to use BTreeSet so the order isn't random.

---

As far as testing, one pattern we use elsewhere with code like this is to have a unit test with some example output as a fixture. See the blockdev code for that.

cgwalters (approved):

PR #1095, opened 2025-02-07
lints: More cleanup && lints: Check for invalid /etc/hostname and /etc/resolv.conf

cgwalters:
lints: Use enum instead of bool

This is always clearer. Prep for adding another argument to this function
that's an effective bool.

---

lints: Rework to use linkme

Two goals:

- The global static LINTS array is a conflict point
- It's easier to lay out the lint info when it's next to each
  function; prep for extending the lint data more.

---

lints: Check for invalid /etc/hostname and /etc/resolv.conf

Detect problems from https://github.com/containers/buildah/issues/4242
or similar.

As part of this, add new infrastructure logic for lints that only operate on
non-running roots (we expect these are mounted/written at runtime).

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -213,6 +213,7 @@ dependencies = [
  "libc",
  "liboverdrop",
  "libsystemd",
+ "linkme",
  "openssl",
  "ostree-ext",
  "regex",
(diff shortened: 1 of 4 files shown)


jeckersb (approved):
:+1: definitely nice having the lints inlined with the functions

PR #1096, opened 2025-02-07
store: Clean up broken /etc/hostname and /etc/resolv.conf

cgwalters:
We can pretty safely work around this here; zero sized files for both are useless.

Closes: https://github.com/containers/bootc/issues/1064


jeckersb (approved):

cgwalters:
Followup in https://github.com/containers/bootc/pull/1167

PR #1105, opened 2025-02-11
reinstall: Ensure podman is installed

### LINKED ISSUE #1104: reinstall: Support installing podman
I'd like to allow distro images like Fedora Cloud to add system-reinstall-bootc but not necessarily pull in podman if they weren't already.

So basically:

- Change to just `Recommends: podman` for system-reinstall-podman
- Change the code to install it if not present (this of course raises the "package system abstraction" problem)

omertuc:
See commit messages

ensure_podman_installed_flowchart

Fixes #1104

### DIFF
--- a/system-reinstall-bootc/src/podman.rs
+++ b/system-reinstall-bootc/src/podman.rs
@@ -1,7 +1,9 @@
-use std::process::Command;
-
 use super::ROOT_KEY_MOUNT_POINT;
 use crate::users::UserKeys;
+use anyhow::{ensure, Context, Result};
+use bootc_utils::CommandRunExt;
+use std::process::Command;
+use which::which;
 
 pub(crate) fn command(image: &str, root_key: &Option<UserKeys>) -> Command {
     let mut podman_command_and_args = [
@@ -57,4 +59,43 @@ pub(crate) fn command(image: &str, root_key: &Option<UserKeys>) -> Command {
     command.args(&all_args[1..]);
 
     command
+}
+
+/// Path to the podman installation script. Can be influenced by the build
+/// SYSTEM_REINSTALL_BOOTC_INSTALL_PODMAN_PATH parameter to override. Defaults
+/// to /usr/lib/system-reinstall-bootc/install-podman
+const fn podman_install_script_path() -> &'static str {
+    if let Some(path) = option_env!("SYSTEM_REINSTALL_BOOTC_INSTALL_PODMAN_PATH") {
+        path
+    } else {
+        "/usr/lib/system-reinstall-bootc/install-podman"
+    }
+}
+
+pub(crate) fn ensure_podman_installed() -> Result<()> {
+    if which("podman").is_ok() {
+        return Ok(());
+    }
+
+    tracing::warn!(
+        "Podman was not found on this system. It's required in order to install a bootc image."
+    );
+
+    ensure!(
+        which(podman_install_script_path()).is_ok(),
+        "Podman installation script {} not found, cannot automatically install podman. Please install it manually and try again.",
+        podman_install_script_path()
+    );
+
+    Command::new(podman_install_script_path())
+        .run_with_cmd_context()
+        .context("installing podman")?;
+
+    // Make sure the installation was actually successful
+    ensure!(
+        which("podman").is_ok(),
+        "podman still doesn't seem to be available, despite the installation. Please install it manually and try again."
+    );
+
+    Ok(())
 }
--- a/utils/src/command.rs
+++ b/utils/src/command.rs
@@ -139,7 +139,9 @@ impl CommandRunExt for Command {
     }
 
     fn run_with_cmd_context(&mut self) -> Result<()> {
-        self.run()
+        self.status()?
+            .success()
+            .then_some(())
             // The [`Debug`] output of command contains a properly shell-escaped commandline
             // representation that the user can copy paste into their shell
             .context("Failed to run command: {self:#?}")
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -668,10 +668,22 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0d6ef0072f8a535281e4876be788938b528e9a1d43900b82c2569af7da799125"
 
 [[package]]
+name = "either"
+version = "1.13.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "60b1af1c220855b6ceac025d3f6ecdd2b7c4894bfe9cd9bda4fbb4bc7c0d4cf0"
+
+[[package]]
 name = "encode_unicode"
 version = "0.3.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a357d28ed41a50f9c765dbfe56cbc04a64e53e5fc58ba79fbc34c10ef3df831f"
+
+[[package]]
+name = "env_home"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c7f84e12ccf0a7ddc17a6c41c93326024c42920d7ee630d04950e6926645c0fe"
 
 [[package]]
 name = "env_logger"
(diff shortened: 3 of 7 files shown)


system-reinstall-bootc/src/podman.rs
```diff
@@ -58,3 +59,42 @@ pub(crate) fn command(image: &str, root_key: &Option<UserKeys>) -> Command {
...
+    }
+
+    tracing::warn!("Podman was not found on this system. It's required to install a bootc image.");
+
+    ensure!(
+        Command::new("dnf").arg("--version").can_execute(),
+        "Please install podman and try again."
+    );
+
+    let mut dnf_install_command = Command::new("dnf");
```
cgwalters, line 75: How about we require the base image to have `/usr/lib/system-reinstall-bootc/install-podman` as an executable?

system-reinstall-bootc/src/podman.rs
```diff
@@ -58,3 +59,42 @@ pub(crate) fn command(image: &str, root_key: &Option<UserKeys>) -> Command {
...
+    dnf_install_command
+        .arg("install")
+        // TODO: Is this a bad idea? It answers yes to EVERYTHING. Maybe we should let the
+        // user interact with dnf manually instead... I wish dnf gave us more granular
+        // control over what we can automatically answer yes to.
+        .arg("-y")
+        .arg("podman");
+
+    if prompt::ask_yes_no(
+        format!("Would you like to install podman using dnf? ({dnf_install_command:?})").as_str(),
```
cgwalters, line 85: I'm not opposed to asking, but on the other hand, why would the user doing this answer no?

Combining the above two things I think we could unconditionally just invoke `/usr/lib/system-reinstall-bootc/install-podman`.

Or maybe I guess we allow the `install-podman` script to be interactive, we pass it stdin, and don't use `-y` for dnf by default there.

But I think for our use case here we can just install by default.

utils/src/command.rs
```diff
@@ -33,6 +33,11 @@ pub trait CommandRunExt {
     /// Execute the child process, parsing its stdout as JSON. This uses `run` internally
     /// and will return an error if the child process exits abnormally.
     fn run_and_parse_json<T: serde::de::DeserializeOwned>(&mut self) -> Result<T>;
+
+    /// Check if the command can be executed by running it and ignoring its output. Returns `true`
+    /// if the command exists, `false` otherwise. The exit code is ignored. Only run this with
+    /// commands that have no undesired side effects, e.g. `dnf --version`.
+    fn can_execute(&mut self) -> bool;
```
cgwalters, line 40: I'd prefer searching `$PATH` and looking for the executable vs running it as a no-op.

cgwalters (approved):
I'm OK with this as is and my comments can be followups if preferred

system-reinstall-bootc/src/podman.rs
omertuc: What's `install-podman`?

system-reinstall-bootc/src/podman.rs
omertuc: >why would the user doing this answer no?

IDK, seems polite to ask before installing a package

system-reinstall-bootc/src/podman.rs
cgwalters: It's an entrypoint script we make up. Think of it like a plugin. Its actual contents would just be:

```
#!/bin/bash
exec dnf install podman
```

But it allows bootc to not reference dnf, and makes it a lot easier for people making custom OSes with bootc to do so without modifying the bootc codebase.

utils/src/command.rs
omertuc: If we do that we won't have stuff like:

- Checking that it's executable
- Checking that we have permissions
- Checking that the architecture is right
- Checking that we have all the shared libraries

Also searching the $PATH sounds hard? From quick googling it looks like our best bet is the `which` crate

(I don't mind doing it, just want to understand the motivation/benefits)

system-reinstall-bootc/src/podman.rs
cgwalters: (And to be clear we'd then do a patch for e.g. https://gitlab.com/fedora/bootc/base-images to include that script)

utils/src/command.rs
cgwalters: At a quick glance the `which` crate looks sane, and [its reverse dependencies](https://crates.io/crates/which/reverse_dependencies) include some important crates.

> Checking that it's executable

This one is trivial to check.

> Checking that we have permissions

What permissions are there other than the executable bit? OK if one starts to get into real details there are e.g. ACLs but I don't think we need to care about that

> Checking that the architecture is right
> Checking that we have all the shared libraries

If one of these is wrong the system is just broken and it's not our fault; I mean how would we deal with a `/usr/bin/podman` with the wrong architecture? It's not likely a `dnf install` would fix that if it somehow happened anyways...

system-reinstall-bootc/src/podman.rs
omertuc: Sound reasonable. 

For bootc base-images, it'll be good to cover when someone wants to do system-reinstall-bootc on an already bootc system (see #1100)

But for OS's like RHEL someone else would have to include it

Until we can be sure it's included everywhere, I suggest the following flow for this PR:

image

With the yellow part deprecated until we can be confident install-podman is widely available

cgwalters: Oh duh sorry of course yes I am completely wrong, we can't put it in the base image since the goal is of course to start not from our base image.

I guess if we can just make *some* nod to other package systems somehow in this project...like a minimal `mod packagesystem` that in theory other people could change at build time?

Another alternative is to require the *packaging* for `system-reinstall-bootc` to include this script...the more I think about the more I like that one. IOW like this

```
diff --git a/contrib/packaging/bootc.spec b/contrib/packaging/bootc.spec
index 4af35a17..f17ef758 100644
--- a/contrib/packaging/bootc.spec
+++ b/contrib/packaging/bootc.spec
@@ -102,6 +102,10 @@ sed -i -e '/https:\/\//d' cargo-vendor.txt
 %if %{with ostree_ext}
 make install-ostree-hooks DESTDIR=%{?buildroot}
 %endif
+cat >%{?buildroot}/usr/lib/system-reinstall-bootc/install-dnf <<EOF
+#!/bin/bash
+exec dnf install podman
+EOF
 
 %if %{with check}
 %check
```

omertuc: Alright, so do away with the yellow part, modify the spec file, and put the burden of writing/maintaining `install-podman` (I assume `install-dnf` was a typo) generally on package maintainers

Can you confirm I understood correctly?

cgwalters: yeah

omertuc:
Settled on: ensure_podman_installed_flowchart

omertuc:
If we skip an internal prompt of "Would you like to install podman using your package manager?", the experience is a bit jarring:


https://github.com/user-attachments/assets/5b0a514a-a8b3-4c20-aa4b-f6576e7d1605

What are your thoughts?

cgwalters:
> If we skip an internal prompt of "Would you like to install podman using your package manager?", the experience is a bit jarring:

Agree it is.

But I'm uncertain if we really need to ask them to do this. We already warn them the system will be reinstalled right? Under what scenario are they going to say "no" to installing podman?

It seems to me we after that after the "system will be reinstalled" prompt we could just run this.

omertuc:
> > If we skip an internal prompt of "Would you like to install podman using your package manager?", the experience is a bit jarring:
> 
> Agree it is.
> 
> But I'm uncertain if we really need to ask them to do this. We already warn them the system will be reinstalled right? Under what scenario are they going to say "no" to installing podman?
> 
> So I'm thinking after that after the "system will be reinstalled" prompt we could just run this.

Following your comment here is the latest experience:


https://github.com/user-attachments/assets/55862392-2cf4-4c88-9a30-7f614d14e163

cgwalters (approved):
The demos and flowcharts are *really* helpful for review BTW, nice work!

cgwalters:
CI failure is unrelated, see https://github.com/containers/bootc/pull/1109
Forcing merge over that

PR #1106, opened 2025-02-11
bootloader: Always pass full path to bootupd

cgwalters:
This will depend on https://github.com/coreos/bootupd/pull/839

And cleans up our bootloader code; move the bootloader partition definitions back to `install/baseline`. Conceptually now it's just the `to-disk` flow and bootupd that
know about bootloader partition GUIDs which is how I think it should be - the bootc core is agnostic to bootloaders.

### DIFF
--- a/lib/src/bootloader.rs
+++ b/lib/src/bootloader.rs
@@ -1,5 +1,5 @@
 use anyhow::{anyhow, bail, Context, Result};
-use camino::{Utf8Path, Utf8PathBuf};
+use camino::Utf8Path;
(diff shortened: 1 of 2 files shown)


cgwalters:
This makes us depend on bootupd 0.2.27 on ppc64le, which is in fedora rawhide, c10s and c9s. I queued a build for f42. I may also queue a build of it for f41, or we can probably safely not care about fedora 41 ppc64le.

ckyrouac (approved):

PR #1107, opened 2025-02-11
lints: More tests && add --skip

cgwalters:
lints; Add unit test checking pass/fail counts

We have a bug here that this will catch.

---

lints: Actually skip skipped lints

Oops.

---

lint: Add --skip

This allows people to explicitly skip individual lints; especially
useful for warnings. But maybe people have a legitimate reason
to skip ones we're calling fatal too.

Prep especially for supporting `--fix --skip`.

---


jeckersb (approved):
:+1:

PR #1108, opened 2025-02-12
A few misc prep patches

cgwalters:
utils: Fix compliation standalone

We use `tokio::test` so need the macros feature.

---

utils: Add a method to split an iterator

---

provision-derived: Clean lots more stuff

Yeah, we're going to need a `dnf clean all --everything`...

---

Update cap-std-ext, use new open_dir_noxdev API

I moved the code there; I plan to use open_dir_noxdev in
the tmpfiles code too which can't depend on lib/util.

---

### DIFF
--- a/utils/src/lib.rs
+++ b/utils/src/lib.rs
@@ -6,3 +6,97 @@ mod command;
 mod tracing_util;
 pub use command::*;
 pub use tracing_util::*;
+
+/// Given an iterator that's clonable, split it into two iterators
+/// at a given maximum number of elements.
+pub fn iterator_split<I>(
+    it: I,
+    max: usize,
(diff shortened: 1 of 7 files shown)


utils/src/lib.rs
```diff
@@ -6,3 +6,97 @@ mod command;
...
+    let mut it = it.peekable();
+    if it.peek().is_some() {
+        Some((it.take(max), rest.skip(max).count()))
+    } else {
+        None
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
jeckersb, line 43: :heart:

jeckersb (approved):
lgtm

cgwalters:
Merging over red because these basically can't¹ break the e2es and it will avoid a retest cycle

¹ I mean...I'd be amazed

PR #1109, opened 2025-02-12
tests/pushpull: Work around podman/skopeo interaction bug

cgwalters:
See https://github.com/containers/bootc/pull/1101#issuecomment-2653862974 Basically things break unless "podman" initializes the c/storage instance right now.

### DIFF
--- a/tests/booted/test-image-pushpull-upgrade.nu
+++ b/tests/booted/test-image-pushpull-upgrade.nu
@@ -31,6 +31,10 @@ def initial_build [] {
     let td = mktemp -d
     cd $td
 
+    # Work around https://github.com/containers/bootc/pull/1101#issuecomment-2653862974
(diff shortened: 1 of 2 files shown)


jmarrero (approved):
lgtm

cgwalters:
Ah ok there were two different tests whose first invocation was `bootc image copy-to-storage`...hmmm...actually, it's tempting to work around this in bootc instead of in our tests

cgwalters:
Cool this works. Looks like copr/s390x is out to lunch though.

cgwalters:
> Ah ok there were two different tests whose first invocation was bootc image copy-to-storage...hmmm...actually, it's tempting to work around this in bootc instead of in our tests

Yeah I thought about this more and I think we should indeed work around this directly in bootc so I did https://github.com/containers/bootc/pull/1110 which reverts this change to the tests.

PR #1110, opened 2025-02-12
storage: Ensure global c/storage is initialized via podman

cgwalters:
Two of our tests (and an unknown set of users may) run
`bootc image copy-to-storage` which happens to invoke
skopeo to target /var/lib/containers/storage. If this
happens to be the *first* thing to write to that
c/storage instance it triggers a bug around the podman
image network setup.

### DIFF
--- a/lib/src/image.rs
+++ b/lib/src/image.rs
@@ -14,6 +14,7 @@ use serde::Serialize;
 use crate::{
     boundimage::query_bound_images,
     cli::{ImageListFormat, ImageListType},
+    imgstorage::ensure_floating_c_storage_initialized,
 };
 
 /// The name of the image we push to containers-storage if nothing is specified.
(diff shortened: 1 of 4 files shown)


jmarrero (approved):
lgtm

cgwalters:
Just circling back to this as I referenced it in another issue; I think the best fix here is actually for podman to ship systemd-tmpfiles config for `/var/lib/containers/storage/`. Actually there is already 

```
> grep lib/containers/storage /usr/lib/tmpfiles.d/podman.conf
D! /var/lib/containers/storage/tmp 0700 root root
```

But I'd say we should have also
```
d /var/lib/containers/storage 0700 - -
f /var/lib/containers/storage/defaultNetworkBackend 0644 - - - netavark
```
or so.

PR #1119, opened 2025-02-17
reinstall: Fix the --user option in podman command

omertuc:
See commit messages

### DIFF
--- a/system-reinstall-bootc/src/podman.rs
+++ b/system-reinstall-bootc/src/podman.rs
@@ -16,7 +16,7 @@ pub(crate) fn command(image: &str, root_key: &Option<UserKeys>) -> Command {
         // The container needs to access the host's PID namespace to mount host directories
         "--pid=host",
         // Set the UID/GID to root overwriting any possible USER directive in the Containerfile
-        "--user root:root",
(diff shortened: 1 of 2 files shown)


system-reinstall-bootc/src/podman.rs
```diff
@@ -16,7 +16,8 @@ pub(crate) fn command(image: &str, root_key: &Option<UserKeys>) -> Command {
         // The container needs to access the host's PID namespace to mount host directories
         "--pid=host",
         // Set the UID/GID to root overwriting any possible USER directive in the Containerfile
-        "--user root:root",
+        "--user",
+        "root:root",
```
germag, line 20: you can use `--user=root:root` if you want (I forgot it)

germag: I cannot approve or lgtm in this repo, but LGTM :)

omertuc: > you can use `--user=root:root` if you want (I forgot it)

Yeah that's definitely better for consistency with the other options used

ckyrouac (approved):

PR #1124, opened 2025-02-18
specfile: fix system-reinstall-bootc cargo build invocation

jeckersb:
Previously this was just running a "normal" `cargo build` a second
time, and overwriting the rhsm feature flags in the main bootc binary

Resolves: BIFROST-658

### DIFF
--- a/contrib/packaging/bootc.spec
+++ b/contrib/packaging/bootc.spec
@@ -92,7 +92,17 @@ rm vendor-config.toml
 # Build the system reinstallation CLI binary
 %global cargo_args -p system-reinstall-bootc
 export SYSTEM_REINSTALL_BOOTC_INSTALL_PODMAN_PATH=%{system_reinstall_bootc_install_podman_path}
-%cargo_build
+%if 0%{?fedora} || 0%{?rhel} >= 10
+    # In cargo-rpm-macros, the cargo_build macro does flag processing,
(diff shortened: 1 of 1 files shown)


ckyrouac (approved):
looks sane to me, I'm not sure why the rpm builds are failing in CI, but that appears to be unrelated.

jeckersb:
> looks sane to me, I'm not sure why the rpm builds are failing in CI, but that appears to be unrelated.

I think it's something to do with the vendoring stuff that @cgwalters did in https://github.com/containers/bootc/pull/1068, I need to dig into the job to see how exactly it's handling that when running out of git

jeckersb:
I want to make sure that the el9 builds work before merging this, since I'm not 100% positive that the `%cargo_build` macro is going to happily accept the flags as I've done it here.  I *could* just go manually test it, but I'll fix CI and make sure it works here.

jeckersb:
Looks like #1125 is enough to fix CI, once that lands I'll rebase this and let CI sanity check things.

jeckersb:
> I want to make sure that the el9 builds work before merging this, since I'm not 100% positive that the `%cargo_build` macro is going to happily accept the flags as I've done it here. I _could_ just go manually test it, but I'll fix CI and make sure it works here.

Yep the el9 macro is broken, trying to figure out the right incantation to correct for that...

PR #1130, opened 2025-02-20
WIP/DNM: Use 2024 edition

jeckersb:
We don't actually want to do this right now, but there wasn't much to it so I'll just post it for future-us when we actually do migrate

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -977,7 +977,10 @@ pub fn global_init() -> Result<()> {
         // to invoke this as early as possible. (In practice, that's just the cli's `main.rs`)
         // xref https://internals.rust-lang.org/t/synchronized-ffi-access-to-posix-environment-variable-functions/15475
         // TODO: Audit that the environment access only happens in single-threaded code.
-        unsafe { std::env::set_var("HOME", "/root") };
+        #[allow(unsafe_code)]
+        unsafe {
+            std::env::set_var("HOME", "/root")
+        };
     }
     Ok(())
 }
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -976,7 +976,8 @@ pub fn global_init() -> Result<()> {
         // Setting the environment is thread-unsafe, but we ask calling code
         // to invoke this as early as possible. (In practice, that's just the cli's `main.rs`)
         // xref https://internals.rust-lang.org/t/synchronized-ffi-access-to-posix-environment-variable-functions/15475
-        std::env::set_var("HOME", "/root");
+        // TODO: Audit that the environment access only happens in single-threaded code.
+        unsafe { std::env::set_var("HOME", "/root") };
     }
     Ok(())
 }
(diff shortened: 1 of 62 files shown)


lib/src/kargs.rs
```diff
@@ -229,7 +229,8 @@ match-architectures = ["x86_64", "aarch64"]
         .to_string();
         let parsed_kargs = parse_kargs_toml(&file_content, sys_arch).unwrap();
         assert_eq!(parsed_kargs, ["console=tty0", "nosmt"]);
-        std::env::set_var("ARCH", "aarch64");
+        // TODO: Audit that the environment access only happens in single-threaded code.
```
cgwalters, line 232: It definitely doesn't, rust runs unit tests in multiple threads by default. We need to stop setting this environment variable in the tests at all. At a quick look it's not clear we need this line of code, shouldn't we be just doing `let sys_arch = "aarch64"`?

lib/src/cli.rs
```diff
@@ -1012,7 +1013,7 @@ impl Opt {
         I::Item: Into<OsString> + Clone,
     {
         let mut args = args.into_iter();
-        let first = if let Some(first) = args.next() {
+        let first = match args.next() { Some(first) => {
```
cgwalters: Here and I *think* all the other changes this is `cargo fix` being conservative with the drop ordering change. IMO `if let` is easier to read than `match` and so we should not apply these changes.

ostree-ext/src/chunking.rs
```diff
@@ -136,35 +136,35 @@ struct Generation {
     dirmeta_found: BTreeSet<RcStr>,
 }
 
-fn push_dirmeta(repo: &ostree::Repo, gen: &mut Generation, checksum: &str) -> Result<()> {
-    if gen.dirtree_found.contains(checksum) {
+fn push_dirmeta(repo: &ostree::Repo, r#gen: &mut Generation, checksum: &str) -> Result<()> {
```
cgwalters, line 139: This is fine as an automated fix, but I think it'd be nicer to rename this to `generation` instead

cgwalters (changes requested):

ostree-ext/src/chunking.rs
jeckersb: 100%, I did exactly that manually before I realized that `cargo fix` would do its own thing.

lib/src/cli.rs
jeckersb: For reference since I had to go find it - https://doc.rust-lang.org/nightly/edition-guide/rust-2024/temporary-if-let-scope.html

lib/src/kargs.rs
jeckersb: Yeah it doesn't even reference the environment variable anywhere.  No idea what that was ever about.  I'll go open a separate PR against main to fix just that.

jeckersb: > Yeah it doesn't even reference the environment variable anywhere. No idea what that was ever about. I'll go open a separate PR against main to fix just that.

#1132

cgwalters:
Per discussion here I think we're basically only down to the `gen` keyword bit. The drop order thing is just noise we don't need.

I think we should probably have tracking issues for our PRs in general; I filed https://github.com/bootc-dev/bootc/issues/1414 for that.

Closing this one to get it off the board until we pick it back up.

PR #1131, opened 2025-02-20
reinstall: Stop waiting for newline after prompt is answered

ckyrouac:
When I first used the script, I hit `y` expecting my answer to be accepted. When nothing happened, I hit `y` again and was presented with the same prompt. It took me far too long (I went as far as trying a different shell and terminal) to realize that hitting `y` only toggled the selection to capital `Y`, I had to also hit "enter". This is the first time I've used a script with this behavior. Maybe I'm in the minority but given that this is supposed to be an intuitive tool, should we toggle this flag?

### DIFF
--- a/system-reinstall-bootc/src/prompt.rs
+++ b/system-reinstall-bootc/src/prompt.rs
@@ -57,7 +57,7 @@ pub(crate) fn ask_yes_no(prompt: &str, default: bool) -> Result<bool> {
     dialoguer::Confirm::new()
         .with_prompt(prompt)
         .default(default)
-        .wait_for_newline(true)
+        .wait_for_newline(false)
         .interact()
         .context("prompting")
 }


jeckersb:
(Disclaimer, I haven't actually tried running `system-reinstall-bootc` yet so maybe I'm missing context...)

My initial reaction is that the newline is expected.  The canonical example for me is `dnf install cowsay` and you get the `Is this ok [y/N]:` prompt.  At that point I'm expecting to type `y<enter>`, not just `y`.

ckyrouac:
ah right, of course. I think what threw me off is the way it's presented. Hitting `y` when `y` is the default appears to do nothing in system-reinstall-bootc. 

There's feedback when using dnf, e.g.:

```
Is this ok [y/N]: yyyy
```

This happens when hitting `y` multiple times with system-reinstall-bootc:

```
❮ sudo ./target/release/system-reinstall-bootc quay.io/fedora/fedora-bootc:41
[sudo] password for chris:
None of the users on this system found have authorized SSH keys, if your image doesn't use cloud-init or other mean
None of the users on this system found have authorized SSH keys, if your image doesn't use cloud-init or other mean
None of the users on this system found have authorized SSH keys, if your image doesn't use cloud-init or other mean
None of the users on this system found have authorized SSH keys, if your image doesn't use cloud-init or other means to set up users, you may not be able to log in after reinstalling. Do you want to continue? [Y/n]

```

Looks like the prompt is (partially) repeated only when there is not enough width in the terminal. Otherwise, hitting `y` multiple times does nothing.

I might have been confused due to lack of sleep more than anything. Feel free to close if you think it's fine as-is.

jeckersb:
I guess I should actually try to use the tool on a VM to see exactly what you're talking about :smile:.  But from your explanation the current behavior doesn't look sane to me.  I'll play with it a bit later this afternoon.

omertuc:
>Looks like the prompt is (partially) repeated only when there is not enough width in the terminal. Otherwise, hitting y multiple times does nothing.

The lines repeating is a dialoguer bug

But it doing nothing is a feature (the selection is communicated through the capitalization of y/n, if y is already capitalized, it means its already selected, not that it's the default)

omertuc:
Opened https://github.com/console-rs/dialoguer/pull/323

cgwalters:
We could mitigate the console-rs bug by line wrapping our prompt, right?

If progress there is slow on merging patches I'm also fine to vendor dialoguer in our codebase FWIW. (Or probably simpler, point to your fork via `git` dependencies, or (engineering this slightly more) another option is to make a branch on *this* repository which is actually not bootc, but is just forked dependencies)

As far as the mechanics of not waiting for a newline...I don't have a really strong opinion honestly, but weakly agree with John.

ckyrouac:
> As far as the mechanics of not waiting for a newline...I don't have a really strong opinion honestly, but weakly agree with John.

Agreed, waiting for newline seems standard. The thing that threw me off is the input is not displayed in the terminal. So, pressing "y" on a selection where the default is "y" results in no feedback. Not a big deal, I'm just not used to that behavior. Closing this.

PR #1133, opened 2025-02-20
reinstall: Support legacy images in system-reinstall-bootc

ckyrouac:
Add the "legacy" parameters to the podman run invocation because there are existing bootc images that need these parameters when running install to-existing-root.

---

I tested this with the fedora:40 image which has bootc v1.1.2.


cgwalters (approved):

cgwalters:
> The job was not started because recent account payments have failed or your spending limit needs to be increased. Please check the 'Billing & plans' section in your settings.

Ooh. Fun. So this one definitely relates to the CNCF transition and getting us out of the barely-managed github.com/containers org...

PR #1137, opened 2025-02-22
sysusers: Handle named group references

cgwalters:
The libvirt package has this entry with a numeric user but an explicit named group reference:

`u qemu 107:qemu "qemu user" - -`

Fix the parser to handle this.


jmarrero (approved):
lgtm

PR #1146, opened 2025-02-26
Smarter ssh authorized key search for system-reinstall-bootc

ckyrouac:
Pushing this branch early for visibility. The first commit just parses the `sshd -T` output into a struct.

### DIFF
--- a/system-reinstall-bootc/Cargo.toml
+++ b/system-reinstall-bootc/Cargo.toml
@@ -24,6 +24,7 @@ rustix = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
 serde_yaml = "0.9.22"
+tempfile = "3.10.1"
 tracing = { workspace = true }
 uzers = "0.12.1"
 which = "7.0.2"
--- a/system-reinstall-bootc/src/main.rs
+++ b/system-reinstall-bootc/src/main.rs
@@ -20,13 +20,17 @@ fn run() -> Result<()> {
 
     let config = config::ReinstallConfig::load().context("loading config")?;
 
-    let root_key = &prompt::get_root_key()?;
+    let ssh_key_file = tempfile::NamedTempFile::new()?;
+    let ssh_key_file_path = ssh_key_file
+        .path()
+        .to_str()
+        .ok_or_else(|| anyhow::anyhow!("unable to create authorized_key temp file"))?;
 
-    if root_key.is_none() {
-        return Ok(());
-    }
+    tracing::trace!("ssh_key_file_path: {}", ssh_key_file_path);
 
-    let mut reinstall_podman_command = podman::command(&config.bootc_image, root_key);
+    prompt::get_ssh_keys(ssh_key_file_path)?;
+
+    let mut reinstall_podman_command = podman::command(&config.bootc_image, ssh_key_file_path);
 
     println!();
 
--- a/system-reinstall-bootc/src/prompt.rs
+++ b/system-reinstall-bootc/src/prompt.rs
@@ -1,7 +1,4 @@
-use crate::{
-    prompt,
-    users::{get_all_users_keys, UserKeys},
-};
+use crate::{prompt, users::get_all_users_keys};
 use anyhow::{ensure, Context, Result};
 
 const NO_SSH_PROMPT: &str = "None of the users on this system found have authorized SSH keys, \
@@ -10,7 +7,9 @@ const NO_SSH_PROMPT: &str = "None of the users on this system found have authori
 
 fn prompt_single_user(user: &crate::users::UserKeys) -> Result<Vec<&crate::users::UserKeys>> {
     let prompt = format!(
-        "Found only one user ({}) with {} SSH authorized keys. Would you like to import it and its keys to the system?",
+        "Found only one user ({}) with {} SSH authorized keys.\n\
+            Would you like to import its SSH authorized keys\n\
+            into the root user on the new bootc system?",
         user.user,
         user.num_keys(),
     );
@@ -25,7 +24,10 @@ fn prompt_user_selection(
 
     // TODO: Handle https://github.com/console-rs/dialoguer/issues/77
     let selected_user_indices: Vec<usize> = dialoguer::MultiSelect::new()
-        .with_prompt("Select the users you want to install in the system (along with their authorized SSH keys)")
+        .with_prompt(
+            "Select which user's SSH authorized keys you want to\n\
+            import into the root user of the new bootc system",
+        )
         .items(&keys)
         .interact()?;
 
@@ -62,10 +64,14 @@ pub(crate) fn ask_yes_no(prompt: &str, default: bool) -> Result<bool> {
         .context("prompting")
 }
 
-/// For now we only support the root user. This function returns the root user's SSH
-/// authorized_keys. In the future, when bootc supports multiple users, this function will need to
-/// be updated to return the SSH authorized_keys for all the users selected by the user.
-pub(crate) fn get_root_key() -> Result<Option<UserKeys>> {
+/// Gather authorized keys for all user's of the host system
+/// prompt the user to select which users's keys will be imported
+/// into the target system's root user's authorized_keys file
+///
+/// The keys are stored in a temporary file which is passed to
+/// the podman run invocation to be used by
+/// `bootc install to-existing-root --root-ssh-authorized-keys`
+pub(crate) fn get_ssh_keys(temp_key_file_path: &str) -> Result<()> {
     let users = get_all_users_keys()?;
     if users.is_empty() {
         ensure!(
@@ -73,7 +79,7 @@ pub(crate) fn get_root_key() -> Result<Option<UserKeys>> {
             "cancelled by user"
         );
 
-        return Ok(None);
+        return Ok(());
     }
 
     let selected_users = if users.len() == 1 {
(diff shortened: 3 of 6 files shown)


system-reinstall-bootc/src/users.rs
```diff
@@ -102,11 +102,58 @@ impl Display for UserKeys {
...
+            authorized_keys_files: Vec::new(),
+            authorized_keys_command: String::new(),
+            authorized_keys_command_user: String::new(),
+        };
+
+        for line in sshd_output.lines() {
+            let mut kv = line.splitn(2, char::is_whitespace);
+            if let (Some(key), Some(value)) = (kv.next(), kv.next()) {
+                match key.trim() {
+                    "authorizedkeysfile" => {
```
cgwalters, line 130: I wonder if it's maybe easier to just scrape the whole config into a `HashMap<&str, &str>` (note we don't need to copy the data!) and then to pick out specific values we want after that.

Something like this:

```
let config = sshd_output.lines()
        .filter_map(|l| l.split_once(' '))
        .collect::<BTreeMap<_, _>>();
let authorized_keys_file = config.get("authorizedkeysfile").unwrap_or("");
...
```

or so?

system-reinstall-bootc/src/prompt.rs
```diff
@@ -1,16 +1,16 @@
...
+use std::io::Write;
+
 const NO_SSH_PROMPT: &str = "None of the users on this system found have authorized SSH keys, \
     if your image doesn't use cloud-init or other means to set up users, \
     you may not be able to log in after reinstalling. Do you want to continue?";
 
 fn prompt_single_user(user: &crate::users::UserKeys) -> Result<Vec<&crate::users::UserKeys>> {
     let prompt = format!(
-        "Found only one user ({}) with {} SSH authorized keys. Would you like to import it and its keys to the system?",
+        "Found only one user ({}) with {} SSH authorized keys. Would you like to import it's SSH authorized keys into the root user on the new bootc system?",
```
cgwalters, line 13: Minor nit, "its" here is correct (not "it's") as it is a possessive. While we're here would it also look nicer to line wrap?

system-reinstall-bootc/src/main.rs
```diff
@@ -10,6 +10,7 @@ mod prompt;
 pub(crate) mod users;
 
 const ROOT_KEY_MOUNT_POINT: &str = "/bootc_authorized_ssh_keys/root";
+const SSH_KEY_TEMP_FILE: &str = "/tmp/.system-reinstall-bootc-ssh-keys";
```
cgwalters, line 13: [Predicable filenames in /tmp](https://cwe.mitre.org/data/definitions/377.html) can lead to security issues. Do we really need a hardcoded path here? Can't we just allocate a `tempfile::NamedTempFile()`? 

If we do need a hardcoded path, one mitigation is to use `/run/bootc-ssh-keys` or so, as that's a directory only root should be able to write to.

system-reinstall-bootc/src/prompt.rs
```diff
@@ -62,18 +62,28 @@ pub(crate) fn ask_yes_no(prompt: &str, default: bool) -> Result<bool> {
         .context("prompting")
 }
 
-/// For now we only support the root user. This function returns the root user's SSH
-/// authorized_keys. In the future, when bootc supports multiple users, this function will need to
-/// be updated to return the SSH authorized_keys for all the users selected by the user.
-pub(crate) fn get_root_key() -> Result<Option<UserKeys>> {
+fn write_ssh_keys(keys: String) -> Result<()> {
+    let mut file = File::create(super::SSH_KEY_TEMP_FILE)?;
+    file.write_all(keys.as_bytes())?;
```
cgwalters, line 67: This can just be
`std::fs::write(super::SSH_KEY_TEMP_FILE, keys.as_bytes())`

system-reinstall-bootc/src/prompt.rs
```diff
@@ -82,12 +92,13 @@ pub(crate) fn get_root_key() -> Result<Option<UserKeys>> {
...
 
-    ensure!(
-        selected_users.iter().all(|x| x.user == "root"),
-        "Only importing the root user keys is supported for now"
-    );
+    let keys = selected_users
+        .into_iter()
+        .map(|user_key| user_key.authorized_keys.as_str())
+        .collect::<Vec<&str>>()
+        .join("\n");
```
cgwalters, line 99: I'm fine with this for now but I do think we should link to https://lib.rs/crates/openssh-keys at some point to have a more proper validator/parser/writer

system-reinstall-bootc/src/users.rs
```diff
@@ -102,61 +103,134 @@ impl Display for UserKeys {
...
+    let home_dir = user.home_dir();
+    let mut user_authorized_keys = String::new();
+
+    for keyfile in keyfiles {
+        let user_authorized_keys_path = home_dir.join(keyfile);
+
+        if !user_authorized_keys_path.exists() {
+            tracing::debug!(
+                "Skipping authorized key file {} for user {} because it doesn't exist",
+                user_authorized_keys_path.to_string_lossy(),
```
cgwalters, line 147: See the helper I added in https://github.com/containers/bootc/blob/22c743fef7990f22b3b449742b35290a6567a020/utils/src/path.rs#L5 which I think is nicer.

(I need to do some digging and find out if that reinvented something already on crates.io, I mean it must have, but...)

system-reinstall-bootc/src/users.rs
```diff
@@ -102,61 +103,134 @@ impl Display for UserKeys {
...
+                user.name().to_string_lossy()
+            );
+            continue;
+        }
+
+        // Safety: The UID should be valid because we got it from uzers
+        #[allow(unsafe_code)]
+        let user_uid = unsafe { Uid::from_raw(user.uid()) };
```
cgwalters, line 155: (This is OK for now but I think per previous discussion we actually don't need to change uid for this, we can just use cap-std)

system-reinstall-bootc/src/users.rs
```diff
@@ -102,61 +103,134 @@ impl Display for UserKeys {
...
+
+        // Change the effective uid for this scope, to avoid accidentally reading files we
+        // shouldn't through symlinks
+        let _uid_change = UidChange::new(user_uid)?;
+
+        let key = std::fs::read_to_string(&user_authorized_keys_path)
+            .context("Failed to read user's authorized keys")?;
+        user_authorized_keys.push_str(key.as_str());
```
cgwalters, line 163: Hmm wait and this will work with multiple keys because we still have the newline?

This goes back to my comment that it'd seem cleaner to parse them and have a `Vec<AuthorizedKeys>` or so?

system-reinstall-bootc/src/users.rs
```diff
@@ -102,61 +103,134 @@ impl Display for UserKeys {
...
+    }
+
+    Ok(user_authorized_keys)
+}
+
+fn get_keys_from_command(command: &str, command_user: &str) -> Result<String> {
```
cgwalters, line 169: Looks OK but it's unfortunate we can't unit test this. Going all the way to a tmt test would feel quite heavyweight though.

I bet though we could (and really should) introduce a test framework that runs as a default-unprivileged container (i.e. uid0, where we have full capability to invoke `setuid`).

Although it looks like 
```
$ podman run --rm -ti localhost/fedora-bootc bash
bash-5.2# sshd -T
sshd: no hostkeys available -- exiting.
```

So we'd have to do a bit of work to set that up

system-reinstall-bootc/src/users.rs
```diff
@@ -183,4 +257,22 @@ mod test {
...
+        let second = parts.next().unwrap_or("");
+        Some((first, second))
+    }
+
+    #[test]
+    pub(crate) fn test_me() {
+        println!("ahhh");
+        let input = "foo   bar  bar2 bar3";
+        if let Some((first, rest)) = split_on_first_whitespace(input) {
+            println!("First part: '{}'", first);
```
cgwalters, line 273: For stuff like this I use `dbg!` - it's actually quicker to type *and* we deny shipping it by default in `make validate-rust` so that one can't accidentally commit debug prints (I recently did this in bootupd :see_no_evil: https://github.com/coreos/bootupd/pull/845 )

system-reinstall-bootc/src/prompt.rs
ckyrouac: filed https://github.com/containers/bootc/issues/1148 as a followup

system-reinstall-bootc/src/users.rs
ckyrouac: it works on my system because all my key files have a newline at the end. Added an explicit newline here just in case.

ckyrouac:
This is ready for a final review

system-reinstall-bootc/Cargo.toml
```diff
@@ -24,6 +24,7 @@ rustix = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
 serde_yaml = "0.9.22"
+tempfile = "3.10.1"
```
cgwalters, line 27: nit-for-followup: Let's use `{ workspace = true }`

system-reinstall-bootc/src/prompt.rs
```diff
@@ -10,7 +7,9 @@ const NO_SSH_PROMPT: &str = "None of the users on this system found have authori
 
 fn prompt_single_user(user: &crate::users::UserKeys) -> Result<Vec<&crate::users::UserKeys>> {
     let prompt = format!(
-        "Found only one user ({}) with {} SSH authorized keys. Would you like to import it and its keys to the system?",
+        "Found only one user ({}) with {} SSH authorized keys.\n\
```
cgwalters, line 10: suggestion-for-followup: This would be more readable with `indoc!`

cgwalters (approved):
LGTM, let's ship it - I had two totally minor things that can just be rolled into a subsequent PR.

I'll try to use the prefix "nit-for-followup:" to mean "next time you (or someone else) does a PR let's consider fixing this then"

PR #1150, opened 2025-02-27
Reinstall ssh crate

ckyrouac:
This includes the refactor to use the openssh_keys crate and removes the options field from all ssh keys.

### DIFF
--- a/system-reinstall-bootc/src/prompt.rs
+++ b/system-reinstall-bootc/src/prompt.rs
@@ -1,5 +1,6 @@
 use crate::{prompt, users::get_all_users_keys};
 use anyhow::{ensure, Context, Result};
+use openssh_keys::PublicKey;
 
 const NO_SSH_PROMPT: &str = "None of the users on this system found have authorized SSH keys, \
     if your image doesn't use cloud-init or other means to set up users, \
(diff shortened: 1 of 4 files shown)


cgwalters (approved):
LGTM overall!

system-reinstall-bootc/src/prompt.rs
```diff
@@ -94,7 +94,11 @@ pub(crate) fn get_ssh_keys(temp_key_file_path: &str) -> Result<()> {
         .flat_map(|user| &user.authorized_keys)
         .collect::<Vec<&PublicKey>>()
         .into_iter()
-        .map(|key| key.to_key_format() + "\n")
+        .map(|key| {
+            let mut key_copy = key.clone();
+            key_copy.options = None;
```
cgwalters, line 99: Also super minor nit, while it's in the commit message I think this change is deserving of a comment too. i.e. just copy-paste your commit message here with a `#`

cgwalters (approved):

cgwalters:
The aarch64 failure is a usual periodic provisioning failure; not sure what happened with the x86_64 one, that's weird. It doesn't look obviously related to this change (and I don't think we could break it )

cgwalters:
Ahh it may be https://status.testing-farm.io/issues/2025-02-27-openstack-unreachable/

ckyrouac:
closed in favor of #1153 which has all this + cleanup

PR #1151, opened 2025-02-27
tmpfiles: Support multiple generations

cgwalters:
Way back in rpm-ostree's original code here, it always assumed that it owned The Build.

But in a container-native flow we may have a chain of builds across an organization and want to generate tmpfiles.d entries multiple times.

When we do this, we can't just blow away the existing `/usr/lib/tmpfiles.d/bootc-generated.conf`.

Teach the tmpfiles code to have "generations" - we pick a new file name like `bootc-generated-0.conf`, `bootc-generated-1.conf` etc.

Also, change the generation code to return a struct with more info about what it did, which will be used in the linting code.


tmpfiles/src/lib.rs
```diff
@@ -18,7 +20,22 @@ use rustix::path::Arg;
...
-const BOOTC_GENERATED: &str = "bootc-autogenerated-var.conf";
+/// The path to the file we use for generation
+const BOOTC_GENERATED_PREFIX: &str = "bootc-autogenerated-var";
+
+/// The number of times we've generated a tmpfiles.d
+#[derive(Debug, Default)]
+struct BootcTmpfilesGeneration(u32);
+
+impl BootcTmpfilesGeneration {
+    fn increment(&self) -> Self {
```
jeckersb, line 31: Pedantic nit (take it or leave it), this should probably either consume `self` then hand it back, or take `&mut self` and do the increment in-place, just to avoid having multiple copies of the generation and accidentally using the wrong one.  It's fine right now since in usage you're always overwriting the old generation with the new instance but changing it would let the type system enforce that.

Worst case, derive `clone` and make it explicit that you intended to have two :smile:

jeckersb (approved):
Turned off auto-merge and approving just to let you consider the nit, if you disagree just merge it as-is!

tmpfiles/src/lib.rs
cgwalters: Good point, will do in a followup

PR #1152, opened 2025-02-27
lints: Add --fix, support with tmpfiles

cgwalters:
This needs some docs and some container tests but yeah

```
$ cat Dockerfile
FROM localhost/bootc
RUN <<EORUN
set -xeuo pipefail
mkdir -p /var/lib/some-test-dir
bootc container lint --fix
EORUN
$ podman build ...
+ bootc container lint --fix
tmpfiles.d: Generated: usr/lib/tmpfiles.d/bootc-autogenerated-var-0.conf with entries: 1
Checks passed: 11
Checks skipped: 1
```

But the big thing I realized here that makes things WAY messier is that actually we need to support *retaining* some of the `/var` subdirs...like `/var/tmp` and `/var/lib/rpm-state` as one wants them in the container too. Tentatively thinking something like

```
cat /usr/lib/bootc/tmpfiles.d/foo.conf
# Temporary directories we ensure are always present in the container too
/var/tmp
```

So what'd be cool about this is that at the end of a build one could switch from

`rm -rf /var/tmp/* /var/lib/{dnf,rhsm} /var/cache/*`

to
```
rm /var/*
bootc container lint --fix
```
And that'd automatically regenerate everything in /var needed.

But, mmm, it may drive us really to have something like `bootc container commit` that is defined to do both of those things?


cgwalters:
OK moved the minor commit to https://github.com/bootc-dev/bootc/pull/1257 and closing this since it's clear it needs to be a separate verb.

PR #1157, opened 2025-02-28
install: Add a generic `install finalize`

cgwalters:
Basically I want to get Anaconda to run this, then we can perform arbitrary fixups on whatever it did between the install and reboot without changing Anaconda's code.

Motivated specifically by https://github.com/containers/bootc/issues/971

This also applies to user `%post` scripts for example; maybe those break the bootloader entries in /boot; we have the opportunity to catch such things here.

Or we may choose to start forcibly relabeling the target `/etc`.


tests-integration/src/install.rs
```diff
@@ -64,6 +65,7 @@ fn find_deployment_root() -> Result<Dir> {
 
 // Hook relatively cheap post-install tests here
 fn generic_post_install_verification() -> Result<()> {
+    let sh = xshell::Shell::new()?;
```
jeckersb, line 68: also unused

jeckersb (approved):
Just lint nits, but we've accumulated a fair number of those recently and we can do all of that cleanup in a followup

cgwalters:
> Just lint nits, but we've accumulated a fair number of those recently and we can do all of that cleanup in a followup

OK yeah, we should definitely be gating merges to git main on at least things like `unused_imports` (and probably *most* things that rustc would warn about by default, but I don't like a blanket "deny all warnings in CI" policy either)

I ended up with https://github.com/containers/bootc/pull/1159

PR #1159, opened 2025-03-01
build: Tweak `make validate-rust`

cgwalters:
This took me some thinking and experimenting. Basically we want:

- Hard deny some warnings (this is covered by the Cargo.toml workspace.lints.rust)
- Gate merging to main in CI on an exact set of warnings we want to forbid, but *without* also using a blanket -Dwarning deny policy because that could break our build when the compiler revs.
- A corollary to the previous: allow developing locally without killing the build just because you have an unused import or some dead code (for example). So we don't want to add `dead_code = deny` into the Cargo.toml.
- Be able to easily reproduce locally what CI is gating on in an efficient way.

We already had `make validate-rust` which was intending to navigate this, but what was missing was the "deny extended set of warnings" so we got code committed to git main which hit `unused_imports`.

Clippy upstream docs recommend the `RUSTFLAGS = -Dwarnings` approach in e.g.
https://doc.rust-lang.org/clippy/continuous_integration/github_actions.html but again I think this is a problem because it can break with updated Rust/clippy versions (unless you pin on those, but that becomes a pain in and of itself).

The problem also with doing `RUSTFLAGS = -Dwarnings` *locally* is it blows out the cargo cache.

So here's the solution I came to: We run `cargo clippy -A clippy:all`, and then deny some specific clippy lints *and* the core Rust warnings we want (`unused_imports` type things) at this stage. The advantage is this doesn't blow out the main Cargo cache, and I can easily reproduce locally exactly what CI would gate on.

Also while we're here, add `make fix-rust` which is a handy way to use the existing `clippy --fix` to locally fix things like unused imports as well as other machine-applicable things that are in e.g. `clippy::suspicious`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,15 +59,22 @@ test-bin-archive: all
 test-tmt:
 	cargo xtask test-tmt
 
-# Checks extra rust things (formatting, a few extra rust warnings, and select clippy lints)
+# This gates CI by default. Note that for clippy, we gate on
+# only the clippy correctness and suspicious lints, plus a select
+# set of default rustc warnings.
+# We intentionally don't gate on this for local builds in cargo.toml
+# because it impedes iteration speed.
+CLIPPY_CONFIG = -A clippy::all -D clippy::correctness -D clippy::suspicious -Dunused_imports -Ddead_code
 validate-rust:
 	cargo fmt -- --check -l
-	cargo check
-	(cd lib && cargo check --no-default-features)
 	cargo test --no-run
-	cargo clippy -- -D clippy::correctness -D clippy::suspicious
+	(cd lib && cargo check --no-default-features)
+	cargo clippy -- $(CLIPPY_CONFIG)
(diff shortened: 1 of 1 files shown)


cgwalters:
Wait that's weird...this auto-merged without an approving review.

cgwalters:
The github status even says 
image

cgwalters:
I filed a support ticket asking this to be diagnosed.

cgwalters:
> I filed a support ticket asking this to be diagnosed.

FTR support got back to me and noted this commit was rolled into
https://github.com/containers/bootc/pull/1157
and when that merged, it noted this exact commit was merged and closed it. This is a known behavior, and makes sense. The obvious thing to do on GH's side though is just add pretty visible note in the PR making clear when this happens.

I messed up and I hadn't meant to roll this commit into that PR, that's why I got confused.

Anyways, all good!

PR #1164, opened 2025-03-04
reinstall: Pull podman image early

ckyrouac:
This splits the `podman pull <image>` and the `podman ... bootc install to-existing` command to prepare for future features (e.g. cloud-init) that will require inspecting the image before constructing the bootc install command.

---

Also adds a prompt before installing podman since it needs to be installed before pulling the image which now happens immediately after running the command, before any other user prompts. Open to other ideas on the UX if this doesn't make sense.


cgwalters:
Looks like it needs a `cargo fmt`

cgwalters (approved):

PR #1167, opened 2025-03-05
store: Really remove empty /etc/resolv.conf and /etc/hostname

cgwalters:
The previous change here was a no-op for two reasons:

- It's actually usr/etc at this point
- We were operating on the wrong rootfs

Fixes: https://github.com/containers/bootc/pull/1096/commits/57bd0dc9835669274696998386a547afb6709ff5

### DIFF
--- a/ostree-ext/src/container/store.rs
+++ b/ostree-ext/src/container/store.rs
@@ -466,7 +466,7 @@ fn timestamp_of_manifest_or_config(
 /// Automatically clean up files that may have been injected by container
 /// builds. xref https://github.com/containers/buildah/issues/4242
 fn cleanup_root(root: &Dir) -> Result<()> {
-    const RUNTIME_INJECTED: &[&str] = &["etc/hostname", "etc/resolv.conf"];
+    const RUNTIME_INJECTED: &[&str] = &["usr/etc/hostname", "usr/etc/resolv.conf"];
     for ent in RUNTIME_INJECTED {
         if let Some(meta) = root.symlink_metadata_optional(ent)? {
             if meta.is_file() && meta.size() == 0 {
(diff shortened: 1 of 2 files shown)


cgwalters:
Hmm the LBI test failure there looks like some kind of networking flake between the controller and target. Let's see if the aarch64 runs green, if so I think we can just merge over red since I still don't know how to retry testing farm jobs other than force pushing, which retriggers everything.

And BTW what I *really* messed up here originally is that stuff under tests/readonly doesn't run unless it starts with `test-` so we were just silently not running that test before :see_no_evil:

jeckersb:
> Hmm the LBI test failure there looks like some kind of networking flake between the controller and target. Let's see if the aarch64 runs green, if so I think we can just merge over red since I still don't know how to retry testing farm jobs other than force pushing, which retriggers everything.
> 
> And BTW what I _really_ messed up here originally is that stuff under tests/readonly doesn't run unless it starts with `test-` so we were just silently not running that test before 🙈

I just retriggered it, you can do it from the github UI

jeckersb (approved):
lgtm, hopefully the re-run x86_64 integration passes and this merges :pray:

p5:
After this commit, systems have lost the empty /etc/hostname which was originally (accidentally) shipped in the image. 
 
This only causes problems if you are running services that depend on that file, like an existing Podman container mounting it (e.g. Distrobox).

Luap99:
This broke podman machine images as well.

The booted podman machine no longer has /etc/resolv.conf in it causing all DNS to fail. We use the fedora coreos image as base and disable systemd-resolved like this: https://github.com/containers/podman-machine-os/blob/6a59cfbc22d1a760aa266ca0b2d8260916fbbbc8/podman-image/Containerfile#L78-L86

When I added it was required to add an empty /etc/resolv.conf as without it dns was not working despite NetworkManager running inside the VM. So at least for podman machine it seems NetworkManager does not create/populate /etc/resolv.conf when it doesn't exist already.

cgwalters:
> This broke podman machine images as well.

Just to close the loop on the above, we had a live chat on this and I added a comment here: https://gist.github.com/Luap99/0eb266e861016e52e27001d2c5b3c70b?permalink_comment_id=5560401#gistcomment-5560401

Basically to disable systemd-resolved from Fedora images it's better to explicitly remove the package to ensure the tmpfiles.d snippet doesn't exist either.

PR #1175, opened 2025-03-06
ostree-ext: Update parser to honor `composefs=verity`

cgwalters:
We have duplicate code to parse this between C and Rust unfortunately; update the Rust side to honor what landed in https://github.com/ostreedev/ostree/pull/3354


cgwalters:
Split this out of https://github.com/containers/bootc/pull/935

jeckersb (approved):
LGTM

PR #1176, opened 2025-03-06
docs: Update generated

jeckersb:


jeckersb:
Hm I guess there's still the problem where the labeler doesn't run until after the ci job has started, so for this doc-only change it of course skipped docs while also doing the full ci suite :roll_eyes:

cgwalters (approved):

PR #1178, opened 2025-03-07
internals: Add new (experimental) `bootc internals fsck`

cgwalters:
Split this out of the fsverity PR.

We obviously want a `fsck` command. This starts by doing just two checks:

- A verification of `etc/resolv.conf`; this tests https://github.com/containers/bootc/pull/1167/commits/98995f662b7207858630f72236c9b850e6f07154
- Just run `ostree fsck`

But obvious things we should be adding here are:

- Verifying kargs
- Verifying LBIs

etc.

### DIFF
--- a/lib/src/fsck.rs
+++ b/lib/src/fsck.rs
@@ -1,0 +1,155 @@
+//! # Perform consistency checking.
+//!
+//! This is an internal module, backing the experimental `bootc internals fsck`
+//! command.
+
+// Unfortunately needed here to work with linkme
+#![allow(unsafe_code)]
+
+use std::future::Future;
+use std::pin::Pin;
+use std::process::Command;
+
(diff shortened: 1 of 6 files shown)


lib/src/fsck.rs
```diff
@@ -0,0 +1,128 @@
+//! # Write deployments merging image with configmap
+//!
```
jeckersb, line 2: copy/paste header leftover from deploy.rs

jeckersb (approved):
LGTM, just one nit about the header that is not a blocker

cgwalters:
I tweaked this one a bit because I ended up wanting to use tokio as a threadpool, and so I made the main fsck entrypoint `async` (which it was in an earlier version of this).

Issue #1181, opened 2025-03-08
--progress-fd being hidden makes it unusable


antheas:
If --progress-fd is hidden in `--help`, it is not possible for programs to [query](https://github.com/hhd-dev/hhd/blob/f6469a79ea9d96687320e8c145898380bd69a8b6/src/hhd/plugins/bootc/__init__.py#L218-L219) if bootc has progress support. So the option cannot be used.

I missed `cli: Make --progress-fd hide=true for now`.

So, I cannot migrate us off of the PR into mainline bootc. If i take it on faith that bootc accepts --progress-fd you cannot remove it anymore

It is also not possible to use the 0 and 1 fds anymore for debugging in cli.

If you did not like the syntax, leaving it as `--json-fd`, marked as experimental but not hidden, would have made it easier for you to deprecate and remove.

antheas:
You also made the API stateful which makes it so that if the program misses the version packet it can no longer trust the future ones. But thats a smaller nit I was fine with.

cgwalters:
> If --progress-fd is hidden in --help, it is not possible for programs to [query](https://github.com/hhd-dev/hhd/blob/f6469a79ea9d96687320e8c145898380bd69a8b6/src/hhd/plugins/bootc/__init__.py#L218-L219) if bootc has progress support. So the option cannot be used.

Well, one could parse the output of `bootc --version`, which isn't explicitly documented to be machine-readable today, but would clearly work for this right?

(Tangent, it probably makes sense to add a `--version-yaml` or so (like what `ostree --version` does, including the feature set; or maybe we just do what `systemctl --version` does and include features in a secondary line)

> If you did not like the syntax, leaving it as --json-fd, marked as experimental but not hidden, would have made it easier for you to deprecate and remove.

Hey so what happened here is basically we needed to get a release out, but I was a little uncertain about committing to that exact progress format ~forever. But it was my bad that I didn't get back to looking at the tasks to mark it officially stable and left it.

I think we had rough consensus on the overall idea.

BTW, https://github.com/bootc-dev/bootc/issues/1016 is linked as the tracker for stabilization, so this discussion is probably better over there.

> You also made the API stateful which makes it so that if the program misses the version packet it can no longer trust the future ones. But thats a smaller nit I was fine with.

How could it get missed?

cgwalters:
> Well, one could parse the output of bootc --version, which isn't explicitly documented to be machine-readable today, but would clearly work for this right?

[...]

PR #1182, opened 2025-03-08
rename back to --json-fd and unhide

antheas:
We will use this PR going forward until the API stabilizes.

If the option is hidden, it cannot be used. The option being called `--json-fd` and under our control gives you the option to change the API once more before the next release (this PR should not be merged).

The rename also saves me from having to deal with the drift of migrating to the new version while the old version is still deployed.

Closes: https://github.com/bootc-dev/bootc/issues/1181

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -38,8 +38,10 @@ pub(crate) struct ProgressOptions {
     ///
     /// Interactive progress will be written to this file descriptor as "JSON lines"
     /// format, where each value is separated by a newline.
-    #[clap(long, hide = true)]
-    pub(crate) progress_fd: Option<RawProgressFd>,
+    /// 
+    /// The format is experimental and may be removed after a small deprecation period.
+    #[clap(long)]
+    pub(crate) json_fd: Option<RawProgressFd>,
 }
 
 impl TryFrom<ProgressOptions> for ProgressWriter {
@@ -47,7 +49,7 @@ impl TryFrom<ProgressOptions> for ProgressWriter {
 
     fn try_from(value: ProgressOptions) -> Result<Self> {
         let r = value
-            .progress_fd
+            .json_fd
             .map(TryInto::try_into)
             .transpose()?
             .unwrap_or_default();


cgwalters:
> If the option is hidden, it cannot be used.

I hope this is addressed with my suggestion of parsing `--version`.

> The option being called --json-fd and under our control gives you the option to change the API once more before the next release (this PR should not be merged).

Can you clarify: Are you filing this just to track that you plan to ship a branch of bootc with this patch? If so, sure sounds fine, but maybe we then close this?

Alternatively, while I think `--progress-fd` is a clearer and better name than `--json-fd` for the rationale described in [the commit message](https://github.com/bootc-dev/bootc/commit/38e59c52e45bce21677c031a236832d56511d77b) I'd be tentatively fine to re-add `--json-fd` as an *alias* for `--progress-fd` so we don't break users who are testing it out now (not to mention the docs, etc).

antheas:
Yes, I'm filing this to track this. We will ship this patch instead for the time being so that the existing updater can work with few changes

If I made progress fd visible it would be a bit messy

Unfortunately, this means we will still be shipping a fork

Bootc has been failing in some updates recently so we need to bump it. Rechunk ostree-rs-ext bump might also introduce other regressions with the old version

In regards to which name is better, I would tend towards just `--progress` and using fd 0, but that would require a logging framework change

antheas:
I should also mention that we should ship an updated version now, and the next bootc version would take a while to trickle through Fedora even if we decided on a change. So I came up with this PR to shortcut this and buy us some time.

cgwalters:
> the next bootc version would take a while to trickle through Fedora even if we decided on a change. 

There is https://copr.fedorainfracloud.org/coprs/rhcontainerbot/bootc/ that auto-builds from git main FWIW.

antheas:
Interesting. We currently ship a fork, but I would like to fallback to the normal bootc package so we do not have to maintain that. The `--progress-fd` option being hidden does not inspire confidence and would need to be timed. So i opted for the rename for now.

If there are any quirks with the API I think we should stabilize that soon and since the option being hidden means it is not stable we get a bit of a chance to touch things up.

PR #1185, opened 2025-03-10
status: Drop nanoseconds from time

cgwalters:
This is just irrelevant noise; nanoseconds never matter for container builds. Motivated by just making this look visually nicer.

In the future I'd like to look at rendering this how e.g. systemd does it also including a "; 1 day ago" humantime suffix.


jeckersb (approved):
:clock1:

PR #1186, opened 2025-03-10
docs: add footer with CNCF trademark notice

jmarrero:


cgwalters (approved):

jeckersb:
I'm not seeing the footer anywhere in the built docs.  I don't think that `footer` is valid for the `html` backend?  See https://rust-lang.github.io/mdBook/format/configuration/renderers.html#html-renderer-options.

I do see there's an `embedify` preprocessor that knows how to do footer insertion though - https://mr-addict.github.io/mdbook-embedify/local/footer.html

jmarrero:
> I'm not seeing the footer anywhere in the built docs. I don't think that `footer` is valid for the `html` backend? See https://rust-lang.github.io/mdBook/format/configuration/renderers.html#html-renderer-options.
> 
> I do see there's an `embedify` preprocessor that knows how to do footer insertion though - https://mr-addict.github.io/mdbook-embedify/local/footer.html

yeah that is why is draft. But the current diff adds it. I am not thrilled on how it looks.

jmarrero:
Maybe adding a centralized `---` before that line would make it clear is not part of the doc text.

jmarrero:
ok, now I like how it looks and actually works.

docs/book.toml
```diff
@@ -8,6 +8,12 @@ title = "bootc"
 [preprocessor.mermaid]
 command = "mdbook-mermaid"
 
+[preprocessor.header-footer]
+headers = []
+footers = [
+  { padding = "\n---\n<p style=\"font-size:11px\">The Linux Foundation® (TLF) has registered trademarks and uses trademarks. For a list of TLF trademarks, see <a href=\"https://www.linuxfoundation.org/trademark-usage/\">Trademark Usage.</a></p>" }
```
cgwalters, line 14: This HTML-in-TOML is a bit syntactically awkward, but I guess there's nothing we can do about that.

cgwalters (approved):

PR #1191, opened 2025-03-11
docs/filesystem: Weaken stateoverlay wording and direct to alternatives

cgwalters:
We wrote the code and it is useful. We are probably just going to need to support it into the forseeable future. However, my core problem with it is what I wrote in the docs here: it has no equivalent in the docker/podman ecosystem, which I think cuts against a core principle that we're trying to make the booted host align with containers.

Of course, stateoverlay isn't a lot of code, and in theory it's a bit independent of ostree, so perhaps we could try to split it out at some point and then it could be used by e.g. `podman run --mount=type=stateoverlay` or so.

Before anyone starts just blindly enabling stateoverlays I'd like them to have considered alternatives such as the symlink approach.

### DIFF
--- a/docs/src/filesystem.md
+++ b/docs/src/filesystem.md
@@ -225,12 +225,25 @@ More on prepare-root: <https://ostreedev.github.io/ostree/man/ostree-prepare-roo
 
 This feature enables a writable overlay on top of `/opt` (or really, any
 toplevel or subdirectory baked into the image that is normally read-only).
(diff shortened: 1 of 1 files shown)


jmarrero (approved):
lgtm

cgwalters:
cc @jlebon in case you want to tweak things

jlebon:
Threw something in https://github.com/bootc-dev/bootc/pull/1199

PR #1198, opened 2025-03-13
imgstorage: Set selinux labels for imgstorage

ckyrouac:
Running some containers (e.g. mssql) requires the imgstorage labels to
be identical to the /var/lib/containers/storage. The ideal long term
solution to this is to ship a policy with bootc to be included in each
bootc image that handles the mapping. As a near term solution, this
commit creates each subdirectory of /sysroot/ostree/bootc/storage, sets
the label of each directory directly, then runs `podman --root ...
images` to populate the storage directory with all the correct labels
set. The reason for directly setting the labels rather than inferring
them from /var/lib/containers in an existing policy, is to avoid
requiring a policy when initializing the storage directory in other
places than the install to filesystem path (e.g. anaconda).

---

this ballooned into more changes than I expected. I'm not sure if directly setting the label is what we want, but I was having a hard time finding a way to access a policy to use in the [anaconda](https://github.com/bootc-dev/bootc/blob/57adb67bf0a884a5cd06576713bf9e058dc708e5/lib/src/install/completion.rs#L294) installation path, or the [image cli stuff](https://github.com/bootc-dev/bootc/blob/57adb67bf0a884a5cd06576713bf9e058dc708e5/lib/src/cli.rs#L1101).

### DIFF
--- a/lib/src/imgstorage.rs
+++ b/lib/src/imgstorage.rs
@@ -20,6 +20,7 @@ use cap_std_ext::cap_tempfile::TempDir;
 use cap_std_ext::cmdext::CapStdExtCommandExt;
 use cap_std_ext::dirext::CapStdExtDirExt;
 use fn_error_context::context;
+use ostree_ext::ostree::{self};
 use std::os::fd::OwnedFd;
 use tokio::process::Command as AsyncCommand;
 
@@ -34,6 +35,8 @@ const SUBCMD_ARGV_CHUNKING: usize = 100;
 pub(crate) const STORAGE_ALIAS_DIR: &str = "/run/bootc/storage";
 /// We pass this via /proc/self/fd to the child process.
 const STORAGE_RUN_FD: i32 = 3;
+
+const LABELED: &str = ".bootc_labeled";
 
 /// The path to the image storage, relative to the bootc root directory.
 pub(crate) const SUBPATH: &str = "storage";
@@ -161,24 +164,68 @@ impl Storage {
         Ok(())
     }
 
+    #[context("Labeling imgstorage dirs")]
+    fn label_dirs(root: &Dir, sepolicy: Option<&ostree::SePolicy>) -> Result<()> {
+        if root.try_exists(LABELED)? {
+            return Ok(());
+        }
+        let Some(sepolicy) = sepolicy else {
+            return Ok(());
+        };
+
+        // recursively set the labels because they were previously set to usr_t,
+        // and there is no policy defined to set them to the c/storage labels
+        crate::lsm::ensure_dir_labeled_recurse_policy(
+            &root,
+            ".",
+            Some(Utf8Path::new("/var/lib/containers/storage")),
+            0o755.into(),
+            Some(sepolicy),
+        )
+        .context("labeling storage root")?;
+
+        let paths = ["overlay", "overlay-images", "overlay-layers", "volumes"];
+        for p in paths {
+            let full_label_path = format!("/var/lib/containers/storage/{}", p);
+            crate::lsm::ensure_dir_labeled_recurse_policy(
+                &root,
+                p,
+                Some(Utf8Path::new(full_label_path.as_str())),
+                0o755.into(),
+                Some(sepolicy),
+            )
+            .context(format!("labeling storage subpath: {}", p))?;
+        }
+
+        root.create(LABELED)?;
+
+        Ok(())
+    }
+
     #[context("Creating imgstorage")]
-    pub(crate) fn create(sysroot: &Dir, run: &Dir) -> Result<Self> {
+    pub(crate) fn create(
+        sysroot: &Dir,
+        run: &Dir,
+        sepolicy: Option<&ostree::SePolicy>,
+    ) -> Result<Self> {
         Self::init_globals()?;
         let subpath = &Self::subpath();
 
         // SAFETY: We know there's a parent
         let parent = subpath.parent().unwrap();
+        let tmp = format!("{subpath}.tmp");
         if !sysroot
             .try_exists(subpath)
             .with_context(|| format!("Querying {subpath}"))?
         {
-            let tmp = format!("{subpath}.tmp");
             sysroot.remove_all_optional(&tmp).context("Removing tmp")?;
             sysroot
                 .create_dir_all(parent)
                 .with_context(|| format!("Creating {parent}"))?;
             sysroot.create_dir_all(&tmp).context("Creating tmpdir")?;
             let storage_root = sysroot.open_dir(&tmp).context("Open tmp")?;
+            Self::label_dirs(&storage_root, sepolicy)?;
+
             // There's no explicit API to initialize a containers-storage:
             // root, simply passing a path will attempt to auto-create it.
             // We run "podman images" in the new root.
@@ -192,7 +239,12 @@ impl Storage {
                 .rename(&tmp, sysroot, subpath)
                 .context("Renaming tmpdir")?;
             tracing::debug!("Created image store");
+        } else {
+            // the storage already exists, make sure it has selinux labels
+            let storage_root = sysroot.open_dir(subpath).context("opening storage dir")?;
+            Self::label_dirs(&storage_root, sepolicy)?;
         }
+
         Self::open(sysroot, run)
     }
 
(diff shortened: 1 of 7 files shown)


lib/src/imgstorage.rs
```diff
@@ -161,38 +160,60 @@ impl Storage {
...
+            .create_dir_all(subpath.parent().unwrap())
+            .context("creating imgstorage parent dir")?;
+
+        //directly set the label, instead of deriving it from an existing /var/lib/containers
+        //directory to avoid requiring loading an selinux policy whenever the storage is created
+        //e.g. during the anaconda installation path
+        crate::lsm::ensure_dir_labeled(
+            &sysroot,
+            subpath,
+            "system_u:object_r:container_var_lib_t:s0",
```
cgwalters, line 181: It's quite suboptimal to hardcode specific labels for a few reasons. A simple one is that these things can change over time inside the policy. But a larger reason is that there are actually multiple SELinux policies out there. For example Android has a totally different one (not that we're likely to run there but you never know). More realistically though there are definitely people who want to deploy custom hardened policies too (selinux-policy-targeted is too "loose" for some use cases). See e.g. https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/using_selinux/using-multi-level-security-mls_using-selinux#ref_selinux-roles-in-mls_using-multi-level-security-mls

(Tangent on `selinux-policy-mls` but a related and very topical one; look at those installation instructions targeting a live system - a giant pile of hacks! Whereas if you build a bootc container with the package, you *should* get a transactional in-place install (or upgrade) to the new state...)

Anyways my belief here is that it *should* work to use what is now `ensure_dir_labeled_as_path` and pass `/var/lib/containers` etc here. The way this would work is it looks up `/var/lib/containers` in the file context regexps list and uses it for our target path. Did you try that and it failed?

Now you mention Anaconda...I think making this work will require that we have `container-selinux` (and `selinux-policy-targeted`) in the Anaconda environment.

Actually more generally, it will mean we use the *versions* of those - which is different from the ones in the bootc container in the general case. This isn't a new problem at all - when installing RPMs with Anaconda, the `/usr/bin/rpm` comes from Anaconda today. But for bootc we can do something a bit nicer because the target for installation is a container image - a well-defined self-contained entity. And that is "tracked" in https://github.com/rhinstaller/anaconda/discussions/5197 to be sure we always use the bootc (and policy versions, etc.) from the target system, not whatever is in the ISO.

---

Anyways...again can you say if the change to use `ensure_dir_labeled_as_path()` was actually failing?

Mmmmm. The more I think about this the more I'm thinking that in the near term we should build on the recently added [install finalize](https://github.com/containers/bootc/pull/1157).

- Change Anaconda to invoke it
- Fix up the labels and such based on the *installed* target system using it

lib/src/imgstorage.rs
```diff
@@ -161,38 +160,60 @@ impl Storage {
...
-                .arg("images")
-                .run()
-                .context("Initializing images")?;
-            drop(storage_root);
-            sysroot
-                .rename(&tmp, sysroot, subpath)
-                .context("Renaming tmpdir")?;
-            tracing::debug!("Created image store");
-        }
+        let storage_dir = sysroot.open_dir(subpath).context("opening storage dir")?;
```
cgwalters, line 207: I'm OK with this change but the rationale behind the more complex code that was there before was it was attempting to do an *atomic* initialization of the target directory, i.e. either everything there is created or not. As is right now if the command dies partway through I am not sure that the c/storage stack will handle completing initialization.

Perhaps a theoretical concern but

- It's important to me that bootc gives transactional upgrades, i.e. we should be robust to having e.g. the kernel freeze or the laptop battery fall out or being randomly struck by the OOM killer or whatever.
- While what we're talking about here is for the installation path, this code can also get invoked for *old* bootc systems that predate LBIs (or just systems previously pure ostree) that don't have the c/storage instance.

I guess in the end we should just make it a c/storage bug if it doesn't handle partial initialization.

lib/src/imgstorage.rs
ckyrouac: > It's quite suboptimal to hardcode specific labels for a few reasons.

Agreed, just thought this might be a valid temporary approach if we plan to ship a policy in the base images sooner rather than later.

> Anyways my belief here is that it should work to use what is now ensure_dir_labeled_as_path and pass /var/lib/containers etc here. The way this would work is it looks up /var/lib/containers in the file context regexps list and uses it for our target path. Did you try that and it failed?

The `ImgStorage::init_dirs` function in this PR worked using `ensure_dir_labeled_as_path` in the install-to-filesystem path. However, since `ensure_dir_labeled_as_path` [looks up the label from the ostree.SePolicy](https://github.com/bootc-dev/bootc/blob/57adb67bf0a884a5cd06576713bf9e058dc708e5/lib/src/lsm.rs#L186), it doesn't work when the storage is initialized in other places of the code, e.g. [here](https://github.com/bootc-dev/bootc/blob/57adb67bf0a884a5cd06576713bf9e058dc708e5/lib/src/install/completion.rs#L294), [here](https://github.com/bootc-dev/bootc/blob/57adb67bf0a884a5cd06576713bf9e058dc708e5/lib/src/boundimage.rs#L155), [here](https://github.com/bootc-dev/bootc/blob/57adb67bf0a884a5cd06576713bf9e058dc708e5/lib/src/cli.rs#L1101), etc.

> Mmmmm. The more I think about this the more I'm thinking that in the near term we should build on the recently added https://github.com/containers/bootc/pull/1157.
>
> Change Anaconda to invoke it
> Fix up the labels and such based on the installed target system using it

I think this will work for the anaconda and install-to-filesystem cases.  A problem I was running into was setting the labels after the storage is initialized, since it's not enough to just set the top level directory (and sub-directories). I guess since the update path will need to recursively set all the labels anyways, we can just do the same thing in `install_finalize`?

The other issue is the CLI commands that currently initialize the storage if it doesn't exists, e.g. `bootc image pull-from-default-storage`. I guess if we make sure the storage is initialized on install and update, then we can change the CLI code that needs storage to fail if it isn't initialized?

lib/src/imgstorage.rs
ckyrouac: :+1: makes sense, should be simple to preserve the atomic initialization. c/storage does seem to handle at least some amount of partial initialization since I am able to create and label the sub-directories then let c/storage create the rest after.

cgwalters:
We discussed this for Anaconda

```
diff --git a/lib/src/install/completion.rs b/lib/src/install/completion.rs
index 85983558..81c941b9 100644
--- a/lib/src/install/completion.rs
+++ b/lib/src/install/completion.rs
@@ -14,6 +14,8 @@ use ostree_ext::{gio, ostree};
 use rustix::fs::Mode;
 use rustix::fs::OFlags;
 
+use crate::utils::deployment_fd;
+
 use super::config;
 
 /// An environment variable set by anaconda that hints
@@ -277,6 +279,9 @@ pub(crate) async fn impl_completion(
         .merge_deployment(stateroot)
         .ok_or_else(|| anyhow::anyhow!("Failed to find deployment (stateroot={stateroot:?}"))?;
     let sysroot_dir = crate::utils::sysroot_dir(&sysroot)?;
+    let deployment_fd = &deployment_fd(sysroot, deployment)?;
+
+    let sepolicy = &ostree::SePolicy::new_at(deployment_fd.as_raw_fd(), gio::Cancellable::NONE)?;
 
     // Create a subdir in /run
     let rundir = "run/bootc-install";
```

ckyrouac:
@cgwalters still need to do some testing, but I pushed the latest code that at least works on the install path so you can take a look.

cgwalters (approved):
This looks pretty good to me overall! I'm fine to land as is, all this stuff is optional followup overall.

cgwalters:
One other thing actually, it should be straightforward to add a test case under `tests/booted/readonly/` that does e.g. `getfattr -n security.selinux /usr/lib/bootc/storage/overlay | grep container_ro_file_t` or so?

ckyrouac:
thanks! Added a test and cleaned up label_dirs per your suggestion. Will file issues for followups.

cgwalters (approved):

cgwalters:
So I use bootc from git on my desktop, and I just hit this:

```
ERROR Upgrading: Staging: Pulling bound images: Creating imgstorage: Labeling imgstorage dirs: labeling storage root: Opening ./db.sql: Found non-directory
```

PR #1199, opened 2025-03-13
docs/filesystem: Rework state overlay and `/opt` messaging a bit

jlebon:
Tweak the changes in #1191 to more clearly lay out the tradeoffs of state overlays. Contrast it with transient root (since that was one of the primary motivations for it), and generalize the "alternatives" to be about both approaches.

### DIFF
--- a/docs/src/filesystem.md
+++ b/docs/src/filesystem.md
@@ -231,9 +231,13 @@ The semantics here are somewhat nuanced:
 - Changes persist across reboots by default
 - During updates, new files from the container image override any locally modified version
 
-A disadvantage of this approach is that there is no equivalent to this feature in the
-Docker/podman ecosystem, and while its semantics can often trivially enable many applications that
-install into `/opt`, it can still be prone to "state drift".
+The advantages are:
+- It makes it very easy to make compatible applications that install into `/opt`.
+- In contrast to transient root (above), a smaller surface of the filesystem is mutable.
+
+The disadvantages are:
+- There is no equivalent to this feature in the Docker/Podman ecosystem.
+- It allows for some temporary state drift until the next update.
 
 To enable this feature, instantiate the `ostree-state-overlay@.service`
 unit template on the target path. For example, for `/opt`:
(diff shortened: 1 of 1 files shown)


docs/src/filesystem.md
```diff
@@ -231,9 +231,13 @@ The semantics here are somewhat nuanced:
...
-A disadvantage of this approach is that there is no equivalent to this feature in the
-Docker/podman ecosystem, and while its semantics can often trivially enable many applications that
-install into `/opt`, it can still be prone to "state drift".
+The advantages are:
+- It makes it very easy to make compatible applications that install into `/opt`.
+- In contrast to transient root (above), a smaller surface of the filesystem is mutable.
+
+The disadvantages are:
+- There is no equivalent to this feature in the Docker/Podman ecosystem.
+- It allows for some temporary state drift until the next update.
```
cgwalters, line 240: I'm fine with this as written but I think the thing that specifically bothers me is it feels quite subtle that that persistence is the default for unknown files, but we overwrite known files while both are mutable. That semantic is absolutely closer to "just works" for many cases of course...

Maybe in reality it's not a real problem. 

Maybe what would help is if we offered a clear tooling to show you the diff and help you flush any unwanted files? Not sure.

cgwalters (approved):

docs/src/filesystem.md
cgwalters: I guess maybe I'm objecting to the "state drift until the next update" because in the case you accidentally write something there you will get that drift.

jlebon: > it feels quite subtle that that persistence is the default for unknown files, but we overwrite known files while both are mutable

Though note that's basically the same semantics as RPM for non-config files. Hmm, would it help the mental model to add that to the docs?

> Maybe what would help is if we offered a clear tooling to show you the diff and help you flush any unwanted files? Not sure.

Yeah, that'd definitely be useful. I guess for now `find /var/ostree/state-overlays/$name/upper` is close to that.

> I guess maybe I'm objecting to the "state drift until the next update" because in the case you accidentally write something there you will get that drift.

Hmm, don't quite follow this. If you overwrite something that's owned by the RPM, it will get reconciled in the next update (so temporary drift). Writing anywhere else under the overlay is assumed to be state, just like `/var` is state. So calling that "state drift" feels odd to me.

PR #1200, opened 2025-03-13
lint: Add an api-base-directories lint

cgwalters:
Saw in the wild that someone accidentally did an
`rm -rf /var`, and that only fails pretty late in the process.

Add a lint that hard requires the "API filesystem" directory mount points, plus `/var` - that one is a requirement for ostree right now.


cgwalters:
Dang it, that test run hit https://github.com/bootc-dev/bootc/issues/1204

/packit retest-failed

ckyrouac (approved):

PR #1202, opened 2025-03-14
feat(rollback): add `--apply` flag

### LINKED ISSUE #1029: [RFE] Add option `--apply` to `bootc rollback` subcommand
**User Story:**

As a system administrator, when I need to rollback my image mode host to `current rollback image`, I would like to use the option `--apply` to reboot the the host automatically.

**Current behavior:**

- Currently I need to restart the host manually as there is no option to reboot automatically after issuing the `bootc rollback` command

```
$ sudo bootc rollback
Next boot: rollback deployment
$ sudo reboot NOW
```

**Desired behavior:**

```
$ sudo bootc rollback --apply
Next boot: rollback deployment
…
```

Host reboots automatically.

**Additional info:**

- I would like to have the option `--apply` as it's the same option already used for `bootc upgrade` and `bootc switch`

p5:
Adds an --apply flag to the `bootc rollback` command to implement automated restarts.

Have confirmed this works by building the bootc binary and running `bootc rollback --apply` on my host.  This restarted the machine into the new (rollback) image.

Closes #1029

---

Hope you don't mind I took the easiest issue I could find :) 
This was mainly to familiarise myself with the debugging flow, and to get an environment setup.

The implementation was referencing the existing --apply flag for the switch command.  The error handling _should_ be okay, since if I understand correctly, the `?` prepended to the `crate::deploy::rollback(sysroot).await` will cause the function to work the same as it did previously - returning the error to the calling function inside a Result.

How do the docs changes work?  They look like they would be auto-generated, though I couldn't find a target to do so.

### DIFF
--- a/docs/src/man/bootc-rollback.md
+++ b/docs/src/man/bootc-rollback.md
@@ -7,7 +7,7 @@ queued upgrade) then it will be discarded
 
 # SYNOPSIS
 
-**bootc rollback** \[**-h**\|**\--help**\]
+**bootc rollback**  \[**\--apply**\] \[**-h**\|**\--help**\]
 
 # DESCRIPTION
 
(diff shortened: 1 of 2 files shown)


cgwalters (approved):
Thanks!

PR #1211, opened 2025-03-18
fix(status): enable --booted option to only show the current deployment

p5:
The existing code didn't take into account the --booted option, so always showed the staged, current and rollback deployments.  This correctly wires through the --booted option to only show that deployment.

Stubs have been left in the code should we wish to enable options to show only the rollback or staged options (--rollback / --staged).

I tried to make the implementation not specific to --booted, so added the Slot enum as an input.  This meant making that enum public.  Please let me know if you would prefer a different approach.

No docs changes were required since the flag is already present.

Please find below the different outputs:

<details>

```
bootc-status-booted-only on  bootc-status-booted-only via 🦀 took 2s 
❯ sudo ./target/debug/bootc status --booted
[sudo] password for admin: 
● Booted image: ghcr.io/rsturla/eternal-linux/lumina:stable-nvidia-open
        Digest: sha256:82de1647a00fd1f7674d268c9aaa33127a09139ac3206c5d71c8e0926846304d
       Version: 250317 (2025-03-17T01:45:07Z)

bootc-status-booted-only on  bootc-status-booted-only via 🦀 took 3s 
❯ sudo ./target/debug/bootc status         
● Booted image: ghcr.io/rsturla/eternal-linux/lumina:stable-nvidia-open
        Digest: sha256:82de1647a00fd1f7674d268c9aaa33127a09139ac3206c5d71c8e0926846304d
       Version: 250317 (2025-03-17T01:45:07Z)

  Rollback image: ghcr.io/rsturla/eternal-linux/lumina:stable-nvidia-open
          Digest: sha256:2b6be3a574af416fa7adb8a1217e408aabe18ab15e6e9a401e180a9d1b40bb44
         Version: 250315 (2025-03-15T01:45:07Z)

bootc-status-booted-only on  bootc-status-booted-only via 🦀 
❯ sudo ./target/debug/bootc status --json  
[...]

PR #1213, opened 2025-03-19
tests(status): add tests for --booted

p5:
Follow up to #1211 

One piece of feedback from my previous PR was to add tests to ensure the changes I made are working, and remain working over future iterations.

I am unsure how exactly to write unit tests for the status function itself since it relies on a sysroot, but I've added what I hope to be a working integration test for this.

Also adds tests to the new filter_to_slots function.  This is probably over tested, but wanted to cover my bases.

---

I couldn't figure out how to get a tmt tests running locally, so I'm hoping they'll run as part of the PR test suite.


tests/booted/readonly/001-test-status.nu
```diff
@@ -5,9 +5,18 @@ tap begin "verify bootc status output formats"
...
+
 let st = bootc status --format=yaml | from yaml
 assert equal $st.apiVersion org.containers.bootc/v1
 assert ($st.status.booted.image.timestamp != null)
+
+let st = bootc status --json --booted | from json
+assert equal $st.apiVersion org.containers.bootc/v1
+assert ($st.status.booted.image.timestamp != null)
+assert (($st.status | get rollback | default null) == null)
+assert (($st.status | get staged | default null) == null)
```
p5, line 20: This:
- ensures the booted image timestamp is present
- extracts <status>.rollback from the status, falling back to null if not found, and enforces null
- extracts <status>.staged from the status, falling back to null if not found and enforces null

cgwalters: > extracts status.rollback JSON object, falling back to null if not found, and enforces null

The elements are still present just set to null, right? So I don't think we need this `| default null`.

cgwalters (approved):

PR #1218, opened 2025-03-21
WIP: Refactor recursive lint traversal

cgwalters:
Prep for adding more lints which do a recursive traversal.

In paticular I wanted to add a check for `ostree.usermeta` should never be present.


cgwalters:
Thought I could knock out a quick check for `user.ostreemeta` being present but then I realized we already had a full recursive traversal of the tree in the utf8 check, and wanted to avoid duplicating that and also avoid walking everything twice, which increases complexity. Stashing this here for now.

ckyrouac (approved):

Issue #1204, opened 2025-03-14
`broken pipe` on fetches still


cgwalters:
We still see
`ERROR Switching: Pulling: Importing: Parsing layer blob sha256:43373869bc293673e2dfc59a04c626501893eed1f7c1ec98004f2e7c0e41a245: failed to invoke method FinishPipe: failed to invoke method FinishPipe: write |1: broken pipe` occasionally at least in CI runs though I haven't seen it myself personally/interactively.

Previously: https://github.com/ostreedev/ostree-rs-ext/issues/657

cc also https://github.com/bootc-dev/bootc/issues/963 where this seems to happen with zstd:chunked

jeckersb:
I've been running the [previous reproducer](https://github.com/ostreedev/ostree-rs-ext/issues/657#issuecomment-2430438301) in a loop for a while on the latest code, and for better or worse I don't see any broken pipes.  So at the very least I don't think we've regressed on that particular fix.

Next I'll look at the zstd:chunked case and see how that plays into things.

jeckersb:
Ok verified that ztd:chunked is consistently broken:

```
# ostree --repo=repo refs --delete ostree/container; ./bootc internals ostree-container image pull --quiet repo ostree-unverified-registry:quay.io/jeckersb/fedora:zstdchunked
layers already present: 0; layers needed: 1 (60.9 MB)
ERROR Importing: Parsing layer blob sha256:f5d3e8e90d538ff16bd4d8719929cbeebee96473d79529975ef602c472256960: failed to invoke method FinishPipe: failed to invoke method FinishPipe: write |1: broken pipe
```

Aside, I wasted entirely too much time this morning trying to debug podman before I realized I was hitting https://github.com/containers/storage/issues/2055.  I had to remove podman-machine from the loop entirely and do the compression directly on my host with 64GB in order for it to actually push 😭

jeckersb:
Debug output from `skopeo`, once I realized I had to rebuild `ostree_ext` pointing at git head for `containers-image-proxy` because we never released a new version that contains https://github.com/containers/containers-image-proxy-rs/pull/70 (TODO!)

```
# ostree --repo=repo refs --delete ostree/container; CONTAINERS_IMAGE_PROXY_DEBUG=1 ./bootc internals ostree-container image pull --quiet --insecure-skip-tls-verification repo ostree-unverified-registry:localhost:5000/zstdchunked
DEBU[0000] Executing method Initialize
[...]

PR #1220, opened 2025-03-21
WIP: container image import flow diagram

jeckersb:


jeckersb:
Still needs some work but I can't stare at it anymore today and if I somehow lost all of this I would cry.  A lot.

cgwalters (approved):
Nice!

jeckersb:
Marking this as ready, fortunately I don't see the need for drastic changes to how this works since I appear to have [figured out our broken pipe problem](https://github.com/bootc-dev/bootc/issues/1204#issuecomment-2751591171), so the diagram can stay relevant for the foreseeable future.

PR #1221, opened 2025-03-22
ostree/encapsulate: Support overriding platform (arch)

cgwalters:
Currently the encapsulate code assumes it's generating images that match the platform (os+arch) that's running. However, for rpm-ostree we have a use case of performing "rechunking" in a cross-architecture fashion. In the end we don't run any code in that process, it's just manipulating bits, so it's perfectly fine to operate cross-arch.

Allow passing in an override for the generated image platform to facilitate this.


jmarrero (approved):
lgtm

Issue #963, opened 2024-12-12
bootc/rpm-ostree update fails on Importing: Unencapsulating base: failed to invoke method FinishPipe with zstd:chunked image


travier:
Updating to a zstd:chunk'ed image using bootc/rpm-ostree fails with:

```
$ sudo bootc update
layers already present: 42; layers needed: 24 (1.3 GB)
Fetched layers: 0 B in 2 seconds (0 B/s)
ERROR Upgrading: Pulling: Importing: Unencapsulating base: failed to invoke method FinishPipe: failed to invoke method FinishPipe: write |1: broken pipe
```
```
$ sudo bootc status
No staged image present
Current booted image: quay.io/travier/fedora-kinoite:latest
    Image version: 41.20241207.0 (2024-12-07 04:10:35.798585027 UTC)
    Image digest: sha256:40efa2aa91e2ad4c3b75218cce8463d78193f2b35b2dfc972d412fba5eaaccb7
Current rollback image: quay.io/travier/fedora-kinoite:latest
    Image version: 41.20241202.0 (2024-12-02 04:11:16.290882830 UTC)
    Image digest: sha256:1807aefcb830c8952c1ecebaaab28900ef48ea5f8075e4a6b34a38d788f1da7e
```
```
$ sudo rpm-ostree update
Pulling manifest: ostree-image-signed:docker://quay.io/travier/fedora-kinoite:latest
Importing: ostree-image-signed:docker://quay.io/travier/fedora-kinoite:latest (digest: sha256:1f27336996fc663ad0ef3f448eef21edfa71322931fde0fe9f577068736aa2d5)
ostree chunk layers already present: 43
ostree chunk layers needed: 22 (1.2 GB)
custom layers needed: 1 (38.8 MB)
[0/23] Fetching ostree chunk f3c0f64a5cdf6196496 (50.4 MB)... done
error: Importing: Unencapsulating base: failed to invoke method FinishPipe: failed to invoke method FinishPipe: write |1: broken pipe
```

[...]

PR #1228, opened 2025-03-25
ostree-ext/filter_tar_async: flush decompressor stream

jeckersb:
We need to read anything past the tar stream, otherwise we'll deadlock
via skopeo in FinishPipe.

Also remove the bit where we hold the pipe open longer, we shouldn't
need to do that anymore because we've ensured we've read everything
out of it by this point.

Resolves: #1204


jeckersb:
Hm should have probably run tests locally first :laughing:

cgwalters (approved):
Ahh, that definitely explains why it happens reliably with zstd:chunked.

But...hmmm...we have still seen it occasionally for non zstd:chunked layers though too, could it be that even gzip has a similar thing, just less likely?

jeckersb:
Yeah I'm not familiar with the specifics on gzip but that could be the case.

Looking at the failing tests though, it looks like maybe when we do the filtering that causes us to not always read to the end of even the basic tar stream?  I guess that could also cause the same behavior.  Should be easy enough to test by writing a large enough file at the end of the stream that ends up getting filtered out.

cgwalters:
If the *last* item is filtered out, then that will definitely happen.

But even there, in the [tar format](https://www.gnu.org/software/tar/manual/html_node/Standard.html) there's trailing zeros to indicate end-of-archive, and I wonder if tar-rs is always consuming those bytes? It probably should be, but not 100% sure.

jeckersb:
Another thing re: "why does gzip work better", note that for the gzip case we [explicitly create a BufReader](https://github.com/bootc-dev/bootc/blob/main/ostree-ext/src/container/unencapsulate.rs#L201-L203) since it's required by `bufread:GzDecoder`, whereas in the zstd case it's allocating it's own [BufReader::with_capacity](https://docs.rs/crate/zstd/latest/source/src/stream/read/mod.rs#29).  I haven't run down what the result of `zstd_safe::DCtx::in_size()` ends up being, but I suspect in the zstd case we end up buffering a lot less than the gzip case, which means it would be more likely that we "miss" the end of the stream in our buffer.

ostree-ext/src/tar/write.rs
```diff
@@ -366,22 +366,32 @@ async fn filter_tar_async(
...
     let copier = tokio::io::copy(&mut rx_buf, &mut dest);
     let (r, v) = tokio::join!(tar_transformer, copier);
     let _v: u64 = v?;
-    let (r, src) = r?;
-    // Note that the worker thread took temporary ownership of the input stream; we only close
-    // it at this point, after we're sure we've done all processing of the input.  The reason
-    // for this is that both the skopeo process *or* us could encounter an error (see join_fetch).
-    // By ensuring we hold the stream open as long as possible, it ensures that we're going to
-    // see a remote error first, instead of the remote skopeo process seeing us close the pipe
-    // because we found an error.
```
cgwalters, line 381: But are we sure we don't still need this logic? We don't have any explicit test cases for errors-in-tar stream, but I think we could add one by e.g. adding a non-UTF8 filename into a container build (or corrupt an ostree checksum).

We definitely want to ensure we see the real error in situations like this and not just "broken pipe" because we dropped `src` which went around and caused skopeo to error which then boomerangs back to us as a broken pipe error.

I'd be a bit more confident in landing this change if we didn't touch this section of the code, only did the copy.

ostree-ext/src/tar/write.rs
```diff
@@ -366,22 +366,32 @@ async fn filter_tar_async(
         let dest = tokio_util::io::SyncIoBridge::new(tx_buf);
 
         let r = filter_tar(&mut src, dest, &config, &repo_tmpdir);
+
+        // We need to make sure to flush out the decompressor here,
```
cgwalters, line 370: I'd say "flush out decompressor and tar stream" because as I said in an earlier comment I believe `filter_tar` could also leave trailing data.

ostree-ext/src/tar/write.rs
```diff
@@ -366,22 +366,32 @@ async fn filter_tar_async(
...
+        // skopeo.
+        //
+        // https://github.com/bootc-dev/bootc/issues/1204
+        let mut sink = std::io::sink();
+        let n = std::io::copy(&mut src, &mut sink)?;
+        if n != 0 {
+            tracing::debug!("Read extra {n} bytes at end of decompressor stream");
+        }
+
         // Pass ownership of the input stream back to the caller - see below.
```
jeckersb, line 388: This comment needs removed (If we actually remove this logic)

ostree-ext/src/tar/write.rs
jeckersb: I don't think it's possible to get "broken pipe" anymore, because at this point we've explicitly drained everything out of the pipe.  Once skopeo sends X bytes and we've read X bytes, over in skopeo the `io.Copy` is done and the goroutine cleans up which in turn completes `FinishPipe`.

In other words, in order to get "broken pipe" we have to close the read end while `io.Copy` is still active in skopeo.  But I don't see how that could happen now.

jeckersb:
> @jeckersb So I went to re-test #963 after this landed but I still get the same symptom. I think we should add an integration test that verifies we can fetch from a `zstd:chunked` image as part of fixes here.

We do already have tests in ostree-ext for zstd and zstd:chunked:
https://github.com/bootc-dev/bootc/blob/c94cfdf0aac585b4c016a566a0924ca6751893e3/ostree-ext/tests/it/main.rs#L1674-L1684
but obviously those were passing before this change.  I was only able to get it to break with a sufficiently large container image (empty worked, 50MB worked, 100MB intermittently failed, 200MB always failed).  So I'll add a test to generate a large (1GB?) zstd/zstd:chunked image and import that.

jeckersb:
> @jeckersb So I went to re-test #963 after this landed but I still get the same symptom. I think we should add an integration test that verifies we can fetch from a `zstd:chunked` image as part of fixes here.

I am able to reproduce this as well with `bootc switch`.  Worth noting that it's always failing in `unencapsulate_base`, whereas before I was seeing it in subsequent layers.  So maybe not exactly the same issue, but certainly closely related.  Should be easy (famous last words...) to figure out what the difference is between the base and non-base case with such a simple and reliable reproducer.

jeckersb:
Your image also fails using the same command I was using before (just `container image pull` from ostree-ext):

```
# ostree --repo=repo refs --delete ostree/container; ./bootc internals ostree-container image pull --quiet --insecure-skip-tls-verification repo ostree-unverified-registry:quay.io/cgwalters/ostest:zstdchunked
layers already present: 0; layers needed: 65 (378.5 MB)
ERROR Importing: Unencapsulating base: failed to invoke method FinishPipe: failed to invoke method FinishPipe: write |1: broken pipe
```

jeckersb:
[Followup PR](https://github.com/bootc-dev/bootc/pull/1247) should fix this

PR #1233, opened 2025-03-26
install: Avoid context-per-recursion

cgwalters:
Got a report over private chat of
```
ERROR Installing to filesystem: Removing boot directory content: Removing entries (noxdev): Removing entries (noxdev): Removing entries (noxdev): Read-only file system (os error 30)
```

That's ugly, we don't want a context per recursive directory entry. Only add a context at the toplevel invokers.


jmarrero (approved):
lgtm

Issue #1238, opened 2025-03-28
consider adding arch info to bootc status output


dustymabe:
In Fedora CoreOS one step on our migration path to `bootc` is switching our existing systems to update from a OCI registry rather than OSTree. One nice thing about registry:tag endpoints is that (at least the way we've set it up) it's a manifest list that points to all the supported architectures.

Ultimately what this means is we lose the nice little side effect of knowing the machines architecture just from the `status` output.

For example before and after in the combined status output below:

```
core@localhost:~$ rpm-ostree status 
State: idle
AutomaticUpdatesDriver: Zincati
  DriverState: active; periodically polling for updates (last checked Fri 2025-03-28 20:52:50 UTC)
Deployments:
● ostree-remote-image:fedora:docker://quay.io/fedora/fedora-coreos:testing
                   Digest: sha256:97cc8eaedf41c6f27d57c8dac0b740f91babe24b031bde447e0bec35acc60589
                  Version: 41.20250315.2.0 (2025-03-18T00:35:22Z)

  fedora:fedora/aarch64/coreos/testing
                  Version: 41.20250302.2.0 (2025-03-04T18:45:14Z)
                   Commit: a4a24122d3f48f6ee66b3f62482edf886e69206b50c67d100b88300dd64f7987
             GPGSignature: Valid signature by 466CF2D8B60BC3057AA9453ED0622462E99D6AD1
```

So in the OCI deployment we don't get the nice `aarch64` in the output.

Since `bootc` is the future I figured I'd start the discussion here about if we'd like to add something back here for that OR if we're happy without it in the status output. From that same system:

```
core@localhost:~$ sudo bootc status 
● Booted image: quay.io/fedora/fedora-coreos:testing
        Digest: sha256:97cc8eaedf41c6f27d57c8dac0b740f91babe24b031bde447e0bec35acc60589
       Version: 41.20250315.2.0 (2025-03-18 00:35:22 UTC)

  Rollback ostree
           Commit: a4a24122d3f48f6ee66b3f62482edf886e69206b50c67d100b88300dd64f7987
```

cgwalters:
This is always identical to the output of running the `arch` binary right? How often do you log into a system and not know its hardware architecture?

Also a core principle for bootc is that we want to align to how container engines work where it makes sense, and podman/docker definitely don't show you this data in the default human-readable output like `podman images` etc.

[...]

PR #1239, opened 2025-03-28
spec, status: Add architecture

cgwalters:
The argument here twofold:

- We used to show it with ostree refs because ostree doesn't have manifest lists
- Actually we do want to highlight that the digest of the image is always the digest of the per-arch image; we currently peel and discard manifest lists, which may not be obvious.

Closes: https://github.com/bootc-dev/bootc/issues/1238

### DIFF
--- a/lib/src/fixtures/spec-only-booted.yaml
+++ b/lib/src/fixtures/spec-only-booted.yaml
@@ -14,6 +14,7 @@ status:
       image:
         image: quay.io/centos-bootc/centos-bootc:stream9
         transport: registry
+      architecture: arm64
       version: stream9.20240807.0
       timestamp: null
       imageDigest: sha256:47e5ed613a970b6574bfa954ab25bb6e85656552899aa518b5961d9645102b38
--- a/lib/src/fixtures/spec-ostree-remote.yaml
+++ b/lib/src/fixtures/spec-ostree-remote.yaml
@@ -14,6 +14,7 @@ status:
       image:
         image: quay.io/otherexample/otherimage:latest
         transport: registry
+      architecture: arm64
       version: 20231230.1
       timestamp: 2023-12-30T16:10:11Z
       imageDigest: sha256:b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c
--- a/lib/src/fixtures/spec-ostree-to-bootc.yaml
+++ b/lib/src/fixtures/spec-ostree-to-bootc.yaml
@@ -13,6 +13,7 @@ status:
       image:
         image: quay.io/centos-bootc/centos-bootc:stream9
         transport: registry
+      architecture: s390x
       version: stream9.20240807.0
       timestamp: null
       imageDigest: sha256:47e5ed613a970b6574bfa954ab25bb6e85656552899aa518b5961d9645102b38
--- a/lib/src/fixtures/spec-staged-booted.yaml
+++ b/lib/src/fixtures/spec-staged-booted.yaml
@@ -14,6 +14,7 @@ status:
         image: quay.io/example/someimage:latest
         transport: registry
         signature: insecure
+      architecture: arm64
       version: nightly
       # This one has nanoseconds, which should be dropped for human consumption
       timestamp: 2023-10-14T19:22:15.42Z
@@ -29,6 +30,7 @@ status:
         image: quay.io/example/someimage:latest
         transport: registry
         signature: insecure
+      architecture: arm64
       version: nightly
       timestamp: 2023-09-30T19:22:16Z
       imageDigest: sha256:736b359467c9437c1ac915acaae952aad854e07eb4a16a94999a48af08c83c34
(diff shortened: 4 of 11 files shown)


cgwalters:
Now obviously something to bikeshed further is we use the OCI arch names which come from Docker where they just punted on and said "use whatever Go uses" which...yeah is not the same as `uname` on at least Fedora derivatives. And yeah, what I see looking at https://quay.io/repository/fedora/fedora-bootc is a mix because the uploads use the uname-arch but quay.io of course renders the manifests in each image (e.g. [here](https://quay.io/repository/fedora/fedora-bootc/manifest/sha256:0b546377458079600d8c3aa0b96654a5e66201f935bd86e5726d3ed7c8d6f61a)) with the OCI arch. So...we might as well give in and use what OCI uses.

dustymabe:
> Now obviously something to bikeshed further is we use the OCI arch names

Ouch.. yeah. The split isn't ideal but I don't have a better suggestion than what you're going with.

lib/src/fixtures/spec-only-booted.yaml
```diff
@@ -6,6 +6,7 @@ spec:
   image:
     image: quay.io/centos-bootc/centos-bootc:stream9
     transport: registry
+    architecture: amd64
```
dustymabe, line 9: I'm not sure I understand this one 100%.. In the other test fixture files you didn't add a spec.image.architecture. Is this one here on purpose?

Also is it intentional that this architecture is `amd64` and the one later in this same file is `arm64`?

dustymabe (approved):
LGTM in a local build

```
[core@cosa-devsh ~]$ rpm-ostree status 
State: idle
Deployments:
● ostree-remote-image:fedora:docker://quay.io/fedora/fedora-coreos:rawhide
                   Digest: sha256:fce2f34d80a04c67b01d6c8633e91a18a39c985e6402ebc1d5538b5c2496b674
                  Version: 43.20250328.dev.1 (2025-03-28T19:55:54Z)
[core@cosa-devsh ~]$ 

[core@cosa-devsh ~]$ sudo /usr/local/bin/bootc status 
● Booted image: quay.io/fedora/fedora-coreos:rawhide
        Digest: sha256:fce2f34d80a04c67b01d6c8633e91a18a39c985e6402ebc1d5538b5c2496b674 (amd64)
       Version: 43.20250328.dev.1 (2025-03-28T19:55:54Z)
```

lib/src/fixtures/spec-only-booted.yaml
cgwalters: Nope it's a mistake, we need to add `serde(deny_unknown_fields)` here

vrothberg (commented):
I love the idea!

A small bikeshed:  I would prefer a dedicated field in the human-readable output.

Image, Digest, Architecture and Version all seem like a good candidates.

cgwalters:
> A small bikeshed: I would prefer a dedicated field in the human-readable output.

Hmm - why? Just visual consistency? The rationale behind the current choice (appending to an existing field) is for most use cases the admin is going to know it - and also it's always going to be the same for each deployment, so the redundancy multiplies.

I think the current "conciseness" of status is a feature.

Per the commit message, the architecture is also associated with the digest.

We also group the timestamp with the version today under the idea that it's often similar data.

dustymabe:
I agree with @cgwalters here. Since this is the human readable output I do prefer conciseness here. When I originally opened https://github.com/bootc-dev/bootc/issues/1238 I was struggling myself with how we could implement it without adding a new field and I think what Colin proposed was a good solution.

dustymabe:
> Ouch.. yeah. The split isn't ideal but I don't have a better suggestion than what you're going with.

I guess the other option would be to maintain a map (`amd64 -> x86_64`, `arm64 -> aarch64`) for those two where it's different, which I'm starting to warm up to as an option.

vrothberg:
The field says "digest", so I expect a digest, not an architecture.  I think the architecture is harder to find like that.  Having a dedicated field would help me find the information faster.

vrothberg:
Just a preference, no need to block at all :)

cgwalters:
> I guess the other option would be to maintain a map (amd64 -> x86_64, arm64 -> aarch64) for those two where it's different, which I'm starting to warm up to.

Oh we have that mapping code in a bunch of places; one is https://github.com/coreos/stream-metadata-go/blob/main/arch/arch.go

But I also added mapping from the OCI architecture to the Rust architecture (ref https://docs.rs/oci-spec/latest/src/oci_spec/image/mod.rs.html#435 ) which is *almost* the same thing as the GNU architectures except for the powerpc/ppc64le split (though I didn't look at RISC-V which is probably its own ball of fun).

Anyways though there's I think a strong argument to stick with the OCI (Go) arch here; if you want the other one it's just a `uname` away (and the Rust architecture name isn't as prominent/relevant except it's just what one has by default in a Rust program, not the other two, so we need to deal with mapping *from* it mainly).

cgwalters:
As far as positioning/layout it's for sure bikeshed-able but I prefer this as is.

I'm more interested in aligning higher level tooling on the problem reporting domain, I don't think `bootc status` should grow too much into that space right now; it should just report what it controls mainly.

dustymabe:
> Anyways though there's I think a strong argument to stick with the OCI (Go) arch here

👍

> As far as positioning/layout it's for sure bikeshed-able but I prefer this as is.

👍

dustymabe (approved):
/lgtm

PR #1242, opened 2025-03-30
Update to rustix 1.0

cgwalters:
No major changes required; we get to drop one `unsafe` call which is nice!

Looks like the ecosystem overall will need a fair bit of porting before we can drop the 0.38 version though.


jmarrero (approved):
lgtm

PR #1245, opened 2025-04-01
install: Error out when not enough disk space to pull image

ckyrouac:
Verify there is enough space to pull the bootc image on the ostree repo's disk. Adds pretty_bytes to print the error message in human readable format.

### DIFF
--- a/lib/src/deploy.rs
+++ b/lib/src/deploy.rs
@@ -14,7 +14,7 @@ use fn_error_context::context;
 use ostree::{gio, glib};
 use ostree_container::OstreeImageReference;
 use ostree_ext::container as ostree_container;
-use ostree_ext::container::store::{ImportProgress, PrepareResult};
+use ostree_ext::container::store::{ImageImporter, ImportProgress, PrepareResult, PreparedImport};
 use ostree_ext::oci_spec::image::{Descriptor, Digest};
 use ostree_ext::ostree::Deployment;
 use ostree_ext::ostree::{self, Sysroot};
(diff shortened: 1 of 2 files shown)


cgwalters (changes requested):
Thanks for working on this!

cgwalters:
Forgot to mention on tests...I think we can mock this out with a manifest that just claims to have large layers, without actually creating one a truly large image by fetching from an `oci` directory.

lib/src/deploy.rs
```diff
@@ -332,6 +333,25 @@ pub(crate) async fn pull(
     let layers_total = prep.all_layers().count();
     let bytes_to_fetch: u64 = layers_to_fetch.iter().map(|(l, _)| l.layer.size()).sum();
     let bytes_total: u64 = prep.all_layers().map(|l| l.layer.size()).sum();
+    tracing::trace!("image {imgref:#} bytes_to_fetch: {bytes_to_fetch}");
+    tracing::trace!("image {imgref:#} bytes_total: {bytes_total}");
+
+    {
+        let borrowed_fd = unsafe { std::os::fd::BorrowedFd::borrow_raw(repo.dfd()) };
+        let stat = rustix::fs::fstatvfs(borrowed_fd)?;
```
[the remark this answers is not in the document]
ckyrouac: hmm, I'm probably missing something obvious but calling `repo.as_fd()` results in `method cannot be called on &Repo due to unsatisfied trait bounds`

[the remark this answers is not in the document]
cgwalters: Ah sorry it's `.dfd_borrow()` for `ostree::Repo`.

That said, a repo is always going to be on the root filesystem, and we have a fd for that in `root_setup.physical_root` already.

lib/src/install.rs
```diff
@@ -751,21 +773,30 @@ async fn install_container(
...
-        repo.set_disable_fsync(true);
-        let r = crate::deploy::pull(
+    let spec_imgref = ImageReference::from(src_imageref.clone());
+    let repo = &sysroot.repo();
+    repo.set_disable_fsync(true);
+    let (prepared_image, existing_image) =
+        crate::deploy::prepare_for_pull(repo, &spec_imgref, Some(&state.target_imgref)).await?;
+
+    let pulled_image = if existing_image.is_some() {
+        existing_image.unwrap()
```
cgwalters, line 783: This is better written as
```
    let pulled_image = if let Some(img) = existing_image {
        img
    } else {
      ....
}

Comment (ckyrouac) [reply lib/src/install.rs]: ah thanks, I just force pushed a cleanup of these conditionals. I think I get how this is supposed to work now, but the extra `else { anyhow::bail!}` seems off

File: lib/src/deploy.rs [hunk 3]
```diff
@@ -382,23 +428,26 @@ pub(crate) async fn pull(
...
+
+    if let Some(existing) = existing {
+        return Ok(existing);
+    }
+
+    if let Some(prepared_image) = prepared_image {
+        Ok(pull_from_prepared(repo, imgref, target_imgref, quiet, prog, prepared_image).await?)
+    } else {
+        //this shouldn't be possible
+        anyhow::bail!("Error while preparing image for pull.");
```
cgwalters, line 449: Don't usually include periods in errors

lib/src/deploy.rs
```diff
@@ -301,15 +301,41 @@ async fn handle_layer_progress_print(
...
+
+#[context("Prepare for pull")]
+pub(crate) async fn prepare_for_pull(
     repo: &ostree::Repo,
     imgref: &ImageReference,
     target_imgref: Option<&OstreeImageReference>,
-    quiet: bool,
-    prog: ProgressWriter,
-) -> Result<Box<ImageState>> {
+) -> Result<(Option<PreparedImagePull>, Option<Box<ImageState>>)> {
```
cgwalters, line 338: We discussed converting this to an `enum`; maybe we can either just return the existing one (maybe?) or make a new thin wrapper like this

```
enum PullPreparedState {
   Ready(PreparedImagePull)
   AlreadyPresent(Box<ImageState>)
}
```

ckyrouac: switched this to an enum. I had to use a wrapper to at least include the `ImageImporter` and `Box<PreparedImport>` in the response.

cgwalters (approved):
LGTM, thanks!!

PR #1247, opened 2025-04-01
unencapsulate: Wrap decompressor and implement Drop

jeckersb:
Followup to #1228

We use the decompressor a couple of times in unencapsulate_base in
addition to the previous fix in filter_tar_async.  Instead of
duplicating the decompressor flush logic in each place, wrap the
decompressor in a struct and implement Drop for it so we always
automatically flush the input stream when we're finished with it.

### DIFF
--- a/ostree-ext/src/container/unencapsulate.rs
+++ b/ostree-ext/src/container/unencapsulate.rs
@@ -191,11 +191,47 @@ pub async fn unencapsulate(repo: &ostree::Repo, imgref: &OstreeImageReference) -
     importer.unencapsulate().await
 }
 
+pub(crate) struct Decompressor {
+    inner: Box<dyn Read + Send + 'static>,
+}
+
+impl Read for Decompressor {
+    fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
+        self.inner.read(buf)
+    }
+}
+
+impl Drop for Decompressor {
+    fn drop(&mut self) {
+        // We need to make sure to flush out the decompressor and/or
+        // tar stream here.  For tar, we might not read through the
+        // entire stream, because the archive has zero-block-markers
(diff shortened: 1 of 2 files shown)


ostree-ext/src/container/unencapsulate.rs
```diff
@@ -191,11 +191,47 @@ pub async fn unencapsulate(repo: &ostree::Repo, imgref: &OstreeImageReference) -
...
+        // it here then on the skopeo proxy we'll block trying to
+        // write the end of the stream.  That in turn will block our
+        // client end trying to call FinishPipe, and we end up
+        // deadlocking ourselves through skopeo.
+        //
+        // https://github.com/bootc-dev/bootc/issues/1204
+
+        let mut sink = std::io::sink();
+        match std::io::copy(&mut self.inner, &mut sink) {
+            Err(e) => tracing::debug!("Ignoring error while dropping decompressor: {e}"),
```
cgwalters, line 223: How about as a followup we add a method `fn finish(self) -> std::io::Result<()>` so we can propagate any errors here instead of ignoring them?

This is a general pattern in Rust, a similar reference example is https://doc.rust-lang.org/std/io/struct.BufWriter.html#method.into_inner

Then we update the callers of `decompress` to invoke `.finish()?;` by default.

cgwalters (approved):

cgwalters:
I'd be a bit happier with a dedicated test but LGTM

ostree-ext/src/container/unencapsulate.rs
cgwalters: See https://github.com/bootc-dev/bootc/blob/7593910951af2b027e3752cd9d76b4d120401f82/blockdev/src/blockdev.rs#L221 for prior art

PR #1254, opened 2025-04-07
build(deps): bump openssl to 0.10.72

jmarrero:


cgwalters (approved):

cgwalters:
We chatted about this and I don't understand why renovate hasn't queued an update. The dashboard is here https://github.com/bootc-dev/bootc/issues/335
This may need some debugging

PR #1255, opened 2025-04-08
reinstall: Use containers-storage transport

### LINKED ISSUE #1205: Add one arg like `--target-transport containers-storage` in `system-reinstall-bootc` commnad
To install from private image repo, the `auth.json` has to be built into image. If `system-reinstall-bootc` supports install from local container storage, customer can `poamdn login` frist, then `podman pull`, and run `system-reinstall-bootc`. That can avoid one more image building.

`bootc install to-existing-root` already supports `--target-transport containers-storage`.

ckyrouac:
The image is always pulled into containers-storage first, so let's use it here to avoid requiring registry keys to be baked into the image.

fixes #1205

### DIFF
--- a/system-reinstall-bootc/src/podman.rs
+++ b/system-reinstall-bootc/src/podman.rs
@@ -41,6 +41,9 @@ pub(crate) fn reinstall_command(image: &str, ssh_key_file: &str) -> Command {
         // this binary. Since this is no longer an "arcane" bootc command, we can safely avoid this
         // timed warning prompt. TODO: Discuss in https://github.com/containers/bootc/discussions/1060
         "--acknowledge-destructive",
+        // The image is always pulled first, so let's avoid requiring the credentials to be baked
+        // in the image for this check.
+        "--skip-fetch-check",
     ]
     .map(String::from)
     .to_vec();


system-reinstall-bootc/src/podman.rs
```diff
@@ -41,6 +41,9 @@ pub(crate) fn reinstall_command(image: &str, ssh_key_file: &str) -> Command {
         // this binary. Since this is no longer an "arcane" bootc command, we can safely avoid this
         // timed warning prompt. TODO: Discuss in https://github.com/containers/bootc/discussions/1060
         "--acknowledge-destructive",
+        // The image is always pulled into containers-storage first, so let's use it here to avoid
+        // requiring registry keys to be baked into the image.
```
cgwalters, line 45: We want `--skip-fetch-check` here to avoid that.

And honestly I'm OK to just drop that check by default.

system-reinstall-bootc/src/podman.rs
```diff
@@ -41,6 +41,9 @@ pub(crate) fn reinstall_command(image: &str, ssh_key_file: &str) -> Command {
...
+        "--target-transport containers-storage",
```
cgwalters, line 46: Because this fundamentally changes what it does. Previously we'd look for updates from the registry, now we won't - to do an upgrade the admin would need to `podman pull` subsequent images which would be a huge change.

cgwalters (changes requested):

system-reinstall-bootc/src/podman.rs
ckyrouac: ah, right, I forgot that flag will affect upgrades. I switched to use `skip-fetch-check` for now and can follow up to make it the default.

cgwalters (approved):

system-reinstall-bootc/src/podman.rs
cgwalters: Of course, really this is all extremely confusing and https://github.com/containers/bootc/issues/20 would help a whole lot in all of this

PR #1256, opened 2025-04-08
install: Deprecate skip-fetch-check, replace with run-fetch-check

ckyrouac:
This deprecates skip-fetch-check in favor of the inverse, run-fetch-check. Updates docs and tests to reflect the change.

---

Spawned from https://github.com/bootc-dev/bootc/pull/1255#discussion_r2033391674. Opening this for discussion. Also, I'm not sure how to properly deprecate an argument so I just took a stab at it.

### DIFF
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -124,12 +124,14 @@ pub(crate) struct InstallTargetOpts {
     #[serde(default)]
     pub(crate) enforce_container_sigpolicy: bool,
 
-    /// By default, the accessiblity of the target image will be verified (just the manifest will be fetched).
-    /// Specifying this option suppresses the check; use this when you know the issues it might find
-    /// are addressed.
-    ///
-    /// A common reason this may fail is when one is using an image which requires registry authentication,
-    /// but not embedding the pull secret in the image so that updates can be fetched by the installed OS "day 2".
+    /// Verify the image can be fetched from the bootc image. Updates may fail when the installation
+    /// host is authenticated with the registry but the pull secret is not in the bootc image.
+    #[clap(long)]
+    #[serde(default)]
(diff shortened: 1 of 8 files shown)


lib/src/install.rs
```diff
@@ -1287,10 +1289,14 @@ async fn prepare_install(
     // And continue to init global state
     osbuild::adjust_for_bootc_image_builder(&rootfs, &tempdir)?;
 
-    if !target_opts.skip_fetch_check {
+    if target_opts.run_fetch_check {
         verify_target_fetch(&tempdir, &target_imgref).await?;
     }
 
+    if target_opts.skip_fetch_check {
+        tracing::warn!("--skip-fetch-check is deprecated and will be removed in a future version");
```
cgwalters, line 1297: I think when we called bootc 1.0 it's a promise to basically never remove APIs. Or at least, not unless they have *serious* problems. In this case the user asked for something and we are just doing it by default, I don't think we need to annoy them into dropping the flag.

We can pretty easily carry this little bit of code into the forseeable future I think, right?

One thing I might say though is that at some point in the future (after we can confidently rely on the new version of bootc being everywhere) we could add a little notice near the end of an install that says that `--skip-fetch-check` is no longer necessary.

Although even then we'd need to bear in mind that for many cases e.g. podman-bootc it's not the *user* that typed `--skip-fetch-check` it's the software they're using and so it basically turns into the same problem that happens with Python `DeprecationWarning` that because Python is (usually) intepreted on the end user system, it's end users that see it (and can't action it) in addition to developers, which would be good to avoid...

So in the end let's just drop the warning or downgrade to `debug!`?

cgwalters (changes requested):
Thanks for this!

cgwalters (approved):

PR #1257, opened 2025-04-08
tmpfiles: Minor PR followup nit

cgwalters:
It's clearer if `increment` mutates self.


jeckersb (approved):
Am I crazy or didn't we already do this once before?  Or maybe it was just a very similar change...

jeckersb:
Ok [I'm not crazy](https://github.com/bootc-dev/bootc/pull/1151#discussion_r1974236070), it just never made it in until now :+1:

cgwalters:
integration tests died on quay.io flakes; this really can't break those so just merging over red

PR #1258, opened 2025-04-08
feat(status): display usroverlay state in status

### LINKED ISSUE #474: `bootc status` does not report unlocked status
Running `bootc usoverlay` is not reflected in `bootc status` as far as I could see.

p5:
Closes #474

Displays the status of the usroverlay within bootc status.

Although not ideal, handles the displaying of the host's filesystem within the image/ostree functions since otherwise it may require the introduction of another "host" block within the status output (i.e. having a "host", "staged", "booted" and "rollback").

---

I have yet to add unit or integration tests to this PR as I'd like to get clarity on the approach I've taken.

This builds off Colin's code snippet on the original issue with (what in my mind) something that makes sense.

As previously mentioned, I'm new to Rust and Bootc development, so please do be harsh with the reviews.

### DIFF
--- a/lib/src/spec.rs
+++ b/lib/src/spec.rs
@@ -2,8 +2,8 @@
 
 use std::fmt::Display;
 
-use ostree_ext::container::OstreeImageReference;
 use ostree_ext::oci_spec::image::Digest;
+use ostree_ext::{container::OstreeImageReference, ostree::DeploymentUnlockedState};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
@@ -53,6 +53,41 @@ pub enum Store {
     OstreeContainer,
 }
 
+#[derive(
+    clap::ValueEnum, Serialize, Deserialize, Copy, Clone, Debug, PartialEq, Eq, JsonSchema,
+)]
+#[serde(rename_all = "camelCase")]
+/// The filesystem overlay type
+pub enum FilesystemOverlay {
+    /// Readonly overlay mode
+    Readonly,
+    /// Read-write overlay mode
+    ReadWrite,
+}
+
+impl FilesystemOverlay {
+    /// Convert from the ostree deployment state
+    pub fn from_ostree_deployment_state(state: &DeploymentUnlockedState) -> Option<Self> {
+        match state {
+            DeploymentUnlockedState::None => Some(Self::Readonly),
+            DeploymentUnlockedState::Development
+            | DeploymentUnlockedState::Transient
+            | DeploymentUnlockedState::Hotfix => Some(Self::ReadWrite),
+            // Default to readonly when unknown since it is the default
+            // for bootc deployments.
+            _ => Some(Self::Readonly),
+        }
+    }
+
+    /// Convert the FilesystemOverlay value to a human-readable string
+    pub fn to_human_string(&self) -> String {
+        match self {
+            Self::Readonly => "read-only".to_string(),
+            Self::ReadWrite => "read-write".to_string(),
+        }
+    }
+}
+
 #[derive(Serialize, Deserialize, Default, Debug, Clone, PartialEq, Eq, JsonSchema)]
 #[serde(rename_all = "camelCase")]
 /// The host specification
@@ -62,6 +97,9 @@ pub struct HostSpec {
     /// If set, and there is a rollback deployment, it will be set for the next boot.
     #[serde(default)]
     pub boot_order: BootOrder,
+    /// Matches the `ostree admin unlock` state
+    #[serde(default)]
+    pub usr_overlay: Option<FilesystemOverlay>,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, JsonSchema)]
@@ -157,6 +195,9 @@ pub struct HostStatus {
     /// Set to true if the rollback entry is queued for the next boot.
     #[serde(default)]
     pub rollback_queued: bool,
+    /// Matches the `ostree admin unlock` state
+    #[serde(default)]
+    pub usr_overlay: Option<FilesystemOverlay>,
 
     /// The detected type of system
     #[serde(rename = "type")]
--- a/lib/src/status.rs
+++ b/lib/src/status.rs
@@ -15,6 +15,7 @@ use ostree_ext::oci_spec;
 use ostree_ext::ostree;
 
 use crate::cli::OutputFormat;
+use crate::spec::FilesystemOverlay;
 use crate::spec::{BootEntry, BootOrder, Host, HostSpec, HostStatus, HostType};
 use crate::spec::{ImageReference, ImageSignature};
 use crate::store::{CachedImageStatus, ContainerImageStore, Storage};
@@ -225,6 +226,12 @@ pub(crate) fn get_status(
     } else {
         BootOrder::Default
     };
+    let usr_overlay = FilesystemOverlay::from_ostree_deployment_state(
+        &booted_deployment
+            .as_ref()
+            .expect("Expected a booted deployment")
+            .unlocked(),
+    );
     tracing::debug!("Rollback queued={rollback_queued:?}");
     let other = {
         related_deployments.extend(other_deployments);
@@ -260,6 +267,7 @@ pub(crate) fn get_status(
         .map(|img| HostSpec {
             image: Some(img.image.clone()),
             boot_order,
+            usr_overlay,
         })
         .unwrap_or_default();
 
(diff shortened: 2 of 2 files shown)


lib/src/spec.rs
```diff
@@ -157,6 +195,9 @@ pub struct HostStatus {
     /// Set to true if the rollback entry is queued for the next boot.
     #[serde(default)]
     pub rollback_queued: bool,
+    /// Matches the `ostree admin unlock` state
+    #[serde(default)]
+    pub usr_overlay: Option<FilesystemOverlay>,
```
p5, line 200: I'm debating whether this is needed as part of the HostStatus.

This might be better being _only_ part of the HostSpec, and treated similar to `ty` (`Type`).

```diff
diff --git a/lib/src/spec.rs b/lib/src/spec.rs
index 3a5e063..08ea264 100644
--- a/lib/src/spec.rs
+++ b/lib/src/spec.rs
@@ -53,9 +53,7 @@ pub enum Store {
     OstreeContainer,
 }
 
-#[derive(
-    clap::ValueEnum, Serialize, Deserialize, Copy, Clone, Debug, PartialEq, Eq, JsonSchema,
-)]
+#[derive(Serialize, Deserialize, Copy, Clone, Debug, PartialEq, Eq, JsonSchema)]
 #[serde(rename_all = "camelCase")]
 /// The filesystem overlay type
 pub enum FilesystemOverlay {
@@ -195,9 +193,6 @@ pub struct HostStatus {
     /// Set to true if the rollback entry is queued for the next boot.
     #[serde(default)]
     pub rollback_queued: bool,
-    /// Matches the `ostree admin unlock` state
-    #[serde(default)]
-    pub usr_overlay: Option<FilesystemOverlay>,
 
     /// The detected type of system
     #[serde(rename = "type")]
diff --git a/lib/src/status.rs b/lib/src/status.rs
index 433777b..6bb5b7a 100644
--- a/lib/src/status.rs
+++ b/lib/src/status.rs
@@ -288,7 +288,6 @@ pub(crate) fn get_status(
         booted,
         rollback,
         rollback_queued,
-        usr_overlay,
         ty,
     };
     Ok((deployments, host))
@@ -372,7 +371,7 @@ fn human_render_imagestatus(
     mut out: impl Write,
     slot: Slot,
     image: &crate::spec::ImageStatus,
-    host: &crate::spec::HostStatus,
+    host: &crate::spec::HostSpec,
 ) -> Result<()> {
     let transport = &image.image.transport;
     let imagename = &image.image.image;
@@ -433,7 +432,7 @@ fn human_render_ostree(
     mut out: impl Write,
     slot: Slot,
     ostree_commit: &str,
-    host: &crate::spec::HostStatus,
+    host: &crate::spec::HostSpec,
 ) -> Result<()> {
     // TODO consider rendering more ostree stuff here like rpm-ostree status does
     let prefix = match slot {
@@ -472,9 +471,9 @@ fn human_readable_output_booted(mut out: impl Write, host: &Host) -> Result<()>
                 writeln!(out)?;
             }
             if let Some(image) = &host_status.image {
-                human_render_imagestatus(&mut out, slot_name, image, &host.status)?;
+                human_render_imagestatus(&mut out, slot_name, image, &host.spec)?;
             } else if let Some(ostree) = host_status.ostree.as_ref() {
-                human_render_ostree(&mut out, slot_name, &ostree.checksum, &host.status)?;
+                human_render_ostree(&mut out, slot_name, &ostree.checksum, &host.spec)?;
             } else {
                 writeln!(out, "Current {slot_name} state is unknown")?;
```

lib/src/status.rs
```diff
@@ -407,10 +417,24 @@ fn human_render_imagestatus(
         writeln!(out, "{timestamp}")?;
     }
 
+    if let Some(usr_overlay) = &host.usr_overlay {
+        // Only show the usr filesystem overlay state if we are booted and not the default
+        // read-only mode.
+        if slot == Slot::Booted && *usr_overlay != FilesystemOverlay::Readonly {
+            write_row_name(&mut out, "Usr overlay", prefix_len)?;
```
cgwalters, line 424: How about `/usr overlay`?

lib/src/spec.rs
cgwalters: Note that we currently don't have an official way to go from unlocked back to locked (though in practice `umount -l /usr` will work...but loses the trace that we *did* have an overlay there and it's likely to still be in use unless processes get restarted)...that's its own nest of things.

This PR as is will do the right thing in reacting to a `bootc usroverlay` invocation and showing it in status. But the idea behind having this in a spec is that one should be able to `bootc edit` and have that change apply. That logic would need to live in `cli::edit`.

cgwalters (commented):
Thanks again for working on this!

lib/src/spec.rs
cgwalters: To be clear though this would be the first thing in `edit` that *wasn't* changing the image source...I'm OK to scope that out and only have this in `status` I guess to start.

p5: I'm certainly interested in giving the edit implementation a go.  

Just so I understand the HostSpec and HostStatus differences a bit - you're wanting the status object to contain the _current_ booted configuration, and the spec object to show the persisted desired state of the /usr overlay?  I was struggling to see how they would differ.

For example if I `unmount` while the spec defines I want a read/write overlay, spec should still show readwrite, but status should be readonly?

cgwalters: > Just so I understand the HostSpec and HostStatus differences a bit - you're wanting the status object to contain the current state, and the spec object to show the persisted desired state of the /usr overlay controlled through both unlock and edit? I was struggling to see how they would differ.

They would differ when one is trying to use `bootc edit` to go from default -> unlocked, right?

p5: When a user runs `bootc edit`, is this something that will only be applied after reboot, even if the operation (e.g. unlock) supports being applied on the current boot?

I would expect if I ran `bootc edit` and saved the configuration, that /usr would be unlocked instantly, thus not displaying different values.

Very sorry I'm just not understanding something and it's something very simple.

cgwalters: > When a user runs bootc edit, is this something that will only be applied after reboot, even if the operation (e.g. unlock) supports being applied on the current boot?

Note that today one can use `bootc edit` to do the equivalent of `bootc switch` and that will immediately update the status for the staged deployment.

> I would expect if I ran bootc edit and saved the configuration, that /usr would be unlocked instantly, thus not displaying different values in subsequent bootc status calls.

Yes, you're right that at all points the spec and status as presented by bootc will be the same, assuming an apply occurred.

All of this gets into a deeper, bigger picture question of whether the spec/status approach will scale as we try to do more things in general, and perhaps it won't, and we could just go forward in general with a model of imperative verbs and a declarative status.

p5:
This PR has been updated to also include changing the /usr overlay within `bootc edit`.

Overall, I'm quite pleased with how it turned out as it seems to fit the brief.  Certainly learned a lot from this one, even if it is quite basic.

Please do let me know all the small (or large) things that could be done to improve this.

lib/src/mount.rs
```diff
@@ -129,6 +129,21 @@ pub(crate) fn mount(dev: &str, target: &Utf8Path) -> Result<()> {
     )
 }
 
+pub fn unmount(target: &Utf8Path, lazy: bool) -> Result<()> {
```
cgwalters, line 132: This is OK as is because we are already forking in other places but my thought lately is we should really just switch to direct syscalls.

cgwalters (approved):
Again sorry about the delay in review, it's my bad for not getting back to this earlier. The code looks good!

Same as the other PR though it will need a rebase for the status unit test.

(Add `stateroot: default` to the `ostree` section)

lib/src/status.rs
```diff
@@ -363,6 +372,7 @@ fn human_render_imagestatus(
     mut out: impl Write,
     slot: Slot,
     image: &crate::spec::ImageStatus,
+    host: &crate::spec::HostStatus,
```
p5, line 375: @cgwalters Would you be able to explain what you meant by this?
This was in relation to passing through BootEntry instead of HostStatus to the render functions.

> we could pass down the metadata ultimately from ostree_deployment_get_unlocked() here.
https://github.com/bootc-dev/bootc/pull/1291#discussion_r2065157284

I've been trying to wrap my head around how to get the unlocked state from only the Slot, ImageStatus and BootEntry, but can't figure it out.

p5:
Diff is quite large with this one.
May pick this back up once the composefs stuff fully lands.

PR #1264, opened 2025-04-10
Redeploy bug fix

ckyrouac:
see commit messages. This is to fix the bug where running `install to-existing-root` continually fails after an initial failure due to low disk space.

### DIFF
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -678,16 +678,24 @@ async fn initialize_ostree_root(
         ostree_ext::fsverity::ensure_verity(repo).await?;
     }
 
-    let stateroot_exists = rootfs_dir.try_exists(format!("ostree/deploy/{stateroot}"))?;
-    ensure!(
-        !stateroot_exists,
-        "Cannot redeploy over extant stateroot {stateroot}"
-    );
-    sysroot
-        .init_osname(stateroot, cancellable)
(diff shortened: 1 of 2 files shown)


lib/src/install.rs
```diff
@@ -678,20 +678,27 @@ async fn initialize_ostree_root(
...
+    }
+
+    if let Some(booted) = sysroot.booted_deployment() {
+        if stateroot == booted.stateroot() {
+            anyhow::bail!("Cannot redeploy over booted stateroot {stateroot}");
+        }
+    }
+
+    sysroot
+        .init_osname(stateroot, cancellable)
```
cgwalters, line 699: Ah can you try just making this function call idempotent? Arguably actually it should be so in libostree... but anyways for now basically just:

```
let stateroot_path = format!("ostree/deploy/{stateroot}");
if !sysroot_dir.try_exists(stateroot_path)? {
  sysroot.init_osname(...)
}
``` 

or so? Then I don't think we need to do any extra cleanup in the `cleanup` function.

ckyrouac: nice, thanks! That's simpler

cgwalters (approved):

cgwalters:
Here's a PR I did related to this https://github.com/coreos/rpm-ostree/pull/4259

PR #1266, opened 2025-04-11
various: adapt to new composefs-rs fsverity API

allisonkarlitskaya:
~...and a bunch of probably-unnecessary changes to cargo lockfiles~

~Don't merge this yet: the composefs-rs revision isn't merged yet.~

 - [x] https://github.com/containers/composefs-rs/pull/102
 - [x] investigate version requirements in composefs-rs: https://github.com/containers/composefs-rs/pull/103
 - [x] update composefs-rs revision in this PR

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -113,6 +113,8 @@ dependencies = [
  "memchr",
  "pin-project-lite",
  "tokio",
+ "zstd",
+ "zstd-safe",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


cgwalters:
Regarding the scale of lockfile changes:

For now I'd say probably just drop the "hard version requirement" bump in composefs-rs (except where it's *actually* required like in rustix).

Ideally we update (minor) dependencies on roughly the same schedule e.g. once a month, and *in the same way* (either bump Cargo.toml or don't, currently in bootc we haven't been).

It's also not right of course to *not* update the Cargo.toml requirements...but most library crates are conservative in this. The thing is of course to be really sure you need to [actually test it](https://github.com/taiki-e/cargo-minimal-versions) which some popular crates I think do...

cgwalters (approved):

PR #1268, opened 2025-04-11
Reinstall prompt tweaks

ckyrouac:
See commit messages, just some prompt tweaks to clarify usage of system-reinstall-bootc.

Closes: https://github.com/bootc-dev/bootc/issues/1269

### DIFF
--- a/system-reinstall-bootc/src/main.rs
+++ b/system-reinstall-bootc/src/main.rs
@@ -52,6 +52,10 @@ fn run() -> Result<()> {
         .run_with_cmd_context()
         .context("running reinstall command")?;
 
+    prompt::reboot()?;
+
+    std::process::Command::new("reboot").run()?;
+
     Ok(())
 }
 
--- a/system-reinstall-bootc/src/prompt.rs
+++ b/system-reinstall-bootc/src/prompt.rs
@@ -12,7 +12,8 @@ fn prompt_single_user(user: &crate::users::UserKeys) -> Result<Vec<&crate::users
     let prompt = indoc::formatdoc! {
         "Found only one user ({user}) with {num_keys} SSH authorized keys.
         Would you like to import its SSH authorized keys
-        into the root user on the new bootc system?",
+        into the root user on the new bootc system?
+        Then you can login as root@ using those keys.",
         user = user.user,
         num_keys = user.num_keys(),
     };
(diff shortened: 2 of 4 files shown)


system-reinstall-bootc/src/main.rs
```diff
@@ -52,6 +52,9 @@ fn run() -> Result<()> {
         .run_with_cmd_context()
         .context("running reinstall command")?;
 
+    println!();
+    println!("Reboot to start using the bootc image");
```
cgwalters, line 56: Just thinking about this more...I wonder if we should instead default to rebooting, with the option to press e.g. ctrl-c/escape to exit?

Or even thinking about this more...we already prompt them in `temporary_developer_protection_prompt` "hey this will replace your system" earlier so I think we could actually change it to say...

```
diff --git i/system-reinstall-bootc/src/prompt.rs w/system-reinstall-bootc/src/prompt.rs
index 6308450a..ec73ae85 100644
--- i/system-reinstall-bootc/src/prompt.rs
+++ w/system-reinstall-bootc/src/prompt.rs
@@ -44,7 +44,7 @@ pub(crate) fn temporary_developer_protection_prompt() -> Result<()> {
     // Print an empty line so that the warning stands out from the rest of the output
     println!();
 
-    let prompt = "THIS WILL REINSTALL YOUR SYSTEM! Are you sure you want to continue?";
+    let prompt = "NOTICE: This will replace the installed operating system and reboot. Are you sure you want to continue?";
     let answer = ask_yes_no(prompt, false)?;
 
     if !answer {
```


And then here at the end we have a timer like:

> Operation complete, rebooting in {n} seconds. Press Ctrl-C to cancel reboot, any other key to continue immediately.

With a value of n=10 perhaps?

system-reinstall-bootc/src/prompt.rs
```diff
@@ -9,7 +9,8 @@ fn prompt_single_user(user: &crate::users::UserKeys) -> Result<Vec<&crate::users
     let prompt = indoc::formatdoc! {
         "Found only one user ({user}) with {num_keys} SSH authorized keys.
         Would you like to import its SSH authorized keys
-        into the root user on the new bootc system?",
+        into the root user on the new bootc system?
+        This is the key that will be used to access the bootc system.",
```
cgwalters, line 13: "bootc system" stutters here. I wonder if we even need to clarify "bootc" system in this context, I think it's implicitly understood. Maybe:

> Would you like to copy the SSH authorized keys for the root user on the new system? Then you can log in as `root@` using those keys.

cgwalters (approved):
Nothing blocking really just, OK to merge as is

cgwalters (approved):

cgwalters:
We were playing with this but it isn't quite right

```
+pub(crate) async fn reboot() -> Result<()> {
+    let delay_seconds = 10;
+    println!(
+        "Operation complete, rebooting in {delay_seconds} seconds. Press Ctrl-C to cancel reboot, or press enter to continue immediately.",
+    );
+
+    let event = crossterm::event::EventStream::new();
+    tokio::pin!(event);
+    let event = event.next();
+    let timeout = tokio::time::timeout(Duration::from_secs(delay_seconds), event);
+    timeout.await?;
+    Ok(())
+}
+
```

cgwalters:
quay.io bad gateway is ruining our CI :frowning:

cgwalters (approved):

PR #1279, opened 2025-04-16
test: Use to-existing-root --acknowledge-destructive

evan-goode:
In the test suite, pass the `--acknowledge-destructive` flag to `bootc install to-existing-root` to avoid the 20 second timer.


cgwalters (approved):
Thank you!

PR #1283, opened 2025-04-24
Port from once_cell to std

cgwalters:
No major reason, just a drive by thing after discussion of once_cell in composefs-rs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1701,7 +1701,6 @@ dependencies = [
  "libsystemd",
  "ocidir",
  "olpc-cjson",
- "once_cell",
  "openssl",
  "ostree",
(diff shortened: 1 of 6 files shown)


jeckersb (approved):
:+1:

cgwalters:
> , it looks like [systemd-devel-252-53.el9](https://kojihub.stream.centos.org/koji/rpminfo?rpmID=2092735) isn't available for some reason...

This is because https://gitlab.com/redhat/centos-stream/containers/bootc/-/issues/1174

The real solution is [strictly binding rpms and containers](https://github.com/rpm-software-management/dnf5/issues/833#issuecomment-2000244607).

PR #1288, opened 2025-04-27
Split mount code into separate helper crate

cgwalters:
Prep for using this elsewhere via git dependency, like we're doing now with bootupd for example.


jmarrero (approved):
lgtm

PR #1289, opened 2025-04-28
Add alongside-cleanup systemd service

ckyrouac:
This will remove all rpms and containers/images from the previous install (found in /sysroot) on reboot after running `install to-existing-root.`

---

WIP: using this PR to test the rpm build

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,13 @@ install:
 	# Copy dracut and systemd config files
 	cp -Prf baseimage/dracut $(DESTDIR)$(prefix)/share/doc/bootc/baseimage/dracut
 	cp -Prf baseimage/systemd $(DESTDIR)$(prefix)/share/doc/bootc/baseimage/systemd
+	# Install fedora-bootc-destructive-cleanup in fedora derivatives 
+	ID=$$(. /usr/lib/os-release && echo $$ID); \
+	ID_LIKE=$$(. /usr/lib/os-release && echo $$ID_LIKE); \
+	if [ "$$ID" = "fedora" ] || [[ "$$ID_LIKE" == *"fedora"* ]]; then \
+	ln -s ../bootc-destructive-cleanup.service $(DESTDIR)/$(prefix)/lib/systemd/system/multi-user.target.wants/bootc-destructive-cleanup.service; \
+	install -D -m 0755 -t $(DESTDIR)/$(prefix)/lib/bootc contrib/scripts/fedora-bootc-destructive-cleanup; \
+	fi
 
 # Run this to also take over the functionality of `ostree container` for example.
 # Only needed for OS/distros that have callers invoking `ostree container` and not bootc.
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -69,6 +69,8 @@ const BOOT: &str = "boot";
 const RUN_BOOTC: &str = "/run/bootc";
 /// The default path for the host rootfs
 const ALONGSIDE_ROOT_MOUNT: &str = "/target";
+/// Global flag to signal the booted system was provisioned via an alongside bootc install
+const DESTRUCTIVE_CLEANUP: &str = "bootc-destructive-cleanup";
 /// This is an ext4 special directory we need to ignore.
 const LOST_AND_FOUND: &str = "lost+found";
 /// The filename of the composefs EROFS superblock; TODO move this into ostree
@@ -334,6 +336,11 @@ pub(crate) struct InstallToExistingRootOpts {
     /// Accept that this is a destructive action and skip a warning timer.
     #[clap(long)]
     pub(crate) acknowledge_destructive: bool,
+
+    /// Add the bootc-destructive-cleanup systemd service to delete files from
+    /// the previous install on first boot
+    #[clap(long)]
+    pub(crate) cleanup: bool,
 
     /// Path to the mounted root; this is now not necessary to provide.
     /// Historically it was necessary to ensure the host rootfs was mounted at here
@@ -1460,7 +1467,11 @@ impl BoundImages {
     }
 }
 
-async fn install_to_filesystem_impl(state: &State, rootfs: &mut RootSetup) -> Result<()> {
+async fn install_to_filesystem_impl(
+    state: &State,
+    rootfs: &mut RootSetup,
+    cleanup: Cleanup,
+) -> Result<()> {
     if matches!(state.selinux_state, SELinuxFinalState::ForceTargetDisabled) {
         rootfs.kargs.push("selinux=0".to_string());
     }
@@ -1489,6 +1500,7 @@ async fn install_to_filesystem_impl(state: &State, rootfs: &mut RootSetup) -> Re
     let bound_images = BoundImages::from_state(state).await?;
 
     // Initialize the ostree sysroot (repo, stateroot, etc.)
+
     {
         let (sysroot, has_ostree, imgstore) = initialize_ostree_root(state, rootfs).await?;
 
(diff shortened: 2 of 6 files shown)


contrib/packaging/bootc.spec
```diff
@@ -72,6 +72,7 @@ Recommends: podman
 This package provides a utility to simplify reinstalling the current system to a given bootc image.
 
 %global system_reinstall_bootc_install_podman_path %{_prefix}/lib/system-reinstall-bootc/install-podman
+%global install_alongside_cleanup %{_prefix}/lib/bootc/install-alongside-cleanup
```
cgwalters, line 75: Missing `git add`

contrib/packaging/bootc.spec
```diff
@@ -115,13 +117,26 @@ sed -i -e '/https:\/\//d' cargo-vendor.txt
...
 mkdir -p %{buildroot}/%{dirname:%{system_reinstall_bootc_install_podman_path}}
 cat >%{?buildroot}/%{system_reinstall_bootc_install_podman_path} <<EOF
 #!/bin/bash
 exec dnf -y install podman
 EOF
 chmod +x %{?buildroot}/%{system_reinstall_bootc_install_podman_path}
 
+mkdir -p %{buildroot}/%{dirname:%{install_alongside_cleanup}}
+cat >%{?buildroot}/%{install_alongside_cleanup} <<EOF
+#!/bin/bash
```
cgwalters, line 130: Can you put this in e.g. `contrib/` or so and have it installed by `make install`?

(And we want to detect the OS version in `make` (or `make install`) and only install these things if we're on a Fedora derivative; parse `/usr/lib/os-release` and check `ID_LIKE=fedora`)

systemd/bootc-install-alongside-cleanup.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Cleanup previous the installation after an alongside installation
+Documentation=man:bootc(8)
+ConditionPathExists=/sysroot/.bootc-installed-alongside
```
cgwalters, line 4: How about `ConditionPathExists=/sysroot/etc/bootc-destructive-cleanup`?

Then let's also per discussion change system-reinstall-bootc to create this file (which note from *its* perspective is the plain `/etc/bootc-destructive-cleanup`), so it doesn't apply by default to bootc install to-existing-root?

ckyrouac: so I implemented an extra option (—cleanup) before reading this comment. I pushed that implementation here. What do you think about using a cli flag so the to-existing-root flow can also do the cleanup in a more straightforward way?

contrib/scripts/bootc-destructive-cleanup
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
cgwalters, line 1: WDYT if we called this e.g. `fedora-bootc-destructive-cleanup`, leaving space for someone to put in an arch/deb/etc version later?

contrib/scripts/bootc-destructive-cleanup
```diff
@@ -0,0 +1,8 @@
...
+set -xeuo pipefail
+
+mount -o remount,rw /sysroot
+rpm -qa --root=/sysroot --dbpath=/usr/lib/sysimage/rpm | xargs rpm -e --root=/sysroot --dbpath=/usr/lib/sysimage/rpm
+
+mount --bind -o rw /sysroot/var/lib/containers /var/lib/containers
```
cgwalters, line 7: Also I guess conceptually since podman is OS independent in theory this could be a shared fragment, but let's not overengineer now, just noting.

Makefile
```diff
@@ -37,6 +37,13 @@ install:
 	# Copy dracut and systemd config files
 	cp -Prf baseimage/dracut $(DESTDIR)$(prefix)/share/doc/bootc/baseimage/dracut
 	cp -Prf baseimage/systemd $(DESTDIR)$(prefix)/share/doc/bootc/baseimage/systemd
+	# Install bootc-destructive-cleanup in fedora derivatives 
+	@ID=$$(grep '^ID=' /usr/lib/os-release | cut -d'=' -f2 | tr -d '"'); \
+	ID_LIKE=$$(grep '^ID_LIKE=' /usr/lib/os-release | cut -d'=' -f2 | tr -d '"'); \
+	if [ "$$ID" = "fedora" ] || [ "$$ID_LIKE" = "fedora" ]; then \
```
cgwalters, line 43: `ID_LIKE` needs to be a substring match
```
$ podman run --rm quay.io/centos-bootc/centos-bootc:stream10 grep ID_LIKE /usr/lib/os-release
ID_LIKE="rhel fedora"
$
```

Maybe we can factor out a `lib/install.sh` or something?

Makefile
```diff
@@ -37,6 +37,13 @@ install:
 	# Copy dracut and systemd config files
 	cp -Prf baseimage/dracut $(DESTDIR)$(prefix)/share/doc/bootc/baseimage/dracut
 	cp -Prf baseimage/systemd $(DESTDIR)$(prefix)/share/doc/bootc/baseimage/systemd
+	# Install bootc-destructive-cleanup in fedora derivatives 
+	@ID=$$(grep '^ID=' /usr/lib/os-release | cut -d'=' -f2 | tr -d '"'); \
```
cgwalters, line 41: The os-release file was explicitly designed to be parsable by actually just `.` by a posix shell (i.e. sourcing it). See e.g. `OS_ID=$(. /usr/lib/os-release && echo $ID)` which we do in `installdeps.sh`

lib/src/install.rs
```diff
@@ -1950,7 +1961,17 @@ pub(crate) async fn install_to_filesystem(
         skip_finalize,
     };
 
-    install_to_filesystem_impl(&state, &mut rootfs).await?;
+    let sysroot = install_to_filesystem_impl(&state, &mut rootfs).await?;
+
+    if cleanup_previous_install {
+        let sysroot_dir = crate::utils::sysroot_dir(&sysroot.sysroot)?;
+        if matches!(fsopts.replace, Some(ReplaceMode::Alongside))
+            && !sysroot_dir.try_exists(format!("etc/{}", DESTRUCTIVE_CLEANUP))?
```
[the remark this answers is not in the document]
ckyrouac: yep that check is unnecessary, removed it

cgwalters (approved):

cgwalters:
Hmm, something is still leaking in the install tests. This test passed on other PRs. I'm not seeing the bug though.

lib/src/install.rs
```diff
@@ -1570,7 +1589,15 @@ pub(crate) async fn install_to_disk(mut opts: InstallToDiskOpts) -> Result<()> {
         (rootfs, loopback_dev)
     };
 
-    install_to_filesystem_impl(&state, &mut rootfs).await?;
+    install_to_filesystem_impl(&state, &mut rootfs, Cleanup::Skip).await?;
+
+    // Finalize mounted filesystems
+    if !rootfs.skip_finalize {
```
cgwalters, line 1595: Don't we also need to delete the copy of this in `install_to_filesystem_impl`?

ckyrouac: ah whoops, forgot to push removing those extra calls I was trying out. Updated now.

lib/src/install.rs
```diff
@@ -1502,9 +1514,16 @@ async fn install_to_filesystem_impl(state: &State, rootfs: &mut RootSetup) -> Re
             &imgstore,
         )
         .await?;
+
+        if matches!(cleanup, Cleanup::TriggerOnNextBoot) {
+            let sysroot_dir = crate::utils::sysroot_dir(&sysroot)?;
```
cgwalters, line 1519: This is fine as is, really. It's just interesting to note that it's equivalent to `rootfs.physical_root` and I think that one's clearer. would probably make sense to try to port more things to it.

cgwalters (approved):
OK this version is even simpler! Nice work.

Just thinking about testing...right now we don't quite have a suite that would cover this flow, though we should. But, something for this could go in https://gitlab.com/fedora/bootc/tests/bootc-workflow-test/ probably too.

Hm, I guess really this is just about an e2e test for s-r-b really, since we added this cleanup option by default there.

PR #1290, opened 2025-04-28
lints: move misplaced docstring for LINT_VAR_RUN

jeckersb:
I guess this got misplaced somewhere along the way while reorganizing
the lint code, but I just now noticed it while browsing the code.

Also removed bit about "if it does not exist error", because that's
not accurate with the present behavior of the lint.

### DIFF
--- a/lib/src/lints.rs
+++ b/lib/src/lints.rs
@@ -299,9 +299,6 @@ fn lint_inner<'skip>(
     })
 }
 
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
Could also probably drop the comment entirely now that lints are required to have inline descriptions, but fine as is obviously!

PR #1291, opened 2025-04-28
status: Prep work for https://github.com/bootc-dev/bootc/pull/1285

cgwalters:
See https://github.com/bootc-dev/bootc/pull/1285

### DIFF
--- a/lib/src/status.rs
+++ b/lib/src/status.rs
@@ -362,6 +362,7 @@ fn write_row_name(mut out: impl Write, s: &str, prefix_len: usize) -> Result<()>
 fn human_render_slot(
     mut out: impl Write,
     slot: Slot,
+    entry: &crate::spec::BootEntry,
     image: &crate::spec::ImageStatus,
 ) -> Result<()> {
     let transport = &image.image.transport;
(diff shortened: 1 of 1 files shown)


lib/src/status.rs
p5, line 1: I've quickly changed a local version of #1285 to use the changes in this PR, and it implemented in a much cleaner way.

Although this still doesn't allow us to access data stored within the HostStatus objects, which would be needed for displaying the `/usr` overlay status.  
Unless you're thinking that can also be moved to be stored under BootEntry?  Or am I completely overthinking the implementation again?

cgwalters: > Although this still doesn't allow us to access data stored within the HostStatus objects, which would be needed for displaying the /usr overlay status (https://github.com/bootc-dev/bootc/pull/1258).
Unless you're thinking that can also be moved to be stored under BootEntry? Or am I completely overthinking the implementation again (a common occurrence with my recent PRs here)?

As is today, a usroverlay state is transient by default. It'd only be hotfix ones that we need to maintain persistently, but I've been trying to really discourage use of the hotfix path.

But anyways yes, we could pass down the metadata ultimately from `ostree_deployment_get_unlocked()` here.

jmarrero (approved):
lgtm

Issue #1292, opened 2025-04-29
`Err: EvalSymlinks: too many links` when using skopeo to pull images


Johan-Liebert1:
Skopeo uses `/var/tmp` as a tmpdir while pulling images and in bootc we symlink `/var/tmp` to `/proc/1/root/var/tmp`. 

`EvalSymlinks` function in the Go compiler doesn't understand how to handle this and throws the error `Err: EvalSymlinks: too many links`

cgwalters:
Previously in https://github.com/bootc-dev/bootc/pull/919 we switched to dynamic bind mounts for other cases, I think it would make sense to do the same for `/var/tmp`. That will make it an actual mount and avoid this symlink issue.

cgwalters:
But one thing I don't understand here is; why are we not hitting this code path currently? What changes have you made to trigger it?

Johan-Liebert1:
It's not on the bootc end. This error popped up when integrating with https://github.com/containers/composefs-rs, which uses https://github.com/containers/containers-image-proxy-rs. 

I've only skimmed through the bootc code, but I see that we're always creating a directory inside of `/var/tmp` before passing it onto skopeo; maybe that's why we haven't faced this issue before? ComposefsRs does no such thing and lets skopeo, or rather https://github.com/containers/containers-image-proxy-rs handle everything. 

Again, I might be wrong with the above assumption

cgwalters:
>  I see that we're always creating a directory inside of /var/tmp before passing it onto skopeo

Where is that? I don't see it.

One thing I do see is the `unprivileged_subprocess()` code which could be having an effect here. I think it's a pretty good idea to do this by default; we could refactor to share code there, *or* change composefs-rs to support passing a proxy config.  In fact we will *definitely* need to support passing a proxy config in order to honor `/etc/ostree/auth.json` etc.

Johan-Liebert1:
> Where is that? I don't see it.
I assumed wrong, I guess. 

I'm **guessing**, not a 100% sure, `skopeo::copy` doesn't actually use the tmpdir to do anything, else it would've also run into this issue.

I traced my issue down to https://github.com/containers/skopeo/blob/main/vendor/github.com/containers/image/v5/oci/layout/oci_transport.go#L89

which calls `EvalSymlinks` internally on `/var/tmp` and then fails.

Johan-Liebert1:
[...]

PR #1294, opened 2025-05-02
install: Do a dynamic mount for /var/tmp

cgwalters:
Closes: https://github.com/bootc-dev/bootc/issues/1292

Basically we were doing the `/proc/1/root/var/tmp` trick for `/var/tmp` because we didn't have the dynamic bind mount infrastructure before. Now we do, so use it instead.

The specific motivation is that Go in some cases uses `EvalSymlinks` which gets confused by the `/proc/<pid>/root` magic links.

Also, this deletes a lot of code.

### DIFF
--- a/lib/src/install.rs
+++ b/lib/src/install.rs
@@ -1103,22 +1103,11 @@ fn require_host_userns() -> Result<()> {
     Ok(())
 }
 
-// Ensure the `/var` directory exists.
-fn ensure_var() -> Result<()> {
-    std::fs::create_dir_all("/var")?;
-    Ok(())
-}
-
-/// We want to have proper /tmp and /var/tmp without requiring the caller to set them up
-/// in advance by manually specifying them via `podman run -v /tmp:/tmp` etc.
-/// Unfortunately, it's quite complex right now to "gracefully" dynamically reconfigure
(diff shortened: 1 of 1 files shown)


jmarrero (approved):
lgtm

cgwalters:
Man quay.io 502 Bad Gateway continues to be a big downer in CI

cgwalters:
> ./tmt-test-wrapper-inner.sh-tmt-tests-test-21-logically-bound-switch-2-default-0-default-0: line 1: nu: command not found

Hmmmm....this is really concerning. I wonder if it's the same bug @jmarrero is seeing.

henrywang:
It's not the nu issue, but a timeout. That needs a re-run.

cgwalters:
> It's not the nu issue, but a bootc-install-provision timeout. That needs a re-run.

I don't quite understand how a timeout could lead to "command not found" but I will take your word for it!

henrywang:
The timeout was happened in `bootc-install-provision` and before `nu` installation. So the `nu` was not installed due to timeout. That caused the test failure.
There should be a setting in TMT to `do not run the rest of tests if previous one timeout`. That setting can avoid confused.

PR #1295, opened 2025-05-02
hack: Rework+cleanup container build

cgwalters:
Main motivation: I was looking at making more changes here

- Use an idiom I'd like to standardize more of copy context to `FROM scratch` image which is then mounted and consumed in other phases by mounting. This helps avoid polluting later containers with intermediate copied files.
- Change `build.sh` to handle being run from any directory
- Drop the `dev-rootfs` stuff as it's weird and awkward; instead we should encourage multi-step builds deriving from this image
- Don't make `bootc.tar.zst` only to immediately untar it; just use `COPY` from the build container
- Use heredocs to condense multiple `RUN` invocations to avoid pointless small layers

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,11 @@
-.cosa
+# The big one - this can get HUGE and we don't want
+# to copy it around.
 target
-!target/dev-rootfs
+# This one can have large .qcow2 files written by coreos-assembler
+.cosa
 # These directories don't contribute to our container build
 docs/
+# TMT interprets these, not the container build
 plans/
+# Avoid changes to this blowing out all layer caches
+hack/Containerfile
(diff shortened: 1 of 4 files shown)


cgwalters:
> ERROR Installing to filesystem: Creating ostree deployment: Pulling: Importing: Parsing layer blob sha256:0f299f3f346782473308fd887c25d99af245027c4bf8df55ddf47121b4aa2d36: error: (null): Processing tar: Failed to commit tar: ExitStatus(unix_wait_status(256))

Humm...very confused by this failure

henrywang:
`.git` was added in gitignore, so there's an error `fatal: not a git repository (or any of the parent directories): .git` and when we do `tar -C / --zstd -xvf /tmp/bootc.tar.zst`, some of files report `tar: ./usr/bin/bootc: implausibly old time stamp -9223372036854775808`. That might be the root cause?

henrywang (approved):
LGTM.

PR #1297, opened 2025-05-05
reinstall: Pass RUST_LOG env var into bootc install container

ckyrouac:
This enables `RUST_LOG=trace system-reinstall-bootc <image>` to print trace messages for the bootc install invocation.


cgwalters (approved):

cgwalters:
Related to this but I wonder if s-r-b should gain e.g. `--podman-arg=--env=RUST_LOG=trace` to fully generalize support for arbitrary changes? That'd be equally easy I think.

PR #1298, opened 2025-05-05
reinstall: Cleaner formatting of podman bootc install message

ckyrouac:
This makes it easier to copy/paste (and read) the `podman ... bootc install ...` command from the CLI output.

### DIFF
--- a/utils/src/command.rs
+++ b/utils/src/command.rs
@@ -1,6 +1,7 @@
 //! Helpers intended for [`std::process::Command`] and related structures.
 
 use std::{
+    fmt::Write,
     io::{Read, Seek},
     os::unix::process::CommandExt,
     process::Command,
@@ -33,6 +34,9 @@ pub trait CommandRunExt {
     /// Execute the child process, parsing its stdout as JSON. This uses `run` internally
     /// and will return an error if the child process exits abnormally.
     fn run_and_parse_json<T: serde::de::DeserializeOwned>(&mut self) -> Result<T>;
+
+    /// Print the command as it would be typed into a terminal
+    fn to_string_pretty(&self) -> String;
 }
 
 /// Helpers intended for [`std::process::ExitStatus`].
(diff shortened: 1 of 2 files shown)


system-reinstall-bootc/src/podman.rs
```diff
@@ -6,6 +6,16 @@ use bootc_utils::CommandRunExt;
 use std::process::Command;
 use which::which;
 
+pub(crate) fn print_command(command: &Command) -> String {
+    let program = command.get_program().to_string_lossy();
+    let args = command
+        .get_args()
+        .map(|arg| arg.to_string_lossy())
```
cgwalters, line 13: In the unusual but possible case that command arguments have spaces or any shell metacharacters, this won't render in a way that the user can copy-paste.

We have a helper `PathQuotedDisplay` which was made for file paths, but it's not any different for commandline arguments.

Also, a general pattern to use to join arguments like this is to use `fold()` on an iterator to avoid allocating intermediate strings.

I put together this quick test example

```
#[test]
fn t() {
    fn command_to_string(c: Command) -> String {
        std::iter::once(c.get_program())
            .chain(c.get_args())
            .fold(String::new(), |mut acc, elt| {
                if !acc.is_empty() {
                    acc.push(' ');
                }
                // SAFETY: Writes to string can't fail
                write!(&mut acc, "{}", PathQuotedDisplay::new(&elt)).unwrap();
                acc
            })
    }
    let mut c = Command::new("foo");
    c.args(["--test", "bar baz", "other"]);
    assert_eq!(command_to_string(c), "foo --test 'bar baz' other")
}
```

utils/src/command.rs
```diff
@@ -33,6 +34,9 @@ pub trait CommandRunExt {
     /// Execute the child process, parsing its stdout as JSON. This uses `run` internally
     /// and will return an error if the child process exits abnormally.
     fn run_and_parse_json<T: serde::de::DeserializeOwned>(&mut self) -> Result<T>;
+
+    /// Print the command as it would be typed into a terminal
+    fn to_string_pretty(&mut self) -> String;
```
cgwalters, line 39: This can be just `&self`

cgwalters (approved):

PR #1299, opened 2025-05-06
Reinstall mount warning

ckyrouac:
@cgwalters this is a rough outline of what I'm planning to do to guide the user to handle their mounted filesystems. I'm not sure which mounts we should list though. This code correctly shows the `/home` mount from a Fedora workstation install but I'm sure there are other cases I'm missing.

(I'm also trying to figure out a more functional way to parse the filesystems)

### DIFF
--- a/system-reinstall-bootc/src/lvm.rs
+++ b/system-reinstall-bootc/src/lvm.rs
@@ -1,0 +1,84 @@
+use std::process::Command;
+
+use anyhow::Result;
+use bootc_mount::run_findmnt;
+use bootc_utils::CommandRunExt;
+use serde::Deserialize;
+
+#[derive(Debug, Deserialize)]
+pub(crate) struct Lvs {
+    report: Vec<LvsReport>,
+}
+
+#[derive(Debug, Deserialize)]
+pub(crate) struct LvsReport {
+    lv: Vec<LogicalVolume>,
+}
+
+#[derive(Debug, Deserialize, Clone)]
+pub(crate) struct LogicalVolume {
+    lv_name: String,
+    lv_size: String,
+    lv_path: String,
+    vg_name: String,
+}
+
+pub(crate) fn parse_volumes(group: Option<&str>) -> Result<Vec<LogicalVolume>> {
+    if which::which("podman").is_err() {
+        tracing::debug!("lvs binary not found. Skipping logical volume check.");
+        return Ok(Vec::<LogicalVolume>::new());
+    }
+
+    let mut cmd = Command::new("lvs");
+    cmd.args([
+        "--reportformat=json",
+        "-o",
+        "lv_name,lv_size,lv_path,vg_name",
+    ])
+    .args(group);
+
+    let output: Lvs = cmd.run_and_parse_json()?;
+
+    Ok(output
+        .report
+        .iter()
+        .flat_map(|r| r.lv.iter().cloned())
+        .collect())
+}
+
+pub(crate) fn check_root_siblings() -> Result<Vec<String>> {
+    let all_volumes = parse_volumes(None)?;
+
+    // first look for a lv mounted to '/'
+    // then gather all the sibling lvs in the vg along with their mount points
+    let siblings: Vec<String> = all_volumes
+        .iter()
+        .filter(|lv| {
+            let mount = run_findmnt(&["-S", &lv.lv_path], None).unwrap_or_default();
+            if let Some(fs) = mount.filesystems.first() {
+                &fs.target == "/"
+            } else {
+                false
+            }
+        })
+        .flat_map(|root_lv| parse_volumes(Some(root_lv.vg_name.as_str())).unwrap_or_default())
+        .try_fold(Vec::new(), |mut acc, r| -> anyhow::Result<_> {
+            let mount = run_findmnt(&["-S", &r.lv_path], None)?;
+            let mount_path = if let Some(fs) = mount.filesystems.first() {
+                &fs.target
+            } else {
+                ""
+            };
+
+            if mount_path != "/" {
+                acc.push(format!(
+                    "Type: LVM, Mount Point: {}, LV: {}, VG: {}, Size: {}",
+                    mount_path, r.lv_name, r.vg_name, r.lv_size
+                ))
+            };
+
+            Ok(acc)
+        })?;
+
+    Ok(siblings)
+}
--- a/system-reinstall-bootc/src/prompt.rs
+++ b/system-reinstall-bootc/src/prompt.rs
@@ -1,4 +1,4 @@
-use crate::{prompt, users::get_all_users_keys};
+use crate::{btrfs, lvm, prompt, users::get_all_users_keys};
 use anyhow::{ensure, Context, Result};
 
 use crossterm::event::{self, Event};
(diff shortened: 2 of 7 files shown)


system-reinstall-bootc/src/prompt.rs
```diff
@@ -92,6 +93,42 @@ pub(crate) fn ask_yes_no(prompt: &str, default: bool) -> Result<bool> {
...
+
+    let problematic_filesystems: Vec<&Filesystem> = mounts
+        .filesystems
+        .iter()
+        .find(|mount| mount.target == "/")
+        .map(|root| {
+            root.children
+                .iter()
+                .flatten()
+                .filter(|child| child.source == root.source)
```
cgwalters, line 107: So...this seems like it will basically only find btrfs subvolumes, right? I think that's the only real sane case where the same source device can be mounted multiple times in that way.

In Linux one *can* mount e.g. the same block device multiple times but for filesystems without subvolumes there's not much use case for it.

cgwalters: So for someone who has e.g. `/home` or `/var/log` as a LVM logical volume, the sources are different so this check won't trigger.

However, the hard part here is to generalize this, we're going to need to exclude e.g. the [API filesystems](https://www.freedesktop.org/wiki/Software/systemd/APIFileSystems/) among many other things. I'm sure there's code to do this, but we'd need to do some investigation.

system-reinstall-bootc/src/prompt.rs
```diff
@@ -92,6 +93,42 @@ pub(crate) fn ask_yes_no(prompt: &str, default: bool) -> Result<bool> {
...
+                .collect()
+        })
+        .unwrap_or_default();
+
+    if !problematic_filesystems.is_empty() {
+        println!();
+        println!("Notice: the following filesystems are currently mounted on the root disk. After rebooting into the bootc system, these filesystems will not be mounted. They will be preserved so you will need to manually mount or remove them.");
```
cgwalters, line 114: Well...strictly speaking, someone could inject configuration into their image to mount it. It would be totally sane to me for example to have an OS configuration that auto-discovers a btrfs subvolume named `home` and mounts it at `/home` or whatever.

Getting into this level of detail, we could in theory also scrape `/etc/fstab` and the `.mount` units in the target environment...but, yeah a lot there.

I'm uncertain. I think in the short term given the above, effectively special casing "/home btrfs subvolume" could make sense as a starting point.

> NOTE: Detected mounted btrfs subvolumes: /home
>   These subvolumes are left unchanged by this tool.

?

cgwalters (changes requested):

system-reinstall-bootc/src/lvm.rs
```diff
@@ -0,0 +1,103 @@
...
+    lv_name: String,
+    lv_size: String,
+    lv_path: String,
+    vg_name: String,
+    #[serde(default)]
+    mount_path: String,
+}
+
+pub(crate) fn parse_volumes(group: Option<String>) -> Result<Vec<LogicalVolume>> {
+    let mut cmd = Command::new("lvs");
```
cgwalters, line 31: We probably need to gracefully handle the case where this binary isn't present.

system-reinstall-bootc/src/lvm.rs
```diff
@@ -0,0 +1,103 @@
...
+        })
+        .filter(|lv| lv.mount_path != "/")
+        .collect();
+
+    Ok(siblings)
+}
+
+pub(crate) fn print_warning(volumes: Vec<LogicalVolume>) {
+    println!();
+    println!("NOTICE: the following logical volumes are in the same volume group as root. After reboot, these will not be automatically mounted unless defined in the bootc image. The filesystems will be preserved and continue to consume disk space. Consult the bootc documentation to determine the appropriate action for your system.");
```
cgwalters, line 94: I think we could dedup this text with the btrfs case? Maybe the most obvious way to do it is to have a single function which takes a list of mounts (which are just strings?).

Then what feels even cleaner is to keep the printing code inside s-r-b/main.rs, leaving btrfs.rs and lvm.rs as just information gatherers.

cgwalters (commented):
Thanks! This is looking pretty good to me so far!

I'm thinking rather than having a lot of explanatory in the binary we add it to the docs or so, which is what the text is implying.

BTW I bet we could get some testing of s-r-b by using loopback mounts to set up alternative root installation targets, though that would require passing through the target root instead of assuming `/`. But definitely nonblocking.

system-reinstall-bootc/src/lvm.rs
```diff
@@ -0,0 +1,103 @@
...
+    vg_name: String,
+}
+
+pub(crate) fn binary_available(binary: &str) -> bool {
+    Command::new("which")
+        .arg(binary)
+        .stdout(Stdio::null())
+        .stderr(Stdio::null())
+        .status()
+        .map(|status| status.success())
```
cgwalters, line 32: Hmm it's a bit unusual to *run* the binary to find out if it's present. Hmmmm...this is pretty trivial functionality BUT there is a crate https://crates.io/crates/which that skimming seems to have a not-small set of reverse dependencies, seems actively maintained and skimming the code looks reasonable...

system-reinstall-bootc/src/lvm.rs
```diff
@@ -0,0 +1,103 @@
...
+    let mut cmd = Command::new("lvs");
+    cmd.args([
+        "--reportformat=json",
+        "-o",
+        "lv_name,lv_size,lv_path,vg_name",
+    ]);
+
+    if let Some(group) = group {
+        cmd.arg(group);
+    }
```
cgwalters, line 51: This can just be `cmd.args(group.as_deref());` or more idiomatic make the function arg be `group: Option<&str>`

system-reinstall-bootc/src/lvm.rs
```diff
@@ -0,0 +1,103 @@
...
+        .filter(|lv| {
+            let mount = run_findmnt(&["-S", &lv.lv_path], None).unwrap_or_default();
+            if let Some(fs) = mount.filesystems.first() {
+                &fs.target == "/"
+            } else {
+                false
+            }
+        })
+        .flat_map(|root_lv| {
+            tracing::warn!("inside flat_map");
```
cgwalters, line 78: These should have really been `dbg!` right? I don't think we want to keep `warn` here

system-reinstall-bootc/src/lvm.rs
```diff
@@ -0,0 +1,103 @@
...
+            parse_volumes(Some(root_lv.vg_name.clone())).unwrap_or_default()
+        })
+        .filter_map(|lv| {
+            tracing::warn!("inside map");
+            let mount = run_findmnt(&["-S", &lv.lv_path], None).unwrap();
```
cgwalters, line 83: Why not `?` instead of `unwrap()`?

system-reinstall-bootc/src/lvm.rs
```diff
@@ -0,0 +1,103 @@
...
+            } else {
+                false
+            }
+        })
+        .flat_map(|root_lv| {
+            tracing::warn!("inside flat_map");
+            parse_volumes(Some(root_lv.vg_name.clone())).unwrap_or_default()
+        })
+        .filter_map(|lv| {
+            tracing::warn!("inside map");
```
cgwalters, line 82: Ditto

cgwalters (changes requested):
Thanks! I only have nits basically

system-reinstall-bootc/src/lvm.rs
cgwalters: Tangential but just as far as debugging/testing things, one path is to add static dumps of JSON output from `findmnt` as text fixtures, parse them with `include_str!` and have unit tests on them

system-reinstall-bootc/src/lvm.rs
ckyrouac: I'm not sure how to use `?` inside `filter_map` because it doesn't handle `Result`

cgwalters: Ah right yeah, this is a general trap with combinators. Moving to explicit iteration is one option, but the other option is `try_fold()` which keeps in the combinator flow.

cgwalters: Here's a little test example that shows the idea

```
#[test]
fn t() -> Result<()> {
    let v = ["1", "3", "", "4"];
    let r = v
        .into_iter()
        .try_fold(Vec::new(), |mut acc, r| -> anyhow::Result<_> {
            if r.is_empty() {
                Ok(acc)
            } else {
                acc.push(r.parse::<u32>()?);
                Ok(acc)
            }
        })?;
    dbg!(r);
    Ok(())
}
```

Try changing the array to have a non-integer value like `"oops"` in it to see the error path

cgwalters (approved):

Issue #205, opened 2023-11-12
Better handle no-op updates


cgwalters:
It's possible today to get a no-op update where e.g. just metadata like labels change.  We should definitely detect this and highlight the no-op nature of the change; probably just default to "live updating" the origin?  Should we record that transition at all though (just in the journal)?.

PR #1302, opened 2025-05-09
feat(switch): support tag + digest image reference

### LINKED ISSUE #1165: Support tag + digest image reference
Currently
```
$ bootc switch quay.io/centos-bootc/centos-bootc:stream10@sha256:9735171dc889739be0ee60621b1ee2f74811ab1eaa54b4f63b5a662c1ebe2949
ERROR Switching: Pulling: Creating importer: failed to invoke method OpenImage: failed to invoke method OpenImage: Docker references with both a tag and digest are currently not supported
$
```

This is coming from skopeo

```
$ skopeo inspect docker://quay.io/centos-bootc/centos-bootc:stream10@sha256:9735171dc889739be0ee60621b1ee2f74811ab1eaa54b4f63b5a662c1ebe2949
FATA[0000] Error parsing image name "docker://quay.io/centos-bootc/centos-bootc:stream10@sha256:9735171dc889739be0ee60621b1ee2f74811ab1eaa54b4f63b5a662c1ebe2949": Docker references with both a tag and digest are currently not supported 
```

Yet
```
$ podman pull  quay.io/centos-bootc/centos-bootc:stream10@sha256:9735171dc889739be0ee60621b1ee2f74811ab1eaa54b4f63b5a662c1ebe2949
...
```

works just fine.

There's a lot of prior discussion of this, see https://github.com/containers/image/issues/1736

AFAICS from the change to cri-o here https://github.com/cri-o/cri-o/pull/3060/commits/c862e1fbd69f27cee1b13046c6c0206d15f412e1 the idea is all c/image consumers need to do the tag-stripping first, which is presumably what podman is doing too.

So I guess we should also just do that. To implement this, we basically need to do that tag stripping before we pass the target image to containers-image-proxy.

To be clear the implementation here is that the tag is purely for human consumption, it's not validated (beyond syntactically).

Note *do* want to retain the original full tag+digest form so it appears in `bootc status`.

p5:
Closes #1165

Performs tag-stripping when the image reference contains both 
a tag and digest.  This allows Skopeo to pull the image 
successfully, while still displaying both the tag + digest inside 
`bootc status`.

---

This does still perform unnecessary actions when switching onto the same image selected via digest, and there's an issue to tackle this no-op deployment here - #205

I chose to keep this logic inside bootc itself, rather than pushing it up to ostree_ext since this same logic will likely be required for the upcoming composefs integration. 

Since there's a separate argument for the transport, I am not accounting for the image ref containing the transport. 
e.g. `containers-storage:quay.io/example/someimage:42@sha256:blah`

```
bootc-switch-to-tag-with-digest on  bootc-switch-to-tag-with-digest via 🦀 
❯ sudo ./target/debug/bootc switch ghcr.io/rsturla/eternal-linux/lumina:42-nvidia-open@sha256:fa8a2a7212a110caedb3729d70513fbcd518c4ed3f011ff16f4531b3b7a4d60e
[sudo] password for admin: 
Fetched layers: 0 B in 0 seconds (0 B/s)                                                                                                                                                                            
  Deploying: done (12 seconds)                                                                                                                                                                                      Pruned images: 1 (layers: 0, objsize: 0 bytes)
Queued for next boot: ghcr.io/rsturla/eternal-linux/lumina:42-nvidia-open@sha256:fa8a2a7212a110caedb3729d70513fbcd518c4ed3f011ff16f4531b3b7a4d60e
  Version: 250507
  Digest: sha256:fa8a2a7212a110caedb3729d70513fbcd518c4ed3f011ff16f4531b3b7a4d60e

bootc-switch-to-tag-with-digest on  bootc-switch-to-tag-with-digest via 🦀 took 29s 
❯ sudo ./target/debug/bootc status                                                                                                                            
  Staged image: ghcr.io/rsturla/eternal-linux/lumina:42-nvidia-open@sha256:fa8a2a7212a110caedb3729d70513fbcd518c4ed3f011ff16f4531b3b7a4d60e
        Digest: sha256:fa8a2a7212a110caedb3729d70513fbcd518c4ed3f011ff16f4531b3b7a4d60e (amd64)
       Version: 250507 (2025-05-07T01:44:43Z)

● Booted image: ghcr.io/rsturla/eternal-linux/lumina:42-nvidia-open
        Digest: sha256:fa8a2a7212a110caedb3729d70513fbcd518c4ed3f011ff16f4531b3b7a4d60e (amd64)
       Version: 250507 (2025-05-07T01:44:43Z)

  Rollback image: ghcr.io/rsturla/eternal-linux/lumina:42-nvidia-open
          Digest: sha256:6c4d19f8ecc21c055c81d0993921399f0cc034ec69ad49bece9c12f5bb1dddf8 (amd64)
         Version: 250505 (2025-05-05T01:46:41Z)

bootc-switch-to-tag-with-digest on  bootc-switch-to-tag-with-digest via 🦀 
❯ sudo ./target/debug/bootc switch ghcr.io/rsturla/eternal-linux/lumina:42-nvidia-open                                                                        
No changes in ghcr.io/rsturla/eternal-linux/lumina:42-nvidia-open => sha256:fa8a2a7212a110caedb3729d70513fbcd518c4ed3f011ff16f4531b3b7a4d60e
  Deploying: done (12 seconds)                                                                                                                                                                                      Pruned images: 0 (layers: 0, objsize: 211.6 kB)
Queued for next boot: ghcr.io/rsturla/eternal-linux/lumina:42-nvidia-open
  Version: 250507
  Digest: sha256:fa8a2a7212a110caedb3729d70513fbcd518c4ed3f011ff16f4531b3b7a4d60e

```

### DIFF
--- a/lib/src/deploy.rs
+++ b/lib/src/deploy.rs
@@ -340,7 +340,9 @@ pub(crate) async fn prepare_for_pull(
     imgref: &ImageReference,
     target_imgref: Option<&OstreeImageReference>,
 ) -> Result<PreparedPullResult> {
-    let ostree_imgref = &OstreeImageReference::from(imgref.clone());
+    let imgref_canonicalized = imgref.clone().canonicalize()?;
+    tracing::debug!("Canonicalized image reference: {imgref_canonicalized:#}");
+    let ostree_imgref = &OstreeImageReference::from(imgref_canonicalized);
     let mut imp = new_importer(repo, ostree_imgref).await?;
     if let Some(target) = target_imgref {
         imp.set_target(target);
@@ -420,7 +422,9 @@ pub(crate) async fn pull_from_prepared(
     })
     .await;
     let import = import?;
-    let ostree_imgref = &OstreeImageReference::from(imgref.clone());
+    let imgref_canonicalized = imgref.clone().canonicalize()?;
+    tracing::debug!("Canonicalized image reference: {imgref_canonicalized:#}");
+    let ostree_imgref = &OstreeImageReference::from(imgref_canonicalized);
     let wrote_imgref = target_imgref.as_ref().unwrap_or(&ostree_imgref);
 
     if let Some(msg) =
--- a/lib/src/spec.rs
+++ b/lib/src/spec.rs
@@ -2,8 +2,8 @@
 
 use std::fmt::Display;
 
-use ostree_ext::container::OstreeImageReference;
 use ostree_ext::oci_spec::image::Digest;
+use ostree_ext::{container::OstreeImageReference, oci_spec};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
@@ -87,6 +87,25 @@ pub struct ImageReference {
     /// Signature verification type
     #[serde(skip_serializing_if = "Option::is_none")]
     pub signature: Option<ImageSignature>,
+}
+
+impl ImageReference {
+    /// Returns a canonicalized version of this image reference, preferring the digest over the tag if both are present.
+    pub fn canonicalize(self) -> Result<Self, anyhow::Error> {
+        let reference: oci_spec::distribution::Reference = self.image.parse()?;
+
+        if reference.digest().is_some() && reference.tag().is_some() {
+            let registry = reference.registry();
+            let repository = reference.repository();
+            let digest = reference.digest().expect("digest is present");
+            return Ok(ImageReference {
+                image: format!("{registry}/{repository}@{digest}"),
+                ..self
+            });
+        }
+
+        Ok(self)
+    }
 }
 
 /// The status of the booted image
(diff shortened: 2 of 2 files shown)


lib/src/spec.rs
```diff
@@ -89,6 +89,22 @@ pub struct ImageReference {
     pub signature: Option<ImageSignature>,
 }
 
+impl ImageReference {
+    /// Returns a canonicalized version of this image reference, preferring the digest over the tag if both are present.
+    pub fn canonicalize(&self) -> ImageReference {
+        if let Some((base, digest)) = self.image.split_once('@') {
```
cgwalters, line 95: This is I think fine as is *however* one thing that now exists (but didn't when this code was written) is https://docs.rs/oci-spec/latest/oci_spec/distribution/struct.Reference.html which we could start using now to parse these things.

lib/src/spec.rs
```diff
@@ -89,6 +89,22 @@ pub struct ImageReference {
     pub signature: Option<ImageSignature>,
 }
 
+impl ImageReference {
+    /// Returns a canonicalized version of this image reference, preferring the digest over the tag if both are present.
+    pub fn canonicalize(&self) -> ImageReference {
```
cgwalters, line 94: I think this should take `self` not `&self` and then the no-op case is a passthrough and not a clone.

lib/src/deploy.rs
```diff
@@ -340,7 +340,7 @@ pub(crate) async fn prepare_for_pull(
     imgref: &ImageReference,
     target_imgref: Option<&OstreeImageReference>,
 ) -> Result<PreparedPullResult> {
-    let ostree_imgref = &OstreeImageReference::from(imgref.clone());
+    let ostree_imgref = &OstreeImageReference::from(imgref.canonicalize());
```
cgwalters, line 343: It *might* be worth adding a `tracing::debug!` or so in the places where we canonicalize just so we could track when this happens.

Or perhaps we should start logging more to the systemd journal actually.

cgwalters (approved):
Thank you so much for working on this! There's nothing blocking here and we could probably merge as is...but I'd feel a bit more comfortable if we did the parsing via Reference.

lib/src/spec.rs
p5: Thanks for pointing this out!  I didn't know this existed.

Have updated the code in the latest commit to use the `Reference`.

lib/src/deploy.rs
p5: I've added a trace.  
As to what is logged to the systemd journal, I think this needs some decisions to take.  I don't believe the `canonicalize()` function would be worth journaling, since you're probably already logging the actual imgref during the switch.

lib/src/spec.rs
p5: Done!  
FYI where we're calling this function, we're cloning the return value anyway.

lib/src/spec.rs
p5: Hmm - yeah

So when using oci_spec and trying to `bootc switch` into a local image in containers-storage that doesn't contain a registry as part of the name, it fails.

```
❯ sudo ./target/debug/bootc switch --transport containers-storage fedora-silverblue:local@sha256:c512a2bfbd52a2168c5fea0ccdb7a94047c1119aed78255ce23af19759997223
ERROR Switching: Creating importer: failed to invoke method OpenImage: failed to invoke method OpenImage: reference "[overlay@/var/lib/containers/storage+/run/containers/storage:overlay.imagestore=/usr/lib/containers/storage,overlay.mountopt=nodev,metacopy=on]docker.io/library/fedora-silverblue@sha256:c512a2bfbd52a2168c5fea0ccdb7a94047c1119aed78255ce23af19759997223" does not resolve to an image ID
```

Will try and work around this, but may mean relying on oci_spec a bit less.

p5: This doesn't actually seem to be an issue.

The current `bootc` will only switch to a containers-storage image when used with the `localhost/` prefix, which oci_spec seems to also support.

Will update test-cases to test this localhost image, then this should be good for another round of reviews.

cgwalters: > it fails since oci_spec adds docker.io/library to the image.

Yeah we almost certainly have the converse problem elsewhere with unqualified images because we aren't using Reference.

*Hopefully* no one out there is actually relying on pulling from `containers-storage:some-unqualified-name`...

However actually...that doesn't actually matter here because it will only break if they use the new tag + digest form, right?

cgwalters (approved):
Awesome thanks!

cgwalters:
Minor followup to this in https://github.com/bootc-dev/bootc/pull/1306

cgwalters:
Followup in https://github.com/bootc-dev/bootc/pull/1310

PR #1306, opened 2025-05-09
spec: Minor style reworking to drop an unwrap()

cgwalters:
By using `if let()` we can drop an unwrap. However to do that we have to unlink the conditionals, so rework things in "early return" style.

While we're here, move usage of anyhow up to the top as we do elsewhere.

### DIFF
--- a/lib/src/spec.rs
+++ b/lib/src/spec.rs
@@ -2,6 +2,7 @@
 
 use std::fmt::Display;
 
+use anyhow::Result;
 use ostree_ext::oci_spec::image::Digest;
 use ostree_ext::{container::OstreeImageReference, oci_spec};
 use schemars::JsonSchema;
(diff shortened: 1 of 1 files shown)


p5 (commented):
The changes here make a lot of sense, thanks!

I thought the unwrap was okay because the digest will always be present if it reached that point (the condition ensured it was set before unwrap was called), but I completely understand that unwrap is generally discouraged.  The exit early does look clean too.

And I need to remember to use anyhow here.  Been learning Rust without it in the background, so this was probably due to context switching.

cgwalters:
> I thought the unwrap was okay because the digest will always be present if it reached that point (the condition ensured it was set before unwrap was called), but I completely understand that unwrap is generally discouraged. 

Yeah, that unwrap would always succeed for sure. But the idea is that by destructuring via `if let` or `match` the compiler actually *verifies* this even if the code is refactored later for example. It starts to matter more in bigger functions.

cgwalters:
Included in https://github.com/bootc-dev/bootc/pull/1310

PR #1310, opened 2025-05-12
fix(switch): don't attempt to parse non-oci image ref formats

### LINKED ISSUE #1308: `bootc switch --transport dir /var/tmp/tmt/repo` failed
`bootc switch --transport dir /var/tmp/tmt/repo` failed with error `ERROR Switching: invalid reference format`. Log: https://artifacts.osci.redhat.com/testing-farm/81bb4fbc-c3de-4350-88c2-2a1a0bfed7e4/

Only switch to local folder failed, switch to registry URL does not have this issue.

This issue caused bib image building failure: `ERROR Installing to filesystem: Creating ostree deployment: invalid reference format`. Log: https://artifacts.osci.redhat.com/testing-farm/75e3e31f-1858-4f21-81d8-15a91b0bbaf6/

This issue exists since `bootc-202505092010.g001106e8bc-1.el9.x86_64`, and `bootc-202505061330.g93b22f4dbc-1.el9.x86_64` does not have this issue.

p5:
Fixes #1308

Introduced in 10b66fb was a bug where oci-archive and dir image ref formats were attempted to be parsed, to which an error was returned.  This changes the behaviour to only attempt to parse image refs from an allowlisted selection of transports.

---

This also updates the test cases to catch this.

### DIFF
--- a/lib/src/spec.rs
+++ b/lib/src/spec.rs
@@ -2,6 +2,7 @@
 
 use std::fmt::Display;
 
+use anyhow::Result;
 use ostree_ext::oci_spec::image::Digest;
 use ostree_ext::{container::OstreeImageReference, oci_spec};
 use schemars::JsonSchema;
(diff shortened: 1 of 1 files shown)


lib/src/spec.rs
```diff
@@ -91,20 +92,35 @@ pub struct ImageReference {
...
-            let registry = reference.registry();
-            let repository = reference.repository();
-            let digest = reference.digest().expect("digest is present");
-            return Ok(ImageReference {
-                image: format!("{registry}/{repository}@{digest}"),
-                ..self
-            });
+    pub fn canonicalize(self) -> Result<Self> {
+        match self.transport.as_str() {
+            "containers-storage" | "registry" | "oci" => {
```
cgwalters, line 97: The idea here is that `Transport` is already a proper `enum` that we can match on instead of looking at its string forms.

```
match self.transport {
  Transport::ContainerStorage | Transport::Registry | Transport::OciDir
```
etc

cgwalters (approved):
Thanks for this fix, looks correct. One style nit which we could change after merging too.

lib/src/spec.rs
p5: Ah, I saw ImageReference.transport is of a string type, so didn't realise I could do this

```
pub struct ImageReference {
    /// The container image reference
    pub image: String,
    /// The container image transport
    pub transport: String,
    /// Signature verification type
    #[serde(skip_serializing_if = "Option::is_none")]
    pub signature: Option<ImageSignature>,
}
```

cgwalters: Err sorry I was wrong it's the ostree type that's an enum. There's some larger tech debt here because the json schema stuff couldn't apply to types from foreign crates, but we merged ostree-ext so it's more obvious now how to unify these two things.

cgwalters:
Followup in https://github.com/bootc-dev/bootc/pull/1311

PR #1311, opened 2025-05-12
spec: Note we don't canonicalize oci

cgwalters:
The previous code in trying to parse `oci` was wrong; the syntax for an `oci` transport is the same as oci-archive, which includes a file path (and there's no mechanism to quote `:` note).

The canonical logic for all of this stuff is in Go, there's no canonical Rust library (yet, though I did think about putting this in oci-spec).

Previous to this attempt to handle tagged+digested, we weren't parsing image references at all.

First, factor out a `canonicalize_reference` helper since that's what we're really doing here, it's independent of the *transport*.

Fix the canonicalize function to drop out trying to parse `oci`.

Add a separate test case that incorrectly passes just so it's a bit more obvious to fix this later.

Note that today at least `skopeo` rejects trying to fetch via tagged+digested form from an `oci:` so it's fine if we don't canonicalize here yet, even though it could confuse someone.

### DIFF
--- a/lib/src/spec.rs
+++ b/lib/src/spec.rs
@@ -3,6 +3,8 @@
 use std::fmt::Display;
 
 use anyhow::Result;
+use ostree_ext::container::Transport;
+use ostree_ext::oci_spec::distribution::Reference;
 use ostree_ext::oci_spec::image::Digest;
 use ostree_ext::{container::OstreeImageReference, oci_spec};
 use schemars::JsonSchema;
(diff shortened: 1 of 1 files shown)


cgwalters:
Ohh now that I look more closely it's just the Fedora rawhide provisioning which is failing. I am not sure we should actually be running jobs against Fedora rawhide in a gating fashion here, at least not until that project has more gating.

The same CI checks passed with c10s.

ckyrouac (approved):

Issue #1312, opened 2025-05-12
Unable to use latest system-reinstall-bootc with old bootc images that lack --cleanup option


ckyrouac:
The new `--cleanup` option isn't available in older bootc images. e.g. trying to `system-reinstall-bootc quay.io/centos-bootc/centos-bootc:stream9` from a machine with the latest system-reinstall-boot fails. The simplest fix would be to detect the bootc version in the image and only add `--cleanup` if it is available.

cgwalters:
> detect the bootc version in the image

This also relates to https://github.com/bootc-dev/bootc/discussions/1273 if we start parsing the output of `bootc --version`. I would kind of lean towards parsing the output of `bootc install to-filesystem --help` instead, on the general principle that feature detection is better than version detection in the face of enterprise distros that may backport patches.

PR #1313, opened 2025-05-12
reinstall: Only add --cleanup when bootc version >= 1.1.8

ckyrouac:


system-reinstall-bootc/src/podman.rs
```diff
@@ -6,7 +6,33 @@ use bootc_utils::CommandRunExt;
...
+
+fn bootc_has_clean(image: &str) -> Result<bool> {
+    let output = Command::new("podman")
+        .args(["run", image, "bootc", "--version"])
+        .output()?;
+    let stdout_str = String::from_utf8_lossy(&output.stdout);
+    let version_str = stdout_str
+        .split_whitespace()
+        .nth(1)
+        .ok_or_else(|| anyhow::anyhow!("Failed to parse bootc version"))?;
```
cgwalters: I probably wrote https://github.com/bootc-dev/bootc/issues/1312#issuecomment-2874008011 after you started working on this but what do you think about that?

To be clear I don't have a really strong opinion on this, there's clear advantages to parsing the version number too in that in the future it'd let us query it just once and then determine *other* feature flags based on that too.

system-reinstall-bootc/src/podman.rs
```diff
@@ -6,7 +6,22 @@ use bootc_utils::CommandRunExt;
 use std::process::Command;
 use which::which;
 
-pub(crate) fn reinstall_command(image: &str, ssh_key_file: &str) -> Command {
+fn bootc_has_clean(image: &str) -> Result<bool> {
+    let output = Command::new("podman")
+        .args([
+            "run",
```
cgwalters, line 12: Missing `--rm` here, otherwise we leak a stopped container.

ckyrouac:
switch to look at the output of `—help` per https://github.com/bootc-dev/bootc/issues/1312#issuecomment-2874008011

cgwalters (approved):

Issue #448, opened 2024-03-27
Investigate alignment/intersection with systemd-repart


cgwalters:
Let's look at [systemd repart](https://www.freedesktop.org/software/systemd/man/latest/systemd-repart.service.html) and our intersections with it.

A big picture overall thing here is that `bootc install` defaults to having one rootfs and inherently can work on top of any filesystem or block storage.  A key thing is that the operating system size is dynamic, not a fixed partition A/B setup.

systemd-repart is definitely oriented towards DDIs and the discoverable partition spec and less so to this path, but it is obviously very general.

I think a baseline pattern that would be neat to support is injecting repart configs into the base container image to define a distinct `/var` partition.

A great first target would be e.g. initializing EBS instance store volumes in AWS, or the NTFS partition in Azure.  

(Note, it looks like the service is designed to run from the initrd, so we'll want to document either regenerating or more ideally *appending* to the initrd)

cgwalters:
I was thinking about this one more in the context of https://github.com/bootc-dev/bootc/pull/1314#discussion_r2086652274

If a base image chooses to include content in `/usr/lib/repart.d`, I think we should probably use it by default for `to-disk`. We could also support `--repart-dir` to allow embedding e.g. `/usr/lib/exampleos/repart-liveos.d` or so for alternative partition layouts.

mtalexan:
> If a base image chooses to include content in `/usr/lib/repart.d`, I think we should probably use it by default for `to-disk`. 

Just keep in mind that one of the primary purposes of `repart.d` is to dynamically size to real hardware as determined at run-time of booting the file system/initramfs the files are included in.  

---

There are two big use cases for repart.d:
1. I already have known existing partition states I'd like to ensure additions/modifications to.
2. I want to create everything from scratch.

While there's a lot of discussion about Use-case 2, I don't think `bootc install to-disk` currently supports it. If I'm not mistaken it currently just has a fixed non-configurable partitioning scheme it implements, and you'd use something like `bootc-image-builder` to do a custom partitioning scheme (using `bootc install to-filesystem` under the hood instead).  

In Use-case 1, which is all that's currently possible with `bootc install to-disk`, it's very unlikely a user would include files matching the default `boot` or `xbootldr` partitions, unless they're in the very unlikely scenario of needing to create additional partitions of those types (in which case you have to identify the existing ones first, so subsequent files are matched as new rather than existing).  The most likely sub-use-cases are really just: 
  - Match the existing `root` partition to grow it larger
[...]

PR #1314, opened 2025-05-13
Composefs backend

Johan-Liebert1:
A rough draft for installing and booting using `composefs-rs`. Current only works for `install to-disk` via `bootc install to-disk --composefs [options]`

Issues:
1. Issue with transport type `continers-storage`, viz, [#117](https://github.com/containers/composefs-rs/pull/117). Needs some ironing out.
2. Bootc images need rebuilding with `composfs-rs` components, namely `composfs-setup-root`

I have a few logs being logged as warning, just for visibility. Will remove them when it's ready.
Same with `composefs-rs` patch in `cargo.toml`


lib/src/install.rs
```diff
@@ -1457,34 +1564,47 @@ async fn install_to_filesystem_impl(
...
-            state,
-            rootfs,
-            &sysroot,
-            &boot_uuid,
-            bound_images,
-            has_ostree,
-            &imgstore,
-        )
-        .await?;
+        tracing::warn!(
```
cgwalters, line 1571: Same comment re `dbg!` *or* this one could also make sense as a `tracing::debug!` that gets left in right?

cgwalters (commented):
Thanks for working on this!!

lib/src/install.rs
```diff
@@ -1428,10 +1437,32 @@ impl BoundImages {
...
+
+    // transport's display is already of type "<transport_type>:"
+    composefs_oci_pull(&Arc::new(repo), &format!("{transport}{name}",), None).await
+}
+
 async fn install_to_filesystem_impl(
     state: &State,
     rootfs: &mut RootSetup,
     cleanup: Cleanup,
+    composefs: bool,
```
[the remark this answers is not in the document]
Johan-Liebert1: I was conflicted on where to put this. Thanks for the review. I'll add this to `State`

lib/src/install.rs
```diff
@@ -1428,10 +1437,32 @@ impl BoundImages {
...
+    state: &State,
+    root_setup: &RootSetup,
+) -> Result<(Sha256Digest, impl FsVerityHashValue)> {
+    let rootfs_dir = &root_setup.physical_root;
+
+    rootfs_dir
+        .create_dir_all("composefs")
+        .context("Creating dir 'composefs'")?;
+
+    tracing::warn!("STATE: {state:#?}");
```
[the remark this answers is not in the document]
cgwalters: Yep that makes sense, what I meant is that use `dbg!` *locally*. (Or use `tracing::debug!` and keep the tracing in there and just use the tracing selector to ensure you see what you want...we are really chatty with a big `env RUST_LOG=debug` unfortunately today, mostly in container fetches)

Johan-Liebert1:
> Let's call it "composefs native" storage or so (the branding here is again hard, open to other ideas)
I think this works. Or we could have `--composefs-native-experimental` or something?

> Default to (i.e. require) composefs-native storage if we detect the image is a sealed UKI.
Curious about this. Is there a reason we'd *want* composfs native storage?

cgwalters:
> Curious about this. Is there a reason we'd want composfs native storage in this case?

Isn't the whole idea that we're aiming for the sealed case?

Johan-Liebert1:
I'll be getting back to this PR next Monday. Been working on a separate issue this week...

lib/src/install/baseline.rs
```diff
@@ -403,6 +409,8 @@ pub(crate) fn install_create_rootfs(
         "root",
         opts.wipe,
         mkfs_options.iter().copied(),
+        // TODO: Add cli option for this
+        Some(uuid::uuid!("6523f8ae-3eb1-4e2a-a05a-18b695ae656f")),
```
cgwalters: It looks like you chose the *alpha* architecture uuid; it'd make sense potentially to use the DPS UUIDs by default for `to-disk` but that will require some work.

cgwalters: cc https://github.com/bootc-dev/bootc/issues/448#issuecomment-3005089786

lib/src/install.rs
```diff
@@ -1425,6 +1486,212 @@ impl BoundImages {
...
+        &cmdline_refs,
+    )?;
+
+    Ok(())
+}
+
+fn get_esp_device() -> Option<PathBuf> {
+    let esp_devices = [COREOS_ESP_PART_LABEL, ANACONDA_ESP_PART_LABEL]
+        .into_iter()
+        .map(|p| Path::new("/dev/disk/by-partlabel/").join(p));
```
cgwalters: Two things
- We shouldn't use `/dev/disk/by-partlabel` because that's global state, in the general case we may be creating partitions on a loopback mounted device and don't want to touch any global state
- There's a well known UUID for the ESP to use

Combining these, see https://github.com/coreos/bootupd/blob/92a8fe2326e35d893afbe464857408c403d9e7ea/src/blockdev.rs#L41 which perhaps we could move into *this* codebase and have bootupd use it (because bootupd already uses bootc).

lib/src/install.rs
```diff
@@ -1425,6 +1486,212 @@ impl BoundImages {
...
+
+    Task::new("Unmounting ESP", "umount")
+        .arg(mounted_esp)
+        .run()?;
+
+    let boot_dir = root_setup.physical_root_path.join("boot");
+    create_dir_all(&boot_dir).context("Failed to create boot dir")?;
+
+    // Add the user grug cfg
+    let grub_user_config = format!(
```
cgwalters: In my ideal world grub related things live in bootupd, not in this project.

And even better if we can make this a *static* configuration...where grub basically implements the BLS spec and finds things in `/EFI/Linux` on its own right?

lib/src/install.rs
Johan-Liebert1: Ah, okay I'll update to not use `by-partlabel`

> perhaps we could move into this codebase and have bootupd use it (because bootupd already uses bootc)

I found this little bit weird, as I feel like this should be the other way around. Is there a reason why bootc uses the `bootupctl` command rather than the `bootup` crate?

lib/src/install.rs
Johan-Liebert1: It makes sense to put this in bootupd, but upon discussions with @travier we're still debating on whether we want to push systemd-boot or not, so this is more or less temporary

travier: This part of the GRUB config can not be generated by bootup as it is the deployment boot order in the UKI + GRUB case. See: https://github.com/containers/composefs-rs/issues/38#issuecomment-3007646048

travier: It's not possible to use the BLS spec partial support in GRUB to boot UKI. It does not support the key that we need to do so, thus why we generate GRUB config snippets: https://github.com/ostreedev/ostree/issues/2753#issuecomment-2237118552

lib/src/install.rs
cgwalters: > Is there a reason why bootc uses the bootupctl command rather than the bootup crate?

Well...unless we took over that CLI, that'd mean we duplicate the code.

lib/src/cli.rs
```diff
@@ -1199,7 +1199,7 @@ async fn run_from_opt(opt: Opt) -> Result<()> {
                     let fd =
                         std::fs::File::open(&path).with_context(|| format!("Reading {path}"))?;
                     let digest: fsverity::Sha256HashValue = fsverity::measure_verity(&fd)?;
-                    let digest = hex::encode(digest);
+                    let digest = digest.to_hex();
```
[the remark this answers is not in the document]
cgwalters: Done in https://github.com/bootc-dev/bootc/pull/1403

lib/Cargo.toml
```diff
@@ -57,6 +57,8 @@ tini = "1.3.0"
 comfy-table = "7.1.1"
 thiserror = { workspace = true }
 canon-json = { workspace = true }
+openat = "0.1.21"
```
cgwalters: This project is not actively maintained and should not be added here. We use [cap-std](https://docs.rs/cap-std/latest/cap_std/) or rustix directly.

ostree-ext/Cargo.toml
```diff
@@ -17,7 +17,9 @@ ostree = { features = ["v2025_2"], version = "0.20" }
 anyhow = { workspace = true }
 bootc-utils = { package = "bootc-internal-utils", path = "../utils", version = "0.0.0" }
 camino = { workspace = true, features = ["serde1"] }
-composefs = { git = "https://github.com/containers/composefs-rs", rev = "59299eb8de6c22410f32213cc57901a7ddd3c134", features = ["rhel9"] }
+composefs = { git = "https://github.com/containers/composefs-rs", rev = "d8e285f14c3c9fa969496fd98e12ffba3079842d", package = "composefs" }
+composefs-boot = { git = "https://github.com/containers/composefs-rs", rev = "d8e285f14c3c9fa969496fd98e12ffba3079842d", package = "composefs-boot" }
```
cgwalters: We also need to discuss https://github.com/containers/composefs-rs/pull/143#issuecomment-2977561906 related to this.

ostree-ext/src/container_utils.rs
```diff
@@ -77,6 +77,13 @@ pub fn ostree_booted() -> io::Result<bool> {
     Path::new(&format!("/{OSTREE_BOOTED}")).try_exists()
 }
 
+
+/// Returns true if the system appears to have been booted with composefs.
```
cgwalters: I don't want to belabor this too much but I think it will cause endless confusion if we keep saying "this uses composefs" when again we're *already using "composefs"*.

How about "without ostree"? Or "composefs native" or "pure composefs".

ostree-ext/src/container_utils.rs
```diff
@@ -77,6 +77,13 @@ pub fn ostree_booted() -> io::Result<bool> {
...
+pub fn composefs_booted() -> io::Result<bool> {
+    let cmdline = std::fs::read_to_string("/proc/cmdline")?;
```
cgwalters: We should use a proper kernel argument parser in general. It's actually a bit complicated because of quoting. systemd has a good version of course, and we have one in C in ostree. It'd probably make sense to make a dedicated one.

cgwalters (changes requested):
Some more comments

lib/src/bls_config.rs
jeckersb, line 1: :+1: for a first pass, but skimming through the boot loader specification document there's a lot more to this.  Probably enough so that it would make sense to spin this off into its own crate, I'm sure others would find it to be useful outside of bootc.

cgwalters: > Probably enough so that it would make sense to spin this off into its own crate,

Yes I had the same thought, was one of the motivations for https://github.com/bootc-dev/bootc/pull/1413

> I'm sure others would find it to be useful outside of bootc.

Probably though I'd say we should probably not at the current time try to go all the way into making it a real published crate unless we had a reason to believe that there was another possible concrete user.

lib/src/install.rs
```diff
@@ -1423,6 +1525,502 @@ impl BoundImages {
...
+        .find(|p| p.parttype.as_str() == ESP_GUID)
+        .ok_or(anyhow::anyhow!("ESP not found for device: {device}"))?;
+
+    Ok((esp.node, esp.uuid))
+}
+
+pub(crate) fn get_user_config(uki_id: &str) -> String {
+    let s = format!(
+        r#"
+menuentry "Fedora Bootc UKI: ({uki_id})" {{
```
cgwalters: Should not hardcode Fedora here but instead get this metadata from the UKI

lib/src/install.rs
```diff
@@ -1423,6 +1525,502 @@ impl BoundImages {
...
+
+        // This will be atomically renamed to 'loader/entries' on shutdown/reboot
+        (boot_dir.join("loader/entries.staged"), Some(booted_bls))
+    } else {
+        (boot_dir.join("loader/entries"), None)
+    };
+
+    create_dir_all(&entries_path).with_context(|| format!("Creating {:?}", entries_path))?;
+
+    std::fs::write(
```
cgwalters: I think in general we should use e.g. https://docs.rs/cap-std-ext/latest/cap_std_ext/dirext/trait.CapStdExtDirExt.html#tymethod.atomic_write for things like this

lib/src/install.rs
```diff
@@ -1423,6 +1525,502 @@ impl BoundImages {
...
+            let mut cmdline_options = String::from(root_setup.kargs.join(" "));
+            cmdline_options.push_str(&format!(" composefs={id_hex}"));
+
+            (root_setup.physical_root_path.clone(), cmdline_options)
+        }
+
+        BootSetupType::Upgrade => (
+            Utf8PathBuf::from("/sysroot"),
+            vec![
+                format!("root=UUID={DPS_UUID}"),
```
cgwalters: The point of the DPS is that we can avoid the `root=` karg right?

lib/src/deploy.rs
```diff
@@ -747,6 +753,165 @@ pub(crate) async fn stage(
...
+
+    Ok(())
+}
+
+/// Filename for `loader/entries`
+const CURRENT_ENTRIES: &str = "entries";
+const ROLLBACK_ENTRIES: &str = "entries.staged";
+
+#[context("Getting boot entries")]
+pub(crate) fn get_sorted_boot_entries(ascending: bool) -> Result<Vec<BLSConfig>> {
```
cgwalters: I think stuff like this would be good to cover via unit tests, we have a pattern of doing so using cap-std elsewhere.

lib/src/cli.rs
```diff
@@ -757,6 +762,53 @@ fn prepare_for_write() -> Result<()> {
...
+    //     .ok_or(anyhow::anyhow!("Could not find booted image"))?;
+
+    // tracing::debug!("booted_image: {booted_image:#?}");
+    // tracing::debug!("imgref: {imgref:#?}");
+
+    // let digest = booted_image
+    //     .digest()
+    //     .context("Getting digest for booted image")?;
+
+    let (repo, entries, id) = pull_composefs_repo(&imgref.transport, &imgref.image).await?;
```
cgwalters: BTW almost certainly this will need work to set up the image proxy with the same config as we use for the ostree backend, otherwise we won't honor `/etc/ostree/auth.json` etc.

lib/src/bls_config.rs
```diff
@@ -0,0 +1,88 @@
...
+    for line in input.lines() {
+        let line = line.trim();
+        if line.is_empty() || line.starts_with('#') {
+            continue;
+        }
+
+        if let Some((key, value)) = line.split_once(' ') {
+            map.insert(key.to_string(), value.trim().to_string());
+        }
+    }
```
cgwalters: This would IMO be a bit cleaner as a combinator, https://github.com/bootc-dev/bootc/blob/b290ad25c7fb4550a038614ba3c9a7ec986be659/ostree-ext/src/fixture.rs#L133 is prior art

lib/src/bls_config.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+impl Ord for BLSConfig {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.version.cmp(&other.version)
+    }
+}
+
+impl BLSConfig {
+    pub(crate) fn to_string(&self) -> String {
```
cgwalters: Why not `impl Display` instead?

lib/src/bls_config.rs
```diff
@@ -0,0 +1,88 @@
...
+        let mut out = String::new();
+
+        if let Some(title) = &self.title {
+            out += &format!("title {}\n", title);
+        }
+
+        out += &format!("version {}\n", self.version);
```
cgwalters: And if we did `impl Display` we can just then here do `write!(out, "version {}\n", self.version)?` which will avoid allocating a temporary string for each line.

lib/src/bls_config.rs
```diff
@@ -0,0 +1,88 @@
...
+        if line.is_empty() || line.starts_with('#') {
+            continue;
+        }
+
+        if let Some((key, value)) = line.split_once(' ') {
+            map.insert(key.to_string(), value.trim().to_string());
+        }
+    }
+
+    let value = serde_json::to_value(map)?;
```
cgwalters: Hum, a little wild to do parsing by converting it to JSON so serde can directly process it, but I get why you did it that way.

OTOH since we don't have any use cases for parsing it from JSON or any other serialization format (do we?) we could also probably just live with a manual implementation of parsing and skip the `Deserialize`.

lib/src/install.rs
```diff
@@ -522,6 +604,24 @@ impl FromStr for MountSpec {
...
+impl InstallToDiskOpts {
+    pub(crate) fn validate(&self) -> Result<()> {
+        if !self.composefs_native {
+            // Reject using --boot without --composefs
+            if self.composefs_opts.boot != BootType::default() {
+                anyhow::bail!("--boot must not be provided without --composefs");
+            }
+        }
+
+        // Can't add kargs to UKI
```
cgwalters: This is complicated but I think actually what we'll need to do at some point is ensure that when e.g. `ukify` is run that we participate in assembling the kernel commandline - and that part should *consume* (i.e. delete) the `/usr/lib/bootc/kargs.d` directory.

lib/src/install.rs
```diff
@@ -1423,6 +1525,502 @@ impl BoundImages {
...
+    rootfs_dir: &Dir,
+) -> Result<ComposefsRepository<Sha256HashValue>> {
+    ComposefsRepository::open_path(rootfs_dir, "composefs")
+        .context("Failed to open composefs repository")
+}
+
+async fn initialize_composefs_repository(
+    state: &State,
+    root_setup: &RootSetup,
+) -> Result<(Sha256Digest, impl FsVerityHashValue)> {
```
cgwalters: I think we should return the repository object too.

lib/src/install.rs
```diff
@@ -1423,6 +1525,502 @@ impl BoundImages {
...
+    id: &Sha256HashValue,
+    entry: ComposefsBootEntry<Sha256HashValue>,
+) -> Result<()> {
+    let id_hex = id.to_hex();
+
+    let (root_path, cmdline_refs) = match setup_type {
+        BootSetupType::Setup(root_setup) => {
+            // root_setup.kargs has [root=UUID=<UUID>, "rw"]
+            let mut cmdline_options = String::from(root_setup.kargs.join(" "));
+            cmdline_options.push_str(&format!(" composefs={id_hex}"));
```
cgwalters: We should definitely be centralizing via e.g. `const` our references to this kernel commandline.

lib/src/install.rs
```diff
@@ -1423,6 +1525,502 @@ impl BoundImages {
...
+
+    let boot_dir = root_path.join("boot");
+
+    let bls_config = match &entry {
+        ComposefsBootEntry::Type1(..) => todo!(),
+        ComposefsBootEntry::Type2(..) => todo!(),
+        ComposefsBootEntry::UsrLibModulesUki(..) => todo!(),
+
+        ComposefsBootEntry::UsrLibModulesVmLinuz(usr_lib_modules_vmlinuz) => {
+            // Write the initrd and vmlinuz at /boot/<id>/
```
cgwalters: And here `<id>` is the composefs digest? Note that that will mean that we duplicate the kernel/initrd for every userspace change. ostree has this concept of a "bootcsum" to avoid that.

Duplicating it is pretty unavoidable for sealed UKIs of course.

But I wonder if we may run into this for the unsealed case. On this topic xref https://github.com/coreos/fedora-coreos-tracker/issues/1465

lib/src/install.rs
```diff
@@ -1423,6 +1525,502 @@ impl BoundImages {
...
+    /// For initial setup, i.e. install to-disk
+    Setup(&'a RootSetup),
+    /// For `bootc upgrade`
+    Upgrade,
+}
+
+#[context("Setting up BLS boot")]
+pub(crate) fn setup_composefs_bls_boot(
+    setup_type: BootSetupType,
+    // TODO: Make this generic
```
cgwalters: Yeah but you know the problem here is...it will basically infect most of the bootc codebase with a generic - the same way it did in composefs-rs. That's going to be super painful.

We had part of this debate in https://github.com/containers/composefs-rs/pull/105 and I'm still kind of skeptical that generics make sense overall. (The especially important argument here is that we may want to defer until *runtime* i.e. via some repository configuration or other state which digest we use) and that will mean we basically double 70% (yes number made up) of the code generated.

A potential thing in the future is to support migrating from sha256 to sha512 for example, and then we'd need two instantations of Repository for each.

lib/src/install.rs
```diff
@@ -1423,6 +1525,502 @@ impl BoundImages {
...
+            let initrd_path = path.join("initrd");
+
+            std::fs::write(
+                &vmlinuz_path,
+                read_file(&usr_lib_modules_vmlinuz.vmlinuz, &repo).context("Reading vmlinuz")?,
+            )
+            .context("Writing vmlinuz to path")?;
+
+            if let Some(initramfs) = &usr_lib_modules_vmlinuz.initramfs {
+                std::fs::write(
```
cgwalters: Again should use atomic writes - I think we also want to ensure we `fsync` here for sure.

ostree-ext/src/container_utils.rs
Johan-Liebert1: Ah yes, definitely. I'll update these throughout the repo

ostree-ext/src/container_utils.rs
Johan-Liebert1: We could build our own parser? @travier suggested https://github.com/rust-bakery/nom for parsing Grub menuentries. Maybe we could use it for parsing kernel cmdline as well?

lib/Cargo.toml
Johan-Liebert1: Ah, I see. I took inspiration from bootupd. We should remove it from there as well

lib/src/install.rs
Johan-Liebert1: Yes, but grub still requires it iirc

lib/src/install.rs
Johan-Liebert1: Yeah, I missed it

lib/src/install.rs
Johan-Liebert1: You're right about the duplication. We should do it similarly to ostree

lib/src/bls_config.rs
Johan-Liebert1: Makes sense. I didn't `impl Display` because `to_string` conveyed the intention better in my mind

lib/src/cli.rs
Johan-Liebert1: The image pulling code in `composefs-rs` definitely needs some improvement. I'm not sure if it's an issue with skopeo or containers-image-proxy but pulling an image from containers-storage sometimes fails with 

```
ERROR Composefs Switching: Pulling composefs repository: Pulling composefs repo: Unable to pull container image containers-storage:localhost/bootc-composefs-bls-upgrade: Failed to pull config Desc
riptor { media_type: ImageConfig, digest: Digest { algorithm: Sha256, value: "sha256:3c6248f829437b55cb57355f8fdefc5f4f35c2379c726a0872383e113a4455fa", split: 6 }, size: 12533, urls: None, annotat
ions: None, platform: None, artifact_type: None, data: None }: failed to invoke method GetBlob: locating item named "sha256:8570667f03be297afea2c40859006418ec85d4dc6ad2e1258f5412fcdd82ffff" for im
age with ID "3c6248f829437b55cb57355f8fdefc5f4f35c2379c726a0872383e113a4455fa" (consider removing the image to resolve the issue): file does not exist
```

lib/src/bls_config.rs
Johan-Liebert1: I do like the expressiveness of a simple for loop; but we could definitely go with a functional pattern. I don't have a *strong* opinion either way

ostree-ext/src/container_utils.rs
jeckersb: I have written what feels like a million various `nom` parsers for Advent of Code over the years, I could probably throw something together for kernel cmdline in short order.  Seems to be well-documented at https://www.kernel.org/doc/html/latest/admin-guide/kernel-parameters.html

lib/src/bls_config.rs
```diff
@@ -0,0 +1,88 @@
...
+{
+    let s: Option<String> = Option::deserialize(deserializer)?;
+
+    match s {
+        Some(s) => Ok(s.parse::<u32>().map_err(D::Error::custom)?),
+        None => Err(D::Error::custom("Version not found")),
+    }
+}
+
+pub(crate) fn parse_bls_config(input: &str) -> Result<BLSConfig> {
```
jeckersb: This sort of thing I would normally `impl TryFrom<&str> for BLSConfig`.  I think it makes it more obvious what is going on when you use it like `let foo = BLSConfig::try_from(&some_str)?;`  Plus for free you also get the inverse of `let foo: BLSConfig = some_str.try_into()?;`

ostree-ext/src/container_utils.rs
cgwalters: Sure, I'm good to make a new small internal crate for this. But I wouldn't try to just greenfield it though, but reference existing parsers. I think https://github.com/ostreedev/ostree/blob/main/src/libostree/ostree-kernel-args.c is correct, as is the systemd version.

cgwalters: Sorry I forgot we already have https://github.com/bootc-dev/bootc/blob/main/lib/src/kernel.rs and to combine with the chat yes let's for now just take the code from https://github.com/containers/composefs-rs/blob/d5455222f282200913902f29c66d5f625ca5661f/crates/composefs-boot/src/cmdline.rs#L9 but what we IMO must do when parsing `/proc/cmdline` is use `to_utf8_lossy` or so so we don't bomb out if the kernel commandline somehow has non-UTF8 data.

cgwalters: https://github.com/bootc-dev/bootc/issues/1425

lib/src/install.rs
cgwalters: https://github.com/containers/composefs-rs/issues/159

cgwalters:
We had a [live chat](https://github.com/containers/composefs-rs/discussions/158) about this and decided to make a branch in this repo for it so multiple people can collaborate!

cgwalters:
Created in https://github.com/bootc-dev/bootc/tree/composefs-backend
I have also sent an invite to @Johan-Liebert1 to have write access to the repository.

(That said we still need to convert to e.g. https://clowarden.io/ )

crates/lib/src/install.rs
```diff
@@ -1416,6 +1514,615 @@ impl BoundImages {
...
+) -> Result<()> {
+    let state_path = root_path.join(format!("{STATE_DIR_RELATIVE}/{}", deployment_id.to_hex()));
+
+    create_dir_all(state_path.join("etc/upper"))?;
+    create_dir_all(state_path.join("etc/work"))?;
+
+    let actual_var_path = root_path.join(format!("state/os/fedora/var"));
+    create_dir_all(&actual_var_path)?;
+
+    symlink(Path::new("../../os/fedora/var"), state_path.join("var"))
```
cgwalters: Another hardcoded `fedora`

crates/ostree-ext/Cargo.toml
```diff
@@ -16,7 +16,9 @@ ostree = { features = ["v2025_2"], version = "0.20" }
 anyhow = { workspace = true }
 bootc-utils = { package = "bootc-internal-utils", path = "../utils", version = "0.0.0" }
 camino = { workspace = true, features = ["serde1"] }
-composefs = { git = "https://github.com/containers/composefs-rs", rev = "d5455222f282200913902f29c66d5f625ca5661f", features = ["rhel9"] }
+composefs = { git = "https://github.com/containers/composefs-rs", rev = "28d4721f77f973f0e394d60d6a69d9b39cb38d7f", package = "composefs", features = ["rhel9"] }
+composefs-boot = { git = "https://github.com/containers/composefs-rs", rev = "28d4721f77f973f0e394d60d6a69d9b39cb38d7f", package = "composefs-boot" }
+composefs-oci = { git = "https://github.com/containers/composefs-rs", rev = "28d4721f77f973f0e394d60d6a69d9b39cb38d7f", package = "composefs-oci" }
```
cgwalters: We could probably split this into a separate PR

crates/ostree-ext/src/container_utils.rs
```diff
@@ -77,6 +77,13 @@ pub fn ostree_booted() -> io::Result<bool> {
     Path::new(&format!("/{OSTREE_BOOTED}")).try_exists()
 }
 
+
+/// Returns true if the system appears to have been booted with composefs.
+pub fn composefs_booted() -> io::Result<bool> {
+    let cmdline = std::fs::read_to_string("/proc/cmdline")?;
+    Ok(cmdline.contains("composefs="))
```
cgwalters: Would still like to centralize this `composefs` in a `const` somewhere

crates/ostree-ext/Cargo.toml
Johan-Liebert1: Created PR https://github.com/bootc-dev/bootc/pull/1445

crates/lib/src/install.rs
Johan-Liebert1: Updated this to `deault`

cgwalters:
[Two](https://martinfowler.com/bliki/TwoHardThings.html) things:

- @Johan-Liebert1 where are we on this PR vs the branch on main https://github.com/bootc-dev/bootc/pull/1444 ? 
- Can you (or someone) review https://github.com/bootc-dev/bootc/pull/1471 ? I want to try rebasing the branch on that
- Also in the vein of rebasing I think when we do that we can probably do some squashing *and* also try to factor out some prep commits (maybe the GRUB parser?)

cgwalters (approved):
Overall looks sane, we will definitely have some splitting/squashing/cleanup/testing to do as we target main though

PR #1321, opened 2025-05-15
install: Fallback to no sepolicy when init imgstore

### LINKED ISSUE #1303: No label found in policy 'None'
When applying an update I get the error

```
ERROR Upgrading: Staging: Pulling bound images: Creating imgstorage: Labeling imgstorage dirs: labeling storage root: No label found in policy 'None' for /var/lib/containers/storage
```

Note that I have disabled SELinux (no time or desire to debug the many problems caused by this over-complicated mess).

ckyrouac:
Check for the policy csum to make sure there is actually a policy available to lookup when trying to label the bootc container storage.

Fixes #1303

### DIFF
--- a/lib/src/store/mod.rs
+++ b/lib/src/store/mod.rs
@@ -92,17 +92,27 @@ impl Storage {
         let sepolicy = if self.sysroot.booted_deployment().is_none() {
             // fallback to policy from container root
             // this should only happen during cleanup of a broken install
+            tracing::trace!("falling back to container root's selinux policy");
             let container_root = Dir::open_ambient_dir("/", cap_std::ambient_authority())?;
             &ostree::SePolicy::new_at(container_root.as_raw_fd(), gio::Cancellable::NONE)?
         } else {
             // load the sepolicy from the booted ostree deployment so the imgstorage can be
             // properly labeled with /var/lib/container/storage labels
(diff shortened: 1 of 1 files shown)


lib/src/store/mod.rs
```diff
@@ -92,17 +92,27 @@ impl Storage {
...
             let dep_fs = deployment_fd(&self.sysroot, &dep)?;
             &ostree::SePolicy::new_at(dep_fs.as_raw_fd(), gio::Cancellable::NONE)?
         };
 
-        let imgstore = crate::imgstorage::Storage::create(&sysroot_dir, &self.run, Some(sepolicy))?;
+        let sepolicy = if sepolicy.csum().is_none() {
+            None
+        } else {
+            Some(sepolicy)
+        };
```
cgwalters, line 111: This is also `let sepolicy = sepolicy.filter(|policy| policy.csum.is_some());` too

cgwalters (approved):
Thanks! 

I do feel like this is a bug probably in the ostree sepolicy code. It'd feel better to just return `None` in this case probably?

But...mmm...I think I have a vague memory now that the reason it's this way is historical because I in *C* code it was convenient to just pass around a `*OstreeSepolicy` and not always check it for `null` or something?

PR #1325, opened 2025-05-19
Rewrite containers org links to bootc-dev org

jeckersb:
Literally just find and replace github.com/containers/bootc with
github.com/bootc-dev/bootc

We have the redirect in place so none of this is really important, but
I figure removing as many instances of the old repo and replacing them
with the current can't hurt for things like search engine
optimization.  Plus some non-zero number of people might assume one is
a fork of the other or something.


cgwalters:
Needs a rebase

cgwalters (approved):

PR #1333, opened 2025-05-27
feat: add luks-password option

hanthor:


lib/src/install/baseline.rs
```diff
@@ -73,9 +74,14 @@ pub(crate) struct InstallBlockDeviceOpts {
     ///
     /// direct: Filesystem written directly to block device
     /// tpm2-luks: Bind unlock of filesystem to presence of the default tpm2 device.
+    /// password-luks: Encrypt filesystem with LUKS using a password.
     #[clap(long, value_enum)]
     pub(crate) block_setup: Option<BlockSetup>,
 
+    /// Password for LUKS encryption when `block_setup` is `password-luks`.
+    #[clap(long)]
+    pub(crate) luks_password: Option<String>,
```
cgwalters, line 83: The reason `cryptsetup` reads from stdin (as opposed to having a parameter) is because command line arguments can be globally visible on Unix. It's probably simplest for us to have `--luks-password-file` that reads from a file say. (And when we do that we should sanity check the file isn't world readable either)

lib/src/install/baseline.rs
```diff
@@ -367,6 +381,27 @@ pub(crate) fn install_create_rootfs(
             ];
             (rootdev, Some(kargs))
         }
+        BlockSetup::PasswordLuks => {
+            let uuid = uuid::Uuid::new_v4().to_string();
+            let password = opts.luks_password.as_ref().unwrap();
+            let root_devpath = root_partition.path();
+
+            Task::new("Initializing LUKS for root", "cryptsetup")
```
cgwalters, line 389: This looks reasonable, however for consistency it could make sense to use `systemd-cryptenroll` for this too.

cgwalters (commented):
Thanks for working on this! 

I will try to clean up our docs in this area but the current status quo is that `to-disk` is intended more for "demo" purposes. It's precisely configuring things like LUKS that get quite complex in the general case, and instead of having a OS/distribution-independent installer in bootc that does partitioning, I'd like to encourage multiple installers that can reuse `bootc install to-filesystem` as a backend.

There's a wide world of this stuff; beyond just LUKS there's LVM/RAID, multiple filesystems with different options, etc.

In your use case how will you be doing installs? You should be able to accomplish what you want via scripting `to-filesystem` relatively easily.

hanthor:
Yes I could replicate the same behaviour with `to-filesystem`. I think the usecase of nuking a disk and setting up encryption is it and I feel like that's all install to-dosk really need to to. I've been using install to-disk with no issues to install on my daily drivers.

I completely understand to-disk isn't meant to be load-bearing and agree that the scope quickly goes to infinity adding installer scenarios. So I leave it up to you to decide if luks-passphrase is a bridge too far. I can probably work on a script to do it, it would just be nice to literally have a flash drive with a live OS and podman to be able to easily setup a new machine.

PR #1336, opened 2025-05-27
ostree-ext: tests: Don't call export_container twice

jeckersb:
Just something weird I noticed when looking at #1328


cgwalters:
Yeah this makes sense. However there *must* have been some sort of underlying race condition where most of the time we get the same manifest, but not always.

PR #1347, opened 2025-06-02
More reproducibility fixes

champtar:
Passing HashMap to serde_json gives you random json :(

I'm adding to_canonical_json_string that uses olpc_cjson::CanonicalFormatter (already used by ocidir), not sure it's the best choice has it  can produce invalid json

Maybe we should write some more helpers and get rid of all direct serde_json::to_* calls


ostree-ext/src/utils.rs
```diff
@@ -32,3 +32,16 @@ impl<T, E: std::fmt::Display> ResultExt<T, E> for Result<T, E> {
...
+#[context("Serializing to canonical JSON")]
+pub(crate) fn to_canonical_json_string<T>(value: T) -> anyhow::Result<String>
+where
+    T: serde::Serialize,
+{
+    let mut ser = serde_json::Serializer::with_formatter(Vec::new(), olpc_cjson::CanonicalFormatter::new());
+    value.serialize(&mut ser)?;
+    let str = String::from_utf8(ser.into_inner())?;
+    Ok(str)
+}
```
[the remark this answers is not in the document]
champtar: olpc_cjson::CanonicalFormatter breaks things:
`error: Installing to filesystem: Creating ostree deployment: Pulling: Importing: control character (\u0000-\u001F) found while parsing a string at line 2 column 0`

This is used in ocidir since https://github.com/ostreedev/ostree-rs-ext/commit/d655a166e4be37ff60b949bd24616f173df0cd67

@cgwalters what do you think about https://docs.rs/cjson/0.1.2/cjson/ ?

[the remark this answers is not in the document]
champtar: never mind this is what you removed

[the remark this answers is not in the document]
champtar: Linking https://github.com/containers/ocidir-rs/issues/10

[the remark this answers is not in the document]
champtar: And https://github.com/youki-dev/oci-spec-rs/issues/249

champtar:
Should I move `JsonOrderedSerialize` to `bootc-utils` and cleanup the whole codebase of `serde_json::to_*` ?

ostree-ext/src/json.rs
```diff
@@ -0,0 +1,46 @@
...
+    where
+        W: io::Write;
+}
+
+impl<S> JsonOrderedSerialize for S
+where
+    S: Serialize,
+{
+    fn to_json_ordered_string(&self) -> Result<String> {
+        let val = serde_json::to_value(self)?;
```
cgwalters: Not that the JSON we serialize here is large, but this is inefficient. It's also not really clear to me why this solves the problem...actually it looks like it only does if a not-on-by-default feature is passed https://github.com/serde-rs/json/blob/c1826ebcccb1a520389c6b78ad3da15db279220d/Cargo.toml#L61

I think we really just need to fix the olpc json formatter crate. In order to not block on upstream I'd tentatively be OK forking it into ocidir or so if that helps.

cgwalters (changes requested):

ostree-ext/src/json.rs
champtar: Indeed it rely on `preserve_order` being off / HashMap being converted to BtreeMap, and is inefficient, but nothing compared to all the single threaded gzip compression / decompression we are doing in between the milliseconds of json processing ;).

Even if a lesser issue, we also need a solution for pretty printing BTW

Let's continue the discussion in https://github.com/containers/ocidir-rs/issues/10

champtar: > I think we really just need to fix the olpc json formatter crate. In order to not block on upstream I'd tentatively be OK forking it into ocidir or so if that helps.

https://github.com/containers/ocidir-rs/pull/39

ostree-ext/src/container/store.rs
```diff
@@ -603,9 +604,13 @@ impl ImageImporter {
             Self::CACHED_KEY_MANIFEST_DIGEST,
             manifest_digest.to_string(),
         );
-        let cached_manifest = serde_json::to_string(manifest).context("Serializing manifest")?;
+        let cached_manifest = manifest
+            .to_canon_json_string()
```
cgwalters, line 608: Side note: this isn't a new problem, but we probably should have always been storing the original manifest instead of re-serializing.

lib/src/progress_jsonl.rs
```diff
@@ -199,8 +200,8 @@ impl TryFrom<RawProgressFd> for ProgressWriter {
 impl ProgressWriter {
     /// Serialize the target value as a single line of JSON and write it.
     async fn send_impl_inner<T: Serialize>(inner: &mut ProgressWriterInner, v: T) -> Result<()> {
-        // serde is guaranteed not to output newlines here
-        let buf = serde_json::to_vec(&v)?;
+        // canon_json is guaranteed not to output newlines here
+        let buf = v.to_canon_json_vec()?;
```
cgwalters, line 204: I don't think the progress stuff needs reproducibility, but I guess it doesn't hurt either.

champtar: No, but ideally i would like to just ban `serde_json::to_{vec,string,writer}` with some linter (haven't looked how to do it yet)

champtar: @cgwalters ok to add those 3 functions to clippy disallowed-methods ? (in a follow up PR)

cgwalters: I'm not *opposed* to it but not really excited by it either...it doesn't seem to me we will have that many places where we really need the reproducibility that we need to explicitly ban it.

champtar: ok, will leave it as is then

PR #1348, opened 2025-06-03
Initial set of system-reinstall-bootc integration tests

ckyrouac:
This adds a few basic integration tests for system-reinstall-bootc, adds a system-reinstall option to tests-integration to run them, and executes them as part of the github action.


tests-integration/src/system_reinstall.rs
```diff
@@ -0,0 +1,158 @@
...
+            // Basic flow stdout verification
+            p.exp_string("Select which user's SSH authorized keys you want to import into")?;
+            p.exp_string("the root user of the new bootc system.")?;
+            p.exp_string("Then you can login as root@ using those keys.")?;
+            p.exp_string("(arrow keys to move, space to select):")?;
+            p.send_line("a")?;
+
+            p.exp_string("Going to run command:")?;
+
+            p.exp_regex(format!("podman run --privileged --pid=host --user=root:root -v /var/lib/containers:/var/lib/containers -v /dev:/dev --security-opt label=type:unconfined_t -v /:/target -v /tmp/(.*):/bootc_authorized_ssh_keys/root {image} bootc install to-existing-root --acknowledge-destructive --skip-fetch-check --cleanup --root-ssh-authorized-keys /bootc_authorized_ssh_keys/root").as_str())?;
```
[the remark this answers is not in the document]
cgwalters: Heh cool, both AIs came up with minor but still reasonable review comments.

I think actually we can just do `([^:]+)` or so (anything but a colon) or so?

tests-integration/src/system_reinstall.rs
```diff
@@ -0,0 +1,158 @@
...
+
+            let output_str = String::from_utf8_lossy(&output.stdout);
+            let lines: Vec<&str> = output_str.lines().collect();
+            let root_partition_info = lines.get(1).expect("Failed to parse df output");
+            let available_space_gb: u64 = root_partition_info
+                .split_whitespace()
+                .nth(3)
+                .expect("Failed to get available space")
+                .parse()
+                .expect("Failed to parse available space");
```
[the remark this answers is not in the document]
cgwalters: Yeah let's just use https://docs.rs/rustix/latest/rustix/fs/fn.statfs.html or so - to get space in bytes need to multiply `f_blocks * f_bsize` btw

tests-integration/src/system_reinstall.rs
```diff
@@ -0,0 +1,158 @@
...
+            let file_size_gb = available_space_gb.saturating_sub(1);
+
+            let tempfile = tempfile::Builder::new().tempfile_in("/")?;
+            let tempfile_path = tempfile.path();
+
+            let fallocate_status = Command::new("fallocate")
+                .arg("-l")
+                .arg(format!("{}GiB", file_size_gb))
+                .arg(tempfile_path)
+                .status()?;
```
cgwalters: In tests (and even in some prod code paths) I don't have any problem running subprocesses. But for this case:

- It's almost the same amount of code to open a `File` and then invoke https://docs.rs/rustix/latest/rustix/fs/fn.ftruncate.html I'd suspect
- If we were to invoke a subprocess I'd suggest `/bin/truncate -s` instead - the difference here is lazy vs eager space allocation, and there's no reason to be eager here (fallocate)

ckyrouac: Switched to use the rust apis however I think this still needs to be eager so the disk space check fails.

cgwalters: >  I think this still needs to be eager so the disk space check fails.

Ah sorry yes!

PR #1355, opened 2025-06-05
reinstall: Only pull the image if it's not already present

ckyrouac:
This enables using a local image with system-reinstall-bootc. A couple drive by cleanups to the integration tests are included.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,8 +115,8 @@ jobs:
           sudo mkdir -p /run/sshd
 
           sudo install -m 0755 target/release/system-reinstall-bootc /usr/bin/system-reinstall-bootc
-          sudo podman pull quay.io/fedora/fedora-bootc:42
-          sudo bootc-integration-tests system-reinstall quay.io/fedora/fedora-bootc:42 --test-threads=1
+          # These tests may mutate the system live so we can't run in parallel
+          sudo bootc-integration-tests system-reinstall localhost/bootc --test-threads=1
 
           # And the fsverity case
           sudo podman run --privileged --pid=host localhost/bootc-fsverity bootc install to-existing-root --stateroot=other \
--- a/system-reinstall-bootc/src/podman.rs
+++ b/system-reinstall-bootc/src/podman.rs
@@ -96,10 +96,33 @@ pub(crate) fn reinstall_command(image: &str, ssh_key_file: &str) -> Result<Comma
     Ok(command)
 }
 
-pub(crate) fn pull_image_command(image: &str) -> Command {
+fn pull_image_command(image: &str) -> Command {
     let mut command = Command::new("podman");
     command.args(["pull", image]);
     command
+}
+
+fn image_exists_command(image: &str) -> Command {
+    let mut command = Command::new("podman");
+    command.args(["image", "exists", image]);
+    command
+}
+
+pub(crate) fn pull_if_not_present(image: &str) -> Result<()> {
+    let result = image_exists_command(image).status()?;
+
+    if result.success() {
+        println!("Image {} is already present locally, skipping pull.", image);
+        return Ok(());
+    } else {
+        println!("Image {} is not present locally, pulling it now.", image);
+        println!();
+        pull_image_command(image)
+            .run_with_cmd_context()
+            .context(format!("pulling image {}", image))?;
+    }
+
+    Ok(())
 }
 
 /// Path to the podman installation script. Can be influenced by the build
--- a/tests-integration/src/system_reinstall.rs
+++ b/tests-integration/src/system_reinstall.rs
@@ -10,7 +10,7 @@ use std::{
 
 use crate::install;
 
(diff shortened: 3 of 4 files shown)


system-reinstall-bootc/src/podman.rs
```diff
@@ -96,12 +96,38 @@ pub(crate) fn reinstall_command(image: &str, ssh_key_file: &str) -> Result<Comma
...
-pub(crate) fn pull_image_command(image: &str) -> Command {
+fn pull_image_command(image: &str) -> Command {
     let mut command = Command::new("podman");
     command.args(["pull", image]);
     command
 }
 
+fn inspect_image_command(image: &str) -> Command {
+    let mut command = Command::new("podman");
+    command.args(["inspect", image]);
```
cgwalters: Nowadays there's `podman image exists` which is designed for this use case

system-reinstall-bootc/src/podman.rs
```diff
@@ -96,12 +96,38 @@ pub(crate) fn reinstall_command(image: &str, ssh_key_file: &str) -> Result<Comma
...
+}
+
+pub(crate) fn pull_if_not_present(image: &str) -> Result<()> {
+    let result = inspect_image_command(image)
+        .stdout(Stdio::null())
+        .stderr(Stdio::null())
+        .status()?;
+
+    if result.success() {
+        println!("Image {} is already present locally, skipping pull.", image);
```
cgwalters, line 115: I wonder if we need this, feels like a "silence is golden" type of thing?

system-reinstall-bootc/src/podman.rs
```diff
@@ -96,12 +96,38 @@ pub(crate) fn reinstall_command(image: &str, ssh_key_file: &str) -> Result<Comma
...
+        return Ok(());
+    } else {
+        println!("Image {} is not present locally, pulling it now.", image);
```
cgwalters, line 118: Maybe I like brevity too much but how about just: `println!("Pulling: {image}")`

.github/workflows/ci.yml
```diff
@@ -115,8 +115,7 @@ jobs:
           sudo mkdir -p /run/sshd
 
           sudo install -m 0755 target/release/system-reinstall-bootc /usr/bin/system-reinstall-bootc
-          sudo podman pull quay.io/fedora/fedora-bootc:42
-          sudo bootc-integration-tests system-reinstall quay.io/fedora/fedora-bootc:42 --test-threads=1
+          sudo bootc-integration-tests system-reinstall localhost/bootc --test-threads=1
```
cgwalters, line 119: Perhaps worth a comment like `// These tests may mutate the system live so we can't run in parallel` ?

tests-integration/src/system_reinstall.rs
```diff
@@ -62,11 +62,12 @@ pub(crate) fn run(image: &str, testargs: libtest_mimic::Arguments) -> Result<()>
             )?;
 
             // Basic flow stdout verification
+            p.exp_string(format!("Image {image} is already present locally, skipping pull.").as_str())?;
             p.exp_regex("Found only one user ([^:]+) with ([\\d]+) SSH authorized keys.")?;
             p.exp_string("Would you like to import its SSH authorized keys")?;
             p.exp_string("into the root user on the new bootc system?")?;
             p.exp_string("Then you can login as root@ using those keys. [Y/n]")?;
-            p.send_line("a")?;
```
cgwalters, line 69: This was "a" preexisting bug?

cgwalters (commented):
Thanks! Looks sane to me, I just have nits

system-reinstall-bootc/src/podman.rs
ckyrouac: I'm usually in favor of brevity too but my reasoning for this is since system-reinstall-bootc is supposed to be a user friendly tool then being super clear about what is happening might be right. Not a strong opinion if you think less output is better

tests-integration/src/system_reinstall.rs
ckyrouac: sort of, "a" still works because `send_line` sends a newline and the default is "Y". Just more correct to send a "y".

system-reinstall-bootc/src/podman.rs
cgwalters: I don't have a strong opinion either honestly on this specific one

I do though have a more strongly held opinion that at some point we probably want to make the whole output here more consistent/formatted.

One possibility here is basically how Github Actions/Ansible and other things do it with clearly delineated "tasks" that have their own description.
(We could probably get a double win by doing this for `install.rs` too)

cgwalters (approved):

cgwalters:
Looks like it needs an unused import cleanup and a `cargo fmt`

system-reinstall-bootc/src/podman.rs
ckyrouac: filed this to track: https://github.com/bootc-dev/bootc/issues/1356

PR #1359, opened 2025-06-09
test: add e2e test

henrywang:
Add two anaconda and two bib tests

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,101 @@
+name: build bootc and bootc image
+
+on:
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, ubuntu-24.04-arm]
+        # os: [ubuntu-latest]
+        # distro: [fedora-41, fedora-42, fedora-r43, centos-stream-9, centos-stream-10]
+        distro: [centos-stream-9]
+    runs-on: ${{ matrix.os }}
+
+    # Required to push container image to ghcr.io
+    # https://github.com/orgs/community/discussions/57724
+    permissions:
+      contents: read
+      packages: write
+      attestations: write
+      id-token: write
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Build bootc RPM
+        env:
+          DISTRO: ${{ matrix.distro }}
+          BUILD_IMAGE: quay.io/fedora/fedora:42
+        run: |
+          podman run \
+            --rm \
+            --privileged \
+            -v $(pwd):/workdir:z \
+            -e DISTRO=$DISTRO \
+            --workdir /workdir \
+            $BUILD_IMAGE \
+            contrib/scripts/mock_build.sh
+
+      - name: Re-build bootc image with new bootc PRM
+        env:
+          DISTRO: ${{ matrix.distro }}
+        run: |
+          set -xeu
+          if [[ $DISTRO =~ fedora ]]; then
+            VERSION_ID=$(cut -d'-' -f2 <<<"$DISTRO")
+            BASE_IMAGE="quay.io/fedora/fedora-bootc:${VERSION_ID}"
+          fi
+          if [[ $DISTRO =~ centos ]]; then
+            VERSION_ID=$(cut -d'-' -f3 <<<"$DISTRO")
+            BASE_IMAGE="quay.io/centos-bootc/centos-bootc:stream${VERSION_ID}"
+          fi
+
+          tee target/build/Containerfile >/dev/null <<CONTAINERFILEEOF
+          FROM $BASE_IMAGE
+          RUN dnf -y upgrade /rpms/*.rpm && dnf -y clean all && rm -rf /var/cache /var/lib/dnf
+          CONTAINERFILEEOF
+          cat target/build/Containerfile
+
+          IMAGE_NAME="bootc-image:${{ github.sha }}-$(uname -m)"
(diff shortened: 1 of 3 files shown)


.packit.yaml
```diff
@@ -81,7 +83,31 @@ jobs:
...
+      - centos-stream-9-x86_64
+      - centos-stream-9-aarch64
+      - centos-stream-10-x86_64
+      - centos-stream-10-aarch64
+      - fedora-41-x86_64
+      - fedora-41-aarch64
+      - fedora-42-x86_64
+      - fedora-42-aarch64
+      - fedora-rawhide-x86_64
+      - fedora-rawhide-aarch64
```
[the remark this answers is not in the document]
cgwalters: Maybe we could use [yaml anchors](https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/) to avoid the duplication?

cgwalters:
Is testing farm still using ec2 `metal` instances or internal openstack for virt? I still really think that we should expand that infrastructure set. In particular both Azure and GCP support nested KVM and I think that's all we need for basic sanity testing. Actually even the default GHA runners have it.

Speaking of GHA, also as part of the CNCF migration we now have access to larger runners and I think it would make sense to use those for at least some workload.

In general (as this relates to https://github.com/bootc-dev/bootc/pull/1358 too) I believe for the forseeable future we will have multiple CI systems and while it's complicated I think it makes sense to try to ensure that we've factored out our testing such that it can be somewhat infrastructure agnostic. At least for baseline testing flows, which a lot of this is right?

See https://cloud-native.slack.com/archives/C014YQ8CDCG/p1749065398627899 for a chat around the CNCF GHA runners.

henrywang:
Will try `oracle-vm` as test runner.

cgwalters:
To me one of the biggest problems here is that `tmt` is designed to run from a host system and it's only packaged for Fedora even! But if we get to the point of running `tmt` from a container image, that makes our "infrastructure" way more portable - us and others can use the default GHA Ubuntu runners, or the new CNCF Oracle runners etc.

To flesh this out a bit more what I think would work well is:

- Start a GHA Ubuntu runner
- Do `apt install libvirt`
- podman run --privileged -v /run/libvirt:/run/libvirt quay.io/bootc-sdk-container ...

Here basically that SDK/test container (which would have e.g. `tmt` inside it) would actually be talking to the host libvirtd to script it.

cgwalters:
I granted organization-wide GITHUB_TOKEN access to write to `packages`.

henrywang:
> I granted organization-wide GITHUB_TOKEN access to write to `packages`.

Thanks!

I'll add workflow task to deploy GCP instance, then run anaconda and bib tests on it. And will add a clean up job to destroy GCP instance.

cgwalters:
It sounds like the GCP route isn't going to pan out. Can we invest in having the e2e test itself support being run in an (optionally privileged) container image? In chat one thing that came up on this is of course we invoke `podman run` in some cases (e.g. bootc-image-builder), but for those we can "escape" and spawn containers in the host environment. One mechanism to do this is to bind mount in the podman socket.

henrywang:
> It sounds like the GCP route isn't going to pan out. Can we invest in having the e2e test itself support being run in an (optionally privileged) container image? In chat one thing that came up on this is of course we invoke `podman run` in some cases (e.g. bootc-image-builder), but for those we can "escape" and spawn containers in the host environment. One mechanism to do this is to bind mount in the podman socket.

Thanks @cgwalters , already added into JIRA ticket and will work on this this week.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,68 @@
+
+name: e2e
+
+# triggered by labeled PR
+on:
+  pull_request_target:
```
cgwalters: This one is super dangerous in general, xref https://blog.gitguardian.com/github-actions-security-cheat-sheet/

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,68 @@
...
+        os: [ubuntu-latest]
+        # distro: [fedora-41, fedora-42, fedora-r43, centos-stream-9, centos-stream-10]
+        distro: [centos-stream-10]
+    runs-on: ${{ matrix.os }}
+
+    # Required to push container image to ghcr.io
+    # https://github.com/orgs/community/discussions/57724
+    permissions:
+      contents: read
+      packages: write
```
cgwalters, line 21: Per discussion let's create a default flow in Github Actions that builds a bootc container image *tagged with the sha digest* and pushes to `ghcr.io` under the repo namespace.

Then we can run further tests on that container image.

I think we'll need a scheduled workflow to prune the images though.

henrywang: We can use https://github.com/marketplace/actions/ghcr-io-cleanup-action action to make clean up container registry, like https://github.com/marketplace/actions/ghcr-io-cleanup-action#older-than.

cgwalters: @henrywang you're saying the push permission here wasn't working, can you link to a failing run?

I am not aware of anything special at all here, a ton of projects use a similar flow so it *should* work - we definitely have this enabled at the repo level

image

cgwalters:
So we chatted about this and one option that came up is to push the built container image as an artifact, and then process it in a separate workflow triggered by that one. I think this is a standard pattern.

Another model though I think would work well is basically:

- PR submitted
- Trusted workflow allocates temporary remote resources (e.g. `quay.io/bootc-dev/test-<shadigest>` repository) with pull secret/`auth.json` that can only write to that temporary repository
- trigger a build that can push there
- trigger e2e test flow (separate workflow, not `pull-request-target`) targeting that container image

PR #1365, opened 2025-06-12
ostree-ext: specify zeroed patch version for ostree

jeckersb:
This was changed previously in #1328, this reverts to the old format.

### DIFF
--- a/ostree-ext/Cargo.toml
+++ b/ostree-ext/Cargo.toml
@@ -14,7 +14,7 @@ version = "0.15.3"
 # semver here you must also bump our semver.
 containers-image-proxy = "0.7.0"
 # We re-export this library too.
-ostree = { features = ["v2025_2"], version = "0.20" }
+ostree = { features = ["v2025_2"], version = "0.20.0" }
 
 # Private dependencies
 anyhow = { workspace = true }


jeckersb:
I'm confused as to why this is required, since `0.20` and `0.20.0` should be [equivalent when specifying via cargo](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#default-requirements), but without it when trying to build rpm-ostree against main here I get:

```
    Updating git repository `https://github.com/containers/bootc`
    Updating git repository `https://github.com/containers/composefs-rs`
    Updating crates.io index
error: failed to select a version for `ostree`.
    ... required by package `ostree-ext v0.15.3 (https://github.com/containers/bootc?rev=a9f01fe59ff656142704c7e1181d43f9d63b5d56#a9f01fe5)`
    ... which satisfies git dependency `ostree-ext` of package `rpmostree-rust v0.1.0 (/var/home/jeckersb/git/rpm-ostree)`
versions that meet the requirements `^0.20` (locked to 0.20.0) are: 0.20.0

the package `ostree-ext` depends on `ostree`, with features: `v2025_2` but `ostree` does not have these features.


failed to select a version for `ostree` which could resolve this conflict
```

:shrug:

jeckersb:
Actually now it's not working even with this, so I think maybe I just have some weird cargo state locally or something.

I'm marking as draft and then going to sleep, this is a sign I should stop messing with this when it's almost midnight :laughing:

cgwalters:
I hit something like this in the past and it was super subtle, basically the caching in the `build.rs` for pkg-config doesn't correctly invalidate when you update system libraries; i.e. this flow:

- have bootc local build with older `libostree` (C shared library) with populated `target/` build cache
- update ostree
- `cargo build` won't correctly re-probe for an updated ostree features

two hard problems in computer science etc

cgwalters:
Actually sorry a problem may be that you also need to `cargo update ostree`?

jeckersb:
> Actually sorry a problem may be that you also need to `cargo update ostree`?

Yeah that was most of it :facepalm:.  Closing this.

PR #1368, opened 2025-06-13
README.md: Add details for Friday Zoom meeting

jeckersb:

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ start by looking at [ADOPTERS.md](ADOPTERS.md).
 
 - [Github discussion forum](https://github.com/containers/bootc/discussions) for async discussion
 - [#bootc-dev on CNCF Slack](https://cloud-native.slack.com/archives/C08SKSQKG1L) for live chat
+- Recurring live meeting hosted on [CNCF Zoom](https://zoom-lfx.platform.linuxfoundation.org/meeting/96540875093?password=7889708d-c520-4565-90d3-ce9e253a1f65) each Friday at 15:30 UTC.
 
 This project is also tightly related to the previously mentioned Fedora/CentOS bootc project,
 and many developers monitor the relevant discussion forums there. In particular there's a
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -99,7 +99,7 @@ This requires two votes from active maintainers.
 ## Meetings
 
 Time zones permitting, Maintainers are expected to participate in the public
-developer meeting, which occurs at 3:30 PM ET on Thursdays via [Zoom](TODO).
+developer meeting, which occurs at 15:30 UTC on Fridays via [Zoom](https://zoom-lfx.platform.linuxfoundation.org/meeting/96540875093?password=7889708d-c520-4565-90d3-ce9e253a1f65).
 
 Maintainers will also have closed meetings in order to discuss security reports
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -44,6 +44,7 @@ start by looking at [ADOPTERS.md](ADOPTERS.md).
 
 - [Github discussion forum](https://github.com/containers/bootc/discussions) for async discussion
 - [#bootc-dev on CNCF Slack](https://cloud-native.slack.com/archives/C08SKSQKG1L) for live chat
+- Recurring [Jitsi live meeting](https://meet.jit.si/YourOperatingSystemIsAContainer) each Friday at 11:30 EDT/ 15:30 UTC.  (Currently ambiguous what happens during daylight saving time switch, if in doubt ask in Slack!)
```
jeckersb: Unclear if we want to bikeshed what happens when dst switches here.  Also this currently references summer time.

Timezones are hard.

cgwalters: Hmm I think what we maybe should do for this is randomly change the Jitsi instance name, and say we'll announce it in Slack beforehand. The problem is that I think if no one else is there the first person who joins can "own" the meeting.

jeckersb: I was the first to join (I think so at least?) on Friday and it told me I couldn't start the meeting because I had to wait for the owner to join or something like that.  So I don't think just anyone can become the owner but we can test it at some point.

dustymabe: suggest to use UTC and not change with DST switches. We've been doing that for FCOS for some time and it seems to work well.

jmarrero: I think we should move to Zoom and give varios maintainer's ownership of the meeting. To avoid the random meeting links.

Also It looks like recordings in jitsi are not super simple. Requires either live streaming to youtube or other software.

For the time, I think Dusty's suggestion makes sense.

In any case given that we need something to point to socialize this meeting or make it official maybe we merge as is and bikeshed in an issue?

cgwalters: my votes are 

- use UTC
- Use CNCF zoom

cgwalters: And skip recording for now but we keep a summary of things and then be sure we've updated any issues/PRs that were discussed

jeckersb: Do we still need to sort out getting access to the CNCF zoom?  I see https://github.com/cncf/sandbox/issues/339#issuecomment-2730445555 but nothing else in that thread about it.  Not sure if it was handled elsewhere and I'm just missing it.

jmarrero: They gave me access and I can give access to as many hosts as needed it seems. The link is: https://zoom-lfx.platform.linuxfoundation.org/meeting/96540875093?password=7889708d-c520-4565-90d3-ce9e253a1f65  I also added @jeckersb and @cgwalters as hosts. I could not find @ckyrouac and @gursewak1997 profiles on the CNCF system.

jeckersb: Ok I've updated this to use the Zoom link and to refer to the time only as UTC.

jmarrero:
@jeckersb can you update the Governence file too? we had a TODO there: https://github.com/bootc-dev/bootc/pull/1387/files

jeckersb:
> @jeckersb can you update the Governence file too? we had a TODO there: https://github.com/bootc-dev/bootc/pull/1387/files

Done

cgwalters (approved):

Issue #1353, opened 2025-06-05
Create `/run/reboot-required` by default


cgwalters:
https://github.com/kubereboot/kured expects this and I can't imagine there's downsides to us just doing this by default and avoiding people needing to create bridging manually.

vlaborie:
Hi,

I use Kured on Kubernetes with `bootc` node, for info it's work just by setting the `reboot-sentinel` param to `/run/ostree/staged-deployment`, eg. with Helm values:

~~~
values:
  configuration:
    rebootSentinel: /run/ostree/staged-deployment
~~~

But create `/run/reboot-required` can be usefull for a working Kured config by default.

PR #1378, opened 2025-06-17
Create /run/reboot-required by default

gursewak1997:
When staging a new deployment, create /run/reboot-required to signal that a reboot is needed. This file is monitored by kured (Kubernetes Reboot Daemon) and other tools to detect when a system needs to be rebooted.

This makes it easier to integrate bootc with kured and similar tools without requiring manual configuration or bridging.

Fixes: [#1353](https://github.com/bootc-dev/bootc/issues/1353)

### DIFF
--- a/lib/src/deploy.rs
+++ b/lib/src/deploy.rs
@@ -737,6 +737,13 @@ pub(crate) async fn stage(
     })
     .await;
 
+    // Unconditionally create or update /run/reboot-required to signal a reboot is needed.
+    // This is monitored by kured (Kubernetes Reboot Daemon).
+    let run_dir = Dir::open_ambient_dir("/run", cap_std::ambient_authority())?;
+    run_dir
+        .atomic_write("reboot-required", b"")
+        .context("Creating /run/reboot-required")?;
+
     Ok(())
 }
 
--- a/tmt/tests/booted/test-image-pushpull-upgrade.nu
+++ b/tmt/tests/booted/test-image-pushpull-upgrade.nu
@@ -65,6 +65,10 @@ RUN echo test content > /usr/share/blah.txt
     let progress = open --raw $progress_json | from json -o
     sanity_check_switch_progress_json $progress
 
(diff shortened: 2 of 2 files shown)


lib/src/deploy.rs
```diff
@@ -737,6 +737,13 @@ pub(crate) async fn stage(
     })
     .await;
 
+    // Create /run/reboot-required to signal that a reboot is needed
+    // This file is monitored by kured (Kubernetes Reboot Daemon)
+    let run_dir = Dir::open_ambient_dir("/run", cap_std::ambient_authority())?;
+    if !run_dir.try_exists("reboot-required")? {
+        run_dir.atomic_write("reboot-required", b"").context("Creating /run/reboot-required")?;
+    }
```
cgwalters: I think I'd say we should drop the conditional and *always* do the write - there's no harm to doing it, and updating the timestamp on the file also acts a signal that something changed again.

cgwalters (commented):
Thanks! One nit, and can you also add an integration test in say `test-image-pushpull-upgrade.nu` around like 68?

tmt/tests/booted/test-image-pushpull-upgrade.nu
```diff
@@ -65,6 +65,12 @@ RUN echo test content > /usr/share/blah.txt
     let progress = open --raw $progress_json | from json -o
     sanity_check_switch_progress_json $progress
 
+    # Check that /run/reboot-required exists and is a zero-byte file
+    let rr_meta = ls /run/reboot-required
+    assert equal $rr_meta.name "/run/reboot-required"
```
cgwalters: I think this is failing because `ls` returns a table. It should work if you do `ls /run/reboot-required | first` instead.

cgwalters:
Yeah confusingly nushell has a dedicated "size" type that's not an integer, I fixed the test

cgwalters (approved):

PR #1380, opened 2025-06-19
deploy: short-circuit pull if digest pullspec already exists

jlebon:
If we're pulling by digest and the pullspec already exists, then there's no need to reach out to the registry or even spawn skopeo.

Detect this case and exit early in the pull code.

This allows RHCOS to conform better to the PinnedImageSet API[1], where the expectation is that once an image is pulled, the registry will not be contacted again. In a future with unified storage, the MCO's pre-pull would work just the same for the RHCOS image as any other.

Framing this more generally: this patch allows one to pre-pull an image into the store, and making the later deployment operation be fully offline. E.g. this could be used to implement a `bootc switch --download-only` option.

[1] https://github.com/openshift/enhancements/blob/26ce3cd8a0c7ce650e73bc5393a3605022cb6847/enhancements/machine-config/pin-and-pre-load-images.md

### DIFF
--- a/lib/src/deploy.rs
+++ b/lib/src/deploy.rs
@@ -4,6 +4,7 @@
 
 use std::collections::HashSet;
 use std::io::{BufRead, Write};
+use std::str::FromStr;
 
 use anyhow::Ok;
 use anyhow::{anyhow, Context, Result};
@@ -15,6 +16,7 @@ use ostree::{gio, glib};
 use ostree_container::OstreeImageReference;
 use ostree_ext::container as ostree_container;
 use ostree_ext::container::store::{ImageImporter, ImportProgress, PrepareResult, PreparedImport};
+use ostree_ext::oci_spec::distribution::Reference as OciReference;
 use ostree_ext::oci_spec::image::{Descriptor, Digest};
 use ostree_ext::ostree::Deployment;
 use ostree_ext::ostree::{self, Sysroot};
@@ -343,6 +345,20 @@ pub(crate) async fn prepare_for_pull(
     let imgref_canonicalized = imgref.clone().canonicalize()?;
     tracing::debug!("Canonicalized image reference: {imgref_canonicalized:#}");
     let ostree_imgref = &OstreeImageReference::from(imgref_canonicalized);
+
+    let maybe_oci_ref = ostree_imgref.imgref.name.parse::<OciReference>().ok();
+    if let Some(digest) = maybe_oci_ref.as_ref().and_then(|oci_ref| oci_ref.digest()) {
+        // This is a digested pull. If that digested image already exists in
+        // the store, we can short-circuit a lot here and e.g. not spawn skopeo
+        // at all.
+        if let Some(c) = ostree_container::store::query_image(repo, &ostree_imgref.imgref)? {
+            let digest = Digest::from_str(digest)?;
+            assert_eq!(digest, c.manifest_digest);
+            println!("Digest-based pullspec {imgref:#} already present");
+            return Ok(PreparedPullResult::AlreadyPresent(Box::new((*c).into())));
+        }
+    }
+
     let mut imp = new_importer(repo, ostree_imgref).await?;
     if let Some(target) = target_imgref {
         imp.set_target(target);


jlebon:
If we do this, I'll probably add the same to rpm-ostree (... will be nice once we get to https://github.com/coreos/rpm-ostree/issues/4994).

Also this needs a test.

lib/src/deploy.rs
```diff
@@ -343,6 +345,20 @@ pub(crate) async fn prepare_for_pull(
...
     let ostree_imgref = &OstreeImageReference::from(imgref_canonicalized);
+
+    let maybe_oci_ref = ostree_imgref.imgref.name.parse::<OciReference>().ok();
+    if let Some(digest) = maybe_oci_ref.as_ref().and_then(|oci_ref| oci_ref.digest()) {
+        // This is a digested pull. If that digested image already exists in
+        // the store, we can short-circuit a lot here and e.g. not spawn skopeo
+        // at all.
+        if let Some(c) = ostree_container::store::query_image(repo, &ostree_imgref.imgref)? {
+            let digest = Digest::from_str(digest)?;
+            assert_eq!(digest, c.manifest_digest);
```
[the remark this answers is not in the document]
jlebon: I guess so. That really should never happen because the image was looked up in the store by a digested pullspec, though it doesn't hurt either.

cgwalters (approved):
Thanks!

cgwalters:
- I downgraded the `println!` to a `tracing::debug!` since I think silence-is-golden in this case
- Now that I look closer I think this would be better in the ostree-ext code

jlebon:
> I downgraded the println! to a tracing::debug! since I think silence-is-golden in this case

WFM. Just to note though, this was an attempt to match the "No changes" code lower down for the `AlreadyPresent` case. (But using a different string to make the cases distinguishable for e.g. testing/debugging.)

> Now that I look closer I think this would be better in the ostree-ext code

Heh, that was actually what I did at first, trying to make it directly part of `ImageImporter::prepare()` since it's what already returns `AlreadyPresent`. The problem is that `new_importer()` already forks skopeo, and crucially does `OpenImage` which will reach out to the registry, even before we get there.

Or do you mean just shoving that code into a non-member utility function in ostree-ext?

jlebon:
Alternatively I guess we switch `proxy_img` to be an `Option<OpenedImage>` and delay initialization until `prepare()`.

cgwalters:
Yep let me take this one

cgwalters:
OK I redid this, can you review?

ostree-ext/src/container/store.rs
```diff
@@ -682,8 +691,35 @@ impl ImageImporter {
...
+        {
+            if let Some(previous_state) = previous_state {
+                // A digested pull spec, and our existing state matches.
+                if previous_state.manifest_digest == target_digest {
+                    tracing::debug!("Digest-based pullspec {:?} already present", self.imgref);
+                    return Ok(PrepareResult::AlreadyPresent(previous_state));
+                }
+                Some(previous_state)
+            } else {
+                None
```
jlebon, line 713: Shouldn't this be `previous_state`?

ostree-ext/src/container/store.rs
```diff
@@ -682,8 +691,35 @@ impl ImageImporter {
             _ => {}
         }
 
+        // Check if we have an image already pulled
         let previous_state = try_query_image(&self.repo, &self.imgref.imgref)?;
 
+        // Parse the target reference to see if it's a digested pull
+        let target_reference = self.imgref.imgref.name.parse::<Reference>().ok();
+        let previous_state = if let Some(target_digest) = target_reference
```
jlebon, line 699: Yeah... I know why we do this (to not lose ownership of `previous_state`). What I had done in my first attempt which was in this same spot was:

```rust
let oci_ref = self.imgref.imgref.name.parse::<OciReference>().ok();
if let Some(digest) = oci_ref.as_ref().and_then(|oci_ref| oci_ref.digest()) {
    let digest = Digest::from_str(digest)?;
    if previous_state.as_ref()
        .map(|state| state.manifest_digest == digest)
        .unwrap_or_default()
    {
        // SAFETY: It can't be None if the map().unwrap() above gave
        // true. We do it this way so that we only consume the Option if
        // we return here. That way it can be reused below if not.
        return Ok(PrepareResult::AlreadyPresent(previous_state.unwrap()));
    }
}
```

which yes, does an unwrap (though is guaranteed safe), but has a cleaner flow.

ostree-ext/src/container/store.rs
cgwalters: No, because we destructured above - here we know `previous_state` is None.

ostree-ext/src/container/store.rs
cgwalters: That version is definitely a lot cleaner to read, but I usually try pretty hard to avoid `unwrap()`.

jlebon:
I can't officially LGTM, but LGTM. Thanks for picking this up! Having it in ostree-ext means rpm-ostree should also get this behaviour.

jeckersb (approved):
LGTM on a cursory glance, mostly trusting @jlebon's +1 and making it official :smile:

PR #1386, opened 2025-06-26
ci: periodically run update-generated

miabbott:
This introduces a GitHub workflow to periodically run `cargo xtask update-generated` in order to keep the docs up-to-date with the latest changes. It will create a PR if changes are detected.

Signed-off-by: Micah Abbott <miabbott@redhat.com>


miabbott:
I put this together with some help from Gemini.

I don't know if auto-generated PRs are desired in this repo, so feel free to close it if not.

.github/workflows/udpate-generated.yml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: 'Install pandoc'
+        run: dnf -y install pandoc
+        shell: bash
+
+      - name: 'Update generated docs'
+        run: cargo xtask update-generated
+        shell: bash
+
+      - name: 'Create Pull Request'
```
cgwalters, line 37: I think this is missing a step like https://github.com/coreos/fedora-coreos-config/blob/ba564a93c463f9bd051cf3333822138f38bf0a69/.github/workflows/bump-fedora-bootc.yaml#L36 to do a diff right?

cgwalters (changes requested):
Thanks!

.github/workflows/udpate-generated.yml
miabbott: Based on my read of how the action behaves, it will silently exit if there are no changes detected - https://github.com/peter-evans/create-pull-request?tab=readme-ov-file#action-behaviour

> How the action behaves:
> 
>   - If there are changes (i.e. a diff exists with the checked-out base branch), the changes will be pushed to a new branch and a pull request created.
>   - If there are no changes (i.e. no diff exists with the checked-out base branch), no pull request will be created and the action exits silently.

When I tested this in my fork, it correctly created a PR that looked like https://github.com/bootc-dev/bootc/pull/1385

Hmm...I should apply those changes to my `main` branch and see if the action just no-ops.  Will report back here with results!

cgwalters: Ohh hmm, then maybe some of these FCOS actions have been unnecessarily doing a diff? Either way thanks so much for testing this out!

miabbott: Looks like my understanding was correct; after applying the generated changes to my `main` branch, the next time I ran the action there was no PR generated.

https://github.com/miabbott/bootc/actions/runs/15917569924/job/44897979972

> Ohh hmm, then maybe some of these FCOS actions have been unnecessarily doing a diff? 

In the example you linked, I think you could simplify the "Check if there are new commits" step  to just do a `git submodule update`. If there was a new submodule commit, the PR action would generate the PR otherwise it would no-op.

.github/workflows/udpate-generated.yml
```diff
@@ -0,0 +1,52 @@
+name: 'Update Generated Files'
+
+# This workflow is triggered on a schedule (every day at 3:15 AM UTC)
+# and can also be triggered manually from the GitHub Actions UI.
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '15 3 * * *'
```
miabbott, line 8: Another option would be to have this fire on every push to `main`

```
on:
  push:
    branches:
      - main
```

cgwalters: That said, we could also just do this as a check on PRs and require the submitter to do it...

But yeah since you already wrote it this way I'm fine to run with it for now.

PR #1387, opened 2025-06-26
GOVERNANCE: Update meeting details

jmarrero:
Make sure we link to the jit.si meeting we have been using.

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -99,7 +99,7 @@ This requires two votes from active maintainers.
 ## Meetings
 
 Time zones permitting, Maintainers are expected to participate in the public
-developer meeting, which occurs at 3:30 PM ET on Thursdays via [Zoom](TODO).
+developer meeting, which occurs at 11:30 AM ET on Fridays via [jit.si](https://meet.jit.si/YourOperatingSystemIsAContainer).
 
 Maintainers will also have closed meetings in order to discuss security reports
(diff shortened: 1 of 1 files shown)


GOVERNANCE.md
```diff
@@ -99,7 +99,7 @@ This requires two votes from active maintainers.
 ## Meetings
 
 Time zones permitting, Maintainers are expected to participate in the public
-developer meeting, which occurs at 3:30 PM ET on Thursdays via [Zoom](TODO).
+developer meeting, which occurs at 11:30 AM ET on Fridays via [jit.si](https://meet.jit.si/YourOperatingSystemIsAContainer).
```
cgwalters, line 102:
```suggestion
developer meeting, which occurs at 11:30 AM EST on Fridays via [jit.si](https://meet.jit.si). The actual URL will be posted on the Slack channel a few minutes before the meeting.
```

That said it'd be better to have the time be in UTC not EST to be a bit more timezone friendly and not have people need to remember when EST changes daylight savings etc.

That said also I think the best thing here is to have an ical entry or so? Surely there's some best practice here.

cgwalters:
Wait this PR just duplicates  https://github.com/bootc-dev/bootc/pull/1368 right?

jmarrero:
Yeah, sorry when trying to socialize the meeting I was looking for where to point and did not see our meeting in the doc. Let's merge the other one and bikeshed later?

PR #1389, opened 2025-06-30
install: Add reset

cgwalters:
This is a nondestructive variant of `to-existing-root`.


cgwalters:
This continues in https://github.com/bootc-dev/bootc/pull/1588

PR #1391, opened 2025-06-30
packaging/rpm: Be sure we pull in chcon (and util-linux)

cgwalters:
We execute it, and Anaconda was somehow omitting this.

Also pull in util-linux-core because we also depend on that.

https://github.com/bootc-dev/bootc/discussions/1383#discussioncomment-13585806


jmarrero (approved):
lgtm

Issue #1350, opened 2025-06-03
Support for soft reboots


cgwalters:
This will build on https://github.com/ostreedev/ostree/pull/3420 and we're hoping to land this by at least the Fedora 43 timeframe.

---

TODO:
 
- [x] update to newer ostree rust APIs
- [x] Expose in status whether we are capable of soft reboots to the staged deployment
- [x] use api to detect soft reboot capable (ostree)
- [ ] handle case where kargs are different on the container image than local state `/usr/lib/bootc/kargs.d`. (bootc)
- [ ] Debug https://github.com/ostreedev/ostree/issues/3365
- [ ] Debug `[  388.410231] systemd[1]: Failed to load new SELinux policy. Continuing with old policy.` (this may be a serious problem!)
- [x] UNDO: `bootc upgrade --apply` should automatically use soft reboots if capable (detect via new API); though we may need an opt-out flag for unusual cases `bootc upgrade --apply --skip-soft-reboot`?
- [x] Handle cases (RHEL9) where soft reboot is not supported by systemd (detect missing `soft-reboot.target`)
- [x] Opt-in and opt-out mechanisms (see below)

### Detailed API

We need more sophistication for agent-driven cases that want precise control over the flow of every step in:

- download image (also xref https://github.com/bootc-dev/bootc/issues/1375 )
- queue for update (xref https://github.com/bootc-dev/bootc/issues/1320)
- detect when update can or cannot be soft-rebooted (on bare metal, the difference in downtime can be *minutes*)
- Control whether or not soft reboot is used

#### Opt-in soft reboots

- `systemctl enable bootc-soft-reboot.service`
- CLI flag `bootc upgrade --queue-soft-reboot`? (Problem: tools like flightctl, gnome-software etc. would all need to learn how to opt in to this) - this just calls `systemctl enable --runtime bootc-soft-reboot.service`

Precedent: `rhsm-publish-facts.service` which monitors bootc state for modifications

dustymabe:
> To start, probably default to soft reboots for non-apply too.

I was doing some thinking on this. As far as defaulting to soft-reboot it would be interesting to be able to base this on how long the machine takes to boot (we can look at logs for this, like how long did it take for the system to get to default.target or some other target)? My feeling is that if your machine boots quickly you'd almost always prefer hard reboot (at least I would, because that tests every part of process that's needed to bring your machine fully up). In the case where your machine takes a long time to boot, then yeah, give me soft-reboot please.

[...]

PR #1392, opened 2025-06-30
lib: add support for soft-reboots

jmarrero:
Closes: https://github.com/bootc-dev/bootc/issues/1350

depends on

- [x] new ostree release
- [x] new ostree rust bindings update
- [x] another ostree release with https://github.com/ostreedev/ostree/commit/f97feced6512767c5f890aa67be39d1747107311
- [x] https://gitlab.com/redhat/centos-stream/containers/bootc/-/issues/1193

### DIFF
--- a/crates/lib/src/cli.rs
+++ b/crates/lib/src/cli.rs
@@ -80,6 +80,13 @@ pub(crate) struct UpgradeOpts {
     #[clap(long, conflicts_with = "check")]
     pub(crate) apply: bool,
 
+    /// Configure soft reboot behavior.
+    ///
+    /// 'required' will fail if soft reboot is not available.
+    /// 'auto' will use soft reboot if available, otherwise fall back to regular reboot.
+    #[clap(long = "soft-reboot", conflicts_with = "check")]
+    pub(crate) soft_reboot: Option<SoftRebootMode>,
+
     #[clap(flatten)]
     pub(crate) progress: ProgressOptions,
 }
@@ -98,6 +105,13 @@ pub(crate) struct SwitchOpts {
     /// a userspace-only restart.
     #[clap(long)]
     pub(crate) apply: bool,
+
+    /// Configure soft reboot behavior.
+    ///
+    /// 'required' will fail if soft reboot is not available.
+    /// 'auto' will use soft reboot if available, otherwise fall back to regular reboot.
+    #[clap(long = "soft-reboot")]
+    pub(crate) soft_reboot: Option<SoftRebootMode>,
 
     /// The transport; e.g. oci, oci-archive, containers-storage.  Defaults to `registry`.
     #[clap(long, default_value = "registry")]
@@ -142,6 +156,13 @@ pub(crate) struct RollbackOpts {
     /// a userspace-only restart.
     #[clap(long)]
     pub(crate) apply: bool,
+
+    /// Configure soft reboot behavior.
+    ///
+    /// 'required' will fail if soft reboot is not available.
+    /// 'auto' will use soft reboot if available, otherwise fall back to regular reboot.
+    #[clap(long = "soft-reboot")]
+    pub(crate) soft_reboot: Option<SoftRebootMode>,
 }
 
 /// Perform an edit operation
@@ -165,6 +186,15 @@ pub(crate) enum OutputFormat {
     Yaml,
     /// Output in JSON format.
     Json,
+}
+
+#[derive(Debug, Clone, Copy, ValueEnum, PartialEq, Eq)]
+#[clap(rename_all = "lowercase")]
+pub(crate) enum SoftRebootMode {
+    /// Require a soft reboot; fail if not possible
+    Required,
+    /// Automatically use soft reboot if possible, otherwise use regular reboot
+    Auto,
 }
 
 /// Perform an status operation
@@ -562,7 +592,7 @@ pub(crate) enum Opt {
         Note on Rollbacks and the `/etc` Directory:
 
         When you perform a rollback (e.g., with `bootc rollback`), any
-        changes made to files in the `/etc` directory won’t carry over
+        changes made to files in the `/etc` directory won't carry over
         to the rolled-back deployment.  The `/etc` files will revert
         to their state from that previous deployment instead.
 
@@ -741,6 +771,104 @@ pub(crate) fn require_root(is_container: bool) -> Result<()> {
     Ok(())
 }
 
+/// Check if a deployment has soft reboot capability
+fn has_soft_reboot_capability(deployment: Option<&crate::spec::BootEntry>) -> bool {
+    deployment.map(|d| d.soft_reboot_capable).unwrap_or(false)
+}
+
+/// Prepare a soft reboot for the given deployment
+#[context("Preparing soft reboot")]
+fn prepare_soft_reboot(
+    sysroot: &crate::store::Storage,
+    deployment: &ostree::Deployment,
+) -> Result<()> {
+    let cancellable = ostree::gio::Cancellable::NONE;
+    sysroot
+        .sysroot
+        .deployment_set_soft_reboot(deployment, false, cancellable)
+        .context("Failed to prepare soft-reboot")?;
+    Ok(())
+}
+
+/// Handle soft reboot based on the configured mode
+#[context("Handling soft reboot")]
+fn handle_soft_reboot<F>(
+    soft_reboot_mode: Option<SoftRebootMode>,
+    entry: Option<&crate::spec::BootEntry>,
+    deployment_type: &str,
+    execute_soft_reboot: F,
+) -> Result<()>
+where
+    F: FnOnce() -> Result<()>,
+{
+    let Some(mode) = soft_reboot_mode else {
+        return Ok(());
+    };
+
+    let can_soft_reboot = has_soft_reboot_capability(entry);
+    match mode {
+        SoftRebootMode::Required => {
+            if can_soft_reboot {
+                execute_soft_reboot()?;
+            } else {
+                anyhow::bail!(
+                    "Soft reboot was required but {} deployment is not soft-reboot capable",
+                    deployment_type
+                );
+            }
+        }
+        SoftRebootMode::Auto => {
+            if can_soft_reboot {
+                execute_soft_reboot()?;
+            }
+        }
+    }
+    Ok(())
+}
+
+/// Handle soft reboot for staged deployments (used by upgrade and switch)
+#[context("Handling staged soft reboot")]
+fn handle_staged_soft_reboot(
+    sysroot: &crate::store::Storage,
+    soft_reboot_mode: Option<SoftRebootMode>,
+    host: &crate::spec::Host,
+) -> Result<()> {
+    handle_soft_reboot(
+        soft_reboot_mode,
+        host.status.staged.as_ref(),
+        "staged",
+        || soft_reboot_staged(sysroot),
+    )
+}
+
+/// Perform a soft reboot for a staged deployment
+#[context("Soft reboot staged deployment")]
+fn soft_reboot_staged(sysroot: &crate::store::Storage) -> Result<()> {
+    println!("Staged deployment is soft-reboot capable, preparing for soft-reboot...");
+
+    let deployments_list = sysroot.deployments();
+    let staged_deployment = deployments_list
+        .iter()
+        .find(|d| d.is_staged())
+        .ok_or_else(|| anyhow::anyhow!("Failed to find staged deployment"))?;
+
+    prepare_soft_reboot(sysroot, staged_deployment)?;
+    Ok(())
+}
+
+/// Perform a soft reboot for a rollback deployment
+#[context("Soft reboot rollback deployment")]
+fn soft_reboot_rollback(sysroot: &crate::store::Storage) -> Result<()> {
+    println!("Rollback deployment is soft-reboot capable, preparing for soft-reboot...");
+
+    let deployments_list = sysroot.deployments();
+    let target_deployment = deployments_list
+        .first()
+        .ok_or_else(|| anyhow::anyhow!("No rollback deployment found!"))?;
+
+    prepare_soft_reboot(sysroot, target_deployment)
+}
+
 /// A few process changes that need to be made for writing.
 /// IMPORTANT: This may end up re-executing the current process,
 /// so anything that happens before this should be idempotent.
@@ -813,6 +941,7 @@ async fn upgrade(opts: UpgradeOpts) -> Result<()> {
     let booted_image = host
         .status
         .booted
+        .as_ref()
         .map(|b| b.query_image(repo))
         .transpose()?
         .flatten();
@@ -859,7 +988,7 @@ async fn upgrade(opts: UpgradeOpts) -> Result<()> {
             .unwrap_or_default();
         if staged_unchanged {
             println!("Staged update present, not changed.");
-
+            handle_staged_soft_reboot(sysroot, opts.soft_reboot, &host)?;
             if opts.apply {
                 crate::reboot::reboot()?;
             }
@@ -880,6 +1009,13 @@ async fn upgrade(opts: UpgradeOpts) -> Result<()> {
     }
     if changed {
         sysroot.update_mtime()?;
+
+        if opts.soft_reboot.is_some() {
+            // At this point we have new staged deployment and the host definition has changed.
+            // We need the updated host status before we check if we can prepare the soft-reboot.
+            let updated_host = crate::status::get_status(sysroot, Some(&booted_deployment))?.1;
+            handle_staged_soft_reboot(sysroot, opts.soft_reboot, &updated_host)?;
+        }
 
         if opts.apply {
             crate::reboot::reboot()?;
@@ -956,6 +1092,13 @@ async fn switch(opts: SwitchOpts) -> Result<()> {
 
     sysroot.update_mtime()?;
 
+    if opts.soft_reboot.is_some() {
+        // At this point we have staged the deployment and the host definition has changed.
+        // We need the updated host status before we check if we can prepare the soft-reboot.
+        let updated_host = crate::status::get_status(sysroot, Some(&booted_deployment))?.1;
+        handle_staged_soft_reboot(sysroot, opts.soft_reboot, &updated_host)?;
+    }
+
     if opts.apply {
         crate::reboot::reboot()?;
     }
@@ -968,6 +1111,18 @@ async fn switch(opts: SwitchOpts) -> Result<()> {
 async fn rollback(opts: RollbackOpts) -> Result<()> {
     let sysroot = &get_storage().await?;
     crate::deploy::rollback(sysroot).await?;
+
+    if opts.soft_reboot.is_some() {
+        // Get status of rollback deployment to check soft-reboot capability
+        let host = crate::status::get_status_require_booted(sysroot)?.2;
+
+        handle_soft_reboot(
+            opts.soft_reboot,
+            host.status.rollback.as_ref(),
+            "rollback",
+            || soft_reboot_rollback(sysroot),
+        )?;
+    }
 
     if opts.apply {
         crate::reboot::reboot()?;
--- a/crates/lib/src/status.rs
+++ b/crates/lib/src/status.rs
@@ -90,6 +90,11 @@ impl From<ImageReference> for OstreeImageReference {
     }
 }
 
+/// Check if a deployment has soft reboot capability
+fn has_soft_reboot_capability(sysroot: &Storage, deployment: &ostree::Deployment) -> bool {
+    ostree_ext::systemd_has_soft_reboot() && sysroot.deployment_can_soft_reboot(deployment)
+}
+
 /// Parse an ostree origin file (a keyfile) and extract the targeted
 /// container image reference.
 fn get_image_origin(origin: &glib::KeyFile) -> Result<Option<OstreeImageReference>> {
@@ -187,11 +192,13 @@ fn boot_entry_from_deployment(
         (CachedImageStatus::default(), false)
     };
 
+    let soft_reboot_capable = has_soft_reboot_capability(sysroot, deployment);
     let store = Some(crate::spec::Store::OstreeContainer);
     let r = BootEntry {
         image,
         cached_update,
         incompatible,
+        soft_reboot_capable,
         store,
         pinned: deployment.is_pinned(),
         ostree: Some(crate::spec::BootEntryOstree {
@@ -425,6 +432,27 @@ fn render_verbose_ostree_info(
     Ok(())
 }
 
+/// Helper function to render if soft-reboot capable
+fn write_soft_reboot(
+    mut out: impl Write,
+    entry: &crate::spec::BootEntry,
+    prefix_len: usize,
+) -> Result<()> {
+    // Show soft-reboot capability
+    write_row_name(&mut out, "Soft-reboot", prefix_len)?;
+    writeln!(
+        out,
+        "{}",
+        if entry.soft_reboot_capable {
+            "yes"
+        } else {
+            "no"
+        }
+    )?;
+
+    Ok(())
+}
+
 /// Write the data for a container image based status.
 fn human_render_slot(
     mut out: impl Write,
@@ -507,6 +535,9 @@ fn human_render_slot(
                 }
             }
         }
+
+        // Show soft-reboot capability
+        write_soft_reboot(&mut out, entry, prefix_len)?;
     }
 
     tracing::debug!("pinned={}", entry.pinned);
@@ -544,6 +575,9 @@ fn human_render_slot_ostree(
         if let Some(ostree) = &entry.ostree {
             render_verbose_ostree_info(&mut out, ostree, slot, prefix_len)?;
         }
+
+        // Show soft-reboot capability
+        write_soft_reboot(&mut out, entry, prefix_len)?;
     }
 
     tracing::debug!("pinned={}", entry.pinned);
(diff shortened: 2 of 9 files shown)


lib/src/reboot.rs
```diff
@@ -6,16 +6,29 @@ use fn_error_context::context;
...
+#[context("Initiating reboot")]
+pub(crate) fn reboot() -> anyhow::Result<()> {
+    perform_reboot("Rebooting system", "reboot")
+}
+
+/// Initiate a soft reboot (userspace-only restart).
+/// This function will only return in case of error.
+#[context("Initiating soft reboot")]
+pub(crate) fn soft_reboot() -> anyhow::Result<()> {
+    perform_reboot("Performing soft-reboot", "systemctl soft-reboot")
```
cgwalters: We shouldn't use commands-as-strings (as would be parsed by a shell etc). I think the cleanest fix here is to change `perform_reboot` to take `enum RebootType { Hard, Soft }` or so, and then it invokes the right command

lib/src/reboot.rs
```diff
@@ -6,16 +6,29 @@ use fn_error_context::context;
...
 /// This function will only return in case of error.
-#[context("Initiating reboot")]
-pub(crate) fn reboot() -> anyhow::Result<()> {
+fn perform_reboot(task_name: &str, command: &str) -> anyhow::Result<()> {
     // Flush output streams
     let _ = std::io::stdout().flush();
     let _ = std::io::stderr().flush();
-    Task::new("Rebooting system", "reboot").run()?;
-    tracing::debug!("Initiated reboot, sleeping forever...");
+    Task::new(task_name, command).run()?;
```
cgwalters: Let's stop using `Task` here actually and just directly `Command::new("systemctl")` or so

lib/src/status.rs
```diff
@@ -415,6 +422,11 @@ fn human_render_slot(
         writeln!(out, "yes")?;
     }
 
+    if entry.soft_reboot_capable {
+        write_row_name(&mut out, "Soft-reboot", prefix_len)?;
```
cgwalters: I think we may want to add a verbose mode and only show this there?

jmarrero: we already have --format yaml or --format json, I am ok just removing this too if it's too much data to show by default since if someone wants details they can use the other formats.

lib/src/reboot.rs
jmarrero: reboot does the correct thing so no need to modify this module anymore.

lib/src/status.rs.orig
```diff
@@ -0,0 +1,637 @@
...
+                  Pinned: yes
+
+             Other image: quay.io/centos-bootc/centos-bootc:stream9
+                  Digest: sha256:47e5ed613a970b6574bfa954ab25bb6e85656552899aa518b5961d9645102b37 (arm64)
+                 Version: stream9.20240807.0
+                  Pinned: yes
+        "};
+        similar_asserts::assert_eq!(w, expected);
+    }
+}
```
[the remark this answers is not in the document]
cgwalters: @jmarrero can you mark comments resolved when you fix them, like this one?

lib/src/cli.rs
```diff
@@ -835,6 +835,31 @@ async fn upgrade(opts: UpgradeOpts) -> Result<()> {
...
+                    .unwrap_or(false);
+
+                if can_soft_reboot {
+                    println!("Staged deployment is soft-reboot capable, performing soft-reboot...");
+
+                    // Find the staged deployment
+                    let deployments_list = sysroot.deployments();
+                    let staged_deployment = deployments_list
+                        .iter()
+                        .find(|d| d.is_staged())
```
cgwalters: I think this is what we need to do at the moment, the code is fine. 

However just looking at things overall, we should probably try to refactor in the future so that we have a clear staged state that has both the status object and the deployment and we shouldn't then need the `unwrap_or` and error checking here.

lib/src/cli.rs
```diff
@@ -931,6 +956,33 @@ async fn switch(opts: SwitchOpts) -> Result<()> {
...
+        let (_updated_deployments, updated_host) =
+            crate::status::get_status(sysroot, Some(&booted_deployment))?;
+        let can_soft_reboot = updated_host
+            .status
+            .staged
+            .as_ref()
+            .map(|s| s.soft_reboot_capable)
+            .unwrap_or(false);
+
+        if can_soft_reboot {
```
cgwalters: As gemini suggested I think a shared helper function for this would make sense (that also invokes `reboot`)

lib/src/reboot.rs
```diff
@@ -13,7 +13,9 @@ pub(crate) fn reboot() -> anyhow::Result<()> {
     // Flush output streams
     let _ = std::io::stdout().flush();
     let _ = std::io::stderr().flush();
-    Task::new("Rebooting system", "reboot").run()?;
+    Task::new("Rebooting system", "systemd-run")
```
cgwalters: Could do this as a separate commit or even PR just to clarify why we need this bit

jmarrero: https://github.com/bootc-dev/bootc/pull/1398

lib/src/cli.rs
```diff
@@ -931,7 +959,36 @@ async fn switch(opts: SwitchOpts) -> Result<()> {
...
+            let status = cmd.status().context("Failed to run ostree admin prepare-soft-reboot")?;
+            if !status.success() {
+                anyhow::bail!("ostree admin prepare-soft-reboot failed");
+            }
+            
+                         // Perform the soft-reboot
+             crate::reboot::soft_reboot()?;
+        } else {
+            crate::reboot::reboot()?;
+        }
```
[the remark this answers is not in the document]
jmarrero: this is resolved in the latest push

lib/src/cli.rs
```diff
@@ -931,7 +959,36 @@ async fn switch(opts: SwitchOpts) -> Result<()> {
...
+            // Prepare the soft-reboot using ostree admin prepare-soft-reboot
+            let mut cmd = std::process::Command::new("ostree");
+            cmd.args(["admin", "prepare-soft-reboot", &staged_index.to_string()]);
+            let status = cmd.status().context("Failed to run ostree admin prepare-soft-reboot")?;
+            if !status.success() {
+                anyhow::bail!("ostree admin prepare-soft-reboot failed");
+            }
+            
+                         // Perform the soft-reboot
+             crate::reboot::soft_reboot()?;
```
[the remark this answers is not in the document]
jmarrero: ran cargo fmt on latest push

lib/src/status.rs
jmarrero: https://github.com/bootc-dev/bootc/pull/1399 for verbose

lib/src/cli.rs
jmarrero: done in the latest push

lib/src/cli.rs
```diff
@@ -941,10 +999,22 @@ async fn switch(opts: SwitchOpts) -> Result<()> {
...
+        // Get status before rollback to check soft-reboot capability
+        let (_booted_deployment, _deployments, host) =
+            crate::status::get_status_require_booted(sysroot)?;
+        let can_soft_reboot = can_perform_soft_reboot(host.status.rollback.as_ref());
+
+        // Perform the rollback
+        crate::deploy::rollback(sysroot).await?;
+
+        if can_soft_reboot {
+            soft_reboot_rollback(sysroot)?;
```
cgwalters: Since this function is only called once so it could be inlined, but OK as is

lib/src/status.rs
```diff
@@ -444,6 +456,11 @@ fn human_render_slot_ostree(
...
+        writeln!(out, "capable")?;
+    }
```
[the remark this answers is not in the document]
jmarrero: removed for now will be added with --verbose on https://github.com/bootc-dev/bootc/pull/1399

cgwalters:
Moving to draft since this is blocked until we either:

- ship an updated ostree
- switch to build/runtime conditionals

cgwalters:
That said one thing that would be cool would be us keeping cached container images for PRs on `ghcr.io/bootc-dev/bootc-pr-<PRNUMBER>` or so.

In this PR (actually the dependency one) we could explicitly bump the ostree version even before it's shipped $everywhere.

cgwalters:
This will require https://github.com/ostreedev/ostree/pull/3487

jmarrero:
On.. 
[ ] debug why centos stream 9 still has old ostree 

it's https://gitlab.com/redhat/centos-stream/containers/bootc/-/merge_requests/641#b1795a50b9b140a9b6dcd2d5c10d39bc56d39805 that has not merged because ppc build does not pass with:
https://konflux-ui.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/ns/centos-bootc-tenant/applications/centos-bootc-c9s/pipelineruns/centos-bootc-c9s-on-pull-request-m26gv/logs?task=build-container-ppc64le

```
>>>> Executing `postprocess` inline script '9' <<<<

+ mkdir -p /usr/lib/ostree
+ cat

>>>> Executing `postprocess` inline script '10' <<<<

+ /usr/bin/bootupctl backend generate-update-metadata

>>>> Executing `postprocess` inline script '11' <<<<

+ rm -vrf /usr/lib/ostree-boot/loader
rm: write error: Bad file descriptor
error: Executing compose install: Subprocess failed: ExitStatus(unix_wait_status(256))
rpm-ostree version: 2025.9
warning: Signature not supported. Hash algorithm SHA1 not available.

(/proc/self/exe compose install:733): libdnf-WARNING **: 05:30:13.060: failed to parse public key for /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-SIG-Extras
error: Installing packages: Executing inline postprocessing script: bwrap(/usr/bin/rpmostree-postprocess-inline-11): Child process exited with code 1

Error executing command: Command '['rpm-ostree', 'compose', 'rootfs', '--source-root=/', '/usr/share/doc/bootc-base-imagectl/manifests/standard.yaml', '/target-rootfs']' returned non-zero exit status 1.
subprocess exited with status 1
subprocess exited with status 1
Error: building at STEP "RUN --mount=type=cache,target=/workdir . /cachi2/cachi2.env &&     /src/build.sh": exit status 1
step-prepare-sboms :-
```

which I think is the issue https://github.com/coreos/rpm-ostree/issues/5436 @cgwalters was debugging this week right? Should I just force merge that MR since the new rpm-ostree is on the way?

cgwalters:
Sorry yes I edited that TODO item to link to the issue.

> Should I just force merge that MR since the new rpm-ostree is on the way?

Which MR?

jmarrero:
> Sorry yes I edited that TODO item to link to the issue.
> 
> > Should I just force merge that MR since the new rpm-ostree is on the way?
> 
> Which MR?

I was talking about https://gitlab.com/redhat/centos-stream/containers/bootc/-/merge_requests/641which is a bad idea either way since it will break ppc until we get rpm-ostree.

crates/lib/src/cli.rs
```diff
@@ -734,9 +736,21 @@ pub(crate) fn require_root(is_container: bool) -> Result<()> {
 }
 
 /// Check if a deployment can perform a soft reboot
+#[allow(dead_code)]
 fn can_perform_soft_reboot(deployment: Option<&crate::spec::BootEntry>) -> bool {
     deployment.map(|d| d.soft_reboot_capable).unwrap_or(false)
 }
+/// If there is staged deployment, check if soft reboot is possible
+fn should_soft_reboot() -> bool {
```
jmarrero: This method would figureout if there is anything staged, if so will try to prepare it for soft-reboot, now the more complicated part I think is checking if we are rebooting to a rollback deployment.

crates/lib/src/cli.rs
```diff
@@ -733,6 +735,59 @@ pub(crate) fn require_root(is_container: bool) -> Result<()> {
...
+/// Prepare and execute a soft reboot for the given deployment
+#[context("Preparing soft reboot")]
+fn prepare_soft_reboot(
+    sysroot: &crate::store::Storage,
+    deployment: &ostree::Deployment,
+) -> Result<()> {
+    let cancellable = ostree::gio::Cancellable::NONE;
+    sysroot
+        .sysroot
+        .deployment_set_soft_reboot(deployment, false, cancellable)
```
bshephar, line 788: I know this is in draft, so my comment might be premature. But I think I'm missing where this executes the soft reboot. I see the `deployment_set_soft_reboot()` sets the ostree deployment:
https://github.com/ostreedev/ostree/blob/v2025.4/rust-bindings/src/auto/sysroot.rs#L178

Does this just setup the ostree deployment to be the next reboot target and all things required for the soft reboot, but not actually perform the soft reboot yet? Or am I missing something obvious? Probably the latter.

jmarrero: Yeah, the idea here is to have bootc prepare the soft-reboot and depend on systemd to decide to do the soft-reboot if the system is prepared for it.

So if a user just calls reboot they will get a soft-reboot if we can do it and the bootc-soft-reboot service is active.

crates/lib/src/cli.rs
```diff
@@ -733,6 +735,59 @@ pub(crate) fn require_root(is_container: bool) -> Result<()> {
...
+    if can_perform_soft_reboot(updated_host.status.staged.as_ref()) {
+        soft_reboot_staged(sysroot)?;
+        return Ok(true);
+    }
+    // TODO check if this reboot is doing a rollback...
+    // TODO then perform a soft reboot for the rollback deployment
+    Ok(false)
+}
+
+/// Prepare and execute a soft reboot for the given deployment
```
bshephar: Maybe we just want to update this comment then to remove the "execute" reference, since this is all just preparing the system to be able to perform a soft reboot. But the actual execution of said reboot is handled externally to this.

jmarrero: done

jmarrero:
If I call ```bootc internals prepare-soft-reboot && reboot``` it does what we expect. However if I call `reboot` the service runs but it still goes thru a full reboot. I wonder if by calling reboot that is too late to prepare /run/nextroot. Maybe I should change the service to monitors /boot and /ostree for rollback/new staged deployments and then run when notice changes on directory?

cgwalters:
> I wonder if by calling reboot that is too late to prepare /run/nextroot. 

Oof yes from a quick skim of the systemd source I think that's the case, see
https://github.com/systemd/systemd/blob/bebaaddd41e1aa86662032d9b1e8969e41faf195/src/systemctl/systemctl-logind.c#L99

Basically `systemctl reboot` looks for `/run/nextroot` before switching target at all...

cgwalters:
We'll hopefully sort out the c9s situation in the next few days, we can ignore that test failure for now

Dockerfile
```diff
@@ -10,7 +10,7 @@ COPY . /src
 
 FROM $base as base
 # Set this to anything non-0 to enable https://copr.fedorainfracloud.org/coprs/g/CoreOS/continuous/
-ARG continuous_repo=0
+ARG continuous_repo=1
```
cgwalters: We can drop this now I believe

crates/lib/src/status.rs
```diff
@@ -463,6 +467,18 @@ fn human_render_slot(
                 }
             }
         }
+
+        // Show soft-reboot capability
+        write_row_name(&mut out, "Soft-reboot", prefix_len)?;
+        writeln!(
+            out,
+            "{}",
+            if entry.soft_reboot_capable {
```
cgwalters: Minor but how about a `bool_yesno(b: bool) -> &str` helper? (Or a `write_soft_reboot` helper?)

crates/lib/src/cli.rs
```diff
@@ -961,6 +1032,22 @@ async fn rollback(opts: RollbackOpts) -> Result<()> {
     let sysroot = &get_storage().await?;
     crate::deploy::rollback(sysroot).await?;
 
+    if opts.queue_soft_reboot {
+        // Get status before rollback to check soft-reboot capability
+        let host = crate::status::get_status_require_booted(sysroot)?.2;
```
cgwalters, line 1117: Right so this is a bit tricky; the `OstreeSysroot` object has "interior mutability" even though we only have a `&` and not a `&mut` and so when we call APIs like `write_deployments` it ends up changing some of its caches and reloads them.

But the `Host` is a bootc side thing in Rust and so follows Rust defaults and we only have a `&` (it also doesn't monitor the `ostree::Sysroot`) so we do need to re-fetch it after making changes.

A followup cleanup here is probably for the APIs that do mutations to return a new `Host by default.

But anyways this looks sane to me as is

crates/lib/src/status.rs
```diff
@@ -144,10 +144,14 @@ fn boot_entry_from_deployment(
         (None, CachedImageStatus::default(), false)
     };
 
+    let soft_reboot_capable =
+        ostree_ext::systemd_has_soft_reboot() && sysroot.deployment_can_soft_reboot(deployment);
```
cgwalters: We may want a shared helper for this?

crates/lib/src/status.rs
```diff
@@ -228,17 +232,17 @@ pub(crate) fn get_status(
         other,
     };
 
-    let staged = deployments
-        .staged
+    let booted = booted_deployment
```
cgwalters: Why the swap of booted vs staged?

crates/lib/src/cli.rs
cgwalters: Actually the inverse of this is important: We need to audit we're not using a stale `Host` object after a mutation in the upgrade/switch cases

crates/lib/src/status.rs
jmarrero: probably a product of using Claude a few times here 🤖. It makes 0 sense and I have no memory of making this manually. Ill make sure to comb over the changes once more before moving this out of draft.

crates/lib/src/cli.rs
jeckersb: I'm walking through this with claude's help, and it seems to be getting a bit confused here, and then looking into this myself I also got a bit confused.  So let me ask for clarification here.

1.  We do the actual rollback (`crate::deploy::rollback(sysroot).await?;`)
1.  Then we reload the host status from the current (post-rollback) state
1.  We check `host.status.rollback` to see if it's soft-reboot-capable
1.  Then we take the first deployment in the list and prepare soft-reboot

The part that claude initially got hung up on is the fact that the first deployment in the list may not necessarily be the rollback deployment.  But upon asking it to expand on it's thinking, it concedes that although it's not really a "type-safe" relationship, after the rollback operation the rollback deployment is going to be at the front of the list:

```
  During rollback, the deployment ordering changes:

  1. Before rollback: [booted_deployment, rollback_deployment, ...other]
  2. After rollback: [rollback_deployment, booted_deployment, ...other]

  The rollback function at deploy.rs:784 explicitly reorders deployments:
  let new_deployments = if reverting {
      [booted_deployment, rollback_deployment]  // reverting: keep current order
  } else {
      [rollback_deployment, booted_deployment]  // rollback: swap order
  };

  After crate::deploy::rollback(sysroot).await? completes, the deployment list has been rewritten, so:
  - deployments_list.first() now points to the rollback deployment (correct)
```

Ok, that seems sane.

However, *after* we do the rollback, is the rolled-back-to deployment still `host.status.rollback`?  Or at this point is the rolled-back-to deployment now `host.status.staged` and the *current* booted deployment is now `host.status.rollback`?

jmarrero: Rollback deployments don't do staging, it just switches the bootloader order. host.status.rollback will not change until you reboot. I guess here the comment is a bit misleading. With the `after` word. It maybe should just say 

```
// Get status of rollback deployment to check soft-reboot capability
```
because in this function status is not called before that.

crates/lib/src/status.rs
jmarrero: done

crates/lib/src/status.rs
jmarrero: done

tmt/tests/booted/test-soft-reboot.nu
```diff
@@ -0,0 +1,89 @@
...
+" | save Dockerfile
+    # Build it
+    podman build -t localhost/bootc-derived .
+
+    bootc switch --queue-soft-reboot --transport containers-storage localhost/bootc-derived
+    let st = bootc status --json | from json
+    assert $st.status.staged.softRebootCapable
+
+    # And reboot into it
+    tmt-reboot
```
jmarrero, line 42: I bet this does not support soft-reboots.... because is not coming back. `TMT_REBOOT_COUNT` probably is not increased.

crates/lib/src/cli.rs
cgwalters: Yes, this is confusing. Basically ostree maintains an ordered list of deployments at the lowest level. The concept of "rollback" is a more invented/higher level one determined by looking at that ordering. (Also it takes stateroots into account)

tmt/tests/booted/test-soft-reboot.nu
cgwalters: Hmm...maybe. tmt is so huge, I have trouble digging around in the code but I did find https://github.com/teemtee/tmt/blob/42bf194a5a198b9213df8ed46aa1e657ad32e6ed/tmt/steps/execute/__init__.py#L416

Probably worth trying to minimize a tmt test around this

tmt/tests/booted/test-soft-reboot.nu
```diff
@@ -34,25 +32,16 @@ RUN echo test content > /usr/share/testfile-for-soft-reboot.txt
     let st = bootc status --json | from json
     assert $st.status.staged.softRebootCapable
 
-    # And reboot into it
-    tmt-reboot
+    # And reboot into it using systemctl to ensure soft-reboot get's called
+    # https://tmt.readthedocs.io/en/stable/stories/features.html#reboot-during-test
+    tmt-reboot -c "systemctl soft-reboot"
```
jmarrero: Still fighting this.

tmt/tests/booted/test-soft-reboot.nu
jmarrero: @cgwalters I keep getting different results locally vs on CI with tmt. I updated the tests to do some "hacking" around the constraints but still tests some core functionality for now. WDYT if we merge and do a follow up Kola tests to no block on this? (If the only thing blocking atm is the test)

jmarrero:
man... this is so sad:
```
 Problem: cannot install both ostree-libs-2025.4-1.el9.x86_64 from appstream and ostree-libs-2025.4-2.el9.x86_64 from @System
  - package ostree-devel-2025.4-1.el9.x86_64 from crb requires ostree-libs(x86-64) = 2025.4-1.el9, but none of the providers can be installed
  - package ostree-2025.4-2.el9.x86_64 from @System requires ostree-libs(x86-64) = 2025.4-2.el9, but none of the providers can be installed
  - cannot install the best candidate for the job
```
But at least https://gitlab.com/redhat/centos-stream/containers/bootc/-/merge_requests/685/diffs is merged and should resolve this hopefully today.

jmarrero:
https://asciinema.org/a/dKMWpMuUF7MmT5t2siErefJaw

crates/lib/src/cli.rs
```diff
@@ -859,7 +914,11 @@ async fn upgrade(opts: UpgradeOpts) -> Result<()> {
             .unwrap_or_default();
         if staged_unchanged {
             println!("Staged update present, not changed.");
-
+            if opts.queue_soft_reboot {
+                if has_soft_reboot_capability(host.status.staged.as_ref()) {
+                    soft_reboot_staged(sysroot)?;
+                }
```
cgwalters: The silent fallthrough here makes me I wonder if we should actually have `--soft-reboot=required|auto` or so where `required` fails if the reboot wouldn't be soft? Implementing this would change from a `bool` to an `Option<enum SoftRebootRequest { Auto, Required }>`

That'd allow higher level tooling which might operate under a "minimize downtime" constraint to impose that requirement.

jmarrero: Latest push attempts this, It seems to work I need still to comb over all the changes to make sure I am not missing something but should good for today's demo.
https://asciinema.org/a/PvBJXqlNiqRdf3suYLukLYsSu

crates/lib/src/cli.rs
```diff
@@ -80,6 +80,13 @@ pub(crate) struct UpgradeOpts {
     #[clap(long, conflicts_with = "check")]
     pub(crate) apply: bool,
 
+    /// Configure soft reboot behavior.
+    ///
+    /// 'required' will fail if soft reboot is not available.
+    /// 'auto' will use soft reboot if available, otherwise fall back to regular reboot.
+    #[clap(long = "soft-reboot", conflicts_with = "check")]
+    pub(crate) soft_reboot: Option<SoftRebootMode>,
```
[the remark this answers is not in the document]
jmarrero: This is the same pattern we use for apply, so feels like would need more refactoring for this. Also it feels to me a like a readability vs code duplication thing which I prefer readability if it's not a crazy big function.

crates/lib/src/cli.rs
```diff
@@ -741,6 +771,104 @@ pub(crate) fn require_root(is_container: bool) -> Result<()> {
...
+                    deployment_type
+                );
+            }
+        }
+        SoftRebootMode::Auto => {
+            if can_soft_reboot {
+                execute_soft_reboot()?;
+            }
+        }
+    }
```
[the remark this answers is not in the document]
jmarrero: The other code allows us to expand what happens on Auto too in the future. Also, I think is easier to follow or add new options. But this seems preference too.

jmarrero:
https://asciinema.org/a/PvBJXqlNiqRdf3suYLukLYsSu

PR #1394, opened 2025-07-01
Rework utils and blockdev crates as -internals in prep for publishing

cgwalters:
See https://github.com/coreos/bootupd/pull/963

Basically I just want a way to keep these in sync with bootupd; we're not attempting to make a truly public library.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -177,8 +177,8 @@ version = "0.0.0"
 dependencies = [
  "anstream",
  "anyhow",
+ "bootc-internal-utils",
  "bootc-lib",
- "bootc-utils",
  "log",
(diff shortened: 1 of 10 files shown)


jmarrero (approved):
lgtm

HuijingHei:
The error is not related to this PR, it is infrastructure issue
```
Errors during downloading metadata for repository 'baseos':
  - Downloading successful, but checksum doesn't match. Calculated:
```

cgwalters:
For the record I published these manually just now via a direct `cargo publish`. I'm not totally sure in the medium term it will make sense to have these as crates but for now this unblocks https://github.com/coreos/bootupd/issues/964

PR #1397, opened 2025-07-08
chunking: Support exclusive chunks defined via xattrs

ckyrouac:
This adds a new "specific_components" parameter to process_mapping to enable mapping individual files to specific layers. This will be used by the rechunker in rpm-ostree.

Related to: https://github.com/coreos/rpm-ostree/pull/5431

### DIFF
--- a/ostree-ext/src/chunking.rs
+++ b/ostree-ext/src/chunking.rs
@@ -49,7 +49,7 @@ pub(crate) struct Chunk {
     pub(crate) packages: Vec<String>,
 }
 
-#[derive(Debug, Deserialize, Serialize)]
+#[derive(Debug, Clone, Deserialize, Serialize)]
 /// Object metadata, but with additional size data
 pub struct ObjectSourceMetaSized {
     /// The original metadata
@@ -276,9 +276,10 @@ impl Chunking {
         meta: &ObjectMetaSized,
         max_layers: &Option<NonZeroU32>,
         prior_build_metadata: Option<&oci_spec::image::ImageManifest>,
+        specific_contentmeta: Option<&ObjectMetaSized>,
     ) -> Result<Self> {
         let mut r = Self::new(repo, rev)?;
-        r.process_mapping(meta, max_layers, prior_build_metadata)?;
+        r.process_mapping(meta, max_layers, prior_build_metadata, specific_contentmeta)?;
         Ok(r)
     }
 
@@ -294,6 +295,7 @@ impl Chunking {
         meta: &ObjectMetaSized,
         max_layers: &Option<NonZeroU32>,
         prior_build_metadata: Option<&oci_spec::image::ImageManifest>,
+        specific_contentmeta: Option<&ObjectMetaSized>,
     ) -> Result<()> {
         self.max = max_layers
             .unwrap_or(NonZeroU32::new(MAX_CHUNKS).unwrap())
@@ -314,6 +316,25 @@ impl Chunking {
             rmap.entry(Rc::clone(contentid)).or_default().push(checksum);
         }
 
+        // Create exclusive chunks first if specified
+        let mut processed_specific_components = BTreeSet::new();
+        if let Some(specific_meta) = specific_contentmeta {
+            for component in &specific_meta.sizes {
+                let mut chunk = Chunk::new(&component.meta.name);
+                chunk.packages = vec![component.meta.name.to_string()];
+
+                // Move all objects belonging to this exclusive component
+                if let Some(objects) = rmap.get(&component.meta.identifier) {
+                    for &obj in objects {
+                        self.remainder.move_obj(&mut chunk, obj);
+                    }
+                }
+
+                self.chunks.push(chunk);
+                processed_specific_components.insert(&*component.meta.identifier);
+            }
+        }
+
         // Safety: Let's assume no one has over 4 billion components.
         self.n_provided_components = meta.sizes.len().try_into().unwrap();
         self.n_sized_components = sizes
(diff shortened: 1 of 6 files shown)


ostree-ext/src/chunking.rs
```diff
@@ -1003,4 +1054,310 @@ mod test {
         assert_eq!(structure_derived, v2_expected_structure);
         Ok(())
     }
+
+    #[test]
+    fn test_exclusive_chunks() -> Result<()> {
+        // Test that exclusive chunks are created first and get their own layers
+        let contentmeta: Vec<ObjectSourceMetaSized> = [
+            vec![1, 100, 50000],
```
cgwalters: I think it'd be cleaner to make these tuples instead of vectors.

ostree-ext/src/chunking.rs
```diff
@@ -1003,4 +1054,310 @@ mod test {
...
+        .iter()
+        .map(|data| ObjectSourceMetaSized {
+            meta: ObjectSourceMeta {
+                identifier: RcStr::from(format!("pkg{}.0", data[0])),
+                name: RcStr::from(format!("pkg{}", data[0])),
+                srcid: RcStr::from(format!("srcpkg{}", data[0])),
+                change_time_offset: 0,
+                change_frequency: data[1],
+            },
+            size: data[2] as u64,
```
cgwalters: And we can make type inference happy by writing e.g. `40000u64` on the first element instead.

ostree-ext/src/chunking.rs
```diff
@@ -314,6 +316,32 @@ impl Chunking {
             rmap.entry(Rc::clone(contentid)).or_default().push(checksum);
         }
 
+        // Create exclusive chunks first if specified
+        let mut processed_specific_components = BTreeSet::new();
+        if let Some(specific_meta) = specific_contentmeta {
+            for component in &specific_meta.sizes {
+                if self.remaining() == 0 {
```
cgwalters: I wonder if we should have `--max-layers` only apply to automatically generated layers? I think that'd be basically just deleting this check.

ostree-ext/src/chunking.rs
```diff
@@ -323,49 +351,72 @@ impl Chunking {
...
+                    srcid: Rc::clone(&component.meta.srcid),
+                    change_time_offset: component.meta.change_time_offset,
+                    change_frequency: component.meta.change_frequency,
+                },
+                size: component.size,
+            })
+            .collect();
+
+        // Process regular components with bin packing if we have remaining layers
+        if !regular_sizes.is_empty() && self.remaining() > 0 {
```
cgwalters: Why the explicit check for `if !regular_sizes.is_empty()` - if I understand correctly what would happen in the case where *all* content is explicitly specified is that we'd still end up with the "resreved for new packages" one, and that should be it?

I think we could special case that one. 

(The "reserved for new packages" thing I think never worked out...but that's a different issue)

ostree-ext/src/chunking.rs
```diff
@@ -1003,4 +1054,304 @@ mod test {
         assert_eq!(structure_derived, v2_expected_structure);
         Ok(())
     }
+
+    #[test]
+    fn test_exclusive_chunks() -> Result<()> {
```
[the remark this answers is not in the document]
cgwalters: Yes please

ostree-ext/src/chunking.rs
```diff
@@ -276,9 +276,10 @@ impl Chunking {
         meta: &ObjectMetaSized,
         max_layers: &Option<NonZeroU32>,
         prior_build_metadata: Option<&oci_spec::image::ImageManifest>,
+        specific_contentmeta: Option<&ObjectMetaSized>,
```
cgwalters, line 279: Would it be cleaner if we made `Chunk` public, and then this could be `chunks: Vec<Chunk>` or so where the caller has already done part of the work?

ostree-ext/src/chunking.rs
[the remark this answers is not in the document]
ckyrouac: I was hoping gemini would automatically fix this for me, or at least propose the specific changes.

ostree-ext/src/chunking.rs
ckyrouac: I started refactoring this but it felt off because it would fragment the code that builds the chunks vec.

ostree-ext/src/chunking.rs
```diff
@@ -323,49 +351,72 @@ impl Chunking {
...
+                meta: ObjectSourceMeta {
+                    identifier: Rc::clone(&component.meta.identifier),
+                    name: Rc::clone(&component.meta.name),
+                    srcid: Rc::clone(&component.meta.srcid),
+                    change_time_offset: component.meta.change_time_offset,
+                    change_frequency: component.meta.change_frequency,
+                },
+                size: component.size,
+            })
+            .collect();
```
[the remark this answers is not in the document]
cgwalters: You marked this as resolved but it didn't change - did `.clone()` not work here?

ostree-ext/src/chunking.rs
```diff
@@ -323,49 +344,72 @@ impl Chunking {
...
+
+        // Process regular components with bin packing if we have remaining layers
+        if self.remaining() > 0 {
```
cgwalters: This would be cleaner like

```
if let Some(remaining) = NonZeroU32::new(self.remaining()) {
  ...
```

Then below...

ostree-ext/src/chunking.rs
```diff
@@ -323,49 +344,72 @@ impl Chunking {
...
+            let start = Instant::now();
+            let packing = basic_packing(
+                &regular_sizes,
+                NonZeroU32::new(self.remaining()).unwrap(),
```
cgwalters: Just pass `remaining` here

cgwalters (approved):
This is all very complex code (mostly my fault!) and I find it hard to get it all in my head, but overall this looks reasonable to me.

I think what would really help is seeing this more "end to end" in combination with https://github.com/coreos/rpm-ostree/pull/5431 - especially via a Dockerfile build that sets xattrs and verifies the chunking.

Maybe we could try standing up a CI job that clones your rpm-ostree PR and runs a test job from there? But optional from my PoV right now.

ckyrouac:
> This is all very complex code (mostly my fault!) and I find it hard to get it all in my head

Yep, took me awhile to wrap my head around. Might take a stab at refactoring it a bit after this gets in.

> I think what would really help is seeing this more "end to end" in combination with [coreos/rpm-ostree#5431](https://github.com/coreos/rpm-ostree/pull/5431) - especially via a Dockerfile build that sets xattrs and verifies the chunking.
> 
> Maybe we could try standing up a CI job that clones your rpm-ostree PR and runs a test job from there? But optional from my PoV right now.

I've been testing locally using this Containerfile (thanks claude) and using my local checkout of bootc in the rpm-ostree Cargo.toml. I can take a stab at doing this in our CI.

```
FROM quay.io/fedora/fedora-bootc:42

# Install attr package to set extended attributes
RUN dnf install -y attr

# Create some test files with different content
RUN echo "Database component file 1" > /usr/local/db-config.conf
RUN echo "Database component file 2" > /usr/local/db-schema.sql
RUN echo "Web server component file 1" > /usr/local/nginx.conf
RUN echo "Web server component file 2" > /usr/local/web-static.html
RUN echo "Monitoring component file" > /usr/local/metrics.json
RUN echo "Cache component file 1" > /usr/local/redis.conf
RUN echo "Cache component file 2" > /usr/local/cache-policy.json

# Set user.component extended attributes for different components (using user namespace since system may not be supported)
RUN setfattr -n user.component -v "database" /usr/local/db-config.conf
RUN setfattr -n user.component -v "database" /usr/local/db-schema.sql
RUN setfattr -n user.component -v "webserver" /usr/local/nginx.conf
RUN setfattr -n user.component -v "webserver" /usr/local/web-static.html
RUN setfattr -n user.component -v "monitoring" /usr/local/metrics.json
RUN setfattr -n user.component -v "cache" /usr/local/redis.conf
RUN setfattr -n user.component -v "cache" /usr/local/cache-policy.json

# Verify the extended attributes are set correctly
RUN echo "=== Verifying extended attributes ===" && \
    getfattr -n user.component /usr/local/db-config.conf && \
    getfattr -n user.component /usr/local/db-schema.sql && \
    getfattr -n user.component /usr/local/nginx.conf && \
    getfattr -n user.component /usr/local/web-static.html && \
    getfattr -n user.component /usr/local/metrics.json && \
    getfattr -n user.component /usr/local/redis.conf && \
    getfattr -n user.component /usr/local/cache-policy.json

# Also create some files without component xattrs for comparison
RUN echo "No component file 1" > /usr/local/no-component-1.txt
RUN echo "No component file 2" > /usr/local/no-component-2.txt

# Display file listing with extended attributes
RUN echo "=== Final file listing with xattrs ===" && \
    ls -la /usr/local/ && \
    echo "=== Extended attributes summary ===" && \
    find /usr/local/ -name "*.conf" -o -name "*.sql" -o -name "*.html" -o -name "*.json" | \
    xargs -I {} sh -c 'echo "File: {}"; getfattr -n user.component {} 2>/dev/null || echo "  No user.component xattr"'
```

ostree-ext/src/chunking.rs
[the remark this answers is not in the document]
ckyrouac: whoops - updated

cgwalters:
> I can take a stab at doing this in our CI.

I think it could also live in the rpm-ostree CI and then we just run that job from here, but ultimately it's also fine if it just lives in rpm-ostree I think. Hmm, *maybe* what could actually make sense is to move the chunking code there instead since there's 0% chance anything else uses it.

Issue #799, opened 2024-09-19
drop loopback out-of-process


cgwalters:
This one may be a good first issue: today when doing `install --via-loopback`, if we get killed by e.g. `Ctrl-C` (i.e. `SIGINT`) we will leak the loopback device allocation.

I've been thinking here what would be nice in general is to create a little "out of process drop" helper that could handle these external resources; it would:

- mask most signals that would otherwise be fatal
- own the resource and if the parent process died (via `PR_SET_PDEATHSIG`) we'd take care of dropping them

This type of flow would also be nice for temporary directories, which are also easy to leak in this way.

---

Hmm, we may also be able to stop forking `/bin/losetup` and instead use the ioctl APIs and only hold a file descriptor, but that's a bigger change.

dchen5022:
Hi, I'm interested in tackling this. I'm not sure if my understanding is correct, but it seems like we need to:

1. Fork a child process
2. Mask signals in the child process
3. Have the child process create the loopback device and have the parent process wait until it is opened
4. Set up a signal handler to drop the loopback device if the parent process is killed.

Am I missing anything here?

cgwalters:
👋 Thanks for your interest in this! One thing I'd note here is that this problem domain generalizes into a "remote process object holder + drop handler" that is independent of bootc really. Another good example of something to also do this way would be `tempfile::tempdir()`.

I did some searching on crates.io but didn't find something exactly like this, though there are various related crates like https://lib.rs/crates/defer-drop and https://lib.rs/crates/backdrop
Basically we need a process and not a thread so we handle signals (and crashing).

I think instead of a plain fork let's do an actual fork+exec which is a lot safer and easier to manage. The downside is that data passing is a bit more complex.

dchen5022:
Thanks so much for the clarification! So just to make sure, this would have to be a separate binary? And if we're fork+execing could this just be done with std::process::spawn then?

cgwalters:
We don't need a separate binary, but can re-exec our own binary with a new entrypoint. See the handling of `bootc internals` - I did something similar in [this commit](https://github.com/containers/bootc/pull/915/commits/c5852ad405a953ce54d5b777b9e1160df21919a2) especially [this section](https://github.com/containers/bootc/pull/915/commits/c5852ad405a953ce54d5b777b9e1160df21919a2#diff-1bb9c5f23547105e52b5e81bb676cd9c8b75c4d85fc5baa77eec39cd2869d4a2R148).

[...]

PR #1402, opened 2025-07-10
blockdev: Fix loopback device resource leak on signal interruption

gursewak1997:
This commit implements issue #799 by creating a signal-safe cleanup helper for loopback devices to prevent resource leaks when bootc install --via-loopback is interrupted by signals like SIGINT (Ctrl-C).

The solution uses an 'out-of-process drop' helper that:
- Forks a cleanup helper process when creating a loopback device
- Uses PR_SET_PDEATHSIG to detect when the parent process dies
- Masks most signals to avoid being killed accidentally
- Automatically cleans up leaked loopback devices if the parent dies
- Gracefully terminates when the parent performs normal cleanup

This prevents the common issue where interrupting bootc install --via-loopback with Ctrl-C would leave /dev/loopN devices allocated on the system.

Fixes: https://github.com/bootc-dev/bootc/issues/799

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -179,9 +179,13 @@ dependencies = [
  "camino",
  "fn-error-context",
  "indoc",
+ "libc",
  "regex",
+ "rustix 1.0.3",
  "serde",
  "serde_json",
+ "tempfile",
+ "tokio",
  "tracing",
 ]
 
--- a/crates/blockdev/Cargo.toml
+++ b/crates/blockdev/Cargo.toml
@@ -11,9 +11,13 @@ anyhow = { workspace = true }
 bootc-utils = { package = "bootc-internal-utils", path = "../utils", version = "0.0.0" }
 camino = { workspace = true, features = ["serde1"] }
 fn-error-context = { workspace = true }
+libc = { workspace = true }
 regex = "1.10.4"
+rustix = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
+tempfile = { workspace = true }
+tokio = { workspace = true, features = ["signal"] }
 tracing = { workspace = true }
 
 [dev-dependencies]
--- a/crates/blockdev/src/blockdev.rs
+++ b/crates/blockdev/src/blockdev.rs
@@ -181,6 +181,14 @@ pub fn partitions_of(dev: &Utf8Path) -> Result<PartitionTable> {
 
 pub struct LoopbackDevice {
     pub dev: Option<Utf8PathBuf>,
+    // Handle to the cleanup helper process
+    cleanup_handle: Option<LoopbackCleanupHandle>,
+}
+
+/// Handle to manage the cleanup helper process for loopback devices
+struct LoopbackCleanupHandle {
+    /// Child process handle
+    child: std::process::Child,
 }
 
 impl LoopbackDevice {
@@ -208,13 +216,50 @@ impl LoopbackDevice {
             .run_get_string()?;
         let dev = Utf8PathBuf::from(dev.trim());
         tracing::debug!("Allocated loopback {dev}");
-        Ok(Self { dev: Some(dev) })
+
+        // Try to spawn cleanup helper process - if it fails, make it fatal
+        let cleanup_handle = Self::spawn_cleanup_helper(dev.as_str())
+            .context("Failed to spawn loopback cleanup helper")?;
+
+        Ok(Self {
+            dev: Some(dev),
+            cleanup_handle: Some(cleanup_handle),
+        })
     }
 
     // Access the path to the loopback block device.
     pub fn path(&self) -> &Utf8Path {
         // SAFETY: The option cannot be destructured until we are dropped
         self.dev.as_deref().unwrap()
+    }
+
+    /// Spawn a cleanup helper process that will clean up the loopback device
+    /// if the parent process dies unexpectedly
+    fn spawn_cleanup_helper(device_path: &str) -> Result<LoopbackCleanupHandle> {
+        use std::process::{Command, Stdio};
+
+        // Get the path to our own executable
+        let self_exe =
+            std::fs::read_link("/proc/self/exe").context("Failed to read /proc/self/exe")?;
+
+        // Create the helper process
+        let mut cmd = Command::new(self_exe);
+        cmd.args(["loopback-cleanup-helper", "--device", device_path]);
+
+        // Set environment variable to indicate this is a cleanup helper
+        cmd.env("BOOTC_LOOPBACK_CLEANUP_HELPER", "1");
+
+        // Set up stdio to redirect to /dev/null
+        cmd.stdin(Stdio::null());
+        cmd.stdout(Stdio::null());
+        cmd.stderr(Stdio::null());
+
+        // Spawn the process
+        let child = cmd
+            .spawn()
+            .context("Failed to spawn loopback cleanup helper")?;
+
+        Ok(LoopbackCleanupHandle { child })
     }
 
     // Shared backend for our `close` and `drop` implementations.
@@ -224,6 +269,13 @@ impl LoopbackDevice {
             tracing::trace!("loopback device already deallocated");
             return Ok(());
         };
+
+        // Kill the cleanup helper since we're cleaning up normally
+        if let Some(mut cleanup_handle) = self.cleanup_handle.take() {
+            // Send SIGTERM to the child process
+            let _ = cleanup_handle.child.kill();
+        }
+
         Command::new("losetup").args(["-d", dev.as_str()]).run()
     }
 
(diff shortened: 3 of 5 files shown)


cgwalters:
Thanks for working on this! While it will be a bit more awkward can you try doing it this way https://github.com/bootc-dev/bootc/issues/799#issuecomment-2568286821 - that should 100% avoid all the unsafe code.

Basically instead of a raw `fork()` (source of basically all the unsafe in general) we fork+exec our own binary `/proc/self/exe`. Look at e.g. `reexec_with_guardenv`.

blockdev/src/blockdev.rs
```diff
@@ -240,6 +303,123 @@ impl Drop for LoopbackDevice {
...
+pub fn run_loopback_cleanup_helper(device_path: &str, parent_pid: u32) -> Result<()> {
+    use std::os::unix::process::CommandExt;
+    use std::process::Command;
+
+    // Check if we're running as a cleanup helper
+    if std::env::var("BOOTC_LOOPBACK_CLEANUP_HELPER").is_err() {
+        anyhow::bail!("This function should only be called as a cleanup helper");
+    }
+
+    // Close stdin, stdout, stderr and redirect to /dev/null
```
cgwalters: This is better done in the parent process setup above

blockdev/src/blockdev.rs
```diff
@@ -217,13 +239,54 @@ impl LoopbackDevice {
...
+    fn spawn_cleanup_helper(device_path: &str) -> Result<LoopbackCleanupHandle> {
+        use std::os::unix::process::CommandExt;
+        use std::process::Command;
+
+        // Get the path to our own executable
+        let self_exe = std::fs::read_link("/proc/self/exe")
+            .context("Failed to read /proc/self/exe")?;
+
+        // Create the helper process using exec
+        let mut cmd = Command::new(self_exe);
```
cgwalters: Set up std{in,out,err} here via https://doc.rust-lang.org/std/process/struct.Stdio.html#method.null

blockdev/src/blockdev.rs
```diff
@@ -240,6 +303,123 @@ impl Drop for LoopbackDevice {
...
+    let null_fd = std::fs::File::open("/dev/null")?;
+    let null_fd = null_fd.as_raw_fd();
+    unsafe {
+        libc::dup2(null_fd, 0);
+        libc::dup2(null_fd, 1);
+        libc::dup2(null_fd, 2);
+    }
+    
+    // Set up death signal notification - we want to be notified when parent dies
+    unsafe {
```
cgwalters: There's a safe version of this in https://docs.rs/rustix/latest/rustix/process/fn.set_parent_process_death_signal.html

blockdev/src/blockdev.rs
```diff
@@ -240,6 +303,123 @@ impl Drop for LoopbackDevice {
...
+        if libc::prctl(libc::PR_SET_PDEATHSIG, libc::SIGUSR1) != 0 {
```
cgwalters: It seems cleaner to me to use `SIGTERM` and react to that

blockdev/src/blockdev.rs
```diff
@@ -240,6 +303,123 @@ impl Drop for LoopbackDevice {
...
+            std::process::exit(1);
+        }
+    }
+
+    // Mask most signals to avoid being killed accidentally
```
cgwalters: So https://docs.rs/tokio/latest/tokio/signal/index.html is one way to handle this in a safe way (will require making the function `async`)

I think the tokio API will replace about 50 lines of unsafe code with 5 lines of safe code.

blockdev/src/blockdev.rs
```diff
@@ -240,6 +303,123 @@ impl Drop for LoopbackDevice {
...
+    if result > 0 {
+        if siginfo.si_signo == libc::SIGUSR1 {
+            // Parent died unexpectedly, clean up the loopback device
+            let status = std::process::Command::new("losetup")
+                .args(["-d", device_path])
+                .status();
+
+            match status {
+                Ok(exit_status) if exit_status.success() => {
+                    // Write to stderr since we closed stdout
```
cgwalters: Mmm my vote here is probably to not inherit stderr at all; if we write to stderr then we have the possibility to intermix the child process writes with the parent's.

One option is to explicitly log to the systemd journal.

I guess speaking of systemd...a whole possibility I hadn't considered until just now is that we fork off via `systemd-run`. That would have some nice advantages but will be trickier to get right the lifecycle binding, so let's leave that for the future.

(There's this whole giant topic in bootc overall defaulting to running via systemd in some cases, which would similarly have a lot of advantages but be a big nontrivial change)

blockdev/src/blockdev.rs
```diff
@@ -311,82 +491,42 @@ pub fn parse_size_mib(mut s: &str) -> Result<u64> {
...
-            }
-         }
-        "# };
-        let table: SfDiskOutput = serde_json::from_str(&fixture).unwrap();
-        assert_eq!(
-            table.partitiontable.find("/dev/loop0p2").unwrap().size,
-            20961247
-        );
-        Ok(())
+        let data = fs::read_to_string("tests/fixtures/lsblk.json").unwrap();
```
cgwalters: This change looks unrelated? I mean it's probably fine to do but let's break it into a separate commit

cgwalters (changes requested):
Thanks! Looking closer

lib/src/cli.rs
```diff
@@ -459,6 +459,15 @@ pub(crate) enum InternalsOpts {
...
         args: Vec<OsString>,
     },
+    /// Loopback device cleanup helper (internal use only)
+    LoopbackCleanupHelper {
+        /// Device path to clean up
+        #[clap(long)]
+        device: String,
+        /// Parent process ID to monitor
+        #[clap(long)]
+        parent_pid: u32,
```
cgwalters: This isn't necessary, with PDEATHSIG we automatically monitor the parent. It can be dropped.

blockdev/src/blockdev.rs
```diff
@@ -217,13 +237,59 @@ impl LoopbackDevice {
...
+        let self_exe =
+            std::fs::read_link("/proc/self/exe").context("Failed to read /proc/self/exe")?;
+
+        // Create the helper process
+        let mut cmd = Command::new(self_exe);
+        cmd.args([
+            "loopback-cleanup-helper",
+            "--device",
+            device_path,
+            "--parent-pid",
```
cgwalters: This can be removed

blockdev/src/blockdev.rs
```diff
@@ -181,6 +181,14 @@ pub fn partitions_of(dev: &Utf8Path) -> Result<PartitionTable> {
...
 pub struct LoopbackDevice {
     pub dev: Option<Utf8PathBuf>,
+    // Handle to the cleanup helper process
+    cleanup_handle: Option<LoopbackCleanupHandle>,
+}
+
+/// Handle to manage the cleanup helper process for loopback devices
+struct LoopbackCleanupHandle {
+    /// Process ID of the cleanup helper
+    helper_pid: u32,
```
cgwalters: This should actually hold the [Child](https://doc.rust-lang.org/stable/std/process/struct.Child.html) instance. And on `Drop` we should send it `SIGTERM` - note *not* [kill](https://doc.rust-lang.org/stable/std/process/struct.Child.html#method.kill) but [rustix kill](https://docs.rs/rustix/latest/rustix/process/fn.kill_process.html) with `SIGTERM`.

blockdev/src/blockdev.rs
```diff
@@ -208,7 +216,19 @@ impl LoopbackDevice {
             .run_get_string()?;
         let dev = Utf8PathBuf::from(dev.trim());
         tracing::debug!("Allocated loopback {dev}");
-        Ok(Self { dev: Some(dev) })
+
+        // Try to spawn cleanup helper process - if it fails, continue without it
+        let cleanup_handle = Self::spawn_cleanup_helper(dev.as_str())
+            .map_err(|e| {
```
cgwalters: I'm uncertain about ignoring errors here by default. It seems better to make it fatal.

blockdev/src/blockdev.rs
```diff
@@ -217,13 +237,59 @@ impl LoopbackDevice {
...
         // SAFETY: This is the only place we take the option
         let Some(dev) = self.dev.take() else {
             tracing::trace!("loopback device already deallocated");
             return Ok(());
         };
+
+        // Kill the cleanup helper since we're cleaning up normally
+        if let Some(cleanup_handle) = self.cleanup_handle.take() {
+            // Kill the helper process since we're doing normal cleanup
+            let _ = std::process::Command::new("kill")
```
cgwalters: See above, we can directly send a signal instead of invoking an external command.

Also, crucially by holding a reference to the `Child` we avoid pid reuse problems if the child exits early.

blockdev/src/blockdev.rs
cgwalters: This still needs fixing, can you move this section of the change to a different PR?

cgwalters (changes requested):
Thanks, looking a lot closer!

cgwalters:
This is looking pretty good to me now! Is there a reason it's still in draft? Can you also mark issues in the discussion as resolved when they're fixed?

gursewak1997:
> Can you also mark issues in the discussion as resolved when they're fixed?

Every issue in discussion should be fixed now.

blockdev/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
```
cgwalters: This looks like a merge conflict artifact. I think we can just drop it - the path is now in `crates/blockdev/Cargo.toml`.

cgwalters (changes requested):
LGTM aside from one last bit

cgwalters (approved):

cgwalters:
Fallout in https://github.com/bootc-dev/bootc/issues/1439

PR #1413, opened 2025-07-16
Move all crates under `crates/`

cgwalters:
Just organizational cleanup, a number of Rust projects do this:
https://github.com/search?q=+%22crates%2F*%22+path%3ACargo.toml+-is%3Afork&type=code&p=5


cgwalters:
Note to self and others, run `make validate-rust` locally before doing PRs like this

jeckersb:
Manually rebased, not sure why github was claiming there was conflicts because it rebased just fine :shrug:

jeckersb:
Hm I think you pushed over my fixup

Edit: to be more specific:

https://github.com/bootc-dev/bootc/compare/8e0c7723e71a0a054382185896e7a72dbea390ad..5fd342e948a99ffc15e6d0911325253e2aaa5a77

cgwalters:
Thanks! Yes I messed that up, I think I fetched from my branch but forgot to actually reset my local to the remote. Should be fixed now

cgwalters:
Hmm Github seems to be treating any modifications to a renamed file as a conflict, when the git client definitely is happy to just do the obvious thing and update the new name.

Issue #1407, opened 2025-07-15
panic on unexpected EOF during pull


codegod100:
```
thread 'tokio-runtime-worker' panicked at ostree-ext/src/container/unencapsulate.rs:217:24:
Decompressor::finish MUST be called: unexpected end of file
Fetching layers ██░░░░░░░░░░░░░░░░░░ 9/75 
 └ Fetching █████░░░░░░░░░░░░░░░ 20.57 MiB/69.38 MiB (0 B/s) ostree chunk a15970c0d9d743e9f5fbd                                                                                                                             
  zsh: IOT instruction  sudo bootc switch ghcr.io/codegod100/template:latest
```

cgwalters:
> Decompressor::finish MUST be called

So we got here https://github.com/bootc-dev/bootc/blob/c8e7a6831690b6f468602232d08b6dc542c50904/ostree-ext/src/container/unencapsulate.rs#L217

> unexpected end of file

If all that happened here was an unexpected EOF we should definitely not panic on that. I think we probably need to just drop this panic because on error paths exactly like this it may be hit. Ultimately the calling code should be verifying checksums and the like for integrity.

And eventually we should switch to https://github.com/containers/skopeo/pull/2601

jeckersb:
If I had to guess, there's probably an error case somewhere that we missed that should be updated to clear out the decompressor when we get EOF like that.  I would be extremely wary to remove that panic, the comment above there sums up why.

jeckersb:
Actually looking at it again, the `unexpected end of file` is coming from 

https://github.com/bootc-dev/bootc/blob/c8e7a6831690b6f468602232d08b6dc542c50904/ostree-ext/src/container/unencapsulate.rs#L267

So I'm guessing something is wrong with the state of `self.inner` just based on the assumption that a `std::io::sink()` can't give unexpected EOF.

jeckersb:
I think this is probably coming from the gzip stream

https://github.com/rust-lang/flate2-rs/blob/7145603d57b7795c51d8674824163d6d685c9733/src/gz/mod.rs#L243

(give or take)

Which is easily reproduced by just feeding it an empty src:

```
diff --git i/ostree-ext/src/container/unencapsulate.rs w/ostree-ext/src/container/unencapsulate.rs
index 81eb90c7..21f7c23d 100644
--- i/ostree-ext/src/container/unencapsulate.rs
+++ w/ostree-ext/src/container/unencapsulate.rs
[...]

PR #1415, opened 2025-07-16
unencapsulate: use "inner" stream when finishing Decompressor

jeckersb:
For things like GzDecoder, we don't want to read via the actual
decompression reader because we don't care about decompressing at this
point.  Plus, the inner reader may have encountered an error partway
through, and trying to decode via decompression will error with
UnexpectedEof.

Instead, wrap a reader for each content type which knows how to get a
mutable reference to the underlying data stream.  When we finish
decompressing, read directly from this "inner" stream to flush any
data.

Resolves: #1407

### DIFF
--- a/crates/ostree-ext/src/container/store.rs
+++ b/crates/ostree-ext/src/container/store.rs
@@ -7,7 +7,7 @@
 
 use super::*;
 use crate::chunking::{self, Chunk};
-use crate::container::Decompressor;
+use crate::generic_decompress::Decompressor;
 use crate::logging::system_repo_journal_print;
 use crate::refescape;
 use crate::sysroot::SysrootLock;
--- a/crates/ostree-ext/src/container/unencapsulate.rs
+++ b/crates/ostree-ext/src/container/unencapsulate.rs
@@ -38,18 +38,12 @@ use containers_image_proxy::{ImageProxy, OpenedImage};
 use fn_error_context::context;
 use futures_util::{Future, FutureExt};
 use oci_spec::image::{self as oci_image, Digest};
-use std::io::Read;
 use std::sync::{Arc, Mutex};
 use tokio::{
     io::{AsyncBufRead, AsyncRead},
     sync::watch::{Receiver, Sender},
 };
 use tracing::instrument;
-
-/// The legacy MIME type returned by the skopeo/(containers/storage) code
-/// when we have local uncompressed docker-formatted image.
-/// TODO: change the skopeo code to shield us from this correctly
-const DOCKER_TYPE_LAYER_TAR: &str = "application/vnd.docker.image.rootfs.diff.tar";
 
 type Progress = tokio::sync::watch::Sender<u64>;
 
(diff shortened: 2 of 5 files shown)


jeckersb:
This was quite the exercise in fighting the type system, but I think this is generally what we want...

ostree-ext/src/container/unencapsulate.rs
```diff
@@ -191,8 +191,59 @@ pub async fn unencapsulate(repo: &ostree::Repo, imgref: &OstreeImageReference) -
...
+    fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
+        self.0.read(buf)
+    }
+}
+
+impl Decompressable for TransparentDecompressor {
+    fn get_inner_mut(&mut self) -> &mut (dyn Read) {
+        &mut self.0
+    }
+}
```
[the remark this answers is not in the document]
jeckersb: Good bot.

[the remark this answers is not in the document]
jeckersb: Well, most of the way there.  Should also update the instantiation to not create the unnecessary `Box`.

ostree-ext/src/container/unencapsulate.rs
```diff
@@ -362,4 +415,18 @@ mod tests {
         let d = Decompressor::new(&oci_image::MediaType::ImageLayer, empty).unwrap();
         drop(d)
     }
+
+    #[test]
+    fn test_drop_decompressor_with_incomplete_gzip_data() {
```
cgwalters: Thanks for the tests; would also probably be good to have some tests which verify we don't panic if we get garbage input too.

ostree-ext/src/container/unencapsulate.rs
```diff
@@ -191,8 +191,59 @@ pub async fn unencapsulate(repo: &ostree::Repo, imgref: &OstreeImageReference) -
     importer.unencapsulate().await
 }
 
+trait Decompressable: Read + Send + 'static {
```
cgwalters: I think this would be a bit cleaner as its own little `mod generic_decompress.rs` or so and also include a little bit of rationale in the module docs there, something like

```
//! Trait which extends the `Read` trait with another method to get mutable access to the inner reader
```

(And we may as well make it not specific to decompression, so it could be `ReadAndGetMut` or something? And the module name `read_and_get_mut.rs` or so)

cgwalters: Though I guess needing concrete impls of the trait per compressor argue for keeping it to just be `generic_decompress.rs`

ostree-ext/src/container/unencapsulate.rs
```diff
@@ -191,8 +191,59 @@ pub async fn unencapsulate(repo: &ostree::Repo, imgref: &OstreeImageReference) -
...
+    fn get_inner_mut(&mut self) -> &mut (dyn Read);
```
cgwalters: The concrete versions (e.g. [this one](https://docs.rs/zstd/latest/zstd/stream/read/struct.Decoder.html#method.get_mut)) spell it `&mut R` which seems clearer to me.

jeckersb: But in the trait definition I can't make the method generic or it's not dyn compatible anymore:

```
error[E0038]: the trait `Decompressable` is not dyn compatible
   --> ostree-ext/src/container/unencapsulate.rs:198:16
    |
198 |     inner: Box<dyn Decompressable>,
    |                ^^^^^^^^^^^^^^^^^^ `Decompressable` is not dyn compatible
    |
note: for a trait to be dyn compatible it needs to allow building a vtable
      for more information, visit <https://doc.rust-lang.org/reference/items/traits.html#dyn-compatibility>
   --> ostree-ext/src/generic_decompress.rs:6:8
    |
5   | pub(crate) trait Decompressable: Read + Send + 'static {
    |                  -------------- this trait is not dyn compatible...
6   |     fn get_inner_mut<R: Read>(&mut self) -> &mut R;
    |        ^^^^^^^^^^^^^ ...because method `get_inner_mut` has generic type parameters
    = help: consider moving `get_inner_mut` to another trait
```

or am I misunderstanding?  Do you have a more exact suggestion?

jeckersb:
Trying something new... added a semi-casual commit just to keep things clear as feedback is addressed during the review process.  We'll squash those out at the end.

ostree-ext/src/container/unencapsulate.rs
jeckersb: Added garbage input tests for both gz and zstd

ostree-ext/src/container/unencapsulate.rs
jeckersb: Moved this out to generic_decompress

ostree-ext/src/container/unencapsulate.rs
cgwalters: Ahh right of course. Yep it needs to be as is then.

ostree-ext/src/generic_decompress.rs
```diff
@@ -0,0 +1,66 @@
...
+//!
+//! We need a common way to access this stream so that we can flush
+//! the data during cleanup.
+//!
+//! See: https://github.com/bootc-dev/bootc/issues/1407
+
+use std::io::Read;
+
+/// Extends the `Read` trait with another method to get mutable access to the inner reader
+pub(crate) trait Decompressable: Read + Send + 'static {
```
cgwalters: Rust trait naming tends to be imperative (I think because usually one invokes methods on them), so it'd argue for this to be `Decompressor` or so.

But then it conflicts with the fact we actually already *have* a `Decompressor` that this is an inner implementation detail of...

Maybe then the right fix is to move that `Decompressor` in here, and rename the trait to `ReadWithGetInnerMut` or so?

jeckersb: Done in latest :+1:

cgwalters (approved):
Nice! There's more code motion here but I think it really makes things a good bit cleaner to have all the gory details hidden inside that module.

ostree-ext/src/generic_decompress.rs
```diff
@@ -0,0 +1,227 @@
...
+        if n > 0 {
+            tracing::debug!("Read extra {n} bytes at end of decompressor stream");
+        }
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
cgwalters: These all look great!

ostree-ext/src/generic_decompress.rs
```diff
@@ -0,0 +1,227 @@
...
+            return;
+        }
+
+        // We really should not get here; users are required to call
+        // `finish()` to clean up the stream.  But we'll give
+        // best-effort to clean things up nonetheless.  If things go
+        // wrong, then panic, because we're in a bad state and it's
+        // likely that we end up with a broken pipe error or a
+        // deadlock.
+        self._finish().expect("Decompressor::finish MUST be called")
```
cgwalters: (Sorry this is an unrelated pre-existing issue, but I'm noticing it as it's revealed by the reported bug; we can add another commit on top here, or just do another PR)

In error paths, we still expect `finish()` to not be called, right?

Bigger picture, this whole thing gets into [linear types](https://faultlore.com/blah/linear-rust/) if we were to really try to enforce it, and it's basically not possible today.

As is basically what we've done is removed all real possibility for errors from this invocation as it should boil down to `read()` from a pipe fd which should basically never fail.

That said still it's not clear to me that we should have ever been panic'ing on this error at all.

So bottom line I think we should pick one of:

- Change the error message to `Failed to flush pipe`
- Change it to an explicit `// Ignore errors here, shouldn't fail let _ = self.finish(); ` (or maybe `tracing::warn!` on the error?)

jeckersb: Yeah in error paths we're going to short-circuit return and then the decompressor will be dropped.

I'm leaning towards still panic'ing and changing the error message to `Failed to flush pipe`.  At this point I *think* it's basically impossible to ever hit that panic... but if somebody somewhere *does* hit it I'd like it to be very loud and obvious so we have a chance to maybe hear about it (much like the report that spawned this PR).  I'm afraid if we ignore the error and log somewhere that nobody will ever see, we're going to have instances where things go back to hanging/deadlocking and then we'll have an even worse experience.

cgwalters: Yep OK by me to just change the error message.

cgwalters: That said on all of this the *real* fix is for us to switch to GetRawBlob - I'll file an issue

:arrow_right: https://github.com/bootc-dev/bootc/issues/1424

jeckersb: Added a commit to address this

cgwalters:
Let's avoid merging main into PRs, can you just squash everything?

jeckersb:
> Trying something new... added a semi-casual commit just to keep things clear as feedback is addressed during the review process. We'll squash those out at the end.

^^^

Always intended on squashing it, which is why I just merged main in for now.  More to the theme of making it easier to follow what's going on as we iterate here instead of force pushing repeatedly.

cgwalters:
In general let's follow a SOP of keeping things draft if they're not ready to merge - and being "ready to merge" should include having a reasonably clean commit history where fixups are squashed.

PR #1417, opened 2025-07-17
lib/src/reboot: use valid systemd-run arguments

p5:
The current `--message` argument was a Gemini hallucination, and doesn't actually exist.
The closest thing I could find was the `--description` argument, which I think does what we're expecting it to.

```
❯ systemd-run --description="Hello World" echo "hi"
Running as unit: run-p58325-i58326.service; invocation ID: 86e9ddd7f9bd470fa8bca10f64710d8d

❯ journalctl -xe  
...
Jul 17 11:05:30 UNKNOWN audit[1]: SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='unit=fprintd comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? termi>
Jul 17 11:05:32 UNKNOWN audit[58368]: USER_AUTH pid=58368 uid=1000 auid=1000 ses=3 subj=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 msg='op=PAM:authentication grantors=pam_unix acct="admin" exe="/usr/l>
Jul 17 11:05:32 UNKNOWN audit[58368]: USER_ACCT pid=58368 uid=1000 auid=1000 ses=3 subj=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 msg='op=PAM:accounting grantors=pam_unix acct="admin" exe="/usr/lib/p>
Jul 17 11:05:32 UNKNOWN systemd[1]: Started run-p58325-i58326.service - Hello World.
```


cgwalters:
Ugh having that previous PR get through CI is a *very* serious regression we need to root cause.

What's going on here is that `--message` *is* a valid argument in `systemctl --message=x reboot` (as distinct from the plain `reboot` wrapper) but we incorrectly passed it to `systemd-run`.

Also in this we should separate the arguments via `--` explicitly.

cgwalters:
> Would this bug warrant a new patch release? Concerned that if users are relying on the auto-update-reboot feature, their machines will not be updated until they manually reboot.

Yes we must fix this immediately

p5:
Pushed a new commit passing it in as so:

```
systemd-run -- reboot --message="Initiated by bootc"
```

This successfully rebooted my machine.

Starting to look at tests, but this can probably be merged before then.

lib/src/reboot.rs
```diff
@@ -14,7 +14,7 @@ pub(crate) fn reboot() -> anyhow::Result<()> {
     let _ = std::io::stdout().flush();
     let _ = std::io::stderr().flush();
     Task::new("Rebooting system", "systemd-run")
-        .args(["--message=Initiated by bootc", "reboot"])
+        .args(["reboot", "--", "--message=Initiated by bootc"])
```
cgwalters: As `reboot --help` says

> This is a compatibility interface, please use the more powerful 'systemctl reboot' command instead.

And to this point 

```
+ reboot --message foo
reboot: unrecognized option '--message'
```

So I don't see how this can work.

What we want is:

`systemd-run systemctl reboot --message=Initiated by bootc`

p5: Ah, I'm very sorry.  I confirmed it _did_ reboot, but not that there was the supplied message associated with the reboot.

My lack of basic Bash knowledge showing up here...

cgwalters:
Getting sidetracked by a testing framework, still looking at that

Issue #1346, opened 2025-05-31
ostree container image pull to bare-user repo silent metadata corruption & non reproducibility


champtar:
This is with rpm-ostree-2025.5-1.el9.x86_64, will need to redo the tests with main at some point

Trying to use `ostree container image pull` in a container (for my CI) with all modes of ostree repo:
- `bare` (in container): `Importing regfile small: Writing content object: Setting xattrs: fsetxattr(security.selinux): Invalid argument`
- `bare-user-only`: `invalid mode 0104755 with bits 04000`
- `bare-split-xattrs`: `Importing regfile small: Writing content object: Not allowed due to repo mode`
- `archive`: `Importing regfile: Cannot currently use ostree_repo_write_regfile() on an archive mode repository`
- `bare-user`: seems to work ...

```
# compose an image
# rpm-ostree compose image --initialize-mode=always --cachedir=build/aaa -l org.opencontainers.image.version=my-version build/ostree-tmp-ba/ba.yaml build/test.oci
...
Wrote commit: 67235f74d871ad8803dabdb58bdb13b222bcd20d679f7ebe0c2c19cf0a82d178
Pushed digest: sha256:0dca95e35c6d4925d537e8713b8e1c5f13f747f3ad8c9b57cec05a21da49ebca
```

```
# rm -rf build/bbb                                                                                                                                                  
# ostree init --mode=bare-user --repo=build/bbb                                                                                                                     
# ostree container image pull build/bbb ostree-unverified-image:oci-archive:build/test.oci                                                                          
layers already present: 0; layers needed: 65 (5.5 GB)
 2.13 MiB [████████████████████] (0s) Fetched ostree chunk 91f16f4fb72a6a5d137                                                                                                              
Wrote: ostree-unverified-image:oci-archive:build/test.oci => cf9fb0c3e462e7c8cb0c6c4ed93fd6d9af32bd8b200947c8b0a5cfdd19417b2c
# ostree diff --repo=build/bbb 67235f74d871ad8803dabdb58bdb13b222bcd20d679f7ebe0c2c19cf0a82d178 cf9fb0c3e462e7c8cb0c6c4ed93fd6d9af32bd8b200947c8b0a5cfdd19417b2c
M    /usr/bin/sudo
M    /usr/etc/gshadow
M    /usr/etc/shadow
M    /usr/etc/polkit-1/rules.d
M    /usr/libexec/utempter
M    /usr/share/polkit-1/rules.d
```

```
[...]

PR #1421, opened 2025-07-17
Make `ostree container image pull` merge commit reproducible

jlebon:
ostree-ext/store: use BTreeMap for merge commit metadata

`HashMap` is unordered which means that the conversion to GVariant (and thus its serialization) is not reproducible since the conversion basically just maps to an GVariant array. That in turns means the commit hash itself is not reproducible.

All we really need to do is just swap it for a `BTreeMap` since all the types involved here implement ordering.

---

ostree-ext/store: don't include filtered files in metadata

Filtered files are only determined at the time we import a layer. So if that layer is already imported, we won't have that information available. That in turn means that the metadata is state-dependent, which in turn means that the commit digest is not reproducible.

We still want to provide the filtered files warning though. Just make this information part of the LayeredImageState object instead. The obvious downside of that is that now we only get that warning the first time the layer is imported and it's no longer part of the commit object itself.

One way to make this more sticky is to attach it to the individual layers' commits instead, and then the merge commit can coalesce them.

Related: https://github.com/bootc-dev/bootc/issues/1346

### DIFF
--- a/ostree-ext/src/cli.rs
+++ b/ostree-ext/src/cli.rs
@@ -238,6 +238,10 @@ pub(crate) enum ContainerImageOpts {
         #[clap(value_parser = parse_imgref)]
         imgref: OstreeImageReference,
 
+        /// File to which to write the resulting OSTree commit digest
+        #[clap(long)]
+        ostree_digestfile: Option<Utf8PathBuf>,
+
         #[clap(flatten)]
         proxyopts: ContainerProxyOpts,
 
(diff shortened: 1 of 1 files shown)


ostree-ext/src/cli.rs
```diff
@@ -238,6 +238,10 @@ pub(crate) enum ContainerImageOpts {
         #[clap(value_parser = parse_imgref)]
         imgref: OstreeImageReference,
 
+        /// File to which to write the resulting OSTree commit digest
+        #[clap(long)]
+        digestfile: Option<Utf8PathBuf>,
```
cgwalters: How about `ostree_digestfile` to be crystal clear which one it is, since there's also the container image digest?

jlebon: Done!

Issue #1016, opened 2025-01-07
json-fd issue tracker


cgwalters:
Followup to https://github.com/containers/bootc/pull/921 which we merged, and I was hoping to do some post-merge followup on. This issue will track some of that followup.

- [ ] docs https://github.com/containers/bootc/pull/961
- [ ] integration tests (I'm working on that now)
- [ ] size concerns

## size concerns

In the PR we kept going back and forth on whether the progress data should require the client to keep track of state or not. @antheas was calling this "normalization". There's clear simplicity in having the protocol be stateless.

However...looking at the layer progress (one of the most important phases) in a local run here during a download, we keep accumulating and emitting more data (completed subtasks) for each layer. In this run the final `ProgressBytes` I got was 12 KiB which...is a bit more than I was expecting.

Click for (prettified) example:

<details>
{
  "type": "ProgressBytes",
  "apiVersion": "org.containers.bootc.progress/v1",
  "task": "pulling",
  "description": "Pulling Image: sha256:127c32e1418aebdd9a6cbeb6517605e11c1aa4004689e4a396307babe801f309",
  "id": "sha256:127c32e1418aebdd9a6cbeb6517605e11c1aa4004689e4a396307babe801f309",
  "bytesCached": 0,
  "bytes": 1212857687,
  "bytesTotal": 1212857687,
  "stepsCached": 0,
  "steps": 74,
  "stepsTotal": 74,
  "subtasks": [
    {
      "subtask": "ostree chunk",
      "description": "ostree chunk: bca923230b7dde927d4f2",
      "id": "bca923230b7dde927d4f2",
      "bytesCached": 0,
      "bytes": 434743772,
      "bytesTotal": 434743772
    },
    {
      "subtask": "ostree chunk",
      "description": "ostree chunk: 7787da09b798ff4c245e5",
      "id": "7787da09b798ff4c245e5",
      "bytesCached": 0,
      "bytes": 23673835,
      "bytesTotal": 23673835
    },
    {
      "subtask": "ostree chunk",
      "description": "ostree chunk: 46b215c7346346e749145",
      "id": "46b215c7346346e749145",
      "bytesCached": 0,
      "bytes": 35716920,
      "bytesTotal": 35716920
    },
    {
      "subtask": "ostree chunk",
      "description": "ostree chunk: 5ce7d0f46f7ce0107eaab",
      "id": "5ce7d0f46f7ce0107eaab",
      "bytesCached": 0,
      "bytes": 14600588,
[...]

PR #1423, opened 2025-07-17
Remove exp progressfd

jmarrero:


cgwalters:
I know this is just starting things, but for reference a lot of discussion in https://github.com/bootc-dev/bootc/issues/1016 which should be clearly linked here in the PR description etc. Did you have any feedback/thoughts there?

Overall...yeah we could just basically do this and move on for now...but...

Actually though, at least for Anaconda this won't be useful until we do https://github.com/bootc-dev/bootc/pull/982

jmarrero:
> I know this is just starting things, but for reference a lot of discussion in #1016 which should be clearly linked here in the PR description etc. Did you have any feedback/thoughts there?

Yeah, have had to read it a couple of times to understand the expectations here. Added a comment there and got some answers.

> Actually though, at least for Anaconda this won't be useful until we do #982

OK, let's talk about that in #1016

PR #1429, opened 2025-07-18
build: Add actions to automate release

ckyrouac:
This adds two github actions, "Create Release PR" and "release". The first is scheduled to run every 3 weeks to automatically create a release PR that bumps the versions. The "release" action is triggered when the release PR is merged. It will create a draft release with the tars attached.

Assited-by: Claude Code

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,132 @@
+name: Release
+
+on:
+  pull_request:
+    types: [closed]
+
+permissions:
+  contents: write
+
+jobs:
+  release:
+    name: Create Release
+    if: |
+      (github.event_name == 'pull_request' && 
+       github.event.pull_request.merged == true &&
+       contains(github.event.pull_request.labels.*.name, 'release'))
+    runs-on: ubuntu-latest
+    container: quay.io/coreos-assembler/fcos-buildroot:testing-devel
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          token: ${{ secrets.GITHUB_TOKEN }}
+      
+      - name: Extract version
+        id: extract_version
+        run: |
+          # Extract version from crates/lib/Cargo.toml
+          VERSION=$(grep '^version' crates/lib/Cargo.toml | sed 's/version = "//;s/"//')
+          
+          # Validate version format
+          if ! echo "$VERSION" | grep -E '^[0-9]+\.[0-9]+\.[0-9]+$' >/dev/null; then
+            echo "Error: Invalid version format in Cargo.toml: $VERSION"
+            exit 1
+          fi
+          
+          echo "Extracted version: $VERSION"
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
+          echo "TAG_NAME=v$VERSION" >> $GITHUB_OUTPUT
+      
+      - name: Install deps
+        run: ./ci/installdeps.sh
+      
+      - name: Mark git checkout as safe
+        run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
+      
+      - name: Import GPG key
+        if: github.event_name != 'push'
+        uses: crazy-max/ghaction-import-gpg@v6
+        with:
+          gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
+          passphrase: ${{ secrets.GPG_PASSPHRASE }}
+          git_user_signingkey: true
+          git_commit_gpgsign: true
+          git_tag_gpgsign: true
+      
+      - name: Create and push tag
+        if: github.event_name != 'push'
+        run: |
+          VERSION="${{ steps.extract_version.outputs.version }}"
+          TAG_NAME="v$VERSION"
+          
+          # Check if tag already exists
+          if git rev-parse "$TAG_NAME" >/dev/null 2>&1; then
+            echo "Tag $TAG_NAME already exists"
+            exit 0
+          fi
+          
+          # Create signed annotated tag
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -99,3 +99,48 @@ vendor:
...
+# Create a release commit with updated version
+# Usage: make release-commit VERSION=1.5.0
+release-commit:
+	@if [ -z "$(VERSION)" ]; then \
+		echo "Error: VERSION is required. Usage: make release-commit VERSION=1.5.0"; \
+		exit 1; \
+	fi
+	@echo "Creating release commit for version $(VERSION)..."
+	# Update version in lib/Cargo.toml
+	sed -i 's/^version = ".*"/version = "$(VERSION)"/' crates/lib/Cargo.toml
```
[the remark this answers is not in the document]
cgwalters: Wild. But actually what we really want here is to use a proper TOML editor.

Actually wait, there's already a tool for this: https://github.com/killercup/cargo-edit/?tab=readme-ov-file#cargo-set-version

Makefile
```diff
@@ -99,3 +99,48 @@ vendor:
...
+	# Create commit
+	git commit -m "Release $(VERSION)"
+	# Create signed tag
+	git tag -s -m "Release $(VERSION)" v$(VERSION)
+.PHONY: release-commit
+
+# Get the next minor version by bumping the minor version from crates/lib/Cargo.toml
+# Outputs the new version string (e.g., "1.3.0" if current version is "1.2.4")
+next-minor-version:
+	@VERSION=$$(grep '^version' crates/lib/Cargo.toml | sed 's/version = "//;s/"//'); \
```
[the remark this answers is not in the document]
cgwalters: We can actually get it from `cargo metadata`.

I guess in general...WDYT about putting some of this in xtask.rs in Rust instead of shell script?

Makefile
```diff
@@ -99,3 +99,48 @@ vendor:
...
+	@echo "Creating release commit for version $(VERSION)..."
+	# Update version in lib/Cargo.toml
+	sed -i 's/^version = ".*"/version = "$(VERSION)"/' crates/lib/Cargo.toml
+	# Update Cargo.lock with new version
+	cargo update --workspace
+	# Run cargo xtask update-generated to update generated files
+	cargo xtask update-generated
+	# Stage all changes
+	git add crates/lib/Cargo.toml Cargo.lock
+	git add -A  # Add any files updated by cargo xtask update-generated
```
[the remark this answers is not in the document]
cgwalters: Also the `git add -A` overlaps with the previous `git add`. I'd probably just use `git commit -a` instead below.

cgwalters:
Also thanks for working on this, very overdue!

.github/workflows/release.yml
```diff
@@ -0,0 +1,127 @@
...
+        run: ./ci/installdeps.sh
+      
+      - name: Mark git checkout as safe
+        run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
+      
+      - name: Import GPG key
+        if: github.event_name != 'push'
+        uses: crazy-max/ghaction-import-gpg@v6
+        with:
+          gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
```
cgwalters, line 52: We need to make this keypair and add to secrets in this repo right?

And when we do that we'll also want to sign it via web of trust and/or just document the public key explicitly in the repo.

.github/workflows/release.yml
```diff
@@ -0,0 +1,127 @@
...
+      
+      - name: Install deps
+        run: ./ci/installdeps.sh
+      
+      - name: Mark git checkout as safe
+        run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
+      
+      - name: Import GPG key
+        if: github.event_name != 'push'
+        uses: crazy-max/ghaction-import-gpg@v6
```
cgwalters, line 50: I had never heard of "crazy-max" before but his action does seem to be widely used at least, and referenced in e.g. https://github.com/goreleaser/goreleaser/blob/2813d4734d2381c6a3ea1d926d41da2eee6b70b4/www/docs/ci/actions.md?plain=1#L92

So that's good.

.github/workflows/scheduled-release.yml
```diff
@@ -0,0 +1,162 @@
...
+            # Validate version format strictly
+            if ! echo "$VERSION" | grep -E '^[0-9]+\.[0-9]+\.[0-9]+$' >/dev/null; then
+              echo "Error: Invalid version format. Expected X.Y.Z (e.g., 1.5.1)"
+              exit 1
+            fi
+            echo "VERSION=$VERSION" >> $GITHUB_OUTPUT
+            exit 0
+          fi
+          
+          NEW_VERSION=$(cargo xtask next-version minor)
```
cgwalters: This code seems to overlap with the `cargo edit` below?

Can't we just bump via `cargo edit` and then use `cargo metadata` to get the version back into this `VERSION` variable?

.github/workflows/release.yml
ckyrouac: I was hesitant to use that action too but it's pretty frustrating to setup the gpg key manually and yea that action is widely used.

cgwalters:
LGTM

.github/workflows/release.yml
cgwalters: @ckyrouac  Did we ever do anything here? here's my proposal:

- Encrypt the release key using the ssh (or gpg as preferred) keys of all existing maintainers
- Commit that encrypted result to https://github.com/bootc-dev/infra (yes publicly... https://github.com/getsops/sops looks like maybe a good choice of tool)
- Move the secret to the org level

My goal here is to be able to reuse exactly the same release workflow (including signing) for bcvk.

PR #1431, opened 2025-07-18
install: add progress-fd for install

jmarrero:
Assisted by: Claude Code

This supersedes #982 and is part of #1016 

Draft because I have not tested this actually does what we want other than making the tests pass.


cgwalters:
For tests I'd add to `tests-integration/src/install.rs` and for verifying that we get JSON out of it at least look at test-image-pushpull-progress.nu. A problem right now is the two test frameworks are totally different.

(tangent but I'm not totally happy with the nushell stuff in the end, it's not bad but it's obscure enough that AI tools IME confabulate a lot. The Rust tests are OK but one general issue there is frameworks like TMT don't make it easy (afaik) to use tests that are compiled, the framework wants to just copy arch-independent scripts/yaml from a git repo)

PR #1442, opened 2025-07-23
chunking: Fix specific component file allocation

ckyrouac:
This fixes the bug where the specific layers were looking at the wrong meta object to allocate files to the layer.

### DIFF
--- a/crates/ostree-ext/src/chunking.rs
+++ b/crates/ostree-ext/src/chunking.rs
@@ -233,6 +233,17 @@ impl Chunk {
 }
 
 impl Chunking {
+    /// Creates a reverse map from content IDs to checksums
+    fn create_content_id_map(
+        map: &IndexMap<String, ContentID>,
+    ) -> IndexMap<ContentID, Vec<&String>> {
+        let mut rmap = IndexMap::<ContentID, Vec<&String>>::new();
+        for (checksum, contentid) in map.iter() {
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
Hmm, how hard would a unit test for this be? AIUI this bug would be provoked when there's a common object between layers?

ckyrouac:
Yea I'll write a unit test. This was exposed by the integration test I'm working on in rpm-ostree. The specific components were being created but the files were landing in the extra stuff layer because this code was looking for them in the wrong bucket when doing the layer mapping.

crates/ostree-ext/src/chunking.rs
```diff
@@ -1221,4 +1230,183 @@ mod test {
...
+            setup_exclusive_test(&packages, 8, None)?;
+
+        // Create object mappings
+        // - system_objects_map: Contains ALL objects in the system (both specific and regular)
+        // - specific_components_objects_map: Contains ONLY objects from specific components
+        let mut system_objects_map = IndexMap::new();
+        let mut specific_components_objects_map = IndexMap::new();
+
+        // SPECIFIC COMPONENT 1 (pkg1): owns 3 objects
+        let pkg1_objects = ["checksum_1_a", "checksum_1_b", "checksum_1_c"];
```
cgwalters, line 1260: OK as is but in theory we should actually require these to be at least string SHA256 digests.

cgwalters (approved):

PR #1445, opened 2025-07-24
Bump composefs-rs

Johan-Liebert1:
Also pull in `composefs-boot` and `composefs-oci`


crates/lib/src/cli.rs
```diff
@@ -1227,7 +1227,7 @@ async fn run_from_opt(opt: Opt) -> Result<()> {
                 FsverityOpts::Enable { path } => {
                     let fd =
                         std::fs::File::open(&path).with_context(|| format!("Reading {path}"))?;
-                    fsverity::enable_verity::<fsverity::Sha256HashValue>(&fd)?;
+                    fsverity::enable_verity_raw::<fsverity::Sha256HashValue>(&fd)?;
```
cgwalters, line 1230: This is the most direct change and is fine for now but we should do a followup that uses the retry variants

PR #1447, opened 2025-07-24
ci: Drop ppc64le COPR builds from presubmits

cgwalters:
These are failing due to infra issues apparently, and we have coverage at release time, we really don't
need to build every single PR by default.


jmarrero (approved):
nuke it

cgwalters:
Added another commit here to drop Fedora 41

jmarrero (approved):

Issue #1439, opened 2025-07-23
`install to-disk` failed with error `Failed to spawn loopback cleanup helper`


henrywang:
`bootc-202507222002.ged787b4cee-1.el10.x86_64` have this issue, but `bootc-202507181316.g03fa72b194-1.el10.x86_64` does not.
According to the commit `03fa72b194` -> `ed787b4cee`, I guess the issue should be related with https://github.com/bootc-dev/bootc/commit/c2c918cc805fe5cfcc9434e4da419b98460abb60

```
error: Installing to disk: Failed to spawn loopback cleanup helper: Failed to spawn loopback cleanup helper: No such file or directory (os error 2)
```

Test log: https://artifacts.osci.redhat.com/testing-farm/2dccf9e7-3148-4ed4-b8ba-f07d1bdab255/

cgwalters:
I'm OK to just back out the PR for now

cgwalters:
(Should have been closed by https://github.com/bootc-dev/bootc/pull/1440 )

PR #1449, opened 2025-07-24
blockdev: implement signal-safe loopback device cleanup helper

gursewak1997:
Adds a signal-safe loopback device cleanup mechanism for `bootc install --via-loopback` to prevent device leaks on interruption (e.g., SIGINT). It introduces a `loopback-cleanup-helper` subcommand and a robust helper binary path resolution strategy to ensure reliable spawning across environments. The update improves error handling, logging, and includes comprehensive tests.


crates/blockdev/src/blockdev.rs
```diff
@@ -197,18 +209,51 @@ impl LoopbackDevice {
...
+        let output = Command::new("losetup")
             .args([
                 "--show",
                 format!("--direct-io={direct_io}").as_str(),
                 "-P",
                 "--find",
+                path.as_str(),
             ])
-            .arg(path)
-            .run_get_string()?;
```
cgwalters, line 208: I think the AI tooling doesn't know about our helper `run_get_string()` which handles the `.success()` and error printing.

IOW this code was probably fine as is, can we just keep it that way please?

crates/blockdev/src/blockdev.rs
```diff
@@ -217,13 +262,81 @@ impl LoopbackDevice {
...
+            .spawn()
+            .context("Failed to spawn loopback cleanup helper")?;
+
+        Ok(LoopbackCleanupHandle { child })
+    }
+
+    /// Find the bootc binary using multiple strategies
+    fn find_bootc_binary() -> Result<PathBuf> {
+        // Strategy 1: Try /proc/self/exe (works in most cases)
+        if let Ok(exe_path) = std::fs::read_link("/proc/self/exe") {
```
cgwalters, line 276: OK so this is trying to handle the original failure...but it's really unclear to me *why* it was failing and specifically how that test case run is different than the existing loopback tests we have.

(original issue in https://github.com/bootc-dev/bootc/issues/1439 )

As far as I know `/proc/self/exe` should really always work; we're using it in `reexec_with_guardenv`.

I think what would be good here is that if `read_link` fails, let's show what value it points to.

crates/blockdev/src/blockdev.rs
```diff
@@ -217,13 +262,81 @@ impl LoopbackDevice {
...
+
     // Shared backend for our `close` and `drop` implementations.
     fn impl_close(&mut self) -> Result<()> {
         // SAFETY: This is the only place we take the option
         let Some(dev) = self.dev.take() else {
             tracing::trace!("loopback device already deallocated");
             return Ok(());
         };
+
+        // Kill the cleanup helper since we're cleaning up normally
```
cgwalters, line 319: Hmm. We could also send it `SIGTERM` and let it do the cleanup. Feels like it's perhaps better that way as then we're always exercising the cleanup via child process.

crates/blockdev/src/blockdev.rs
```diff
@@ -194,6 +194,10 @@ struct LoopbackCleanupHandle {
 impl LoopbackDevice {
     // Create a new loopback block device targeting the provided file path.
     pub fn new(path: &Path) -> Result<Self> {
+        let path = Utf8PathBuf::from_path_buf(path.to_path_buf())
```
cgwalters: If we wanted to impose this requirement it'd be cleaner to make the API just take a `Utf8Path` instead. But I don't think we need the requirement so I think we should drop this change.

crates/blockdev/src/blockdev.rs
```diff
@@ -205,25 +209,50 @@ impl LoopbackDevice {
...
 
-        let dev = Command::new("losetup")
+        // Create the loopback device
+        let output = Command::new("losetup")
             .args([
                 "--show",
                 format!("--direct-io={direct_io}").as_str(),
                 "-P",
                 "--find",
+                path.as_str(),
```
cgwalters: The utf8 thing probably came from this, but I don't think the change is necessary.

crates/blockdev/src/blockdev.rs
```diff
@@ -311,22 +372,32 @@ pub async fn run_loopback_cleanup_helper(device_path: &str) -> Result<()> {
...
+    match output {
+        Ok(output) if output.status.success() => {
             // Log to systemd journal instead of stderr
             tracing::info!("Cleaned up leaked loopback device {}", device_path);
             std::process::exit(0);
         }
-        Ok(_) => {
-            tracing::error!("Failed to clean up loopback device {}", device_path);
+        Ok(output) => {
+            let stderr = String::from_utf8_lossy(&output.stderr);
```
cgwalters, line 371: This child process has its stdout/stderr going to /dev/null by default so all of this is just lost unfortunately.

crates/blockdev/src/blockdev.rs
cgwalters: Hmm random idea as to the root cause - what if this is happening just when we need to re-exec the current process *twice*? And inside the container env?

I'm starting to suspect it has to do with that...

gursewak1997: Def could be. Adding some changes that handles the case where `/proc/self/exe` fails (which could happen during re-execution in container environments)

cgwalters: Ah yes it's this almost for sure https://github.com/bootc-dev/bootc/blob/f4b01ab08cb8097afabfcc6ff4c8d5179df38a2f/crates/lib/src/lsm.rs#L91

In some cases we're hitting that fallback, but not others.

So...this is getting a bit complicated. Your fallback code here will definitely work in general....but I have some ideas to do something more sophisticated.

cgwalters: The more complex fix I think is to make an new internal "out of process drop helper" crate that we initialize *before* we do any re-exec dances. I.e. we fork/re-exec that helper and talk to it over IPC or so. It would be more involved though.

cgwalters:
OK it turns out this never worked https://github.com/bootc-dev/bootc/issues/1510

PR #1466, opened 2025-07-30
docs: add Playtron GameOS

LukeShortCloud:
to the ADOPTERS.md. We have builds dating all the way back to 2022.


cgwalters (approved):
Thanks!

PR #1471, opened 2025-07-31
store: Add support for composefs-rs

cgwalters:


crates/lib/src/store/mod.rs
```diff
@@ -111,6 +153,31 @@ impl Storage {
...
+        let ostree_repo = &self.sysroot.repo();
+        let ostree_verity = ostree_ext::fsverity::is_verity_enabled(ostree_repo)?;
+        if !ostree_verity.enabled {
+            tracing::debug!("Setting insecure mode for composefs repo");
+            composefs.set_insecure(true);
+        }
+        let composefs = Arc::new(composefs);
+        let r = Arc::clone(self.composefs.get_or_init(|| composefs));
+        Ok(r)
+    }
```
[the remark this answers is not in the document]
cgwalters: Yeah...in practice this code isn't called from multiple threads, we just want at most one init.

Also get_or_try_init requires the external crate...I guess we can add it back though.

crates/lib/src/store/mod.rs
```diff
@@ -111,6 +153,31 @@ impl Storage {
...
     }
 
+    pub(crate) fn get_ensure_composefs(&self) -> Result<Arc<ComposefsRepository>> {
+        if let Some(composefs) = self.composefs.get() {
+            return Ok(Arc::clone(composefs));
+        }
+
+        let mut db = DirBuilder::new();
+        db.mode(COMPOSEFS_MODE.as_raw_mode());
+        self.physical_root.ensure_dir_with(COMPOSEFS, &db)?;
```
bshephar, line 163: Do we not care about the returned bool for anything here? We just rely on the fact that if there's an error, we're unwraping it and we will return the error. The returned bool isn't really useful for us in this context I guess?

crates/lib/src/store/mod.rs
```diff
@@ -111,6 +153,31 @@ impl Storage {
...
+        // default right now.
+        let ostree_repo = &self.sysroot.repo();
+        let ostree_verity = ostree_ext::fsverity::is_verity_enabled(ostree_repo)?;
+        if !ostree_verity.enabled {
+            tracing::debug!("Setting insecure mode for composefs repo");
+            composefs.set_insecure(true);
+        }
+        let composefs = Arc::new(composefs);
+        let r = Arc::clone(self.composefs.get_or_init(|| composefs));
+        Ok(r)
```
bshephar, line 178: I noted your response to Gemini below, you mentioned you only expect this to be called once and don't necessarily require thread safety here. I'm just curious about your use of Arc over rc in that case? Since Arc is slightly more computationally expensive, I'm curious to understand your thought process. 

There's every possibility that my lack of Rust experience is causing me to overlook something. In which case, consider this a question for my own educational purposes rather than a functional critique of your PR per se.  :)

crates/lib/src/store/mod.rs
cgwalters: Yes, it just says whether or not the directory existed already which is information we don't need.

crates/lib/src/store/mod.rs
cgwalters: The underlying API wants Arc https://github.com/containers/composefs-rs/blob/126751dee8a71aa54c9666e69645d2fd1eccb176/crates/composefs/src/repository.rs#L112

Ultimately we definitely don't care about the refcounting overhead here, and in general Arc is needed for things like tokio with work stealing.

cgwalters: >  In which case, consider this a question for my own educational purposes rather than a functional critique of your PR per se. :)

Yes that's fine! Thanks for looking at the PR!

bshephar: Ahh, that makes sense then. Thanks for the clarification.

crates/lib/src/store/mod.rs
bshephar: Yeah, that makes sense. It would never make it to the next line if there was an error returned from ensure_dir_with(). There's no logical need to store and reference the returned value.

Issue #1425, opened 2025-07-17
Clean up kernel argument parsing


cgwalters:
ref https://github.com/bootc-dev/bootc/pull/1314#discussion_r2214204846_

Basically while we can use OstreeKernelArgs here, it'd probably be good to add a new native rust implementation in general.

We already have `kernel.rs` so let's add this stuff there.

bshephar:
I was reading the comment that you linked on that thread, and you mentioned that we should use `to_utf8_lossy`, I assume you mean `from_utf8_lossy` (https://doc.rust-lang.org/std/string/struct.String.html#method.from_utf8_lossy). I just wonder whether we _would_ actually want those replacement chars in the parsed output. A possible implementation of that could look like:
https://github.com/bshephar/bootc/commit/eea1bf2cb015e2aa5378e69e70d0e2c1e1e30437

But I'm not convinced we actually _do_ want that. Maybe it would be better to just unwrap the Err from `read_to_string` if there are non-UTF8 chars in the file?

cgwalters:
I think we need to be pretty conservative with what types of errors we make fatal. For data *we* control (or is directly provided as input to us) we can reasonably impose hard constraints.

The obvious case is a malformed TOML for one of our config file, we wouldn't want to just try to stumble on with a potentially invalid config.

But it also needs to be relatively easy for users to work around in emergency situations (and for a config file they could "hot patch" it with a `usroverlay` or so).

I would really hope no one in reality has non-UTF8 in the kernel command line, and in theory one could hot patch there too with a bind mount over `/proc/cmdline`.

But the key difference is that for `/proc/cmdline` it is a big bag of data, only a small portion of which we read, so we shouldn't fail on unrelated issues.

bshephar:
Right, yeah, that makes sense. So if the non UTF8 char happened to be in data that _we_ didn't care about, then ideally we shouldn't fail. I see what you're going for. So, in that case, then yeah, we probably do want to parse the file like I proposed in that commit. Then just make sure we have validated all of the `cmdline` options that we care about.

That would allow us to be more deliberate about our specific failure scenarios.

jeckersb:
I started on this briefly before I disappeared on PTO, but the general idea I ran with is that everything is just `&[u8]` because that's how the kernel parses it.  The range of allowable things is actually crazy when reading through the code.

jeckersb:
[...]

Issue #1469, opened 2025-07-31
/etc & 3 way merge vs overlayfs for composefs native integration


travier:
We currently do overlayfs for /etc in composefs-rs.

This has two major downsides:
- no official API to reset state (i.e. reset a file to the default from the image) without manually doing changes in the upper layer of overlayfs
- there is only one copy of /etc (instead of one copy of /etc per deployment in the current ostree case), thus you can not easily rollback to a previous deployment with the previous config

We touched a little bit on this in:
- https://github.com/containers/composefs-rs/issues/38
- https://github.com/bootc-dev/bootc/issues/1190

Thus we would like to implement in bootc support for an ostree-like 3 way merge for the composefs native case that mirrors how ostree works today: https://ostreedev.github.io/ostree/deployment/#contents-of-a-deployment

cgwalters:
(We discussed this on a chat call, I'm adding here what came up there)

> there is only one copy of /etc (instead of one copy of /etc per deployment in the current ostree case),

Yeah per discussion this bit is a HUGE implicit API/ABI surface of ostree systems today and is really extremely useful for cases where you have arbitrary things mutating `/etc`.

In the end we may need to bite the bullet and make this configurable.

(BTW...one thing we should definitely fix at some point is that with `etc.transient = true` we can definitely default to the overlayfs thing instead of making reflink copies)

cgwalters:
Closed by https://github.com/bootc-dev/bootc/pull/1485

PR #1476, opened 2025-08-01
Improve kernel argument parsing

jeckersb:
This adds a new `kernel::Cmdline` struct, which is populated either
via `Cmdline::from` (borrowed) or `Cmdline::from_proc` (owned).

This attempts to follow the same behavior as the kernel, which is
mostly covered in:

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/lib/cmdline.c?id=e8d780dcd957d80725ad5dd00bab53b856429bc0#n227

The algorithm is basically:
- Scan forward until you find the first unquoted isspace()
  byte. That's the end of the param.
- If you encounter an `=` along the way, note where. That's where it
  will terminate the key and split for the value. Any future `=` are
  not treated as special.
- The value can be quoted to allow spaces, but is unquoted only in as
  much as `"` is removed from the first or last byte. You can still
  have `"` in the middle of the value.

This operates on `&[u8]` because the kernel does not enforce any
particular encoding for the cmdline.  Iterating using
`Cmdline::iter()` will emit the `Parameter` type, which has helper
methods `key_lossy()` and `value_lossy()` to convert
potentially-non-UTF8 data into `String`s.

Resolves: #1425

### DIFF
--- a/crates/lib/src/kernel.rs
+++ b/crates/lib/src/kernel.rs
@@ -1,40 +1,132 @@
+use std::borrow::Cow;
+
 use anyhow::Result;
-use fn_error_context::context;
 
 /// This is used by dracut.
-pub(crate) const INITRD_ARG_PREFIX: &str = "rd.";
+pub(crate) const INITRD_ARG_PREFIX: &[u8] = b"rd.";
 /// The kernel argument for configuring the rootfs flags.
-pub(crate) const ROOTFLAGS: &str = "rootflags=";
+pub(crate) const ROOTFLAGS: &[u8] = b"rootflags";
 
-/// Parse the kernel command line.  This is strictly
-/// speaking not a correct parser, as the Linux kernel
-/// supports quotes.  However, we don't yet need that here.
-///
-/// See systemd's code for one userspace parser.
-#[context("Reading /proc/cmdline")]
-pub(crate) fn parse_cmdline() -> Result<Vec<String>> {
-    let cmdline = std::fs::read_to_string("/proc/cmdline")?;
-    let r = cmdline
-        .split_ascii_whitespace()
-        .map(ToOwned::to_owned)
-        .collect();
-    Ok(r)
+pub(crate) struct Cmdline<'a>(Cow<'a, [u8]>);
+
+impl<'a, T: AsRef<[u8]> + ?Sized> From<&'a T> for Cmdline<'a> {
+    fn from(input: &'a T) -> Self {
+        Self(Cow::Borrowed(input.as_ref()))
+    }
 }
 
-/// Return the value for the string in the vector which has the form target_key=value
-pub(crate) fn find_first_cmdline_arg<'a>(
-    args: impl Iterator<Item = &'a str>,
-    target_key: &str,
-) -> Option<&'a str> {
-    args.filter_map(|arg| {
-        if let Some((k, v)) = arg.split_once('=') {
-            if target_key == k {
-                return Some(v);
+impl<'a> Cmdline<'a> {
+    pub fn from_proc() -> Result<Self> {
+        Ok(Self(Cow::Owned(std::fs::read("/proc/cmdline")?)))
+    }
+
+    pub fn iter(&'a self) -> impl Iterator<Item = Parameter<'a>> {
+        let mut in_quotes = false;
+
+        self.0
+            .split(move |c| {
+                if *c == b'"' {
+                    in_quotes = !in_quotes;
+                }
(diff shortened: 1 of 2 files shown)


crates/lib/src/kernel.rs
```diff
@@ -1,51 +1,269 @@
...
+            let value = self.value_lossy();
+
+            if value.chars().any(|c| c.is_ascii_whitespace()) {
+                write!(f, "{key}=\"{value}\"")
+            } else {
+                write!(f, "{key}={value}")
+            }
+        } else {
+            write!(f, "{key}")
+        }
```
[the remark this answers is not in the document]
jeckersb: :-1: keep lossy conversion in the lossy functions.  Also this is just wrong, if the value is `None` we don't call `value_lossy()`.

crates/lib/src/kernel.rs
```diff
@@ -1,51 +1,269 @@
...
-    args: impl Iterator<Item = &'a str>,
-    target_key: &str,
-) -> Option<&'a str> {
-    args.filter_map(|arg| {
-        if let Some((k, v)) = arg.split_once('=') {
-            if target_key == k {
-                return Some(v);
+pub(crate) const ROOTFLAGS: &[u8] = b"rootflags";
+
+const DASH: u8 = 0x2d;
```
cgwalters: I'm usually a fan of constants but can't we just use e.g. `b'-'`? (I mean at least in the constant definition itself)

crates/lib/src/kernel.rs
```diff
@@ -1,51 +1,269 @@
...
+        } else {
+            String::from("")
+        }
+    }
+}
+
+impl<'a, T: AsRef<[u8]> + ?Sized> From<&'a T> for Parameter<'a> {
+    fn from(input: &'a T) -> Self {
+        let input = input.as_ref();
+        let equals = input.iter().position(|b| *b == EQUALS);
```
cgwalters: I think it's a lot cleaner to use `split_once` for this. There are several other examples in the code.

crates/lib/src/kernel.rs
```diff
@@ -1,51 +1,269 @@
...
+                let (key, mut value) = input.split_at(i);
+
+                // skip `=`, we know it's the first byte because we
+                // found it above
+                value = &value[1..];
+
+                // *Only* the first and last double quotes are stripped
+                if let Some(&DOUBLE_QUOTE) = value.first() {
+                    value = &value[1..];
+                }
```
cgwalters: `value = value.strip_prefix(b'"').unwrap_or(value).strip_suffix(b'"').unwrap_or(value)`
or so

cgwalters (commented):
Looks sane just some style nits. Thanks for working on this!

crates/lib/src/kernel.rs
jeckersb: :facepalm: ... yes, thank you.  Somehow I didn't realize that `b''` was a thing, and I was annoyed that `b""` was always a slice.

jeckersb: I'll just remove the constants and put these inline where they go.

crates/lib/src/kernel.rs
jeckersb: `split_once` is nightly-only for slices though :frowning_face: 

https://doc.rust-lang.org/std/primitive.slice.html#method.split_once

cgwalters:
Not a blocker but something to discuss I think raised by this that we will want to address is that we still have overlap/duplication with at least the kargs parsing in composefs-boot. So one option is to move this there (and also make it `pub`). The other possibility (which I keep thinking more about) is a "friendly" fork of composefs-boot that lives in this repo; I think we may end up needing that really for things like https://github.com/bootc-dev/bootc/issues/1469 among others.

crates/lib/src/kernel.rs
cgwalters: Ah yeah, annoying.

You know I think the bigger picture thing here is that almost everyone doing parsing of `u8` is using a higher level library and so there's not much motivation to improve std here. There's a lot of those.

One most relevant for this use case is `BStr` which has https://docs.rs/bstr/latest/bstr/trait.ByteSlice.html#method.split_once_str

bstr is a pretty good fit for this, it's not in our depchain today but that's just because we don't use `regex` among other things.

OK as is but perhaps worth a comment like
```
// https://doc.rust-lang.org/std/primitive.slice.html#method.split_once is nightly only right now
// We might consider using bstr in the future too
```

or so?

cgwalters (approved):

PR #1477, opened 2025-08-01
Add bootc internals cfs

cgwalters:
---

This exposes the current functionality of the cfsctl binary.
It's not a crate right now, and it's not a lot of code, so we just
fork it.

I did take the effort to use `git subtree merge` to do the import.
For the record, here's how I did it:

- In composefs-rs: git subtree split --prefix=crates/cfsctl
- In bootc: git subtree add --prefix=crates/lib/cfsctl ../../containers/composefs-rs/ <git sha1 from above>

In cfsctl I also:

- Adjusted it to accept the bootc-configured composefs repo
  (which note is right now hardcoded to sha512, not sha256)
- Dropped the http stuff since I don't think it really makes
  sense vs OCI


jeckersb (approved):
Went over this with @cgwalters on a call :+1:

PR #1482, opened 2025-08-04
Composefs backend: Various build/CI fixes

cgwalters:
I only compile tested my rebase by default, this ensures `make validate-rust` passes which should let the rest of CI run.


cgwalters:
I messed up and this isn't actually on current tip main, will fix

cgwalters:
I just pushed this stuff to the branch for now, but feel free to post-push review here

Issue #1360, opened 2025-06-09
RFE: Migrate `/etc` diversions to a private directory instead of `/usr/etc`


Conan-Kudo:
The current state of using `/usr/etc` for diverted pristine copies of `/etc` content is a problem.

1. [openSUSE uses `/usr/etc`](https://en.opensuse.org/openSUSE:Packaging_UsrEtc) which makes bootc fully incompatible with openSUSE.
2. Work done by SUSE to use `/usr/etc` for the vendor layer of configuration is being upstreamed, and that software is incompatible with bootc.

Could we please move this to somewhere else? Perhaps something like `/usr/lib/bootc/pristine-etc` or similar?

Conan-Kudo:
This would also help with issues like #942 as well.

cgwalters:
This comes from ostree today and is probably best tracked there by default. That said the implementation of this would probably change by default as part of https://github.com/bootc-dev/bootc/issues/1190 (ref https://github.com/containers/composefs-rs/blob/6884b989eb33b9400b7b96887a074653009eb6b4/crates/composefs-setup-root/src/main.rs#L263 )

One problem is though that it's possible some things have come to rely on it (even though they shouldn't). We can check today by running tests after doing `mount --bind /usr/share/empty /usr/etc` in early boot.

There's other details here like even if the composefs-rs change landed we'd need to probably at least keep `ostree admin config-diff` say working.

Conan-Kudo:
I did file this as a ticket for rpm-ostree as well: https://github.com/coreos/rpm-ostree/issues/5419

travier:
> This comes from ostree today and is probably best tracked there by default. That said the implementation of this would probably change by default as part of [#1190](https://github.com/bootc-dev/bootc/issues/1190) (ref [containers/composefs-rs@`6884b98`/crates/composefs-setup-root/src/main.rs#L263](https://github.com/containers/composefs-rs/blob/6884b989eb33b9400b7b96887a074653009eb6b4/crates/composefs-setup-root/src/main.rs#L263) )

Agree that it's probably the right place and time to change this. We are looking at this right now and the current composefs-rs implementation does not do any special treatment here so `/etc` stays "in place" when using the composefs native backend.

We can implement the `/etc` three way merge in a way that does not rely on it.

> One problem is though that it's possible some things have come to rely on it (even though they shouldn't). We can check today by running tests after doing `mount --bind /usr/share/empty /usr/etc` in early boot.

👍🏻 

[...]

PR #1485, opened 2025-08-05
Implementation for /etc merge

Johan-Liebert1:
PR for `/etc` merge

### DIFF
--- a/crates/etc-merge/src/lib.rs
+++ b/crates/etc-merge/src/lib.rs
@@ -1,0 +1,452 @@
+//! Lib for /etc merge
+
+#![allow(dead_code)]
+
+use std::io::BufReader;
+use std::path::PathBuf;
+use std::rc::Rc;
+
+use anyhow::Context;
+use cap_std_ext::cap_std;
+use cap_std_ext::cap_std::fs::{Dir as CapStdDir, MetadataExt};
+use composefs::fsverity::{FsVerityHashValue, Sha256HashValue};
+use composefs::generic_tree::{Directory, Inode, Leaf, LeafContent, Stat};
+use composefs::tree::ImageError;
+use rustix::fs::readlinkat;
+
+#[derive(Debug)]
+struct CustomMetadata {
+    content_hash: String,
+    verity: Option<String>,
+}
+
+impl CustomMetadata {
+    fn new(content_hash: String, verity: Option<String>) -> Self {
+        Self {
+            content_hash,
+            verity,
+        }
+    }
+}
+
+struct MyStat(Stat);
+
+impl From<&cap_std::fs::Metadata> for MyStat {
+    fn from(value: &cap_std::fs::Metadata) -> Self {
+        Self(Stat {
+            st_mode: value.mode(),
+            st_uid: value.uid(),
+            st_gid: value.gid(),
+            st_mtim_sec: value.mtime(),
+            xattrs: Default::default(),
+        })
+    }
+}
+
+fn stat_eq(this: &Stat, other: &Stat) -> bool {
+    if this.st_uid != other.st_uid {
+        return false;
+    }
+
+    if this.st_gid != other.st_gid {
+        return false;
+    }
+
+    if this.st_mode != other.st_mode {
+        return false;
+    }
+
+    if this.st_mtim_sec != other.st_mtim_sec {
+        return false;
+    }
+
+    if this.xattrs != other.xattrs {
+        return false;
+    }
+
+    return true;
+}
+
+#[derive(Debug)]
+struct Diff {
+    added: Vec<PathBuf>,
+    modified: Vec<PathBuf>,
+    removed: Vec<PathBuf>,
+}
+
+fn collect_all_files(root: &Directory<CustomMetadata>) -> Vec<PathBuf> {
+    fn collect(
+        root: &Directory<CustomMetadata>,
+        mut current_path: PathBuf,
+        files: &mut Vec<PathBuf>,
+    ) {
+        for (path, inode) in root.entries() {
+            current_path.push(path);
+
+            if let Inode::Directory(dir) = inode {
+                collect(dir, current_path.clone(), files);
+            } else {
+                files.push(current_path.clone());
+            }
+
+            current_path.pop();
+        }
+    }
+
+    let mut files = vec![];
+    collect(root, PathBuf::new(), &mut files);
+
+    return files;
+}
+
+fn get_deletions(
+    pristine: &Directory<CustomMetadata>,
+    current: &Directory<CustomMetadata>,
+    mut current_path: PathBuf,
+    diff: &mut Diff,
+) -> anyhow::Result<()> {
+    for (file_name, inode) in pristine.entries() {
+        current_path.push(file_name);
+
+        match inode {
+            Inode::Directory(pristine_dir) => {
+                match current.get_directory(file_name) {
+                    Ok(curr_dir) => {
+                        get_deletions(pristine_dir, curr_dir, current_path.clone(), diff)?
+                    }
+
+                    Err(ImageError::NotFound(..)) => {
+                        // Directory was deleted
+                        diff.removed.push(current_path.clone());
+                    }
+
+                    Err(e) => Err(e)?,
+                }
+            }
+
+            Inode::Leaf(..) => match current.ref_leaf(file_name) {
+                Ok(..) => {
+                    // Empty as all additions/modifications are tracked above
+                }
+
+                Err(ImageError::NotFound(..)) => {
+                    // File was deleted
+                    diff.removed.push(current_path.clone());
+                }
+
+                Err(e) => Err(e)?,
+            },
+        }
+
+        current_path.pop();
+    }
+
+    Ok(())
+}
+
+// 1. Files in the currently booted deployment’s /etc which were modified from the default /usr/etc (of the same deployment) are retained.
+//
+// 2. Files in the currently booted deployment’s /etc which were not modified from the default /usr/etc (of the same deployment)
+// are upgraded to the new defaults from the new deployment’s /usr/etc.
+
+// Modifications
+// 1. File deleted from new /etc
+// 2. File added in new /etc
+//
+// 3. File modified in new /etc
+//    a. Content added/deleted
+//    b. Permissions/ownership changed
+//    c. Was a file but changed to directory/symlink etc or vice versa
+//    d. xattrs changed - we don't include this right now
+fn compute_diff_root(
+    pristine: &mut Directory<CustomMetadata>,
+    current: &Directory<CustomMetadata>,
+    mut current_path: PathBuf,
+    diff: &mut Diff,
+) -> anyhow::Result<()> {
+    for (path, inode) in current.entries() {
+        current_path.push(path);
+
+        match inode {
+            Inode::Directory(curr_dir) => {
+                match pristine.get_directory_mut(path) {
+                    Ok(old_dir) => {
+                        compute_diff_root(old_dir, &curr_dir, current_path.clone(), diff)?
+                    }
+
+                    Err(ImageError::NotFound(..)) => {
+                        // Dir not found in original /etc, dir was added
+                        diff.added.push(current_path.clone());
+
+                        // Also add every file inside that dir
+                        diff.added.extend(collect_all_files(&curr_dir));
+                    }
+
+                    Err(ImageError::NotADirectory(..)) => {
+                        // A file was changed to a directory
+                        diff.modified.push(current_path.clone());
+                    }
+
+                    Err(e) => Err(e)?,
+                }
+            }
+
+            Inode::Leaf(leaf) => match pristine.ref_leaf(path) {
+                Ok(old_leaf) => {
+                    let LeafContent::Regular(current_meta) = &leaf.content else {
+                        unreachable!("File types do not match");
+                    };
+
+                    let LeafContent::Regular(old_meta) = &old_leaf.content else {
+                        unreachable!("File types do not match");
+                    };
+
+                    if old_meta.content_hash != current_meta.content_hash
+                        || !stat_eq(&old_leaf.stat, &leaf.stat)
+                    {
+                        // File modified in some way
+                        diff.modified.push(current_path.clone());
+                    }
+
+                    pristine.remove(path);
+                }
+
+                Err(ImageError::IsADirectory(..)) => {
+                    // A directory was changed to a file
+                    diff.modified.push(current_path.clone());
+                }
+
+                Err(ImageError::NotFound(..)) => {
+                    // File not found in original /etc, file was added
+                    diff.added.push(current_path.clone());
+                }
+
+                Err(e) => Err(e)?,
+            },
+        }
+
+        current_path.pop();
+    }
+
+    Ok(())
+}
+
+fn compute_diff(
+    pristine_etc: &CapStdDir,
+    current_etc: &CapStdDir,
+    new_etc: &CapStdDir,
+) -> anyhow::Result<Diff> {
+    let mut pristine_etc_files = Directory::default();
+    recurse_dir(pristine_etc, &mut pristine_etc_files)
+        .context(format!("Recursing {pristine_etc:?}"))?;
+
+    let mut current_etc_files = Directory::default();
+    recurse_dir(current_etc, &mut current_etc_files)
+        .context(format!("Recursing {current_etc:?}"))?;
+
+    let mut new_etc_files = Directory::default();
+    recurse_dir(new_etc, &mut new_etc_files).context(format!("Recursing {new_etc:?}"))?;
+
+    let mut diff = Diff {
+        added: vec![],
+        modified: vec![],
+        removed: vec![],
+    };
+
+    compute_diff_root(
+        &mut pristine_etc_files,
+        &current_etc_files,
+        PathBuf::new(),
+        &mut diff,
+    )?;
+
+    get_deletions(
+        &pristine_etc_files,
+        &current_etc_files,
+        PathBuf::new(),
+        &mut diff,
+    )?;
+
+    Ok(diff)
+}
+
+fn recurse_dir(dir: &CapStdDir, root: &mut Directory<CustomMetadata>) -> anyhow::Result<()> {
+    for entry in dir.entries()? {
+        let entry = entry.context(format!("Getting entry"))?;
+        let entry_name = entry.file_name();
+
+        let entry_type = entry.file_type()?;
+        let entry_meta = entry
+            .metadata()
+            .context(format!("Getting metadata for {entry_name:?}"))?;
+
+        if entry_type.is_dir() {
+            let dir = dir
+                .open_dir(&entry_name)
+                .with_context(|| format!("Opening dir {entry_name:?} inside {dir:?}"))?;
+
+            let mut directory = Directory::default();
+
+            recurse_dir(&dir, &mut directory)?;
+
+            root.insert(&entry_name, Inode::Directory(Box::new(directory)));
+
+            continue;
+        }
+
+        if !(entry_type.is_symlink() || entry_type.is_file()) {
+            // We cannot read any other device like socket, pipe, fifo.
+            // We shouldn't really find these in /etc in the first place
+            tracing::debug!("Ignoring non-regular/non-symlink file: {:?}", entry_name);
+            continue;
+        }
+
+        // TODO: Another generic here but constrained to Sha256HashValue
+        // Regarding this, we'll definitely get DigestMismatch error if SHA512 is being used
+        let measured_verity =
+            composefs::fsverity::measure_verity_opt::<Sha256HashValue>(entry.open()?)?;
+
+        if let Some(measured_verity) = measured_verity {
+            root.insert(
+                &entry_name,
+                Inode::Leaf(Rc::new(Leaf {
+                    stat: MyStat::from(&entry_meta).0,
+                    content: LeafContent::Regular(CustomMetadata::new(
+                        "".into(),
+                        Some(measured_verity.to_hex()),
+                    )),
+                })),
+            );
+
+            // file has fs-verity enabled. We don't need to check the content/metadata
+            continue;
+        }
+
+        let mut hasher = openssl::hash::Hasher::new(openssl::hash::MessageDigest::sha256())?;
+
+        if entry_type.is_symlink() {
+            let readlinkat_result = readlinkat(&dir, &entry_name, vec![])
+                .context(format!("readlinkat {entry_name:?}"))?;
+
+            hasher.update(readlinkat_result.as_bytes())?;
+        } else if entry_type.is_file() {
+            let file = entry
+                .open()
+                .context(format!("Opening entry {entry_name:?}"))?;
+
+            let mut reader = BufReader::new(file);
+            std::io::copy(&mut reader, &mut hasher)?;
+        };
+
+        let content_digest = hex::encode(hasher.finish()?);
+
+        root.insert(
+            &entry_name,
+            Inode::Leaf(Rc::new(Leaf {
+                stat: MyStat::from(&entry_meta).0,
+                content: LeafContent::Regular(CustomMetadata::new(content_digest, None)),
+            })),
+        );
+    }
+
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use cap_std::fs::PermissionsExt;
+
+    use super::*;
+
+    const FILES: &[(&str, &str)] = &[
+        ("a/file1", "a-file1"),
+        ("a/file2", "a-file2"),
+        ("a/b/file1", "ab-file1"),
+        ("a/b/file2", "ab-file2"),
+        ("a/b/c/fileabc", "abc-file1"),
+        ("a/b/c/modify-perms", "modify-perms"),
+        ("a/b/c/to-be-removed", "remove this"),
+        ("to-be-removed", "remove this 2"),
+    ];
+
+    #[test]
+    fn test_etc_diff() -> anyhow::Result<()> {
+        let tempdir = cap_std_ext::cap_tempfile::tempdir(cap_std::ambient_authority())?;
+
+        tempdir.create_dir("pristine_etc")?;
+        tempdir.create_dir("current_etc")?;
+        tempdir.create_dir("new_etc")?;
+
+        let p = tempdir.open_dir("pristine_etc")?;
+        let c = tempdir.open_dir("current_etc")?;
+        let n = tempdir.open_dir("new_etc")?;
+
+        p.create_dir_all("a/b/c")?;
+        c.create_dir_all("a/b/c")?;
+
+        for (file, content) in FILES {
+            p.write(file, content.as_bytes())?;
+            c.write(file, content.as_bytes())?;
+        }
+
+        let new_files = ["new_file", "a/new_file", "a/b/c/new_file"];
+
+        // Add some new files
+        for file in new_files {
+            c.write(file, b"hello")?;
+        }
+
+        let overwritten_files = [FILES[1].0, FILES[4].0];
+        let perm_changed_files = [FILES[5].0];
+
+        // Modify some files
+        c.write(overwritten_files[0], b"some new content")?;
+        c.write(overwritten_files[1], b"some newer content")?;
+
+        // Modify permissions
+        let file = c.open(perm_changed_files[0])?;
+        // This should be enough as the usual files have permission 644
+        file.set_permissions(cap_std::fs::Permissions::from_mode(0o400))?;
+
+        // Remove some files
+        let deleted_files = [FILES[6].0, FILES[7].0];
+        c.remove_file(deleted_files[0])?;
+        c.remove_file(deleted_files[1])?;
+
+        let res = compute_diff(&p, &c, &n)?;
+
+        // Test added files
+        assert_eq!(res.added.len(), new_files.len());
+        assert!(res.added.iter().all(|file| {
+            new_files
+                .iter()
+                .find(|x| PathBuf::from(*x) == *file)
+                .is_some()
+        }));
+
+        // Test modified files
+        let all_modified_files = overwritten_files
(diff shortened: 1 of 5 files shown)


crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,281 @@
...
+    recurse_dir(pristine_etc, PathBuf::new(), &mut pristine_etc_files)
+        .context(format!("Recursing {pristine_etc:?}"))?;
+
+    let mut current_etc_files = BTreeMap::new();
+    recurse_dir(current_etc, PathBuf::new(), &mut current_etc_files)
+        .context(format!("Recursing {current_etc:?}"))?;
+
+    let mut new_etc_files = BTreeMap::new();
+    recurse_dir(new_etc, PathBuf::new(), &mut new_etc_files)
+        .context(format!("Recursing {new_etc:?}"))?;
```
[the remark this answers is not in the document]
Johan-Liebert1: Not used yet

Johan-Liebert1:
I've tested the output of this implementation with `ostree admin config-diff` and it seems correct. But again, I haven't tested this exhaustively

crates/etc-merge/src/lib.rs
p5, line 1: General question, not something specific about this PR (this looks great BTW!)

---

Would Bootc's ostree backend be refactored to make use of this crate in time, or is the plan for ostree to remain mostly separate?   Obviously the implementation would be completely separate PR to this anyway.

cgwalters: This is a good question indeed. I've been really tempted to basically vendor+rewrite some of the ostree-sysroot stuff in this project. However, it has a ton of implications.

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+
+            let mut file = entry.open().context(format!("Opening entry {path:?}"))?;
+            file.read_exact(&mut buf)
+                .context(format!("Reading {path:?}. Buf: {buf:?}"))?;
+
+            buf
+        } else {
+            // We cannot read any other device like socket, pipe, fifo.
+            // We shouldn't really find these in /etc in the first place
+            unimplemented!("Found file of type {:?}", entry_type)
```
cgwalters: We shouldn't just crash the process. See https://github.com/ostreedev/ostree/pull/3143

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+    let mut modified = vec![];
+
+    for (current_file, current_meta) in current_etc_files {
+        let Some(old_meta) = pristine_etc_files.get(&current_file) else {
+            // File was created
+            added.push(current_file);
+            continue;
+        };
+
+        if old_meta.metadata_hash != current_meta.metadata_hash
```
cgwalters: What's the value here in hashing the metadata vs just comparing it directly?

BTW on this topic, we could instead use https://github.com/containers/composefs-rs/blob/main/crates/composefs/src/generic_tree.rs to hold the generated metadata trees.

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+            let readlinkat_result = readlinkat(&dir, &entry_name, vec![])
+                .context(format!("readlinkat {entry_name:?}"))?;
+
+            readlinkat_result.into()
+        } else if entry_type.is_file() {
+            let mut buf = vec![0u8; entry.metadata()?.size() as usize];
+
+            let mut file = entry.open().context(format!("Opening entry {path:?}"))?;
+            file.read_exact(&mut buf)
+                .context(format!("Reading {path:?}. Buf: {buf:?}"))?;
```
cgwalters: Two key things here. First I think we should incremental hashing instead of reading the whole thing into memory and then hashing the memory buffer.

But second a nice optimization to make here is handling the case where the file has fsverity enabled. Then we can efficiently just query its fsverity hash to do a diff.

It won't be the *default* for files in a mutable `/etc` to have fsverity enabled but it makes sense to do so (dir is mutable, so the write-tmpfile-then-rename pattern continues to work) and we should make things faster for those who do use it.

cgwalters (changes requested):
Thanks for working on this!

crates/etc-merge/src/lib.rs
Johan-Liebert1: Makes sense. Had left it unimplemented because I was unsure how we handled this. Updated

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,343 @@
...
+            loop {
+                let bytes_read = reader
+                    .read(&mut buf)
+                    .context(format!("Reading chunk from {path:?}"))?;
+
+                if bytes_read == 0 {
+                    break;
+                }
+
+                hasher.update(&buf[..bytes_read])?;
```
cgwalters: This can just be
`std::io::copy(&mut hasher, &mut reader)`

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,343 @@
...
+            );
+
+            recurse_dir(&dir, path.clone(), list).context(format!("Recursing {path:?}"))?;
+
+            path.pop();
+            continue;
+        }
+
+        // TODO: Another generic here but constrained to Sha256HashValue
+        // Regarding this, we'll definitely get DigestMismatch error if SHA512 is being used
```
cgwalters, line 483: Ah yeah true...in theory we could try both hashes?

crates/etc-merge/src/lib.rs
Johan-Liebert1: We can definitely go with comparing the metadata directly. In fact that would be better now that I think about it as it would be (maybe a little bit) faster than computing the hash. 

We could use https://github.com/containers/composefs-rs/blob/main/crates/composefs/src/generic_tree.rs for the entire map I guess. It doesn't have a generic param for symlinks, which is okay I guess as we could just compare `readlink`'s contents instead of the hash. OTOH adding more fields might be a challenge

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,343 @@
...
+                if bytes_read == 0 {
+                    break;
+                }
+
+                hasher.update(&buf[..bytes_read])?;
+            }
+        } else {
+            // We cannot read any other device like socket, pipe, fifo.
+            // We shouldn't really find these in /etc in the first place
+            tracing::info!("Ignoring non-regular/non-symlink file: {:?}", path);
```
cgwalters: Let's make this `debug!`

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,343 @@
...
+
+        if let Some(measured_verity) = measured_verity {
+            list.insert(
+                path.clone(),
+                Metadata::new("".into(), "".into(), Some(measured_verity.to_hex())),
+            );
+
+            path.pop();
+
+            // file has fs-verity enabled. We don't need to check the content/metadata
```
cgwalters: Note that fsverity only covers file contents, not metadata. It's totally allowed to do e.g. `chmod` or `chown` on such a file...which we probably want to have tracked as modified right?

Eh sorry...I guess maybe this fsverity suggestion is adding too much complexity at the moment for a non-default corner case.

Let's I guess just leave it as a `// TODO consider also using fsverity here if available`?

crates/etc-merge/src/lib.rs
```diff
@@ -3,33 +3,69 @@
...
+            st_mode: value.mode(),
+            st_uid: value.uid(),
+            st_gid: value.gid(),
+            st_mtim_sec: value.mtime(),
+            xattrs: Default::default(),
+        })
+    }
+}
+
+fn stat_eq(this: &Stat, other: &Stat) -> bool {
```
cgwalters: Let's just `#[derive(PartialEq, Eq)]` in composefs-rs right?

Johan-Liebert1: We wouldn't want to compare `mtime` of a file or directory right? I see that's the way ostree does things. `#derive` will also end up comparing the mtimes, and having a custom impl but just ignoring mtime feels a little weird to me. I did end up implementing `Eq` on `MyStat` in a previous iteration but that required cloning the `Stat` which should be okay as it's not a large struct. What do you think?

cgwalters: Right, OK. But we should then call it `eq_ignore_mtime` or so and also clearly document that.

Johan-Liebert1:
Hmm... tests failed because we can't chown unless we're root. And some issue with `getxattr` arguments

Cargo.toml
```diff
@@ -36,9 +36,9 @@ anyhow = "1.0.82"
 camino = "1.1.6"
 canon-json = "0.2.1"
 cap-std-ext = "4.0.3"
-composefs = { git = "https://github.com/containers/composefs-rs", rev = "28d4721f77f973f0e394d60d6a69d9b39cb38d7f", package = "composefs", features = ["rhel9"] }
-composefs-boot = { git = "https://github.com/containers/composefs-rs", rev = "28d4721f77f973f0e394d60d6a69d9b39cb38d7f", package = "composefs-boot" }
-composefs-oci = { git = "https://github.com/containers/composefs-rs", rev = "28d4721f77f973f0e394d60d6a69d9b39cb38d7f", package = "composefs-oci" }
+composefs = { git = "https://github.com/containers/composefs-rs", rev = "8d1e570275621531d39b8f01681584bcc85ce01c", package = "composefs", features = ["rhel9"] }
```
cgwalters: In order to reduce conflicts let's do separate PRs for bumping crate dependencies.

cgwalters:
> Hmm... tests failed because we can't chown unless we're root. And some issue with getxattr arguments

Yeah, some of this stuff really needs privileged integration tests, not unit tests.

In the short term what I think is easiest is to add an option to the core logic to ignore uid/gid + xattrs and that's what we do in unit tests.

For integration tests what we have in some other places is exposing the core functionality via `bootc internals` and then the tests can drive it, see e.g.
https://github.com/bootc-dev/bootc/blob/8ed1134e9bdf35a6f1b19ba071c481c91f9eb1f4/crates/lib/src/cli.rs#L1237
(actually that's not so good an example because the command hardcodes the test, which bloats production binaries)
This one is maybe a bit better:
https://github.com/bootc-dev/bootc/blob/8ed1134e9bdf35a6f1b19ba071c481c91f9eb1f4/crates/lib/src/cli.rs#L1291 

There's the tests-integration crate which is also intended to hold stuff like this.

Cargo.toml
cgwalters: Done in https://github.com/bootc-dev/bootc/pull/1503

Johan-Liebert1: Okay

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,1037 @@
...
+        let _ = writeln!(stdout, "{} {modified:?}", ModificationType::Modified.cyan());
+    }
+
+    for removed in &diff.removed {
+        let _ = writeln!(stdout, "{} {removed:?}", ModificationType::Removed.red());
+    }
+}
+
+#[context("Collecting xattrs")]
+fn collect_xattrs(etc_fd: &CapStdDir, rel_path: &OsString) -> anyhow::Result<Xattrs> {
```
cgwalters: This one seems like it'd be good to put in its own module with some unit tests etc.

Also prefer `impl AsRef<Path>` over `&OsString`

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,1037 @@
...
+    let link = format!("/proc/self/fd/{}", etc_fd.as_fd().as_raw_fd());
+    let path = Path::new(&link).join(rel_path);
+
+    const DEFAULT_SIZE: usize = 128;
+
+    // Start with a guess for size
+    let mut buf: Vec<u8> = vec![0; DEFAULT_SIZE];
```
cgwalters: It's confusing we have two things called `buf` - how about `xattr_names_buf`

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,1037 @@
...
+        buf = vec![0; size];
+        llistxattr(&path, &mut buf).context("llistxattr")?;
+    }
+
+    let xattrs: Xattrs = RefCell::new(BTreeMap::new());
+
+    for name_buf in buf[..size].split(|&b| b == 0).filter(|x| !x.is_empty()) {
+        let name = OsStr::from_bytes(name_buf);
+
+        let mut buf = vec![0; DEFAULT_SIZE];
```
cgwalters: Then this would be `xattr_values_buf`

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,1037 @@
...
+        let size = lgetxattr(&path, name_buf, &mut buf).context("lgetxattr")?;
+
+        if size > DEFAULT_SIZE {
+            buf = vec![0; size];
```
cgwalters: Dropping and recreating is inefficient, use `buf.resize(size, 0u8)`

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,1037 @@
...
+
+        match res {
+            Ok(_) => {}
+            Err(e) if e.kind() == std::io::ErrorKind::NotFound => {}
+            Err(e) => Err(e).context(format!("Removing {dir_name:?}"))?,
+        };
+    }
+
+    new_etc_fd
+        .create_dir_all(&dir_name)
```
cgwalters: I think we shouldn't `create_dir_all` as we want to be sure to precisely recreate permissions on each dir

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,1037 @@
...
+    dir_name: &PathBuf,
+    stat: &Stat,
+    remove_existing: bool,
+) -> anyhow::Result<()> {
+    if remove_existing {
+        let res = new_etc_fd.remove_all_optional(&dir_name);
+
+        match res {
+            Ok(_) => {}
+            Err(e) if e.kind() == std::io::ErrorKind::NotFound => {}
```
cgwalters: That can't happen, the `_optional` handles it

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,1037 @@
...
+            Err(e) => Err(e).context(format!("Removing {dir_name:?}"))?,
+        };
+    }
+
+    new_etc_fd
+        .create_dir_all(&dir_name)
+        .context(format!("Failed to create dir {dir_name:?}"))?;
+
+    new_etc_fd
+        .set_permissions(&dir_name, Permissions::from_mode(stat.st_mode))
```
cgwalters, line 586: This is ok as is, but since we need to use raw unix stuff we might as well just directly use `rustix::fs::fchmodat`

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,1037 @@
...
+                anyhow::bail!("Symlink {file:?} coverted to dir");
+            };
+
+            // A new file with the same path exists, we delete this
+            new_etc_fd
+                .remove_all_optional(&file)
+                .context(format!("Deleting {file:?}"))?;
+
+            new_etc_fd
+                .symlink(PathBuf::from(os_str), &file)
```
cgwalters: shouldn't need to allocate here

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,1037 @@
...
+                file,
+                Some(Uid::from_raw(leaf.stat.st_uid)),
+                Some(Gid::from_raw(leaf.stat.st_gid)),
+                AtFlags::SYMLINK_NOFOLLOW,
+            )
+            .context(format!("chown {file:?}"))?;
+
+            copy_xattrs(&leaf.stat.xattrs, new_etc_fd, file)?;
+
+            "file"
```
cgwalters: Wait why are we returning this string? It seems unused

If we need a file type enumeration there are many choices for that like https://doc.rust-lang.org/std/fs/struct.FileType.html

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,1037 @@
...
+        current_path.push(file_name);
+
+        match inode {
+            Inode::Directory(pristine_dir) => {
+                match current.get_directory(file_name) {
+                    Ok(curr_dir) => {
+                        get_deletions(pristine_dir, curr_dir, current_path.clone(), diff)?
+                    }
+
+                    Err(ImageError::NotFound(..)) => {
```
cgwalters, line 134: Here and elsewhere, there's `get_directory_opt` which is intended for exactly this use case

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,1037 @@
...
+                        }
+
+                        get_modifications(old_dir, &curr_dir, current_path.clone(), diff)?
+                    }
+
+                    Err(ImageError::NotFound(..)) => {
+                        // Dir not found in original /etc, dir was added
+                        diff.added.push(current_path.clone());
+
+                        // Also add every file inside that dir
```
cgwalters, line 213: I wonder if it'd be cleaner to have `added` actually be a borrowed reference to the inode for the current directory.

Then it's implicit that leaf files are, well, leaves - and directories are recursive.

cgwalters:
Also I think you want to rebase on main right?

crates/etc-merge/src/lib.rs
Johan-Liebert1: ah right. I had used the non optional version first but forgot to remove this arm after changing

crates/etc-merge/src/lib.rs
Johan-Liebert1: This is something I used to debug the type of entity being created. Removing it

Johan-Liebert1:
> Also I think you want to rebase on main right?

Do we want to merge this directly to main? Kind of makes sense as this is entirely disjoint

cgwalters:
Yes please let's keep trying to merge independent things to main; it's OK to just mark them as `#[dead_code]` for now

Johan-Liebert1:
Okay, sounds good

crates/etc-merge/src/lib.rs
Johan-Liebert1: I realised I had used `get_directory` as we would still want to show a diff if a file was changed to a directory and vice versa. Given, we will fail while trying to merge, but I think it makes sense to show it as a diff?

cgwalters:
We had a live chat about this and we're currently thinking that we wouldn't have `/usr/etc` in the filesystem tree as it appears. Instead this code would mount the pristine `/etc` from the underlying composefs (which is what it's doing now).

Note this would fix https://github.com/bootc-dev/bootc/issues/1360 - but I have the same concerns as I had there in that I think `/usr/etc` is kind of an API at this point...but I'm ok to just "break" that API for composefs-native for now. If we need to re-introduce `/usr/etc` I think we could make it a thing done at image build time (by making an empty `/usr/etc` mountpoint dir and having a systemd unit to mount it)

Johan-Liebert1:
I have rebased the branch on main

Johan-Liebert1:
Some tests failed due to "connectivity issues". Assuming this is temporary, I've restarted the failed ones

cgwalters:
Any reason to keep the intermediate commits vs squashing to one? I have a slight preference for squashing in cases like this, but if you prefer it's OK by me to separate.

Things like `etc-merge: Add license to Cargo.toml` in particular just don't seem to stand on their own to me.

Also needs a rebase :surfer: to fix CI

Johan-Liebert1:
Rebased onto main and squashed commits only keeping what I thought were necessary

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,1079 @@
...
+            }
+
+            current_path.pop();
+        }
+    }
+
+    let mut files = vec![];
+    collect(root, current_path, &mut files);
+
+    return files;
```
cgwalters: This is just `files`

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,1079 @@
...
+                        get_modifications(old_dir, &curr_dir, current_path.clone(), diff)?
+                    }
+
+                    Err(ImageError::NotFound(..)) => {
+                        // Dir not found in original /etc, dir was added
+                        diff.added.push(current_path.clone());
+
+                        // Also add every file inside that dir
+                        diff.added
+                            .extend(collect_all_files(&curr_dir, current_path.clone()));
```
cgwalters: It doesn't matter really but here we're allocating a vec and then copying its contents into another; it'd be more efficient for `collect_all_files` to directly push to `diff.added`

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,1079 @@
...
+    for name_buf in xattrs_name_buf[..size]
+        .split(|&b| b == 0)
+        .filter(|x| !x.is_empty())
+    {
+        let name = OsStr::from_bytes(name_buf);
+
+        let mut xattrs_value_buf = vec![0; DEFAULT_SIZE];
+        let size = lgetxattr(&path, name_buf, &mut xattrs_value_buf).context("lgetxattr")?;
+
+        if size > DEFAULT_SIZE {
```
cgwalters: Also nit: `if size > xattrs_value_buf.capacity()`

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,1079 @@
...
+    let link = format!("/proc/self/fd/{}", etc_fd.as_fd().as_raw_fd());
+    let path = Path::new(&link).join(rel_path);
+
+    const DEFAULT_SIZE: usize = 128;
+
+    // Start with a guess for size
+    let mut xattrs_name_buf: Vec<u8> = vec![0; DEFAULT_SIZE];
+    let size = llistxattr(&path, &mut xattrs_name_buf).context("llistxattr")?;
+
+    if size > DEFAULT_SIZE {
```
cgwalters: I'd make this `if size > xattrs_name_buf.capacity()` in order to avoid duplicating the constant.

Also, while we probably don't need to care because it's racy regardless, in theory this should actually be a `loop {}` in order to support concurrent changes.

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,1079 @@
...
+        .filter(|x| !x.is_empty())
+    {
+        let name = OsStr::from_bytes(name_buf);
+
+        let mut xattrs_value_buf = vec![0; DEFAULT_SIZE];
+        let size = lgetxattr(&path, name_buf, &mut xattrs_value_buf).context("lgetxattr")?;
+
+        if size > DEFAULT_SIZE {
+            xattrs_value_buf.resize(size, 0);
+            lgetxattr(&path, name_buf, &mut xattrs_value_buf).context("lgetxattr")?;
```
cgwalters: We should also do `size = lgetxattr(...)` here I think

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,1079 @@
...
+
+    const DEFAULT_SIZE: usize = 128;
+
+    // Start with a guess for size
+    let mut xattrs_name_buf: Vec<u8> = vec![0; DEFAULT_SIZE];
+    let size = llistxattr(&path, &mut xattrs_name_buf).context("llistxattr")?;
+
+    if size > DEFAULT_SIZE {
+        xattrs_name_buf.resize(size, 0);
+        llistxattr(&path, &mut xattrs_name_buf).context("llistxattr")?;
```
cgwalters: And related to the above, we should probably do `size = llistxattr(...)`

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,1079 @@
...
+        &current_etc_files,
+        PathBuf::new(),
+        &mut diff,
+    )?;
+
+    Ok(diff)
+}
+
+/// Prints a colorized summary of differences to standard output.
+pub fn print_diff(diff: &Diff) {
```
cgwalters: Seems a bit more elegant to have the caller pass the stream to write to; this allows easy unit testing among other things

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,1079 @@
...
+                    Err(e) => Err(e)?,
+                }
+            }
+
+            Inode::Leaf(leaf) => match pristine.ref_leaf(path) {
+                Ok(old_leaf) => {
+                    match (&old_leaf.content, &leaf.content) {
+                        (Regular(old_meta), Regular(current_meta)) => {
+                            if old_meta.content_hash != current_meta.content_hash
+                                || !stat_eq_ignore_mtime(&old_leaf.stat, &leaf.stat)
```
cgwalters: I think we could move this conditional to line 228 so it's not repeated for the regfile and symlink cases

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,1079 @@
...
+
+/// Goes through the added, modified, removed files and apply those changes to the new_etc
+/// This will overwrite, remove, modify files in new_etc
+/// Paths in `diff` are relative to `etc`
+#[context("Merging")]
+pub fn merge(
+    current_etc_fd: &CapStdDir,
+    current_etc_dirtree: &Directory<CustomMetadata>,
+    new_etc_fd: &CapStdDir,
+    new_etc_dirtree: &Directory<CustomMetadata>,
```
cgwalters, line 713: In order to do this, the calling code needs to make a copy of the new pristine `etc` already pre-populated into `new_etc_fd`.

But...while it's generally useful to have "compute diff without merging", won't the main use case be to actually just one-shot a combination of diff + merge?

IOW, how about renaming this function `merge_with` and then we have

```
pub fn merge(pristine_etc_fd: &Dir, current_etc_fd: &Dir, new_etc_fd: &Dir)
```

And inside that function we do:

- Copy pristine to new
- Compute diff
- Apply diff

This is kind of what https://github.com/ostreedev/ostree/blob/edfe02d01b78039db67c4247df6948070eae0cbe/src/libostree/ostree-sysroot-deploy.c#L511 is already doing, though https://github.com/ostreedev/ostree/blob/edfe02d01b78039db67c4247df6948070eae0cbe/src/libostree/ostree-sysroot-deploy.c#L511 does the copy beforehand.

cgwalters (approved):
Some nonblocking comments that can be done as followups too.

crates/etc-merge/src/lib.rs
Johan-Liebert1: Don't quite get it... Do you mean removing the `return`?

crates/etc-merge/src/lib.rs
Johan-Liebert1: I think it makes sense to pass `diff.added` to prevent allocating

crates/etc-merge/src/lib.rs
Johan-Liebert1: We shouldn't run into a case where the xattrs change between the two `llistxattr` calls as this would be called before shutdown/reboot right?

crates/etc-merge/src/lib.rs
Johan-Liebert1: Removed the return anyway by passing `diff.added` into the function

cgwalters: Yes in this situation `return files;` is the same as just writing `files`

(cargo clippy will point this out but we have a ton of other minor clippy lints...)

crates/etc-merge/src/lib.rs
cgwalters: Yes, agreed we shouldn't run into that case. However, it'd probably be better to not crash if it does somehow happen.

crates/etc-merge/src/lib.rs
Johan-Liebert1: I had the diff function separate for testing while developing, and because ostree had a command that'd just print the diff between two etcs. I think it makes sense to combine both into a single merge function.

cgwalters: Right. Though I'm just saying we add a new high level helper, not that we need drop the split APIs.

Johan-Liebert1: I see. That makes sense. Thinking about this a bit more, I think the best place to copy files/dirs in the new_etc would be in the function `write_composefs_state`. During a fresh install there will be no merging of etcs; only on subsequent updates will the merge happen, so we'll kinda have two places where we copy the contents of etc depending upon what operation was performed.

crates/etc-merge/src/lib.rs
cgwalters: To expand on this a bit, in the corner case that the value of an xattr shrinks between the two calls, we will end up copying a truncated value which could be really confusing to debug.

crates/etc-merge/src/lib.rs
```diff
@@ -0,0 +1,1079 @@
...
+
+        lsetxattr(path, attr.as_ref(), value, XattrFlags::empty())
+            .context(format!("setxattr for {file:?}"))?;
+    }
+
+    Ok(())
+}
+
+fn recurse_dir(dir: &CapStdDir, root: &mut Directory<CustomMetadata>) -> anyhow::Result<()> {
+    for entry in dir.entries()? {
```
cgwalters, line 431: Optional/for followup: This could probably use https://docs.rs/cap-std-ext/latest/cap_std_ext/dirext/trait.CapStdExtDirExt.html#tymethod.walk (I maintain that crate, a lot of handy helpers there)

cgwalters:
Needs `cargo fmt`

crates/etc-merge/src/lib.rs
cgwalters: > During a fresh install there will be no merging of etcs; only on subsequent updates will the merge happen, so we'll kinda have two places where we copy the contents of etc depending upon what operation was performed.

Yeah, in ostree the copying always happens outside of the merge function. So I'm fine having it always be separate here too.

Johan-Liebert1:
Ran cargo fmt. Weird my editor is also set up to run cargo fmt but for some reason it sorts imports differently

PR #1487, opened 2025-08-05
Fix autovendor workflow CLI directory path

jeckersb:
Update the autovendor GitHub workflow to use the correct path 'crates/cli'
instead of 'cli', which was causing the job to fail with "No such file or
directory" error.

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude <noreply@anthropic.com>


jeckersb:
I noticed this in https://github.com/bootc-dev/bootc/actions/runs/16755360619/job/47436476996 and seemed like a simple enough first task to unleash claude on

cgwalters (approved):

PR #1491, opened 2025-08-06
Add automated PR reviewer rotation system

gursewak1997:
- Add GitHub Action workflow for automatic reviewer assignment
- Create Python script that reads reviewers from MAINTAINERS.md
- Implement 3-week sprint-based rotation cycle
- Preserve existing reviewer assignments (manual/GitHub auto)
- Add configuration for start date and rotation cycle

### DIFF
--- a/.github/scripts/assign_reviewer.py
+++ b/.github/scripts/assign_reviewer.py
@@ -1,0 +1,311 @@
+#!/usr/bin/env python3
+"""
+Auto-reviewer assignment script for GitHub PRs.
+Rotates through a list of reviewers based on a weekly cycle.
+Automatically reads reviewers from MAINTAINERS.md
+"""
+
+import os
+import sys
+import yaml
+import subprocess
+import re
+from datetime import datetime, timezone, timedelta
+from typing import List, Optional, TypedDict
+
+
+class RotationInfo(TypedDict):
+    start_date: Optional[datetime]
+    rotation_cycle_weeks: int
+    weeks_since_start: float
+    before_start: bool
+    error: Optional[str]
+
+
+class SprintInfo(TypedDict):
+    sprint_number: int
+    week_in_sprint: int
+    total_weeks: int
+    before_start: Optional[bool]
+
+
+def load_config(config_path: str) -> dict:
+    """Load the reviewer configuration from YAML file."""
+    try:
+        with open(config_path, 'r') as f:
+            return yaml.safe_load(f)
+    except FileNotFoundError:
+        print(f"Error: Configuration file {config_path} not found", file=sys.stderr)
+        sys.exit(1)
+    except yaml.YAMLError as e:
+        print(f"Error parsing YAML configuration: {e}", file=sys.stderr)
+        sys.exit(1)
+
+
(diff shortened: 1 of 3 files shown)


cgwalters:
(Probably worth having a local AI agent do the review before putting up PRs, I only sometimes remember do that)

gursewak1997:
> (Probably worth having a local AI agent do the review before putting up PRs, I only sometimes remember do that)

I do use the local agent, but it keeps coming up with suggestions even after initial feedback. I’m not sure it’s the right approach; it never seems fully convinced, and it’s a bit frustrating that it doesn’t provide all its suggestions in one go.

cgwalters:
The standard way I test Github Actions is to push to my fork's main branch; did you try that?

gursewak1997:
> The standard way I test Github Actions is to push to my fork's main branch; did you try that?

Spent some time on this only to realize that GitHub doesn’t allow requesting reviewers on PRs within your own fork; not even yourself, which makes testing a bit tricky.
Example Workflow: 
- https://github.com/gursewak1997/bootc/actions/runs/16789052288/job/47546872516 (trying to add myself)
- https://github.com/gursewak1997/bootc/actions/runs/16788150829/job/47545807264 (when trying to add Chris as reviewer)

cgwalters:
So recently I created https://github.com/bootc-dev/ci-sandbox/ which I was using to test some GHA stuff but it's intended for stuff like this, do you have write access there?

.github/workflows/auto-review.yml
```diff
@@ -0,0 +1,38 @@
+name: Auto Assign Reviewer
+
+on:
+  pull_request_target:
+    types: [opened, ready_for_review]
+
+permissions:
+  pull-requests: write
```
cgwalters, line 8: I think this is OK as is, though Chris added a Github app for the releases so that's another option for these kinds of workflows too.

.github/scripts/assign_reviewer.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+def get_existing_reviewers(pr_number: str) -> List[str]:
+    """Get list of reviewers already assigned to the PR."""
+    repo = os.environ.get('GITHUB_REPOSITORY', 'bootc-dev/bootc')
+    result = run_gh_command(
+        ['api', f'repos/{repo}/pulls/{pr_number}', '--jq', '.requested_reviewers[].login'],
+        f"Could not fetch existing reviewers for PR {pr_number}"
+    )
+    return result.stdout.strip().split('\n') if result.stdout.strip() else []
```
cgwalters, line 239: Not that it matters at all but just skimming this stood out to me, it's  inefficient to allocate a string, trim it (allocating a whole new copy) and then re-trim again and allocate *again* but anyways, it doesn't matter

cgwalters (approved):
Still just a superficial skim, I'm ok trying this out...

cgwalters:
Let's try it out! Can always iterate from here

cgwalters:
This seems to be failing https://github.com/bootc-dev/bootc/actions/runs/16933716266/job/47985179886?pr=1503

PR #1493, opened 2025-08-06
ci: Release every 3 weeks starting from 2025-08-04

ckyrouac:
This modifies the scheduled release to happen every 3 weeks regardless of when the previous release happened. This aligns with the team's schedule.

Assisted-by: Claude code


cgwalters (approved):
Looks superficially reasonable

PR #1494, opened 2025-08-06
ci: Refactor release action to use app and create-pull-request

ckyrouac:
This uses the token generated in the GitHub app that has a limited
scope. Also refactors the commit/push steps to use the
create-pull-request action rather than manually committing and pushing.
This simplifies the token usage when committing/pushing.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,40 +11,46 @@ jobs:
   release:
     name: Create Release
     if: |
-      (github.event_name == 'pull_request' && 
+      (github.event_name == 'pull_request' &&
        github.event.pull_request.merged == true &&
        contains(github.event.pull_request.labels.*.name, 'release'))
     runs-on: ubuntu-latest
     container: quay.io/coreos-assembler/fcos-buildroot:testing-devel
     steps:
+      - uses: actions/create-github-app-token@v2
+        id: app-token
+        with:
+          app-id: ${{ secrets.APP_ID }}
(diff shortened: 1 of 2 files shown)


cgwalters:
>  I'm not sure it will work though, you might need to generate it as the owner. 

I am not the owner of the repo anymore, CNCF is right?

> The other option is to enable "Allow GitHub Actions to create and approve pull requests" on the repo and continue using the GITHUB_TOKEN (that's what I was doing on my fork while testing).

Ah. That seems...a bit dangerous to me. AFAIK the more tedious but also more secure/correct way is to make a Github App https://docs.github.com/en/apps/creating-github-apps/about-creating-github-apps/deciding-when-to-build-a-github-app

Last I looked, it wasn't terribly hard and would help unblock other automations too.

ckyrouac:
Ah - looks like we can just create an empty app and use the generated token in the action via this: https://github.com/actions/create-github-app-token. I don't have permissions to the bootc-dev org settings to [add the app](https://docs.github.com/en/apps/creating-github-apps/registering-a-github-app/registering-a-github-app#registering-a-github-app) though. Is that something I need to request from CNCF?

cgwalters:
I was wrong and I am an Owner in the org as are several other people, you were not. I invited you. I am not sure what syncs that (but it's definitely something we want to automate from MAINTAINERS.md) or some other source of truth.

ckyrouac:
thanks for adding me! I reworked this to use the GH app.

cgwalters (approved):

PR #1496, opened 2025-08-07
test: Add integration test running on github runner

henrywang:


henrywang:
@cgwalters , The two failures still need investigate. It's tmt issue, not our code issue.
Do you have any comment on this PR? Thanks.

.github/workflows/integration.yml
```diff
@@ -0,0 +1,86 @@
...
+jobs:
+  build:
+    strategy:
+      matrix:
+        test_os: [fedora-42]
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Install podman for heredoc support
```
cgwalters, line 16: Unrelated to this PR specifically but we should probably factor this stuff out into a shared action helper

.github/workflows/integration.yml
```diff
@@ -0,0 +1,86 @@
+name: bootc integration test
+on:
+  pull_request:
+    branches: [main]
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        test_os: [fedora-42]
```
cgwalters: Let's make this be `quay.io/fedora/fedora:42` e.g. - see below

tests/build.sh
```diff
@@ -0,0 +1,124 @@
...
+
+# Some tmt-* commands to /usr/local/bin
+COPY bin /usr/local/bin
+CONTAINERFILEOF
+
+LOCAL_IMAGE="localhost/bootc:test"
+sudo podman build \
+    --retry 5 \
+    --retry-delay 5s \
+    --tls-verify=false \
```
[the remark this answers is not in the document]
cgwalters: Yeah I don't know why we'd need to disable TLS here

tests/build.sh
```diff
@@ -0,0 +1,120 @@
...
+    -v "$(pwd)":/code:z \
+    -t "$LOCAL_IMAGE" \
+    -f "$CONTAINERFILE" \
+    "$BOOTC_TEMPDIR"
+
+SSH_KEY=${BOOTC_TEMPDIR}/id_rsa
+ssh-keygen -f "${SSH_KEY}" -N "" -q -t rsa-sha2-256 -b 2048
+
+sudo truncate -s 10G "${BOOTC_TEMPDIR}/disk.raw"
```
cgwalters: There's so much `sudo` in this script I think it's way easier to just require it to be invoked via `sudo` from the start via the GHA invocation.

(Of course at some point we will support the disk image generation non-root, and at that time the script should mostly Just Work non-root too)

tests/test.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash
+set -exuo pipefail
+
+# This script runs disk image with qemu-system and run tmt against this vm.
+
+BOOTC_TEMPDIR="/tmp/tmp-bootc-build"
```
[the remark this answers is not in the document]
cgwalters: Yes, I think we could just require this script to be invoked in the context of the temporary directory.

That said also, the only thing that actually binds the two things together right now is the ssh key.

But we can avoid that by not using `--root-ssh-authorized-keys` at `bootc install to-disk` time but instead injecting the SSH key via systemd credentials - that's how podman-bootc does it.

(Of course this whole topic instantly gets into the whole https://gitlab.com/fedora/bootc/tracker/-/issues/2 and whether/how podman-bootc and other virt provisioning tools should be shared underneath different testing frameworks)

Anyways for now though let's just change the GHA to allocate a temporary directory, see below

Makefile
```diff
@@ -66,6 +66,9 @@ test-bin-archive: all
 test-tmt:
 	cargo xtask test-tmt
 
+test:
+	tests/build.sh && tests/test.sh
```
cgwalters, line 87: I think this would then become:

```
tmpd=$(mktemp -d) && cd $tmpd && $srcdir/tests/build.sh && $srcdir/tests/test.sh
```
or so

.github/workflows/integration.yml
henrywang: This `test_os` will be [fedora-41, fedora-42, fedora-43, centos-9, centos-10].

tests/build.sh
henrywang: Two reasons the script can't be run with `sudo` in github workflow:
1. `env` in job will not work for sudo, because the `env` injected env only works with current user, not root.
2. Files generated with `sudo` can't be archived due to `Error: EACCES: permission denied`

cgwalters: > env in job will not work for sudo, because the env injected env only works with current user, not root.

This is fixable with `sudo -E`

> Files generated with sudo can't be archived due to Error: EACCES: permission denied

Right, I think the simplest thing here is to do:

```
if test -n "${SUDO_UID}" then;
  chown -R -h ${SUDO_UID} /path/to/output
fi
```

or so

cgwalters (approved):
I'm OK with this as is if you prefer, we can land further cleanups as followups

cgwalters:
>          package: building container image with dependencies

We had a live chat about this and I think the fact that tmt is dynamically rebuilding the OS image on boot is a problem. In the *long* term ideally the testing framework doesn't require installing any binaries at all on the target (this would let us test minimal systems without python etc) but:

- In the short term how about `tmt print-target-dependencies` or so and then we can pre-install those in our container image, and then tmt shouldn't require rebuilding inside the target system? i.e. in the Dockerfile we do `RUN dnf -y install $(tmt print-target-dependencies)`

henrywang:
`tmt-1.56` fixed the blocker issue. @cgwalters I think we can land this PR. And I'll add `https://github.com/cgwalters/bootc-kit` into CI later in another PR. Thanks.

tests/test.sh
```diff
@@ -0,0 +1,70 @@
...
+SSH_OPTIONS=(-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o ConnectTimeout=5)
+SSH_KEY=${BOOTC_TEMPDIR}/id_rsa
+
+ARCH=$(uname -m)
+case "$ARCH" in
+"aarch64")
+  qemu-system-aarch64 \
```
cgwalters, line 13: Eventually this is going to need to use a shared library of some form. Can't we just point tmt at the disk image and let testcloud run it or so?

.github/workflows/integration.yml
```diff
@@ -0,0 +1,100 @@
...
+        with:
+          name: PR-${{ github.event.number }}-${{ matrix.test_os }}-id_rsa
+          path: /tmp/tmp-bootc-build/id_rsa
+          retention-days: 1
+
+  test:
+    needs: build
+    strategy:
+      matrix:
+        test_os: [fedora-41, fedora-42, fedora-43, centos-9]
```
cgwalters, line 54: See above

.github/workflows/integration.yml
```diff
@@ -0,0 +1,100 @@
+name: bootc integration test
+on:
+  pull_request:
+    branches: [main]
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        test_os: [fedora-41, fedora-42, fedora-43, centos-9]
```
cgwalters, line 10: I think we can stop testing fedora-41 overall and we MUST test centos-10

tests/build.sh
```diff
@@ -0,0 +1,118 @@
...
+BOOTC_TEMPDIR="/tmp/tmp-bootc-build"
+
+# Get OS info from TEST_OS env
+OS_ID=$(echo "$TEST_OS" | cut -d '-' -f 1)
+OS_VERSION_ID=$(echo "$TEST_OS" | cut -d '-' -f 2)
+
+# Base image
+case "$OS_ID" in
+    "centos")
+        TIER1_IMAGE_URL="quay.io/centos-bootc/centos-bootc:stream${OS_VERSION_ID}"
```
cgwalters, line 17: Not new and not directly related to this but we have this legacy of "TIER1" naming - there's only one default image and it's likely to stay that way. I'd probably just say `BASE_IMAGE_URL` going forward for things like this.

.github/workflows/integration.yml
```diff
@@ -0,0 +1,100 @@
...
+          name: PR-${{ github.event.number }}-${{ matrix.test_os }}-id_rsa
+          path: /tmp/tmp-bootc-build/id_rsa
+          retention-days: 1
+
+  test:
+    needs: build
+    strategy:
+      matrix:
+        test_os: [fedora-41, fedora-42, fedora-43, centos-9]
+        tmt_plan: [test-01-readonly, test-20-local-upgrade, test-21-logically-bound-switch, test-22-logically-bound-install, test-23-install-outside-container, test-24-local-upgrade-reboot]
```
cgwalters, line 55: It's going to be a maintenance trap if we have to list all tests here *and* in the tmt directory.

One thing that also seems like it will become a bit painful is having a CI context check per test. (It's not wrong, but it will get really long)

A good middle ground might be "sharding"; basically like https://nexte.st/docs/ci-features/partitioning/?h=shard#hashed-sharding where we run a group of tests per GHA runner. That would give us a good mix of parallelism and efficiency.

And crucially by doing the sharding we would avoid needing to list each test; the allocated runner discovers which tests it should run by just taking an integer from 0-N as input, and looking if the hash of the test modulo the input key is zero.

tests/build.sh
```diff
@@ -0,0 +1,118 @@
...
+        ;;
+    "fedora")
+        TIER1_IMAGE_URL="quay.io/fedora/fedora-bootc:${OS_VERSION_ID}"
+        ;;
+esac
+
+CONTAINERFILE="${BOOTC_TEMPDIR}/Containerfile"
```
cgwalters, line 24: This would create a *third* copy of our Dockerfile, I think we can avoid that by just having the github action matrix be *directly* passed down into e.g.

```
base=$1
shift
podman build --build-arg=base=$base
```

And that's all right?

.github/workflows/integration.yml
```diff
@@ -0,0 +1,100 @@
+name: bootc integration test
```
cgwalters, line 1: Biggest question: Will we *also* continue to run the tests via packit+testing farm? Probably not, right?

But still though, we should bear in mind that we still need to have good bridges/integration with at least tmt/testing-farm since other groups will use that.

tests/test.sh
henrywang: tmt will call libvirt, not qemu. `bootc-kit` can replace those part of code later. I'm working on it now.

.github/workflows/integration.yml
henrywang: This one is required to get different base image.

tests/build.sh
henrywang: Right. Need update the name.

.github/workflows/integration.yml
henrywang: No, test can be run on github action runner and with `make test` locally.

cgwalters:
OK let's get this in and then continue iteration!

.github/workflows/integration.yml
cgwalters: I thought more about this and I think there's an obvious thing to keep doing with the packit+testing-farm flow, which is testing of system-reinstall-bootc.

For that we very precisely want an RPM and a stock cloud image. It's actually almost a good thing if we're *not* testing having a new bootc in the *target* container image just to shake out any cross-version issues.

tests/test.sh
cgwalters: Yep and what I'm hitting now is that these qemu instances just leak after the test script runs, so trying to run multiple tests locally just fails.

I fixed this with a quick strategic use of `setpriv --deathsig SIGTERM` around qemu.

PR #1499, opened 2025-08-07
chore(deps): update cargo dependencies

p5:
An attempt to get Renovate back up and running on this repo.

Looking through the changes, all make sense.  There were some breakages which I have resolved, though I am counting on CI to confirm I haven't broken anything.

The vast majority of the breaking changes were straight forward lifetimes, but a couple features were renamde and functions deprecated.

Split minor and patch bumps into a separate commit (still within this PR), but can break them out to a separate PR if preferred.


p5:
Remaining changes (to be pushed):
- [x] Cargo format
- [ ] ~~Switch from GzipLayerWriter to LayerWriter in ostree-ext's ocidir~~ 

> Switch from GzipLayerWriter to LayerWriter in ostree-ext's ocidir

GitHub Copilot gave up on this too, and ended up removing the ocidir crate in frustration 😆  
Need help from a Rust expert...  Maybe GPT-5 will have better luck

crates/lib/src/imgstorage.rs
```diff
@@ -104,7 +104,7 @@ fn bind_storage_roots(cmd: &mut Command, storage_root: &Dir, run_root: &Dir) ->
                 Mode::empty(),
             )?;
             rustix::process::fchdir(&storage_root)?;
-            rustix::thread::unshare(rustix::thread::UnshareFlags::NEWNS)?;
+            rustix::thread::unshare_unsafe(rustix::thread::UnshareFlags::NEWNS)?;
```
p5, line 107: TODO: Investigate this further.  Is this the right thing to do?  It was the recommended way to handle this deprecation in the crate's docs, but we should try and avoid unsafe code.

crates/ostree-ext/src/tar/export.rs
```diff
@@ -75,13 +75,13 @@ fn map_path_inner<'p>(
 }
 
 /// Convert /usr/etc back to /etc
-fn map_path(p: &Utf8Path) -> std::borrow::Cow<Utf8Path> {
+fn map_path<'a>(p: &'a Utf8Path) -> std::borrow::Cow<'a, Utf8Path> {
```
p5, line 78: The majority of these were recommendations by cargo itself, and passed tests, so I think they are okay.

crates/ostree-ext/src/fixture.rs
```diff
@@ -340,13 +340,13 @@ fn ensure_parent_dirs(
         .map_err(Into::into)
 }
 
-fn relative_path_components(p: &Utf8Path) -> impl Iterator<Item = Utf8Component> {
+fn relative_path_components(p: &Utf8Path) -> impl Iterator<Item = Utf8Component<'_>> {
```
p5: The majority of these were directly applied via the recommendations by cargo itself, and passed tests, so I think they are okay.  

I took a quick look into _why_ they are showing, and it seems a recent Rust update became a bit stricter, preferring explicit lifetimes.  I do not believe these were caused by any dependency updates, but were easy fixes and made my build output look cleaner.

crates/lib/src/imgstorage.rs
p5: It appears `unshare` itself is considered unsafe, and this name change simply clarifies it as such.
https://github.com/bytecodealliance/rustix/pull/1482

Oh - it turns out Allison was the reporter of the issue!  

Since this PR isn't meant to change logic, I vote to potentially refactor this separately.

crates/blockdev/Cargo.toml
```diff
@@ -12,7 +12,7 @@ bootc-utils = { package = "bootc-internal-utils", path = "../utils", version = "
 camino = { workspace = true, features = ["serde1"] }
 fn-error-context = { workspace = true }
 libc = { workspace = true }
-regex = "1.10.4"
+regex = "1.11.1"
```
cgwalters: Historically we haven't bumped these unless the semver requirement is wrong.

And I would bet we aren't *actually* using any new regex APIs.

How did you write this PR? Did you run some tool which also bumped these requirements in `Cargo.toml`?

crates/blockdev/Cargo.toml
```diff
@@ -21,7 +21,7 @@ tokio = { workspace = true, features = ["signal"] }
 tracing = { workspace = true }
 
 [dev-dependencies]
-indoc = "2.0.5"
+indoc = "2.0.6"
```
cgwalters: In a quick glance there's basically no interesting changes between 2.0.5 and 2.0.6 so I'd prefer to avoid the churn so we can highlight just the semver-incompat bumps.

crates/lib/src/utils.rs
```diff
@@ -32,7 +32,7 @@ pub(crate) fn origin_has_rpmostree_stuff(kf: &glib::KeyFile) -> bool {
 
 // Access the file descriptor for a sysroot
 #[allow(unsafe_code)]
-pub(crate) fn sysroot_fd(sysroot: &ostree::Sysroot) -> BorrowedFd {
+pub(crate) fn sysroot_fd(sysroot: &ostree::Sysroot) -> BorrowedFd<'_> {
```
cgwalters: We already had https://github.com/bootc-dev/bootc/pull/1497 which just merged so I think this should be rebased

crates/system-reinstall-bootc/Cargo.toml
```diff
@@ -21,16 +21,16 @@ clap = { workspace = true, features = ["derive"] }
...
 dialoguer = "0.11.0"
 indoc = { workspace = true }
-log = "0.4.21"
+log = "0.4.27"
 openssh-keys = "0.6.4"
 rustix = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
-serde_yaml = "0.9.22"
+serde_yaml = "0.9.34"
```
cgwalters: I think we should actually be moving more of these common dependencies into the workspace - that's a different PR

cgwalters (changes requested):
Thanks for working on this!

cgwalters:
> An attempt to get Renovate back up and running on this repo.

Yeah we need to fix that for sure soon...

crates/blockdev/Cargo.toml
p5: > How did you write this PR? Did you run some tool which also bumped these requirements in Cargo.toml?

I used two Cargo tools - cargo-outdated and cargo-upgrade.  The former lists available updates and the latter allows you to apply version bumps in bulk.  Cargo didn't seem to have anything built-in, like other frameworks do.

The first commit was to cargo-upgrade bumping compatible versions in the lock files.  The second was updating potentially breaking packages (and dealing with any incompatibilities).  The third is undoing one of the updates that I couldn't figure out and the final commit is re-running the first to bump any packages that can be updated after the breaking changes are resolved.

p5: > And I would bet we aren't actually using any new regex APIs.

No, we are not.

> Historically we haven't bumped these unless the semver requirement is wrong.

I'm slightly confused by your approach of not bumping dependencies unless we're using new features.

The point of this PR was to give Renovate a jumpstart so that it can start suggesting updates once again.  **Renovate (and Dependabot) won't care about whether we're using features in new versions or not**.  It will always suggest updating them.  Plus I don't think we'd be reverting updates in Renovate PRs each week just because there's no useful changes for us.

If this were a distinct PR to "update regex" and "update indoc", I would agree, it probably wouldn't be merged because it adds useless commits and a waste of the reviewer's time.  But for grouped updates where you increase the overall baseline of dependencies without adding useless commits, I don't see a need to follow the same.

---

If I've not convinced you, I can remove the regex and indoc version bumps.

crates/lib/src/utils.rs
p5: PR has been rebased!

crates/blockdev/Cargo.toml
cgwalters: > I'm slightly confused by your approach of not bumping dependencies unless we're using new features.

There's two version numbers - in `Cargo.toml` and `Cargo.lock`. The versions in `Cargo.toml` only influence selection of versions in the latter (here and in other projects that use this code as crates, of which there are only a select few).

We *definitely* need to keep things in `Cargo.lock` up to date - that's the *real* change. But that's distinct from the question of whether to bump things in `Cargo.toml`.

Some people do prefer to keep those aggressively bumped to just enforce things, but to me it adds a lot of unnecessary noise.

p5: Ah, right.  So my understanding of Rust dependencies was wrong here.

I was under the assumption it was the same as NPM etc - where if you specify a version in the main package list (package.json/Cargo.toml), that is the version you will get in the lock file.  And if you wanted a range or a "loose" version, you'd need include a comparison operator ("^", ">=" etc).

Will have a look at which to revert in the Cargo.toml.

p5: Sorry about that - this has been actioned.  

Will cleanup commit history.

crates/system-reinstall-bootc/Cargo.toml
cgwalters: :arrow_right: https://github.com/bootc-dev/bootc/pull/1501

PR #1500, opened 2025-08-08
Add initramfs infrastructure

cgwalters:
This adds scaffolding to install a stub binary which can optionally be added into the initramfs;
prep for us doing real work during setup as we aim to move to the native composefs backend.

The binary is *built* but is only installed by a
new `Makefile` target, so existing build system
users won't pick it up. Our development-only
`Dockerfile` gains a build option to use it
(and also ensures the initramfs is regenerated).

However previously we also discussed moving the fstab logic into the initramfs:
https://github.com/bootc-dev/bootc/pull/1113

I might try doing that once this lands.

One notable thing is that even this trivial nearly-no-op binary is still 4MB which I think is mostly due
to linking in a whole copy of prebuilt rust `std`. In theory we could try going to `#[no_std]` but I
don't think it'll be viable once we start doing more here. Probably most practical thing re size is `-Z build-std` + LTO.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -186,6 +186,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "bootc-initramfs-setup"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+]
+
+[[package]]
 name = "bootc-internal-blockdev"
 version = "0.0.0"
 dependencies = [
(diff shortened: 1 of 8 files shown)


jeckersb:
FWIW on 1.89 with release build, `bootc-initramfs-setup` comes in at 2.3M for me.

Edit: and after stripping, it's down to 378K

jeckersb:
With build-std and LTO it only drops to 354K but takes waaaaaay longer to build.

cgwalters:
Oh duh of course I forgot we had `debug = true`; 378k is way more reasonable of course.

PR #1502, opened 2025-08-12
composefs: Add missed commits when rebasing

cgwalters:
I messed up the previous rebase/squash of the composefs branch. This adds commits since https://github.com/bootc-dev/bootc/commit/c89efab8ba14683682093f6f730a696b05491783 

There were definitely some messy conflicts, but this at least compiles


crates/lib/src/install.rs
```diff
@@ -1637,73 +1775,55 @@ pub(crate) fn setup_composefs_bls_boot(
             vec![
                 format!("root=UUID={DPS_UUID}"),
                 RW_KARG.to_string(),
-                format!("composefs={id_hex}"),
+                format!("{COMPOSEFS_CMDLINE}{id_hex}"),
```
Johan-Liebert1: `=` missing here

crates/lib/src/bls_config.rs
```diff
@@ -34,24 +35,22 @@ impl Ord for BLSConfig {
     }
 }
 
-impl BLSConfig {
-    pub(crate) fn to_string(&self) -> String {
-        let mut out = String::new();
-
+impl Display for BLSConfig {
```
Johan-Liebert1: This should've been moved to `lib/src/parsers/bls_config.rs` and indeed it has been. This piece seems leftover

crates/lib/src/parsers/grub_menuconfig.rs
```diff
@@ -13,14 +13,11 @@ use nom::{
 /// Body content of a GRUB menuentry containing parsed commands.
 #[derive(Debug, PartialEq, Eq)]
 pub(crate) struct MenuentryBody<'a> {
-    /// Kernel modules to load
```
Johan-Liebert1, line 16: @cgwalters The changes you made, and the ones we pushed to main, seem to have been overwritten here. I think we should keep your changes for this particular file

crates/lib/src/install.rs
cgwalters: Good catch, Gemini found the other one of this.

crates/lib/src/parsers/grub_menuconfig.rs
cgwalters: Just the added comments right? The rest of this was changing things here to use an owned value for chainloader per your previous commit.

(That said, it's weird to have a mix of borrowed and owned data here...)

cgwalters:
OK, updated

crates/lib/src/parsers/grub_menuconfig.rs
Johan-Liebert1: Yes, it is kinda weird, but getting rust to return a local ref without `Box::leak` is pretty complicated

Johan-Liebert1:
centos-10 tests seem to be failing due to package conflicts. Else LGTM

```
+ dnf -y builddep contrib/packaging/bootc.spec
Updating Subscription Management repositories.
Unable to read consumer identity

This system is not registered with an entitlement server. You can use "rhc" or "subscription-manager" to register.

CentOS Stream 10 - BaseOS                        35 MB/s | 6.7 MB     00:00    
CentOS Stream 10 - AppStream                     23 MB/s | 3.3 MB     00:00    
CentOS Stream 10 - CRB                          7.0 MB/s | 748 kB     00:00    
CentOS Stream 10 - Extras packages               54 kB/s | 6.9 kB     00:00    

Package ostree-2025.4-3.el10.x86_64 is already installed.
Package skopeo-2:1.20.0-1.el10.x86_64 is already installed.
Package systemd-257-11.el10.x86_64 is already installed.
Error: 
 Problem: cannot install both ostree-libs-2025.4-2.el10.x86_64 from appstream and ostree-libs-2025.4-3.el10.x86_64 from @System
  - package ostree-devel-2025.4-2.el10.x86_64 from crb requires ostree-libs(x86-64) = 2025.4-2.el10, but none of the providers can be installed
  - package ostree-2025.4-3.el10.x86_64 from @System requires ostree-libs(x86-64) = 2025.4-3.el10, but none of the providers can be installed
  - cannot install the best candidate for the job
(try to add '--allowerasing' to command line to replace conflicting packages or '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
Error: building at STEP "RUN <<EORUN": while running runtime: exit status 1
```

cgwalters:
Yeah that's https://gitlab.com/redhat/centos-stream/containers/bootc/-/issues/1174 - unrelated to this PR

crates/lib/src/parsers/grub_menuconfig.rs
cgwalters: There's https://lib.rs/crates/yoke among others

PR #1507, opened 2025-08-13
install: Use read_file from composefs-boot

cgwalters:
Just reducing code here.


cgwalters:
Yeah it'd probably be messy right now to gate it via a feature, but...the grub stuff is super messy itself and it would be nice to cordon it off a bit in the future...

cgwalters:
> LGTM, with note that make validate-rust is failing in the composefs-backend branch unrelated to this change:

Fixed in  https://github.com/bootc-dev/bootc/pull/1512

PR #1508, opened 2025-08-13
Resolve GH_TOKEN conflict in auto-reviewer workflow

gursewak1997:


.github/workflows/auto-review.yml
```diff
@@ -27,12 +28,18 @@ jobs:
           pip install pyyaml
           
       - name: Authenticate with GitHub
-        run: echo "${{ secrets.GH_PAT }}" | gh auth login --with-token
-        env:
-          GH_TOKEN: ${{ secrets.GH_PAT }}
+        run: |
+          if [ -z "${{ secrets.GH_PAT }}" ]; then
```
cgwalters: Is this a PAT associated with your account?

I think it'd make sense to cut over to the "bootc actions token" app right?

.github/workflows/auto-review.yml
```diff
@@ -27,12 +28,16 @@ jobs:
           pip install pyyaml
           
       - name: Authenticate with GitHub
-        run: echo "${{ secrets.GH_PAT }}" | gh auth login --with-token
-        env:
-          GH_TOKEN: ${{ secrets.GH_PAT }}
+        run: |
+          if [ -z "${{ secrets.REVIEWER_TOKEN }}" ]; then
```
cgwalters: This looks like a newly created token? How is it managed? Is there a process for rotating it etc?


Did you see https://github.com/bootc-dev/bootc/pull/1494 ? My understanding is that we can use that app credentials consistently - then it's not tied to any particular user, there's no additional secrets to create and manage etc.

gursewak1997: I’m now using the default GITHUB_TOKEN provided by GitHub Actions, with permissions: pull-requests: write. Previously, testing on other repos or forked branches failed due to access restrictions. At this point, no personal tokens are used and the workflow relies entirely on the org/repo-owned token, as intended.

Issue #1510, opened 2025-08-14
install --via-loopback: Loopback-cleanup-helper never invoked due to missing "internals" subcommand


SamD2021:
On current tip of main, `bootc install --via-loopback` attempts to run `bootc loopback-cleanup-helper` instead of `bootc internals loopback-cleanup-helper`.

Because the helper is only available under the `internals` subcommand, this call fails with `error: unrecognized subcommand 'loopback-cleanup-helper'`.

The likely reason this was not caught earlier is that the spawn itself succeeds and the failure can't be easily checked due to the fire and forget nature required by the out of process helper. Because of this, the install still completes and the system boots normally even though the command is incorrect resulting in the cleanup helper never running, leaving loop devices uncleared and not reaching the loopback cleanup work added in #1402 / #1449.

---

**Expected behavior:**
The helper should be invoked as:

```bash
bootc internals loopback-cleanup-helper --device <path>
```

so that it runs and cleans up loop devices as intended.

---

**Actual behavior:**
Bootc spawns:

```bash
bootc loopback-cleanup-helper --device <path>
```

and fails with:

```bash
error: unrecognized subcommand 'loopback-cleanup-helper'
```

---

**Steps to reproduce:**

1. Clone the bootc repository at current `main` (tip) and build the container with bootc.

   ```bash
   git clone https://github.com/bootc-dev/bootc
   cd bootc
   just build
   ```
2. Run podman-bootc to trigger a bootc install --via-loopback:

```bash
   podman-bootc --log-level debug run localhost/bootc
```
3. Observe the log where `bootc` tries to invoke `loopback-cleanup-helper` without `internals`.

---

**Environment:**

* bootc commit: current tip (#3e63596)
* OS/Distro: Fedora
* Podman version: 5.5.2

---

**Related issues/PRs:**

* \#799 (starter issue for loopback cleanup)
* \#1402 (original PR adding helper)
* \#1449 (follow-up PR)

PR #1511, opened 2025-08-14
fix(blockdev): include "internals" when spawning loopback-cleanup-helper

SamD2021:
Adds the missing `internals` subcommand when spawning `loopback-cleanup-helper` so it runs as intended during `bootc install --via-loopback`. Without `internals`, the helper fails with `unrecognized subcommand 'loopback-cleanup-helper'`, preventing loop device cleanup. This doesn’t block install/boot because the helper is fire-and-forget, but it leaves loop devices uncleared.

Fixes: #1510

### DIFF
--- a/crates/blockdev/src/blockdev.rs
+++ b/crates/blockdev/src/blockdev.rs
@@ -251,7 +251,12 @@ impl LoopbackDevice {
 
         // Create the helper process
         let mut cmd = Command::new(bootc_path);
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
Ouch. Thanks for finding and fixing this.

SamD2021:
Glad to help! :)

PR #1512, opened 2025-08-14
Map grub feature to composefs-backend and import parser/bls

cgwalters:
Some prep work for merging the composefs-backend branch.


cgwalters:
There's definitely more to do for bls, in particular we actually need to handle filenames too.

PR #1513, opened 2025-08-14
rechunker: Use (path,checksum) when creating exclusive components

ckyrouac:
This allows files with duplicate contents (and checksums) to be stored in different layers when using the rechunker.

### DIFF
--- a/crates/ostree-ext/src/chunking.rs
+++ b/crates/ostree-ext/src/chunking.rs
@@ -15,7 +15,7 @@ use crate::objectsource::{ContentID, ObjectMeta, ObjectMetaMap, ObjectSourceMeta
 use crate::objgv::*;
 use crate::statistics;
 use anyhow::{anyhow, Result};
-use camino::Utf8PathBuf;
+use camino::{Utf8Path, Utf8PathBuf};
 use containers_image_proxy::oci_spec;
 use gvariant::aligned_bytes::TryAsAligned;
 use gvariant::{Marker, Structure};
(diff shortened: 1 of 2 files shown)


crates/ostree-ext/src/chunking.rs
```diff
@@ -230,6 +230,28 @@ impl Chunk {
             false
         }
     }
+
+    pub(crate) fn move_path(&mut self, dest: &mut Self, checksum: &str, path: Utf8PathBuf) -> bool {
```
cgwalters: Style nit, accepting `path: &Utf8Path` is more idomatic

crates/ostree-ext/src/chunking.rs
```diff
@@ -230,6 +230,28 @@ impl Chunk {
...
+        if let Some((_size, paths)) = self.content.get_mut(checksum) {
```
cgwalters, line 235: Style nit, I prefer doing `let Some(...) else { return false; }` for things like this to avoid rightward drift

crates/ostree-ext/src/chunking.rs
```diff
@@ -324,21 +346,18 @@ impl Chunking {
...
+                chunk.packages = vec![component.to_string()];
 
                 // Move all objects belonging to this exclusive component
-                if let Some(objects) = specific_map.get(&component.meta.identifier) {
-                    for &obj in objects {
-                        self.remainder.move_obj(&mut chunk, obj);
-                    }
+                for (path, checksum) in files {
+                    self.remainder
+                        .move_path(&mut chunk, checksum.as_str(), path.clone());
```
cgwalters: This function returns `bool` but is unused? No problem with that, I'm just curious if there's a reason

ckyrouac:
> Doesn't it open the question a bit whether we should also stop doing that for non-exclusive?

The way I've been thinking about it is when a user explicitly defines an exclusive component, the expectation is the files go exactly where they want. The package based rechunking is an automatic process that the user doesn't really control so it's less surprising when duplicate files are in the same layer. I'll add docs to explain this in a followup.

PR #1518, opened 2025-08-15
store: Remove dynamic abstraction

cgwalters:
A while ago we added a trait in preparation for multiple backends. The current composefs branch ignores it and has a bunch of `if {} else {}`.

Looking at this, what I think will work better in the end is actually a more refined version of the `if {} else {}` model instead of trying to really flesh out this trait. It's hard to say of course until we get there, but the trait approach forces a high level of abstraction vs just trying to factor out common code between two backends.


jeckersb (approved):
:+1: 

yeah the backend thing never really panned out, this definitely makes things a lot simpler

PR #1519, opened 2025-08-15
system-reinstall-bootc: Ensure stdout is flushed

cgwalters:
We're seeing a CI failure where the final reinstall message isn't seen, I think this might fix it.


jmarrero (approved):
lgtm

PR #1522, opened 2025-08-15
Add systemd structured logging for bootc state changes

gursewak1997:
Log key operations (deploy, upgrade, switch, install, etc.) to systemd journal with structured fields including image references, digests, and operation types. Uses existing journal infrastructure and follows ostree logging patterns.

### DIFF
--- a/crates/lib/src/boundimage.rs
+++ b/crates/lib/src/boundimage.rs
@@ -39,7 +39,23 @@ pub(crate) struct ResolvedBoundImage {
 
 /// Given a deployment, pull all container images it references.
 pub(crate) async fn pull_bound_images(sysroot: &Storage, deployment: &Deployment) -> Result<()> {
-    let bound_images = query_bound_images_for_deployment(sysroot.get_ostree()?, deployment)?;
+    // Log the bound images operation to systemd journal
+    const BOUND_IMAGES_JOURNAL_ID: &str = "1a0b9c8d7e6f5a4b3c2d1e0f9a8b7c6d5";
+    tracing::info!(
+        message_id = BOUND_IMAGES_JOURNAL_ID,
+        bootc.deployment.osname = deployment.osname().as_str(),
+        bootc.deployment.checksum = deployment.csum().as_str(),
+        "Starting pull of bound images for deployment"
+    );
+
+    let ostree = sysroot.get_ostree()?;
+    let bound_images = query_bound_images_for_deployment(ostree, deployment)?;
+    tracing::info!(
+        message_id = BOUND_IMAGES_JOURNAL_ID,
+        bootc.bound_images_count = bound_images.len(),
+        "Found {} bound images to pull",
+        bound_images.len()
+    );
     pull_images(sysroot, bound_images).await
 }
 
--- a/crates/lib/src/cli.rs
+++ b/crates/lib/src/cli.rs
@@ -1067,6 +1067,26 @@ async fn switch(opts: SwitchOpts) -> Result<()> {
         println!("Image specification is unchanged.");
         return Ok(());
     }
+
+    // Log the switch operation to systemd journal
+    const SWITCH_JOURNAL_ID: &str = "7a6b5c4d3e2f1a0b9c8d7e6f5a4b3c2d1";
+    let old_image = host
+        .spec
+        .image
+        .as_ref()
+        .map(|i| i.image.as_str())
+        .unwrap_or("none");
+
+    tracing::info!(
+        message_id = SWITCH_JOURNAL_ID,
+        bootc.old_image_reference = old_image,
+        bootc.new_image_reference = &target.image,
+        bootc.new_image_transport = &target.transport,
+        "Switching from image {} to {}",
+        old_image,
+        target.image
+    );
+
     let new_spec = RequiredHostSpec::from_spec(&new_spec)?;
 
     let fetched = crate::deploy::pull(repo, &target, None, opts.quiet, prog.clone()).await?;
--- a/crates/lib/src/deploy.rs
+++ b/crates/lib/src/deploy.rs
@@ -429,11 +429,24 @@ pub(crate) async fn pull_from_prepared(
     let imgref_canonicalized = imgref.clone().canonicalize()?;
     tracing::debug!("Canonicalized image reference: {imgref_canonicalized:#}");
 
+    // Log successful import completion
+    const IMPORT_COMPLETE_JOURNAL_ID: &str = "4d3e2f1a0b9c8d7e6f5a4b3c2d1e0f9a8";
+
+    tracing::info!(
+        message_id = IMPORT_COMPLETE_JOURNAL_ID,
+        bootc.image.reference = &imgref.image,
+        bootc.image.transport = &imgref.transport,
+        bootc.manifest_digest = import.manifest_digest.as_ref(),
+        bootc.ostree_commit = &import.merge_commit,
+        "Successfully imported image: {}",
+        imgref
+    );
+
     if let Some(msg) =
         ostree_container::store::image_filtered_content_warning(&import.filtered_files)
             .context("Image content warning")?
     {
-        crate::journal::journal_print(libsystemd::logging::Priority::Notice, &msg);
+        tracing::info!("{}", msg);
     }
     Ok(Box::new((*import).into()))
 }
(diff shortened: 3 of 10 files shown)


crates/lib/src/boundimage.rs
```diff
@@ -39,7 +40,36 @@ pub(crate) struct ResolvedBoundImage {
...
+            ("BOOTC_OPERATION", "pull_bound_images"),
+            ("BOOTC_DEPLOYMENT_OSNAME", &deployment.osname()),
+            ("BOOTC_DEPLOYMENT_CHECKSUM", &deployment.csum()),
+        ]
+        .into_iter(),
+    );
+
     let bound_images = query_bound_images_for_deployment(sysroot, deployment)?;
+
+    if !bound_images.is_empty() {
```
bshephar: Seems like this would be the default case and this conditional check could probably be omitted? 

Or, maybe `if bound_images.is_empty()` we could return early and leave this logging as the default path. Seems like `pull_images()` needs to run anyway to initialize the image store and labels, so we still want to run that, and we also check for it being empty there:
https://github.com/bootc-dev/bootc/blob/main/crates/lib/src/boundimage.rs#L153-L155

Maybe something like:
```
if bound_images.is_empty() {
  return pull_images(sysroot, bound_images).await
 }

<logging stuff by default>

```

The `!is_empty()` is a bit awkward, but this has been raised before in Rust:
https://internals.rust-lang.org/t/the-is-not-empty-method-as-more-clearly-alternative-for-is-empty/10612/3

Your implementation appears to be idiomatic Rust, so I'm happy for this comment to be ignored if you want to keep it as is.

crates/lib/src/boundimage.rs
```diff
@@ -39,7 +40,36 @@ pub(crate) struct ResolvedBoundImage {
 
 /// Given a deployment, pull all container images it references.
 pub(crate) async fn pull_bound_images(sysroot: &Storage, deployment: &Deployment) -> Result<()> {
+    // Log the bound images operation to systemd journal
+    const BOUND_IMAGES_JOURNAL_ID: &str = "1a0b9c8d7e6f5a4b3c2d1e0f9a8b7c6d5";
```
bshephar, line 43: Where do these come from? Is this just a randomly generated string to ensure a unique ID for each log message?

bshephar (commented):
This is a nice improvement. It will be good to have these messages nicely in the Journal!

crates/lib/src/boundimage.rs
gursewak1997: Yes, these are randomly generated unique identifiers for each log message type. They enable filtering the systemd journal for specific operations and follow systemd best practices for structured logging correlation and debugging.

crates/lib/src/status.rs
```diff
@@ -331,6 +332,33 @@ pub(crate) fn get_status(
 /// Implementation of the `bootc status` CLI command.
 #[context("Status")]
 pub(crate) async fn status(opts: super::cli::StatusOpts) -> Result<()> {
+    // Log the status query operation to systemd journal
```
cgwalters: I don't think we should log status operations, that feels like it'll be noisy - who would ever want to know that a status was requested?

It's a bit like Unix atime - having a read-only operation perform a write is weird.

crates/lib/src/install/completion.rs
```diff
@@ -275,6 +276,20 @@ pub(crate) async fn impl_completion(
...
 ) -> Result<()> {
+    // Log the completion operation to systemd journal
+    const COMPLETION_JOURNAL_ID: &str = "0f9a8b7c6d5e4f3a2b1c0d9e8f7a6b5c4";
+    let msg = "Starting bootc installation completion";
+    crate::journal::journal_send(
+        Priority::Info,
+        msg,
+        [
+            ("MESSAGE_ID", COMPLETION_JOURNAL_ID),
+            ("BOOTC_OPERATION", "install_completion"),
```
cgwalters: This `BOOTC_OPERATION` bit feels slightly ad-hoc; it has a pretty high overlap with the `CODE_FUNC` emitted by the default C library - but that one isn't 

I think I'd vote to drop it for now, and actually fix things so that we have a macro that injects `CODE_FUNC`.

crates/lib/src/status.rs
gursewak1997: Made changes to make logs more focused so we get actionable information about system state changes without journal clutter from routine queries.

crates/lib/src/install/completion.rs
gursewak1997: Dropped for now (in other files too).

cgwalters (approved):

cgwalters:
I looked at the test failures and did https://github.com/bootc-dev/bootc/pull/1535 but not sure why this is just showing up now.

crates/lib/src/install/completion.rs
```diff
@@ -276,6 +277,19 @@ pub(crate) async fn impl_completion(
...
+    let msg = "Starting bootc installation completion";
+    crate::journal::journal_send(
+        Priority::Info,
+        msg,
+        [
+            ("MESSAGE_ID", COMPLETION_JOURNAL_ID),
+            ("BOOTC_STATEROOT", stateroot.unwrap_or("default")),
+        ]
+        .into_iter(),
+    );
```
cgwalters: I'm OK to merge as is but one thing that would help readability a good bit here is for us to have a wrapper macro like `journal_send!` that wraps the `format!` infrastructure.

cgwalters: And to go a bit more advanced, having it be something like

```
journal::message!(COMPLETION_JOURNAL_ID, "Starting installation completion", ...<other format args>).with(["BOOTC_STATEROOT", stateroot.unwrap_or("default")).send()
```
or so would look nicer, though definitely a bit more work to have the macro return a struct with methods etc.

crates/lib/src/deploy.rs
```diff
@@ -442,8 +459,38 @@ pub(crate) async fn pull(
     prog: ProgressWriter,
 ) -> Result<Box<ImageState>> {
     match prepare_for_pull(repo, imgref, target_imgref).await? {
-        PreparedPullResult::AlreadyPresent(existing) => Ok(existing),
+        PreparedPullResult::AlreadyPresent(existing) => {
+            // Log that the image was already present (Debug level since it's not actionable)
```
cgwalters, line 464: In looking at things this still does get logged to the journal (which shows it greyed out sure) - as `man journald.conf` says

>  Defaults to "debug" for MaxLevelStore=, MaxLevelSyslog= and MaxLevelSocket=, to ensure that the all messages are stored in the journal, forwarded to syslog and the socket if one exists.

My original take was that since we don't have a way to configure this logging level in our own process, we should do one of:

- Change it to INFO
- Drop it
- Combine it with another message - in a no-op upgrade a *single* "Upgrade request complete, no changes" message seems like it'd be nice.

---

Though thinking about this more led me to realize that actually we *do* already have a logging system wired up with `tracing`...and https://docs.rs/tracing-journald/latest/tracing_journald/ exists although it doesn't appear to be widely used.

So a whole other different option here is to investigate wiring up that tracing-journald crate instead. This would need some careful analysis but I think would probably look a lot nicer.

Among other things, the tracing crate already has its own logging macros so we don't need to carry/invent them.

Also, a compelling feature of tracing is that it supports event contexts. We aren't making heavy use of that yet, but I could imagine doing so.

crates/lib/src/deploy.rs
```diff
@@ -442,8 +459,38 @@ pub(crate) async fn pull(
...
+                    ("BOOTC_IMAGE_TRANSPORT", &imgref.transport),
+                    ("BOOTC_STATUS", "already_present"),
+                ]
+                .into_iter(),
+            );
+            Ok(existing)
+        }
         PreparedPullResult::Ready(prepared_image_meta) => {
+            // Log that we're pulling a new image
+            const PULL_JOURNAL_ID: &str = "5c4d3e2f1a0b9c8d7e6f5a4b3c2d1e0f9";
```
cgwalters: Why duplicate this `const`? Though it does seem to me that "no changes" and "pulling a new image" are *different* events

crates/lib/src/cli.rs
```diff
@@ -992,6 +993,9 @@ async fn upgrade(opts: UpgradeOpts) -> Result<()> {
         } else if booted_unchanged {
             println!("No update available.")
         } else {
+            // Note: Individual operation logging (pull, stage, deploy) will provide detailed information
+            // No need for a separate "upgrading" message here since we're not adding new information
+
```
gursewak1997: This comment can obviously be removed but will wait if anything else is needed and will remove in next iteration.

crates/utils/src/tracing_util.rs
```diff
@@ -1,7 +1,21 @@
...
+        if let Ok(subscriber) = tracing_journald::layer() {
+            tracing_subscriber::registry()
+                .with(subscriber)
+                .with(tracing_subscriber::EnvFilter::from_default_env())
+                .init();
+            return;
+        }
+    }
+
+    // Fall back to the previous setup if journald isn't available
```
cgwalters: Hmm interesting...this would be a pretty big workflow change for those who are used to doing e.g. `env RUST_LOG=trace bootc ...`

I wonder if we can set things up so that we log to the journal always for key messages, but still use stdout/stderr for debug/trace level?

crates/lib/src/install.rs
```diff
@@ -1677,6 +1706,34 @@ pub(crate) async fn install_to_filesystem(
...
+        .source_imgref
+        .as_ref()
+        .map(|s| s.as_str())
+        .unwrap_or("none");
+    let target_path = opts.filesystem_opts.root_path.as_str();
+    let msg = format!(
+        "Starting filesystem installation from {} to {}",
+        source_image, target_path
+    );
+    crate::journal::journal_send(
```
cgwalters: We can convert all of these to `tracing!` right?

crates/utils/src/tracing_util.rs
```diff
@@ -1,19 +1,39 @@
...
+
 /// Initialize tracing with the default configuration.
 pub fn initialize_tracing() {
-    // Don't include timestamps and such because they're not really useful and
-    // too verbose, and plus several log targets such as journald will already
-    // include timestamps.
+    // Always try to use journald subscriber if we're running under systemd
+    // This ensures key messages (info, warn, error) go to the journal
+    let journald_layer = if let Ok(()) = std::env::var("JOURNAL_STREAM").map(|_| ()) {
+        tracing_journald::layer().ok()
```
cgwalters: I think we want to have a filter here so that only info and above goes to the journal, right? I think it's possible with https://docs.rs/tracing-subscriber/latest/tracing_subscriber/layer/trait.Layer.html#method.with_filter - I can take a look if you want.

Otherwise someone debugging with e.g. `env RUST_LOG=trace` will completely spam the journal.

cgwalters (changes requested):
Thanks! This is looking a lot nicer, wouldn't you agree? I just have one outstanding comment

crates/utils/src/tracing_util.rs
cgwalters: Basically same as this https://github.com/bootc-dev/bootc/pull/1522#discussion_r2298883624

cgwalters (approved):

cgwalters:
Ah looks like this needs a rebase now

PR #1523, opened 2025-08-16
Update workflow to use Github App token for reviewer assignment

gursewak1997:


.github/workflows/auto-review.yml
```diff
@@ -27,9 +26,16 @@ jobs:
           python -m pip install --upgrade pip
           pip install pyyaml
 
+      - name: Generate Bootc Actions Token
+        id: bootc_token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ secrets.BOOTC_ACTIONS_APP_ID }}
+          private-key: ${{ secrets.BOOTC_ACTIONS_APP_PRIVATE_KEY }}
```
cgwalters: How are these secrets different from the `secrets.APP_ID` used in the autorelease workflow?

gursewak1997: I originally created this one to include additional permissions needed for the workflow, but it makes more sense to just use `APP_ID` and get the same permissions there. Updating this.

gursewak1997: Will update the permissions on Monday and will confirm them with Chris too.

PR #1525, opened 2025-08-18
Some more refactoring of storage

cgwalters:


crates/lib/src/store/mod.rs
```diff
@@ -82,21 +95,32 @@ impl Storage {
...
+            ostree: sysroot,
             run,
             composefs: Default::default(),
             imgstore: Default::default(),
         })
     }
 
+    /// Access the underlying ostree repository
+    pub(crate) fn get_ostree(&self) -> Result<&SysrootLock> {
+        Ok(&self.ostree)
```
Johan-Liebert1, line 107: I don't see the reason why this is a result...

Johan-Liebert1:
So, if I understand correctly, we want `Store` as an abstraction over all different storage we have and we want to have a single function call instead of `if {} else {}` everywhere? Looks okay to me, just one question above

crates/lib/src/store/mod.rs
cgwalters: https://github.com/bootc-dev/bootc/pull/1509#discussion_r2274516289

cgwalters:
> So, if I understand correctly, we want Store as an abstraction over all different storage we have and we want to have a single function call instead of if {} else {} everywhere? Looks okay to me, just one question above

`Store` holds references to the different storages, but it isn't itself an abstraction yet.

That said, the more I think about things the more I feel what will make sense is to eventually aim for truly "unified storage" where we always pull bootable host containers into the containers-storage instance, and then copy from there into composefs or so. It'd give us a decent mix of the advantages of both.

PR #1529, opened 2025-08-19
images: Discuss `/.autorelabel`

cgwalters:
Came up in chat.

### DIFF
--- a/docs/src/bootc-images.md
+++ b/docs/src/bootc-images.md
@@ -132,6 +132,18 @@ for handling the `security.selinux` extended attribute in tar
 streams; but this can only currently be done with a custom
 build process.
 
+#### Forcing relabeling
+
+Some distributions support a `/.autorelabel` file; this will not work
+on a bootc managed system where `/` is immutable state from the base
+image by default. The labels on the files under `/sysroot/ostree` are
+also managed by bootc/ostree and should not be changed by external
+tools.
+
+Most typically, you instead want to only relabel targeted subdirectories
+of `/etc` and `/var`. This can be done via a systemd unit, e.g.
+`ExecStart=restorecon -R -v /var/lib/mysql`.
+
 ### Toplevel directories
 
 In particular, a common problem is that inside a container image,


HouMinXi:
In QE testing, almost all tests are under the enable selinux env. 
If I want to add an .autorelabel in the container file so that this image can be directly deployed on other machines. 
Once another machine using ostreecontainer in ks to deploy imaging. The selinux context will be relabeled.

bshephar:
Timely topic. I noticed that if I `install to-existing-root` from a el8 host, then I'm not able to login once the node reboots due to AVC denials. I can `rd.break` and then:

```
mount -o remount,rw /sysroot/sysroot/
chroot /sysroot/sysroot/
setfiles /etc/selinux/targeted/contexts/files/file_contexts /
setfiles /etc/selinux/targeted/contexts/files/file_contexts /*
```

Which seems to work because I am then able to login. But presumably, this isn't the ideal approach and is in contrast to the guidance being added by this change. I also know that this isn't really the intended installation method that you would like folks to be using. My use case was more exploratory to understand whether it's possible to in-place upgrade from Rocky 8 to CentOS Stream 10 and how much pain am I willing to endure to do it in-place.

If I try to `chroot` into the `ostree` root immediately after the `install to-existing-root`, and then set the contexts like I did above, then the system is forever broken. :)

cgwalters:
> My use case was more exploratory to understand whether it's possible to in-place upgrade from Rocky 8 to CentOS Stream 10 

In place here means preserving the existing filesystem? It's definitely possible.

> setfiles /etc/selinux/targeted/contexts/files/file_contexts /

Yeah don't do that. We'd need to identify in this case which security contexts are showing up as incorrectly labeled. At the current time a raw `setfiles` won't do the right thing for bootc systems and it'd be really hard for us to fix in general.

As the doc says here only aim to relabel subdirectories of `etc` and `var`.

Digging into specific AVC denials is painful, but it's really what needs to be done.

bshephar:
> > My use case was more exploratory to understand whether it's possible to in-place upgrade from Rocky 8 to CentOS Stream 10
> 
> In place here means preserving the existing filesystem? It's definitely possible.
> 
> > setfiles /etc/selinux/targeted/contexts/files/file_contexts /
> 
> Yeah don't do that. We'd need to identify in this case which security contexts are showing up as incorrectly labeled. At the current time a raw `setfiles` won't do the right thing for bootc systems and it'd be really hard for us to fix in general.
> 
> As the doc says here only aim to relabel subdirectories of `etc` and `var`.
> 
> Digging into specific AVC denials is painful, but it's really what needs to be done.

No worries, I'll reproduce it this week, gather some details and raise a new issue where we can discuss the particulars. Don't want to hijack this thread.

HouMinXi:
On the other hand, Can I understand that directories other than the /sysroot directory can be relabeled by the selinux tool?

cgwalters:
> On the other hand, Can I understand that directories other than the /sysroot directory can be relabeled by the selinux tool?

We'd need to dig in to what specific directories you're talking about but again in general as the docs say, only `/etc` and `/var` in the target deployments are intended for mutation by arbitrary code on the client system.

Conan-Kudo:
On SUSE distributions, the autorelabel file was moved to `/etc/selinux/.autorelabel`, I'm not sure it should be in `/etc` instead of `/var`, but it's something.

PR #1532, opened 2025-08-19
install: Automatically configure aboot

alexlarsson:
This allows you to set a non-default ostree bootloader backend.  In particular, this is needed for aboot images, as otherwise  the deploy will not shell out to aboot-deploy to create the required A/B boot symlinks.

PR includes a fix to make aboot-deploy actually work.

### DIFF
--- a/crates/lib/src/install.rs
+++ b/crates/lib/src/install.rs
@@ -19,6 +19,7 @@ use std::io::Write;
 use std::os::fd::{AsFd, AsRawFd};
 use std::os::unix::process::CommandExt;
 use std::path::Path;
+use std::process;
 use std::process::Command;
 use std::str::FromStr;
 use std::sync::Arc;
@@ -624,7 +625,11 @@ async fn initialize_ostree_root(state: &State, root_setup: &RootSetup) -> Result
     }
 
     let sysroot = {
-        let path = format!("/proc/self/fd/{}", rootfs_dir.as_fd().as_raw_fd());
+        let path = format!(
+            "/proc/{}/fd/{}",
+            process::id(),
+            rootfs_dir.as_fd().as_raw_fd()
+        );
         ostree::Sysroot::new(Some(&gio::File::for_path(path)))
     };
     sysroot.load(cancellable)?;
--- a/crates/lib/src/install.rs
+++ b/crates/lib/src/install.rs
@@ -779,6 +779,25 @@ async fn install_container(
     )?;
     let kargsd = kargsd.iter().map(|s| s.as_str());
 
+    // If the target uses aboot, then we need to set that bootloader in the ostree
+    // config before deploying the commit
+    if ostree_ext::bootabletree::commit_has_aboot_img(&merged_ostree_root, None)? {
+        tracing::debug!("Setting bootloader to aboot");
+        Command::new("ostree")
+            .args([
+                "config",
+                "--repo",
+                "ostree/repo",
+                "set",
+                "sysroot.bootloader",
+                "aboot",
+            ])
+            .cwd_dir(root_setup.physical_root.try_clone()?)
+            .run_capture_stderr()
+            .context("Setting bootloader config to aboot")?;
+        sysroot.repo().reload_config(None::<&gio::Cancellable>)?;
+    }
+
     // Keep this in sync with install/completion.rs for the Anaconda fixups
     let install_config_kargs = state
         .install_config
--- a/crates/ostree-ext/src/bootabletree.rs
+++ b/crates/ostree-ext/src/bootabletree.rs
@@ -12,6 +12,7 @@ use ostree::prelude::*;
 
 const MODULES: &str = "usr/lib/modules";
 const VMLINUZ: &str = "vmlinuz";
+const ABOOT_IMG: &str = "aboot.img";
 
 /// Find the kernel modules directory in a bootable OSTree commit.
 /// The target directory will have a `vmlinuz` file representing the kernel binary.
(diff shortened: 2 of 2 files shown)


crates/lib/src/install/config.rs
```diff
@@ -73,6 +73,8 @@ pub(crate) struct InstallConfiguration {
     pub(crate) kargs: Option<Vec<String>>,
     /// Supported architectures for this configuration
     pub(crate) match_architectures: Option<Vec<String>>,
+    /// Supported architectures for this configuration
```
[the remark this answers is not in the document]
alexlarsson: done

alexlarsson:
Note: The install config file was the most obvious way to allow bootloader to be configured, it may not be the best approach.

crates/lib/src/install.rs
```diff
@@ -624,7 +625,11 @@ async fn initialize_ostree_root(state: &State, root_setup: &RootSetup) -> Result
     }
 
     let sysroot = {
-        let path = format!("/proc/self/fd/{}", rootfs_dir.as_fd().as_raw_fd());
+        let path = format!(
+            "/proc/{}/fd/{}",
```
cgwalters, line 629: I'm OK with this but I think the real fix here is that everything in ostree should be using fd-relative access.

crates/lib/src/install/config.rs
```diff
@@ -73,6 +73,8 @@ pub(crate) struct InstallConfiguration {
     pub(crate) kargs: Option<Vec<String>>,
     /// Supported architectures for this configuration
     pub(crate) match_architectures: Option<Vec<String>>,
+    /// Bootloader backend to use
+    pub(crate) bootloader: Option<String>,
```
cgwalters: Hmm, ideally there's a way we can detect this automatically instead. 

In ostree wouldn't it be as simple as saying "if there's a /usr/bin/aboot-deploy" then use it?

crates/lib/src/install.rs
cgwalters: :arrow_right:  https://github.com/ostreedev/ostree/pull/3507

crates/lib/src/install/config.rs
alexlarsson: I'll have a look at what is possible here. But sure, that would be nice.

alexlarsson: I don't think we want to auto-detect the aboot bootloader if bootloader is set to "none" (maybe if its "auto", but that is not what bootc is doing), because it is fundamentally a pretty destructive operation. I.e. if aboot is enabled the "ostree deploy will potentially dd stuff to a raw partition, and even though its unlikely to happen accidentally I don't fee like it is worth risking it. Not to mention that it is a large change in ostree behavior.

However, I think you're right that we want to automatically detect this by just looking at the image. But we should probably do that in "bootc install" instead. I'll have a look at this.

crates/lib/src/install.rs
alexlarsson: The issue here is that the path passed to ostree here ends up in a commandline when ostree spawns aboot-deploy. I don't see how using fd-relative access inside ostree itself can fix that.

crates/lib/src/install/config.rs
alexlarsson: I pushed a new version that does this instead

crates/lib/src/install.rs
cgwalters: > I don't see how using fd-relative access inside ostree itself can fix that.

Because ostree turns the provided path into a fd which we should be using instead as the above PR does, right?

crates/lib/src/install/config.rs
cgwalters: > I pushed a new version that does this instead

I don't see any changes like that?

> I don't think we want to auto-detect the aboot bootloader if bootloader is set to "none" (maybe if its "auto", but that is not what bootc is doing), because it is fundamentally a pretty destructive operation. 

Yes agree re historical ostree behavior, but note that bootc is basically overriding it by always setting `none` today. I think it does make sense to re-consider how we do autodetection at least in the bootc context because (unlike ostree) there actually *is* an explicit install operation.

cgwalters: This all said, I'm overall OK to just merge this as is

crates/lib/src/install.rs
alexlarsson: What happens currently (without this change) is that the ostree deploy code will spawn `/usr/bin/aboot-deploy -r /proc/self/fd/12`.  And then when aboot-deploy tries to read this, /proc/self/fd/12 no longer points to the correct thing.

For sure, we could make ostree::Sysroot work with a dirfd for the root path, but it would run into the same problem when it tried to spawn aboot-deploy.

crates/lib/src/install/config.rs
alexlarsson: > > I pushed a new version that does this instead
> 
> I don't see any changes like that?

I'm pretty sure I pushed it? Weird. Anyway, pushed it now.

cgwalters:
At first I thought this was a flake

> Error: creating build container: unable to copy from source docker://quay.io/centos-bootc/centos-bootc:stream9: copying system image from manifest list: reading blob sha256:7573f97f5d8045fa19daa4df963f4b59bf4b55738984b06f4dab6923c5e5af1b: Digest did not match, expected sha256:7573f97f5d8045fa19daa4df963f4b59bf4b55738984b06f4dab6923c5e5af1b, got sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855

But thinking about it a bit more I don't know how we'd hit that unless there was actually an incorrect image on the registry. But I can't reproduce it pulling locally.

Issue #1498, opened 2025-08-07
Sealed image build UX


jeckersb:
(Spawned from [composefs-rs integration tracker](https://github.com/bootc-dev/bootc/issues/1190))
(See also: https://github.com/containers/composefs-rs/pull/143)

The repo for composefs-rs has [many examples](https://github.com/containers/composefs-rs/tree/main/examples) that sketch out proof-of-concept image builds.  However, I think we need to improve on the user experience of building sealed images.  It's gotta be something better than "copy and paste these example Containerfiles".  So here's the place where we'll debate and design how this might work "for real".

My initial thought is to follow the prior art of the rechunker.  Currently that's made up of a few parts:

- `bootc-base-imagectl rechunk` at the top-level, which ultimately calls...
- `rpm-ostree compose build-chunked-oci`

And then rechunk just takes `--from` and `--to` imgrefs and does the conversion automatically between the original and rechunked format.  Ideally we should have a similar workflow for building sealed images:

- `bootc-base-imagectl seal --from quay.io/example/my-img:unsealed --to quay.io/example/my-img:sealed` that calls some new code like...
- `bootc internals build-sealed-image` (naming very bikesheddable)

And as a related note (this will eventually be spun-off into its own sub-issue) we'll need to expand `bootc container lint` to be able to lint sealed images.

cgwalters:
We should give an example with the Secure Boot keys involved here but yep I think this is sane!

allisonkarlitskaya:
I sort of like the approach as well.  Huge bonus points if we could avoid having to do the second stage container build.

The actual work required to seal an image is basically to generate the UKI, injecting the appropriate `composefs=` argument into the commandline, sign that UKI, then store it in `/boot`.  There's nothing stopping us from doing any of that outside of the container, assuming the correct tooling is available on the host.  We could even do that as an operation on composefs's tree structure, but getting it back into the container world would be difficult.  We don't even have support for creating layers in composefs.  We've discussed push support in composefs in the past, but this would be a lot of work.

More realistically we should probably just shell out directly to buildah to derive the source image, adding the signed UKI to it, then committing to get our new image with the UKI included.

allisonkarlitskaya:
Related: we also set the label for the (non-boot) fsverity integrity at the time we do this sealing step.  This is the old `cfsctl oci seal` operation, but we'd do it in podman's world instead of our own.

allisonkarlitskaya:
[...]

PR #1533, opened 2025-08-19
examples: Add bootc UKI & BLS examples

jeckersb:
This is a 1-for-1 copy of https://github.com/containers/composefs-rs/pull/143/files, re: https://github.com/bootc-dev/bootc/issues/1498#issuecomment-3201471070 I'm posting this here as a draft so we have somewhere to add public comments so we can discuss where this needs to change.


examples/bootc-uki/build.base
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+set -eux
+
+cd "${0%/*}"
+
+cargo build --release --features=pre-6.15 --bin cfsctl --bin composefs-setup-root
+
+cp ../../target/release/cfsctl .
```
jeckersb: Should use [bootc internals cfs](https://github.com/bootc-dev/bootc/pull/1477) instead

examples/bootc-uki/build.base
```diff
@@ -0,0 +1,18 @@
...
+cp ../../target/release/composefs-setup-root extra/usr/lib/dracut/modules.d/37composefs/
```
jeckersb: Hmm, what to do about this?  Fold it into `bootc internals` as well?  We had discussion recently re: binary size but I can't find it offhand...

examples/bootc-bls/build
```diff
@@ -0,0 +1,17 @@
...
+
+cp ../../target/release/cfsctl .
+cp ../../target/release/composefs-setup-root extra/usr/lib/dracut/modules.d/37composefs/
+
+mkdir -p tmp
+
+sudo podman build \
+    -t quay.io/fedora/fedora-bootc-bls:42 \
+    -f Containerfile \
+    --iidfile=tmp/iid \
```
p5: This command seems to be cut off.  Missing a `.`

examples/bootc-uki/build.base
jeckersb: oh right this is https://github.com/bootc-dev/bootc/pull/1500

examples/bootc-bls/Containerfile
p5, line 1: ~~Would it be possible to start with a quay.io/fedora/fedora image?~~  Or can we remove the ostree "bits" from the image within the Containerfile so we're certain it's not being relied on (e.g. `/ostree`)?

Or maybe longer term, there will be a separate Fedora image dedicated for bootc with cfs?

examples/bootc-uki/extra/etc/dracut.conf.d/no-xattr.conf
```diff
@@ -0,0 +1 @@
+export DRACUT_NO_XATTR=1
```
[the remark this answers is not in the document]
p5: Oh, seems to be issues when building on a btrfs according to dracut GitHub issues.  Does this flag cause xattr to not work in the same way on a booted system, or just the build?

examples/bootc-uki/build.base
cgwalters: Yes I commented on that in https://github.com/containers/composefs-rs/pull/143/files#r2269645005

Johan-Liebert1:
Instead of two containerfiles, we can have a two stage build in a single Containerfile; similar to what the UKI build examples in `composefs-rs` does

examples/bootc-uki/build.final
```diff
@@ -0,0 +1,47 @@
...
+    -f Containerfile.stage2 \
+    --secret=id=key,src=secureboot/db.key \
+    --secret=id=cert,src=secureboot/db.crt \
+    --iidfile=tmp/iid2
+
+rm -rf tmp/efi
+mkdir -p tmp/efi
+./cfsctl --repo tmp/sysroot/composefs oci pull containers-storage:"${IMAGE_ID}"
+./cfsctl --repo tmp/sysroot/composefs oci compute-id --bootable "${IMAGE_ID}"
+./cfsctl --repo tmp/sysroot/composefs oci prepare-boot "${IMAGE_ID}" --bootdir tmp/efi
```
jeckersb: These use `IMAGE_ID` from above but (if I'm following correctly) we should re-compute `IMAGE_ID` based on `tmp/iid2` above and use that instead?

examples/bootc-bls/Containerfile
cgwalters: Dropping `/ostree` is tracked at https://gitlab.com/fedora/bootc/base-images/-/issues/58

That said though note today the https://docs.fedoraproject.org/en-US/bootc/building-from-scratch/ already produces images without it, but that flow has the downside of producing a single large layer, which leads to https://github.com/coreos/rpm-ostree/issues/5383

examples/bootc-uki/build.final
Johan-Liebert1: The verity for both `iid1` and `iid2` should be identical, as they both (essentially) have the same contents. The only difference is that in the second step, we build the UKI and put it inside `/boot` which is ignored by `composefs-rs` while creating the EROFS image. 

Actually, building the UKI might introduce some artifacts which we do not want to take into account

jeckersb:
both the bls and uki examples build successfully now, next item is to wire in tests to ensure these continue building properly in CI

Issue #1190, opened 2025-03-11
composefs-rs integration in bootc


travier:
The [composefs-rs project](https://github.com/containers/composefs-rs) is a Rust implementation of composefs that is capable of generating composefs images from container images.

We should integrate it in bootc as an alternative to the ostree backend. This would help make progress on phasing out ostree, UKI support and unified storage:

- https://github.com/bootc-dev/bootc/issues/806
- https://github.com/bootc-dev/bootc/issues/20

To be able to do that, we need to make bootc capable of handling both repository formats and have it handle the transition from ostree to pure composefs.

A potential layout for this is discussed in https://github.com/containers/composefs-rs/issues/38.

Here are suggested steps for creating a first proof of concept implementation:

- Add an option to bootc switch
    - `bootc switch --composefs quay.io://foo:bar`
    - bootc will import the container image using the composefs-rs library in a dedicated composefs repo
    - bootc will set up the repo as needed
- bootc will create a new "deployment" for this image
    - Do the three way merge for /etc, comparing previous image, new image, current changes
    - Or use overlayfs instead to do that for /etc
    - Setup /var so that it's shared with ostree deployments
- bootc will setup the new deployment for the next boot
    - Non-UKI case:
        - GRUB: Install the kernel & initrd in /boot and setup the BLS config
        -  systemd-boot: Install the kernel & initrd in /boot/efi and setup the BLS config
    - UKI case:
        - GRUB: Generate GRUB config snippet to boot the UKIs in order
        - systemd-boot: Install the UKI in /boot/efi/EFI/Linux (order handling to be confirmed)

cgwalters:
I added the [area/composefs label](https://github.com/bootc-dev/bootc/issues?q=state%3Aopen%20label%3A%22area%2Fcomposefs%22) to this, there's a few related things there, but especially https://github.com/bootc-dev/bootc/pull/935 will step towards this from the other direction and I'd like to try to land that relatively early on.

> progress on phasing out ostree

And yes while this is definitely a longer term plan there's *A LOT* of stuff there; for example, we document configuring ostree-prepare-root today, and so we'll need to continue honoring configs for `/usr/lib/ostree/prepare-root.conf` into the forseeable future.

https://bootc-dev.github.io/bootc/filesystem.html#enabling-state-overlays is another one.

cgwalters:
BTW https://github.com/bootc-dev/bootc/pull/935 is updated and working now at least.

cgwalters:
[...]

PR #1541, opened 2025-08-23
install: store boot assets in ESP for composefs bls install

p5:
Please note: I do not currently have an environment to test this with GRUB.  

The existing code copied the boot entries, kernel and initrd into the rootfs rather than the ESP.  This meant systemd-boot was unable to see them, and is not compliant with the BLS spec.  From my brief investigation, it seems like grub handles this fine, however systemd-boot does not.  

I am aware this is different to how the existing ostree backend works today, however this change makes the implementation more BLS-compliant (which it seems like we're wanting to be for the composefs backend) due to the following requirement:

> When Type #1 boot loader menu entry snippets refer to other files (for linux, initrd, efi, devicetree, and devicetree-overlay), those files must be located on the same partition, and the paths must be absolute paths relative to the root of that file system.

In my mind I would wonder, if this was an intentional decision, why we would want bootc to even need to be aware of whether sd-boot or grub is being used after the initial install as my understanding is that grub _can_ act in a BLS-compliant way.  Therefore, a single BLS-compliant implementation that works across grub and sd-boot would be simpler overall (though I'm more than happy to be corrected on this).

The majority of this code was created referencing the existing `setup_composefs_uki_boot()` function.

Also, please note that I don't really know what I'm doing when it comes to bootloaders, so there may be a better approach to solve this problem.

Assisted by: ChatGPT (for conceptual knowledge)
Assisted by: GitHub Copilot gpt-4.1 (for code completion and rust syntax help)

### DIFF
--- a/crates/lib/src/install.rs
+++ b/crates/lib/src/install.rs
@@ -98,6 +98,8 @@ use bootc_mount::{inspect_filesystem, Filesystem};
 
 /// The toplevel boot directory
 const BOOT: &str = "boot";
+/// The EFI Linux directory
+const EFI_LINUX: &str = "EFI/Linux";
 /// Directory for transient runtime state
 #[cfg(feature = "install-to-disk")]
 const RUN_BOOTC: &str = "/run/bootc";
@@ -1691,7 +1693,7 @@ pub(crate) fn setup_composefs_bls_boot(
 ) -> Result<String> {
     let id_hex = id.to_hex();
 
-    let (root_path, cmdline_refs) = match setup_type {
+    let (esp_device, cmdline_refs) = match setup_type {
         BootSetupType::Setup((root_setup, state)) => {
             // root_setup.kargs has [root=UUID=<UUID>, "rw"]
             let mut cmdline_options = String::from(root_setup.kargs.join(" "));
@@ -1705,23 +1707,53 @@ pub(crate) fn setup_composefs_bls_boot(
                 }
             };
 
-            (root_setup.physical_root_path.clone(), cmdline_options)
+            // Locate ESP partition device
+            let esp_part = root_setup
+                .device_info
+                .partitions
+                .iter()
+                .find(|p| p.parttype.as_str() == ESP_GUID)
+                .ok_or_else(|| anyhow::anyhow!("ESP partition not found"))?;
+
+            (esp_part.node.clone(), cmdline_options)
         }
 
-        BootSetupType::Upgrade => (
-            Utf8PathBuf::from("/sysroot"),
-            vec![
-                format!("root=UUID={DPS_UUID}"),
-                RW_KARG.to_string(),
-                format!("{COMPOSEFS_CMDLINE}={id_hex}"),
-            ]
-            .join(" "),
-        ),
+        BootSetupType::Upgrade => {
+            let sysroot = Utf8PathBuf::from("/sysroot");
+
+            let fsinfo = inspect_filesystem(&sysroot)?;
+            let parent_devices = find_parent_devices(&fsinfo.source)?;
+
+            let Some(parent) = parent_devices.into_iter().next() else {
+                anyhow::bail!("Could not find parent device for mountpoint /sysroot");
+            };
+
+            (
+                get_esp_partition(&parent)?.0,
+                vec![
+                    format!("root=UUID={DPS_UUID}"),
+                    RW_KARG.to_string(),
+                    format!("{COMPOSEFS_CMDLINE}={id_hex}"),
+                ]
+                .join(" "),
+            )
+        }
     };
 
-    let boot_dir = root_path.join("boot");
+    let temp_efi_dir = tempfile::tempdir()
+        .map_err(|e| anyhow::anyhow!("Failed to create temporary directory for EFI mount: {e}"))?;
+    let mounted_efi = temp_efi_dir.path().to_path_buf();
+
+    Command::new("mount")
+        .args([&PathBuf::from(&esp_device), &mounted_efi])
+        .log_debug()
+        .run_inherited_with_cmd_context()
+        .context("Mounting EFI")?;
+
     let is_upgrade = matches!(setup_type, BootSetupType::Upgrade);
 
+    let efi_dir = Utf8PathBuf::from_path_buf(mounted_efi.join(EFI_LINUX))
+        .map_err(|_| anyhow::anyhow!("EFI dir is not valid UTF-8"))?;
     let (bls_config, boot_digest) = match &entry {
         ComposefsBootEntry::Type1(..) => unimplemented!(),
         ComposefsBootEntry::Type2(..) => unimplemented!(),
@@ -1735,16 +1767,16 @@ pub(crate) fn setup_composefs_bls_boot(
             bls_config.title = Some(id_hex.clone());
             bls_config.sort_key = Some("1".into());
             bls_config.machine_id = None;
-            bls_config.linux = format!("/boot/{id_hex}/vmlinuz");
-            bls_config.initrd = vec![format!("/boot/{id_hex}/initrd")];
+            bls_config.linux = format!("/{EFI_LINUX}/{id_hex}/vmlinuz");
+            bls_config.initrd = vec![format!("/{EFI_LINUX}/{id_hex}/initrd")];
             bls_config.options = Some(cmdline_refs);
             bls_config.extra = HashMap::new();
 
             if let Some(symlink_to) = find_vmlinuz_initrd_duplicates(&boot_digest)? {
-                bls_config.linux = format!("/boot/{symlink_to}/vmlinuz");
-                bls_config.initrd = vec![format!("/boot/{symlink_to}/initrd")];
+                bls_config.linux = format!("/{EFI_LINUX}/{symlink_to}/vmlinuz");
+                bls_config.initrd = vec![format!("/{EFI_LINUX}/{symlink_to}/initrd")];
             } else {
-                write_bls_boot_entries_to_disk(&boot_dir, id, usr_lib_modules_vmlinuz, &repo)?;
+                write_bls_boot_entries_to_disk(&efi_dir, id, usr_lib_modules_vmlinuz, &repo)?;
             }
 
             (bls_config, boot_digest)
@@ -1757,43 +1789,55 @@ pub(crate) fn setup_composefs_bls_boot(
 
         // This will be atomically renamed to 'loader/entries' on shutdown/reboot
         (
-            boot_dir.join(format!("loader/{STAGED_BOOT_LOADER_ENTRIES}")),
+            mounted_efi.join(format!("loader/{STAGED_BOOT_LOADER_ENTRIES}")),
             Some(booted_bls),
         )
     } else {
-        (boot_dir.join(format!("loader/{BOOT_LOADER_ENTRIES}")), None)
+        (
+            mounted_efi.join(format!("loader/{BOOT_LOADER_ENTRIES}")),
+            None,
+        )
     };
 
     create_dir_all(&entries_path).with_context(|| format!("Creating {:?}", entries_path))?;
 
-    let loader_entries_dir =
-        cap_std::fs::Dir::open_ambient_dir(&entries_path, cap_std::ambient_authority())
-            .with_context(|| format!("Opening {entries_path}"))?;
+    // Scope to allow for proper unmounting
+    {
+        let loader_entries_dir =
+            cap_std::fs::Dir::open_ambient_dir(&entries_path, cap_std::ambient_authority())
+                .with_context(|| format!("Opening {entries_path:?}"))?;
 
-    loader_entries_dir.atomic_write(
-        // SAFETY: We set sort_key above
-        format!(
-            "bootc-composefs-{}.conf",
-            bls_config.sort_key.as_ref().unwrap()
-        ),
-        bls_config.to_string().as_bytes(),
-    )?;
-
-    if let Some(booted_bls) = booted_bls {
         loader_entries_dir.atomic_write(
             // SAFETY: We set sort_key above
             format!(
                 "bootc-composefs-{}.conf",
-                booted_bls.sort_key.as_ref().unwrap()
+                bls_config.sort_key.as_ref().unwrap()
             ),
-            booted_bls.to_string().as_bytes(),
+            bls_config.to_string().as_bytes(),
         )?;
+
+        if let Some(booted_bls) = booted_bls {
+            loader_entries_dir.atomic_write(
+                // SAFETY: We set sort_key above
+                format!(
+                    "bootc-composefs-{}.conf",
+                    booted_bls.sort_key.as_ref().unwrap()
+                ),
+                booted_bls.to_string().as_bytes(),
+            )?;
+        }
+
+        let owned_loader_entries_fd = loader_entries_dir
+            .reopen_as_ownedfd()
+            .context("Reopening as owned fd")?;
+        rustix::fs::fsync(owned_loader_entries_fd).context("fsync")?;
     }
 
-    let owned_loader_entries_fd = loader_entries_dir
-        .reopen_as_ownedfd()
-        .context("Reopening as owned fd")?;
-    rustix::fs::fsync(owned_loader_entries_fd).context("fsync")?;
+    Command::new("umount")
+        .arg(&mounted_efi)
+        .log_debug()
+        .run_inherited_with_cmd_context()
+        .context("Unmounting EFI")?;
 
     Ok(boot_digest)
 }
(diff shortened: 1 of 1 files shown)


p5:
Can confirm with this patch set (plus https://github.com/coreos/bootupd/pull/978 & the below), we are able to produce a bootable Chainguard Wolfi disk image with sd-boot.

```
local-bootc-patch on  local-bootc-patch [!?] 
❯ sudo losetup -Pf --show bootable.img
/dev/loop0

local-bootc-patch on  local-bootc-patch [!?] 
❯ sudo mount /dev/loop0p2 /var/mnt/esp

local-bootc-patch on  local-bootc-patch [!?] 
❯ tree /var/mnt/esp  
/var/mnt/esp
├── EFI
│   ├── BOOT
│   │   └── BOOTX64.EFI
│   ├── Linux
│   │   └── 4b4eb53d1cab36f96d8f10864d98096a8bdabea944abf1fdf6fb02496fcacf88
│   │       ├── initrd
│   │       └── vmlinuz
│   └── systemd
│       └── systemd-bootx64.efi
└── loader
    ├── entries
    │   └── bootc-composefs-1.conf
    ├── entries.srel
    ├── keys
    ├── loader.conf
    └── random-seed
```

Additional bootc patch, dependent on bootupd:

<details>

```
From 49326238bbc7b3e441050647b023c4032139dde7 Mon Sep 17 00:00:00 2001
From: Robert Sturla <robertsturla@outlook.com>
Date: Sun, 24 Aug 2025 17:49:05 +0100
Subject: [PATCH] install: pass through bootupd bootloader args

Signed-off-by: Robert Sturla <robertsturla@outlook.com>
---
 crates/lib/src/bootloader.rs | 9 +++++++++
 crates/lib/src/install.rs    | 4 ++++
 2 files changed, 13 insertions(+)

diff --git a/crates/lib/src/bootloader.rs b/crates/lib/src/bootloader.rs
index 0f02198a..93fb7a30 100644
--- a/crates/lib/src/bootloader.rs
+++ b/crates/lib/src/bootloader.rs
@@ -28,12 +28,21 @@ pub(crate) fn install_via_bootupd(
     } else {
         vec![]
     };
+
+    // If configopts.bootloader is systemd-boot, pass in the `--bootloader systemd-boot` flag to bootupctl
+    let mut bootloader_args = vec![];
+    if configopts.bootloader == Some("systemd-boot".into()) {
+        bootloader_args.push("--bootloader");
+        bootloader_args.push("systemd-boot");
+    }
+
     let devpath = device.path();
     println!("Installing bootloader via bootupd");
     Command::new("bootupctl")
         .args(["backend", "install", "--write-uuid"])
         .args(verbose)
         .args(bootupd_opts.iter().copied().flatten())
+        .args(bootloader_args)
         .args(src_root_arg)
         .args(["--device", devpath.as_str(), rootfs.as_str()])
         .log_debug()
diff --git a/crates/lib/src/install.rs b/crates/lib/src/install.rs
index c058768f..7e313960 100644
--- a/crates/lib/src/install.rs
+++ b/crates/lib/src/install.rs
@@ -252,6 +252,10 @@ pub(crate) struct InstallConfigOpts {
     #[clap(long)]
     pub(crate) stateroot: Option<String>,
+
+    /// The bootloader to use
+    #[clap(long)]
+    pub(crate) bootloader: Option<String>,
 }

 #[derive(
--
2.50.1
```

</details>

crates/lib/src/install.rs
```diff
@@ -1705,23 +1705,60 @@ pub(crate) fn setup_composefs_bls_boot(
...
+
+    Task::new("Mounting ESP", "mount")
+        .args([&PathBuf::from(&esp_device), &mounted_esp.clone()])
+        .run()?;
+
+    create_dir_all(&mounted_esp).context("Failed to create dir {mounted_esp:?}")?;
+
+    Task::new("Mounting ESP", "mount")
+        .args([&PathBuf::from(&esp_device), &mounted_esp.clone()])
+        .run()?;
```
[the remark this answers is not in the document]
p5: Nice spot!  Removed the duplicate ESP mounting
[fdc7199](https://github.com/bootc-dev/bootc/pull/1541/commits/fdc7199bc542d3d16386854201a8ed72fe7b8c67)

p5:
I tried to implement this in a slightly different way - with the vmlinuz and initrd in rootfs, and the boot entries in ESP that point to them.  Thinking it might not have been a mistake directing these files to rootfs.  But unfortunately this failed to boot.

Unsuccessful patch (applied ontop of [fdc7199](https://github.com/bootc-dev/bootc/pull/1541/commits/fdc7199bc542d3d16386854201a8ed72fe7b8c67)):
```
From 346fa791c9b405a0402a79dac22428b50146b4ea Mon Sep 17 00:00:00 2001
From: Robert Sturla <robertsturla@outlook.com>
Date: Sun, 24 Aug 2025 20:44:51 +0100
Subject: [PATCH] store kernel/initrd in rootfs

---
 crates/lib/src/install.rs | 82 +++++++++++++++++++++------------------
 1 file changed, 45 insertions(+), 37 deletions(-)

diff --git a/crates/lib/src/install.rs b/crates/lib/src/install.rs
index 47228e7a..c058768f 100644
--- a/crates/lib/src/install.rs
+++ b/crates/lib/src/install.rs
@@ -1633,47 +1633,49 @@ fn find_vmlinuz_initrd_duplicates(digest: &str) -> Result<Option<String>> {
     Ok(symlink_to)
 }
 
-#[context("Writing BLS entries to disk")]
-fn write_bls_boot_entries_to_disk(
-    boot_dir: &Utf8PathBuf,
+/// Write vmlinuz and initrd to the rootfs boot directory
+#[context("Writing vmlinuz/initrd to rootfs")]
+fn write_vmlinuz_initrd_to_rootfs(
+    rootfs_boot_dir: &Utf8PathBuf,
     deployment_id: &Sha256HashValue,
     entry: &UsrLibModulesVmlinuz<Sha256HashValue>,
     repo: &ComposefsRepository<Sha256HashValue>,
 ) -> Result<()> {
     let id_hex = deployment_id.to_hex();
-
-    // Write the initrd and vmlinuz at /boot/<id>/
-    let path = boot_dir.join(&id_hex);
+    let path = rootfs_boot_dir.join(&id_hex);
     create_dir_all(&path)?;
-
-    let entries_dir = cap_std::fs::Dir::open_ambient_dir(&path, cap_std::ambient_authority())
+    let dir = cap_std::fs::Dir::open_ambient_dir(&path, cap_std::ambient_authority())
         .with_context(|| format!("Opening {path}"))?;
-
-    entries_dir
-        .atomic_write(
-            "vmlinuz",
-            read_file(&entry.vmlinuz, &repo).context("Reading vmlinuz")?,
-        )
-        .context("Writing vmlinuz to path")?;
-
+    dir.atomic_write(
+        "vmlinuz",
+        read_file(&entry.vmlinuz, &repo).context("Reading vmlinuz")?,
+    ).context("Writing vmlinuz to rootfs")?;
     let Some(initramfs) = &entry.initramfs else {
         anyhow::bail!("initramfs not found");
     };
+    dir.atomic_write(
+        "initrd",
+        read_file(initramfs, &repo).context("Reading initrd")?,
+    ).context("Writing initrd to rootfs")?;
+    let owned_fd = dir.reopen_as_ownedfd().context("Reopen as owned fd")?;
+    rustix::fs::fsync(owned_fd).context("fsync rootfs boot dir")?;
+    Ok(())
+}
 
-    entries_dir
-        .atomic_write(
-            "initrd",
-            read_file(initramfs, &repo).context("Reading initrd")?,
-        )
-        .context("Writing initrd to path")?;
-
-    // Can't call fsync on O_PATH fds, so re-open it as a non O_PATH fd
-    let owned_fd = entries_dir
-        .reopen_as_ownedfd()
-        .context("Reopen as owned fd")?;
-
-    rustix::fs::fsync(owned_fd).context("fsync")?;
-
+/// Write BLS entry to ESP, referencing vmlinuz/initrd in rootfs
+#[context("Writing BLS entry to ESP")]
+fn write_bls_entry_to_esp(
+    esp_dir: &Utf8PathBuf,
+    bls_config: &BLSConfig,
+) -> Result<()> {
+    let entries_dir = cap_std::fs::Dir::open_ambient_dir(esp_dir, cap_std::ambient_authority())
+        .with_context(|| format!("Opening {esp_dir}"))?;
+    entries_dir.atomic_write(
+        format!("bootc-composefs-{}.conf", bls_config.sort_key.as_ref().unwrap()),
+        bls_config.to_string().as_bytes(),
+    )?;
+    let owned_fd = entries_dir.reopen_as_ownedfd().context("Reopen as owned fd")?;
+    rustix::fs::fsync(owned_fd).context("fsync ESP dir")?;
     Ok(())
 }
 
@@ -1768,18 +1770,24 @@ pub(crate) fn setup_composefs_bls_boot(
             bls_config.title = Some(id_hex.clone());
             bls_config.sort_key = Some("1".into());
             bls_config.machine_id = None;
-            bls_config.linux = format!("/EFI/Linux/{id_hex}/vmlinuz");
-            bls_config.initrd = vec![format!("/EFI/Linux/{id_hex}/initrd")];
+            bls_config.linux = format!("/boot/{id_hex}/vmlinuz");
+            bls_config.initrd = vec![format!("/boot/{id_hex}/initrd")];
             bls_config.options = Some(cmdline_refs);
             bls_config.extra = HashMap::new();
 
             if let Some(symlink_to) = find_vmlinuz_initrd_duplicates(&boot_digest)? {
-                bls_config.linux = format!("/EFI/Linux/{symlink_to}/vmlinuz");
-                bls_config.initrd = vec![format!("/EFI/Linux/{symlink_to}/initrd")];
+                bls_config.linux = format!("/boot/{symlink_to}/vmlinuz");
+                bls_config.initrd = vec![format!("/boot/{symlink_to}/initrd")];
             } else {
-                let efi_dir_utf8 = Utf8PathBuf::from_path_buf(efi_dir.clone())
-                    .map_err(|_| anyhow::anyhow!("EFI dir is not valid UTF-8"))?;
-                write_bls_boot_entries_to_disk(&efi_dir_utf8, id, usr_lib_modules_vmlinuz, &repo)?;
+                // Write vmlinuz/initrd to rootfs boot dir
+                let rootfs_boot_dir = root_path.join("boot");
+                write_vmlinuz_initrd_to_rootfs(&rootfs_boot_dir, id, usr_lib_modules_vmlinuz, &repo)?;
+                // Write BLS entry to ESP/loader/entries
+                let loader_entries_dir = mounted_esp.join("loader/entries");
+                create_dir_all(&loader_entries_dir).context("Creating loader/entries in ESP")?;
+                let loader_entries_dir_utf8 = Utf8PathBuf::from_path_buf(loader_entries_dir)
+                    .map_err(|_| anyhow::anyhow!("loader/entries dir is not valid UTF-8"))?;
+                write_bls_entry_to_esp(&loader_entries_dir_utf8, &bls_config)?;
             }
 
             (bls_config, boot_digest)
-- 
2.50.1
```

This does make sense though, as we're aiming for bls-compliance, and the bls specification requires all assets referenced in a entry file to be on the ESP.

> When Type #1 boot loader menu entry snippets refer to other files (for linux, initrd, efi, devicetree, and devicetree-overlay), those files must be located on the same partition, and the paths must be absolute paths relative to the root of that file system.

cgwalters:
> In my mind I would wonder, if this was an intentional decision, why we would want bootc to even need to be aware of whether sd-boot or grub is being used after the initial install as my understanding is that grub can act in a BLS-compliant way. Therefore, a single BLS-compliant implementation that works across grub and sd-boot would be simpler overall (though I'm more than happy to be corrected on this).

Yes I think this is the ideal we want to get to, is that grub can be made to just implement BLS.

cgwalters:
> Assisted by: ChatGPT (for conceptual knowledge)
> Assisted by: GitHub Copilot gpt-4.1 (for code completion and rust syntax help)

Thank you so much for adding these tags! It's very helpful to know.

crates/lib/src/install.rs
```diff
@@ -1705,23 +1705,56 @@ pub(crate) fn setup_composefs_bls_boot(
...
+        }
     };
 
-    let boot_dir = root_path.join("boot");
+    let mounted_esp: PathBuf = root_path.join("esp").into();
+    let esp_mount_point_existed = mounted_esp.exists();
+
+    create_dir_all(&mounted_esp).context("Failed to create dir {mounted_esp:?}")?;
+
+    Task::new("Mounting ESP", "mount")
```
cgwalters: I've been trying to slowly get rid of `Task`, see https://github.com/bootc-dev/bootc/pull/1475 etc

crates/lib/src/install.rs
```diff
@@ -1705,23 +1705,56 @@ pub(crate) fn setup_composefs_bls_boot(
...
+                    RW_KARG.to_string(),
+                    format!("{COMPOSEFS_CMDLINE}={id_hex}"),
+                ]
+                .join(" "),
+            )
+        }
     };
 
-    let boot_dir = root_path.join("boot");
+    let mounted_esp: PathBuf = root_path.join("esp").into();
```
cgwalters: The recommended ESP mount point is `/efi`, not `/esp` https://uapi-group.org/specifications/specs/boot_loader_specification/#mount-points

Regardless though it's a bit ugly for us to mutate the target disk image for this; we can mount it to a temporary directory instead.

(Actually though, what would be even nicer is to use the new mount API to get a detached FD instead...hmm, could add a helper that does that but falls back to `mkdtemp + mount + openat + umount(MNT_DETACH)` if we don't have `open_tree` or so...)

crates/lib/src/install.rs
```diff
@@ -1735,16 +1768,18 @@ pub(crate) fn setup_composefs_bls_boot(
             bls_config.title = Some(id_hex.clone());
             bls_config.sort_key = Some("1".into());
             bls_config.machine_id = None;
-            bls_config.linux = format!("/boot/{id_hex}/vmlinuz");
-            bls_config.initrd = vec![format!("/boot/{id_hex}/initrd")];
+            bls_config.linux = format!("/EFI/Linux/{id_hex}/vmlinuz");
```
cgwalters: Nonblocking but we should probably have a `const` for `EFI/Linux` or so

crates/lib/src/install.rs
```diff
@@ -1735,16 +1768,18 @@ pub(crate) fn setup_composefs_bls_boot(
...
             bls_config.extra = HashMap::new();
 
             if let Some(symlink_to) = find_vmlinuz_initrd_duplicates(&boot_digest)? {
-                bls_config.linux = format!("/boot/{symlink_to}/vmlinuz");
-                bls_config.initrd = vec![format!("/boot/{symlink_to}/initrd")];
+                bls_config.linux = format!("/EFI/Linux/{symlink_to}/vmlinuz");
+                bls_config.initrd = vec![format!("/EFI/Linux/{symlink_to}/initrd")];
             } else {
-                write_bls_boot_entries_to_disk(&boot_dir, id, usr_lib_modules_vmlinuz, &repo)?;
+                let efi_dir_utf8 = Utf8PathBuf::from_path_buf(efi_dir.clone())
```
cgwalters: It'd be cleaner to enforce this above i.e. make `efi_dir` already a `Utf8PathBuf`

p5:
Thanks for the reviews!  
All make sense, I'll need to learn how the new mount API works and how to fallback to the `mkdtemp + mount + openat + umount(MNT_DETACH)` for pre-6.15 kernels.

> Yes I think this is the ideal we want to get to, is that grub can be made to just implement BLS.

Perfect!  Glad there's not going to need to be much forking out based on the bootloader - the majority of the implementation will be identical between the two.

I wanted to get this confirmation first before I spent too much more time looking at this.

cgwalters:
> All make sense, I'll need to learn how the new mount API works and how to fallback to the mkdtemp + mount + openat + umount(MNT_DETACH) for pre-6.15 kernels.

Let's not block on that, just mounting to a tmpdir I think is totally fine.

p5:
Have pushed a new commit with the updates.  
Applied some of the changes to the pre-existing uki boot function too for consistency.

Will squash after I've confirmed this still works as expected.
Will be able to test this tonight, so please don't merge until after squashing.

---
May follow this up with a separate PR after I learn what the new mount API is.

p5:
Verified this created a bootable Wolfi image with sd-boot.  I don't anticipate this breaking UKIs since the same changes were made to both bls and uki functions.

Ready for review!
Will invoke Gemini for a sanity check.

crates/lib/src/install.rs
```diff
@@ -1705,23 +1707,53 @@ pub(crate) fn setup_composefs_bls_boot(
...
+    Command::new("mount")
+        .args([&PathBuf::from(&esp_device), &mounted_efi])
+        .log_debug()
+        .run_inherited_with_cmd_context()
+        .context("Mounting EFI")?;
+
     let is_upgrade = matches!(setup_type, BootSetupType::Upgrade);
 
+    let efi_dir = Utf8PathBuf::from_path_buf(mounted_efi.join("EFI/Linux"))
+        .map_err(|_| anyhow::anyhow!("EFI dir is not valid UTF-8"))?;
```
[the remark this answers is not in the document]
p5: Missed this one.  Updated in latest commit

travier:
> In my mind I would wonder, if this was an intentional decision, why we would want bootc to even need to be aware of whether sd-boot or grub is being used after the initial install as my understanding is that grub _can_ act in a BLS-compliant way. Therefore, a single BLS-compliant implementation that works across grub and sd-boot would be simpler overall (though I'm more than happy to be corrected on this).

Unfortunately GRUB support for BLS is incomplete enough that we can not do that.

There is a lot of discussion about UKI/systemd-boot support in the following issues:
- https://github.com/containers/composefs-rs/issues/38#issuecomment-3047969797
- https://github.com/bootc-dev/bootc/issues/1190

Issue #1544, opened 2025-08-25
Panic during sysroot operations (git main)


p5:
The current bootc version on git main (https://github.com/bootc-dev/bootc/commit/ae8930b1adbeafbc32015a9c8cd53c0a485e1b1f) panics during `bootc update`, `bootc status` and any other bootc operation that interacts with the sysroot lock.

```
❯ rpm -qa | grep bootc  
bootc-202508212033.gae8930b1ad-1.fc42.x86_64
```

```
❯ bootc update                                                                             
⠁ Waiting for sysroot lock...                                                                                                                                                                                       
thread 'main' panicked at /builddir/build/BUILD/bootc-202508212033.gae8930b1ad-build/bootc-202508212033.gae8930b1ad/vendor/gio/src/gio_future.rs:66:13:
Spawning futures only allowed if the thread is owning the MainContext
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I do not yet know which commit introduced this as I've been pinned on an older image for about a week.

p5:
This should also probably be followed up with an investigation as to why CI checks weren't able to catch these

p5:
A clean `cargo build` on ae8930b1adbeafbc32015a9c8cd53c0a485e1b1f does not produce an error.

I'm thinking the cause is likely `rpm-ostree update` running at the same time, although I thought this produced a different, more descriptive error.  Will look at, at least, not making it panic and potentially improving the error

jeckersb:
Hm I see this showing up in CI at the very least on aarch64 - https://artifacts.dev.testing-farm.io/1210de35-5213-4ba1-b8d1-7caa5b58d079/

```
thread 'main' panicked at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/gio-0.20.12/src/gio_future.rs:66:13:
Spawning futures only allowed if the thread is owning the MainContext
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error: nu::shell::core_dumped

  x External command core dumped
     ,-[/var/ARTIFACTS/work-test-20-local-upgradeogm5yztq/tmt/plans/integration/test-20-local-upgrade/discover/default-0/tests/tmt/tests/booted/test-image-pushpull-upgrade.nu:142:14]
 141 |     # Verify we staged an update
 142 |     let st = bootc status --json | from json
     :              ^^|^^
     :                `-- core dumped with SIGABRT (6)
[...]

Issue #1546, opened 2025-08-25
ci: fix/mitigate `/root/buildinfo`


cgwalters:
Our CI now broke with

```
+ bootc container lint --fatal-warnings
error: Linting: Checks failed: 1
Lint warning: var-tmpfiles: Found content in /var missing systemd tmpfiles.d entries:
  d /var/roothome/buildinfo 0755 root root - -
  d /var/roothome/buildinfo/content_manifests 0755 root root - -
Found non-directory/non-symlink files in /var:
  var/roothome/buildinfo/content_manifests/content-sets.json
```

See https://gitlab.com/redhat/centos-stream/containers/bootc/-/issues/1195

cgwalters:
OK and now we have *another* different failure

```

thread 'main' panicked at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/gio-0.20.12/src/gio_future.rs:66:13:
Spawning futures only allowed if the thread is owning the MainContext
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error: nu::shell::core_dumped

  x External command core dumped
     ,-[/var/ARTIFACTS/work-test-20-local-upgradeogm5yztq/tmt/plans/integration/test-20-local-upgrade/discover/default-0/tests/tmt/tests/booted/test-image-pushpull-upgrade.nu:142:14]
 141 |     # Verify we staged an update
 142 |     let st = bootc status --json | from json
     :              ^^|^^
     :                `-- core dumped with SIGABRT (6)
 143 |     let staged_digest = $st.status.staged.image.imageDigest
     `----

Connection to 13.59.156.65 closed.
```

jeckersb:
(xref https://github.com/bootc-dev/bootc/issues/1544#issuecomment-3221160805)

PR #1547, opened 2025-08-25
build-sys: Work around #1546

cgwalters:
While we work to fix the base image (again).

---

Closes: https://github.com/bootc-dev/bootc/issues/1544

### DIFF
--- a/crates/ostree-ext/src/sysroot.rs
+++ b/crates/ostree-ext/src/sysroot.rs
@@ -37,13 +37,9 @@ impl SysrootLock {
     /// immediately, a status message will be printed to standard output.
     /// The lock will be unlocked when this object is dropped.
     pub async fn new_from_sysroot(sysroot: &ostree::Sysroot) -> Result<Self> {
-        if sysroot.try_lock()? {
-            return Ok(Self {
-                sysroot: sysroot.clone(),
-                unowned: false,
-            });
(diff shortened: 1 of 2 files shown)


jeckersb:
re: c10s aarch64 failure - https://github.com/bootc-dev/bootc/issues/1544#issuecomment-3221160805 :thinking:

cgwalters:
Since we now have *two* different things breaking our CI, I've rolled a fix for the locking into this PR.

crates/ostree-ext/src/sysroot.rs
```diff
@@ -37,13 +37,9 @@ impl SysrootLock {
...
-        if sysroot.try_lock()? {
-            return Ok(Self {
-                sysroot: sysroot.clone(),
-                unowned: false,
-            });
-        }
-        async_task_with_spinner("Waiting for sysroot lock...", sysroot.lock_future()).await?;
+        let sysroot_clone = sysroot.clone();
+        let locker = tokio::task::spawn_blocking(move || sysroot_clone.lock());
+        async_task_with_spinner("Waiting for sysroot lock...", locker).await??;
```
cgwalters, line 42: This does for me basically always cause a flicker of "Waiting for sysroot lock..." on a terminal, but if someone cares we can change that.

PR #1551, opened 2025-08-26
crates/*: Fix most clippy lints

cgwalters:
Just a run of `cargo clippy --fix` plus some manual bits.

I keep going back and forth on clippy...it's really a spectrum between useful and noisy.


jeckersb (approved):
So close to having a bare `cargo clippy` be clean!  Last two things maybe for a followup?

```
warning: this function has too many arguments (8/7)
   --> crates/tmpfiles/src/lib.rs:318:1
    |
318 | / fn convert_path_to_tmpfiles_d_recurse<U: uzers::Users, G: uzers::Groups>(
319 | |     out_entries: &mut BTreeSet<String>,
320 | |     out_unsupported: &mut Vec<PathBuf>,
321 | |     users: &U,
...   |
326 | |     readonly: bool,
327 | | ) -> Result<()> {
    | |_______________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#too_many_arguments
    = note: `#[warn(clippy::too_many_arguments)]` on by default
```

Would be easy enough to refactor this to take a struct of options instead, I agree with clippy that 8 arguments is a bit much.

```
    Checking bootc-sysusers v0.1.0 (/var/home/jeckersb/git/bootc/crates/sysusers)
warning: `bootc-tmpfiles` (lib) generated 1 warning
error: use of a disallowed method `str::len`
   --> crates/sysusers/src/lib.rs:132:75
    |
132 |                 let idx = s.find(|c: char| c.is_whitespace()).unwrap_or(s.len());
    |                                                                           ^^^
    |
    = note: use <str>.as_bytes().len() instead
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#disallowed_methods
    = note: requested on the command line with `-D clippy::disallowed-methods`

error: could not compile `bootc-sysusers` (lib) due to 1 previous error
```

We explicitly check for this via clippy.toml, should fix.  Also an existing bug that the target for `validate-rust` does not include `-D clippy::disallowed-methods` which has let this sneak past CI.

cgwalters:
> Would be easy enough to refactor this to take a struct of options instead, I agree with clippy that 8 arguments is a bit much.

I don't have a strong opinion here, but I just allowed it for now.

> We explicitly check for this via clippy.toml, should fix. Also an existing bug that the target for validate-rust does not include -D clippy::disallowed-methods which has let this sneak past CI.

Done in https://github.com/bootc-dev/bootc/pull/1552

PR #1558, opened 2025-08-28
composefs/install/bls: Fix empty version in config

Johan-Liebert1:
Fll `version` field in generated BLS config


crates/lib/src/install.rs
```diff
@@ -2309,7 +2348,7 @@ async fn install_to_filesystem_impl(
         // Load a fd for the mounted target physical root
         let (id, verity) = initialize_composefs_repository(state, rootfs).await?;
 
-        tracing::warn!(
+        tracing::info!(
```
cgwalters, line 2351: Unrelated but shouldn't this actually be `debug!`?

Johan-Liebert1: I think we had a discussion some time ago that we want this to show up, but I might be misremembering. Either (info/debug) is fine with me

PR #1560, opened 2025-08-29
composefs/install: Copy /etc contents to state

Johan-Liebert1:
For bind mounting /etc we copy the contents of the EROFS' /etc to the deployment's state directory

Mounting the EORFS requires help from the initramfs crate, so we also turn it into a library crate.

### DIFF
--- a/crates/lib/src/bootc_composefs/state.rs
+++ b/crates/lib/src/bootc_composefs/state.rs
@@ -1,0 +1,47 @@
+use std::process::Command;
+
+use anyhow::{Context, Result};
+use bootc_utils::CommandRunExt;
+use camino::Utf8PathBuf;
+use fn_error_context::context;
+
+use rustix::{
+    fs::{open, Mode, OFlags, CWD},
+    mount::{unmount, UnmountFlags},
+    path::Arg,
+};
+
+/// Mounts an EROFS image and copies the pristine /etc to the deployment's /etc
+#[context("Copying etc")]
+pub(crate) fn copy_etc_to_state(
+    sysroot_path: &Utf8PathBuf,
+    erofs_id: &String,
+    state_path: &Utf8PathBuf,
+) -> Result<()> {
+    let sysroot_fd = open(
+        sysroot_path.as_std_path(),
+        OFlags::PATH | OFlags::DIRECTORY | OFlags::CLOEXEC,
+        Mode::empty(),
+    )
+    .context("Opening sysroot")?;
+
+    let composefs_fd = bootc_initramfs_setup::mount_composefs_image(&sysroot_fd, &erofs_id, false)?;
+
+    let tempdir = tempfile::tempdir().context("Creating tempdir")?;
+
+    bootc_initramfs_setup::mount_at_wrapper(composefs_fd, CWD, tempdir.path())?;
+
+    // TODO: Replace this with a function to cap_std_ext
+    let cp_ret = Command::new("cp")
+        .args([
+            "-a",
+            &format!("{}/etc/.", tempdir.path().as_str()?),
+            &format!("{state_path}/etc/."),
+        ])
+        .run_capture_stderr();
+
+    // Unmount regardless of copy succeeding
+    unmount(tempdir.path(), UnmountFlags::DETACH).context("Unmounting composefs")?;
+
+    cp_ret
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -253,6 +253,7 @@ dependencies = [
  "anstream",
  "anstyle",
  "anyhow",
+ "bootc-initramfs-setup",
  "bootc-internal-blockdev",
  "bootc-internal-utils",
  "bootc-kernel-cmdline",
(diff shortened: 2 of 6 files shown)


Johan-Liebert1:
I've started out by separating new composefs-backend related code to `lib/src/bootc-composefs`. @cgwalters we could go ahead with this approach or something else if you have something cleaner in mind.

crates/lib/src/bootc_composefs/state.rs
```diff
@@ -0,0 +1,50 @@
+use std::{fs::create_dir_all, process::Command};
+
+use anyhow::{Context, Result};
+use camino::Utf8PathBuf;
+use fn_error_context::context;
+
+use bootc_initramfs_setup;
```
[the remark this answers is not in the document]
Johan-Liebert1: Done

crates/lib/src/bootc_composefs/state.rs
```diff
@@ -0,0 +1,50 @@
...
+            &format!("{state_path}/etc/."),
+        ])
+        .output();
+
+    // Unmount regardless of copy succeeding
+    unmount(&dir, UnmountFlags::DETACH).context("Unmounting composefs")?;
+
+    copy_res?;
+
+    Ok(())
```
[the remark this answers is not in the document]
Johan-Liebert1: Updated

crates/lib/src/bootc_composefs/state.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    remove_dir_all(dir).context("Removing temp directory")?;
+
+    if !output.status.success() {
+        anyhow::bail!(
+            "Copying /etc failed with status {}: {}",
+            output.status,
+            String::from_utf8_lossy(&output.stderr)
+        );
+    }
```
cgwalters: We have a handy https://github.com/bootc-dev/bootc/blob/7434b0fb0c422c02d3d37d8d609e63dc30e68aff/crates/utils/src/command.rs#L38 that does all this

crates/lib/src/bootc_composefs/state.rs
```diff
@@ -0,0 +1,60 @@
...
+        Mode::empty(),
+    )
+    .context("Opening sysroot")?;
+
+    let composefs_fd = bootc_initramfs_setup::mount_composefs_image(&sysroot_fd, &erofs_id, false)?;
+
+    let uuid = uuid::Uuid::new_v4();
+    let dir = format!("/var/tmp/{uuid}");
+
+    create_dir_all(&dir).context("Creating temp directory")?;
```
cgwalters: Confused why we wouldn't just use `tempdir::tempdir`

crates/lib/src/bootc_composefs/state.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    bootc_initramfs_setup::mount_at_wrapper(composefs_fd, CWD, &dir)?;
```
cgwalters: Ah yeah, we have mount helper code in lib/...and in composefs-boot, probably again need to factor out a helper crate

crates/lib/src/bootc_composefs/state.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    let output = Command::new("cp")
```
cgwalters: I'm totally fine with this, but at some point I'd like to add a high level `copy_recursive` to e.g. cap-std-ext

crates/lib/src/bootc_composefs/state.rs
```diff
@@ -0,0 +1,60 @@
...
+        .args([
+            "-a",
+            &format!("{dir}/etc/."),
+            &format!("{state_path}/etc/."),
+        ])
+        .output()?;
+
+    // Unmount regardless of copy succeeding
```
cgwalters, line 43: Yeah I think what we want is a wrapper for tempdir that handles unmount-on-drop

crates/lib/src/bootc_composefs/state.rs
Johan-Liebert1: I see. I'll use that

crates/lib/src/bootc_composefs/state.rs
Johan-Liebert1: Composefs has a struct for exactly this, but again is only visible at crate level. We could make that public. https://github.com/containers/composefs-rs/blob/main/crates/composefs/src/mountcompat.rs#L129

The composefs-rs crate has a lot of helper functions which are private

crates/lib/src/bootc_composefs/state.rs
Johan-Liebert1: yes, that would be ideal

crates/lib/src/bootc_composefs/state.rs
```diff
@@ -0,0 +1,47 @@
...
+    // TODO: Replace this with a function to cap_std_ext
+    Command::new("cp")
+        .args([
+            "-a",
+            &format!("{}/etc/.", tempdir.path().as_str()?),
+            &format!("{state_path}/etc/."),
+        ])
+        .run_capture_stderr()?;
+
+    // Unmount regardless of copy succeeding
```
cgwalters, line 43: But now we lost the "unmount regardless" right?

Need to do

```
let r = run_cp().run_capture_stderr();
umount()
r?;
```

Johan-Liebert1: Ah, you're right. My bad, I missed it

Johan-Liebert1: Updated

PR #1570, opened 2025-09-01
 lint: warn about potential UID/GID drift under /etc; respect tmpfiles chown (z/Z) (#1562)

### LINKED ISSUE #1562: lint: Check for potential uid drift for files in `/etc`
Now that we have parsers for both `tmpfiles.d` and `sysusers.d`, we can detect the case where files in `/etc` are owned by a floating uid/gid and warn.

However we should *not* warn if there's a tmpfiles.d entry which does a chown, which is a standard workaround for components/packages which have floating uid/gids; e.g. we shouldn't warn on if we detect 
`/usr/lib/tmpfiles.d/polkit.conf:Z /etc/polkit-1/localauthority - polkitd polkitd`

(Humm actually this is tangential but it seems that rpm-ostree auto-synthesized 
```
cat /usr/lib/tmpfiles.d/polkit-tmpfiles.conf 
d /etc/polkit-1/rules.d 0750 root polkitd - -
```
which seems kind of wrong because it should actually be `Z` here probably)

rishi-jat:
### Description 
This adds a new container lint (etc-uid-drift) that warns when files or directories under /etc are owned by non-root users or groups that aren’t numerically pinned in systemd-sysusers and aren’t covered by systemd-tmpfiles chown rules. The goal is to help catch ownership that can “drift” across upgrades when /etc persists.

The lint parses usr/lib/sysusers.d to find users/groups with fixed numeric IDs, and scans usr/lib/tmpfiles.d for z/Z entries (chown at boot, either exact or recursive). It walks /etc (skipping symlinks, handling non‑UTF8 paths), and warns on entries where uid/gid is non-root, not pinned via sysusers, and not covered by a tmpfiles chown. For example, if a component uses a floating user but ships a tmpfiles rule like Z /etc/polkit-1/…, the lint will not warn.

This is a warning-only check (can be promoted to fatal with --fatal-warnings) and does not change runtime behavior.

Fixes #1562

### DIFF
--- a/crates/tmpfiles/src/lib.rs
+++ b/crates/tmpfiles/src/lib.rs
@@ -507,15 +507,6 @@ fn tmpfiles_entry_get_path(line: &str) -> Result<PathBuf> {
     unescape_path(&mut it)
 }
 
-/// A parsed tmpfiles entry kind and path of interest.
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub struct TmpfilesEntryRef {
-    /// The entry type character (e.g. 'd', 'L', 'z', 'Z', ...)
-    pub kind: char,
-    /// The target path for the entry
-    pub path: PathBuf,
-}
-
 /// Chown-affecting tmpfiles.d entries summary
 #[derive(Debug, Default, Clone)]
 pub struct TmpfilesChowners {
@@ -528,28 +519,12 @@ pub struct TmpfilesChowners {
 impl TmpfilesChowners {
     /// Returns true if a chown entry would apply to the specified absolute path
     pub fn covers(&self, p: &Path) -> bool {
-        if self.exact.contains(p) {
-            return true;
-        }
-        // For recursive entries, any ancestor match qualifies
-        for anc in p.ancestors() {
-            if self.recursive.contains(anc) {
-                return true;
-            }
-        }
-        false
+        self.exact.contains(p) || p.ancestors().any(|anc| self.recursive.contains(anc))
     }
 }
 
 fn tmpfiles_entry_get_kind(line: &str) -> Option<char> {
-    let mut it = line.bytes();
-    // Skip leading whitespace
-    while let Some(c) = it.next() {
-        if !c.is_ascii_whitespace() {
-            return Some(c as char);
-        }
-    }
-    None
+    line.trim_start().chars().next()
 }
 
 /// Read tmpfiles.d entries and return only those affecting chown operations (z/Z)


crates/lib/src/lints.rs
```diff
@@ -772,6 +772,93 @@ fn check_boot(root: &Dir, config: &LintExecutionConfig) -> LintResult {
...
+        }
+    }
+    let Some(etcd) = root.open_dir_optional("etc")? else {
+        return lint_ok();
+    };
+    // We'll collect problematic items
+    let mut problems: BTreeSet<std::path::PathBuf> = BTreeSet::new();
+    // Depth-first walk under /etc
+    let mut stack: Vec<(Dir, std::path::PathBuf)> = vec![(etcd, std::path::PathBuf::from("/etc"))];
+    while let Some((dir, abspath)) = stack.pop() {
```
cgwalters, line 819: This is probably OK as is but I've been trying to have recursive filesystem walks use https://docs.rs/cap-std-ext/latest/cap_std_ext/dirext/trait.CapStdExtDirExt.html#tymethod.walk

crates/lib/src/lints.rs
```diff
@@ -982,6 +1069,44 @@ mod tests {
...
+            "usr/lib/tmpfiles.d/test.conf",
+            "Z /etc/sub - - - -",
+        )?;
+
+        // Create two paths under /etc: one covered by Z /etc/sub, one not
+        root.create_dir_all("etc/sub/covered")?;
+        root.create_dir_all("etc/uncovered")?;
+
+        let config = &LintExecutionConfig { no_truncate: true };
+        // Since we cannot alter uid/gid in this test easily, fake non-root by relying on the logic
```
cgwalters, line 1091: Right...but we are going to need some "real" tests for this. 

https://github.com/bootc-dev/bootc/blob/df2da1adaf17ad0ec5309ca5dc22118b414676e8/crates/tests-integration/src/hostpriv.rs#L38 is one possible place - that test suite is defined to run as root and so it's supported to make tempdirs with arbitrary owned files.

Another (which is perhaps overdue) is one where we have Dockerfile fixtures and we explicitly just build each one and run `bootc container lint` in there and check the result.

crates/tmpfiles/src/lib.rs
```diff
@@ -507,6 +507,65 @@ fn tmpfiles_entry_get_path(line: &str) -> Result<PathBuf> {
...
+    let Some(tmpfiles_dir) = rootfs.open_dir_optional(TMPFILESD)? else {
+        return Ok(Default::default());
+    };
+    let mut out = TmpfilesChowners::default();
+    for entry in tmpfiles_dir.entries()? {
+        let entry = entry?;
+        // Only process .conf files
+        let name = entry.file_name();
+        let path = Path::new(&name);
+        if path.extension() != Some(OsStr::new("conf")) {
```
cgwalters, line 541: I think it'd be a lot cleaner to have this run `read_tmpfiles()` and then just filter the result.

crates/tmpfiles/src/lib.rs
```diff
@@ -507,6 +507,65 @@ fn tmpfiles_entry_get_path(line: &str) -> Result<PathBuf> {
...
+        let r = BufReader::new(entry.open()?);
+        for line in r.lines() {
+            let line = line?;
+            if line.is_empty() || line.starts_with('#') {
+                continue;
+            }
+            let Some(kind) = tmpfiles_entry_get_kind(&line) else { continue };
+            if kind != 'z' && kind != 'Z' {
+                continue;
+            }
```
cgwalters, line 553: I think it'd be cleaner to just fold this into the `match`

crates/tmpfiles/src/lib.rs
```diff
@@ -507,6 +507,65 @@ fn tmpfiles_entry_get_path(line: &str) -> Result<PathBuf> {
...
+    /// Exact chown entries (non-recursive) i.e. kind 'z'
+    pub exact: BTreeSet<PathBuf>,
+    /// Recursive chown entries (apply to all children) i.e. kind 'Z'
+    pub recursive: BTreeSet<PathBuf>,
+}
+
+impl TmpfilesChowners {
+    /// Returns true if a chown entry would apply to the specified absolute path
+    pub fn covers(&self, p: &Path) -> bool {
+        self.exact.contains(p) || p.ancestors().any(|anc| self.recursive.contains(anc))
```
cgwalters, line 522: I think we could have some unit tests for this pretty easily

cgwalters (changes requested):
Thanks for working on this!

Offhand it looks plausible but this really does need careful review which I'd like to do once we have non-synthetic tests that are actually doing file chowning.

PR #1571, opened 2025-09-01
composefs-native/boot: Handle systemd-boot and grub

Johan-Liebert1:
Fix Grub boot error caused by https://github.com/bootc-dev/bootc/pull/1541. Introduce a `--bootloader` cli option to `--composefs-native`. Depending upon the type of bootloader passed in we write BLS configs respectively

### DIFF
--- a/crates/lib/src/install.rs
+++ b/crates/lib/src/install.rs
@@ -94,7 +94,7 @@ use crate::lsm;
 use crate::parsers::bls_config::{parse_bls_config, BLSConfig};
 use crate::parsers::grub_menuconfig::MenuEntry;
 use crate::progress_jsonl::ProgressWriter;
-use crate::spec::ImageReference;
+use crate::spec::{Bootloader, Host, ImageReference};
 use crate::store::Storage;
 use crate::task::Task;
 use crate::utils::{path_relative_to, sigpolicy_from_opt};
@@ -119,7 +119,7 @@ const OSTREE_COMPOSEFS_SUPER: &str = ".ostree.cfs";
 /// The mount path for selinux
 const SELINUXFS: &str = "/sys/fs/selinux";
 /// The mount path for uefi
-const EFIVARFS: &str = "/sys/firmware/efi/efivars";
+pub(crate) const EFIVARFS: &str = "/sys/firmware/efi/efivars";
 pub(crate) const ARCH_USES_EFI: bool = cfg!(any(target_arch = "x86_64", target_arch = "aarch64"));
 pub(crate) const ESP_GUID: &str = "C12A7328-F81F-11D2-BA4B-00A0C93EC93B";
 pub(crate) const DPS_UUID: &str = "6523f8ae-3eb1-4e2a-a05a-18b695ae656f";
@@ -311,6 +311,10 @@ pub(crate) struct InstallComposefsOpts {
     #[clap(long, default_value_t)]
     #[serde(default)]
     pub(crate) insecure: bool,
+
+    #[clap(long, default_value_t)]
+    #[serde(default)]
+    pub(crate) bootloader: Bootloader,
 }
 
 #[cfg(feature = "install-to-disk")]
@@ -1581,7 +1585,7 @@ pub(crate) enum BootSetupType<'a> {
     /// For initial setup, i.e. install to-disk
     Setup((&'a RootSetup, &'a State, &'a FileSystem<Sha256HashValue>)),
     /// For `bootc upgrade`
-    Upgrade(&'a FileSystem<Sha256HashValue>),
+    Upgrade((&'a FileSystem<Sha256HashValue>, &'a Host)),
 }
 
 /// Compute SHA256Sum of VMlinuz + Initrd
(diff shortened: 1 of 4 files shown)


crates/lib/src/install.rs
```diff
@@ -1765,24 +1794,45 @@ pub(crate) fn setup_composefs_bls_boot(
...
+            let efi_linux_dir = mounted_efi.join(EFI_LINUX);
+
+            BLSEntryPath {
+                entries_path: efi_linux_dir,
+                config_path: mounted_efi.clone(),
+                abs_entries_path: EFI_LINUX,
+                mount_path: Some(mounted_efi),
+            }
+        }
+    };
```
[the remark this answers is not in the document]
Johan-Liebert1: yeah, this sounds right. Updated

crates/lib/src/install.rs
```diff
@@ -1831,44 +1887,66 @@ pub(crate) fn setup_composefs_bls_boot(
...
+                bls_config.linux =
+                    format!("/{}/{symlink_to}/vmlinuz", entry_paths.abs_entries_path);
+
+                bls_config.initrd = vec![format!(
+                    "/{}/{symlink_to}/initrd",
+                    entry_paths.abs_entries_path
+                )];
             } else {
-                write_bls_boot_entries_to_disk(&efi_dir, id, usr_lib_modules_vmlinuz, &repo)?;
+                write_bls_boot_entries_to_disk(
```
p5, line 1909: Apologies for breaking GRUB in the previous PR.  From reading through various docs (Fedora change requests) and issues in the repo, it sounded like GRUB was able to work in a (mostly) BLS-compliant way.  My understanding must have not been correct.

I also think my understanding was further enforced through the function names - `write_**bls**_boot_...` - which made me think we were trying to achieve full [BLS compliance](https://uapi-group.org/specifications/specs/boot_loader_specification/) in the Composefs backend.

---

I greatly appreciate you fixing the implementation rather than reverting the PR!

Johan-Liebert1: Grub doesn't fully support the BLS spec, that's why we have to jump over hoops to support it, which is unfortunate. 

> I greatly appreciate you fixing the implementation rather than reverting the PR!

We want to support systemd-boot eventually :)

crates/lib/src/composefs_consts.rs
```diff
@@ -19,6 +19,8 @@ pub(crate) const ORIGIN_KEY_BOOT: &str = "boot";
 pub(crate) const ORIGIN_KEY_BOOT_TYPE: &str = "boot_type";
 /// Key to store the SHA256 sum of vmlinuz + initrd for a deployment
 pub(crate) const ORIGIN_KEY_BOOT_DIGEST: &str = "digest";
+/// Bootloader, Grub or Systemd
+pub(crate) const ORIGIN_KEY_BOOTLOADER: &str = "bootloader";
```
cgwalters: We can very reliably detect if systemd-boot is in use by looking at the EFI variables it sets, or parsing `bootctl --json` or so...I'm not excited by carrying a stateful thing for this.

Also this heavily overlaps with what bootupd is doing. For the bootupd case of course since it *installs* the bootloader it may need some explicit configuration.

---

This heavily relates to my comment in https://github.com/ostreedev/ostree/pull/3359#issuecomment-2555621129

Basically though, since we require bootc to manage some out of band config files for grub, how about we default to systemd boot, but query `/boot` and if we detect our own grub config files then we use that?

If we *did* need a config knob, see also https://github.com/bootc-dev/bootc/pull/1532 which changed the install config.

cgwalters (changes requested):

crates/lib/src/composefs_consts.rs
cgwalters: > how about we default to systemd boot, 

Really I mean, default to the Boot Loader Spec...and detect grub by looking for the config files we currently need.

So the enum would be
```
enum Bootloader {
 BLS,
 Grub
}
```
or so

Johan-Liebert1: It makes sense to default to systemd when we have full support for it in bootupd. Right now, since we don't have it, I think it makes sense to keep Grub as the default. When we do have full support, it's just a matter of updating the code to change the default

Making it stateless is definitely cleaner than relying on the origin file.

cgwalters: > Making it stateless is definitely cleaner than relying on the origin file.

Right, can we do that here please?

Johan-Liebert1: Updated to get the bootloader from files we we might / might not have in /boot. I have made an assumption in https://github.com/bootc-dev/bootc/pull/1571/files#diff-b8d81da4f5dcb6d2bf0366e6a0663a5e784b62f1c49682a851e0322c590632faR428 that we will only have one bootloader, either grub or systemd

Johan-Liebert1: Okay, I changed it to check from `/sys/firmware/efi/efivars/LoaderInfo-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f` according to https://man.archlinux.org/man/systemd-boot.7#EFI_VARIABLES

cgwalters (approved):
LGTM as is however...I still think we should detect grub instead via looking for its config files

Johan-Liebert1:
> I still think we should detect grub instead via looking for its config files

Is there some peculiar reason why we should be looking for grub config files? I think the EFI vars approach is a bit better as it also lets us distinguish BIOS/UEFI

cgwalters:
> Is there some peculiar reason why we should be looking for grub config files? I think the EFI vars approach is a bit better as it also lets us distinguish BIOS/UEFI

In theory we should work with *any* bootloader that implements the spec instead of specifically detecting systemd-boot - right?

Johan-Liebert1:
Right. AFAIK bootloaders on UEFI will populate the LoaderInfo file with the bootloader name, which we check right now. Though, we do default to Grub if we can't find `/sys/firmware/efi/efivars`, in which case it does become necessary to look for bootlaoder specific config files.

PR #1573, opened 2025-09-02
composefs-native/uki: Handle UKI addons

Johan-Liebert1:
If we find UKI addons in the boot entries list, write them to ESP along  with the UKI

Right now if a UKI Addon also has the `composefs=` cmdline param, it's ignored.

Supporting: https://github.com/containers/composefs-rs/issues/126

Needs https://github.com/containers/composefs-rs/pull/178

### DIFF
--- a/crates/lib/src/bootc_composefs/boot.rs
+++ b/crates/lib/src/bootc_composefs/boot.rs
@@ -8,7 +8,10 @@ use bootc_blockdev::find_parent_devices;
 use bootc_mount::inspect_filesystem;
 use bootc_utils::CommandRunExt;
 use camino::{Utf8Path, Utf8PathBuf};
-use cap_std_ext::{cap_std, dirext::CapStdExtDirExt};
+use cap_std_ext::{
+    cap_std::{ambient_authority, fs::Dir},
+    dirext::CapStdExtDirExt,
+};
 use clap::ValueEnum;
 use composefs::fs::read_file;
 use composefs::tree::{FileSystem, RegularFile};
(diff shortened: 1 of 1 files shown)


Johan-Liebert1:
I thought `needs-ok-to-test` skipped CI as the CI won't currently pass anyway

crates/lib/src/install.rs
```diff
@@ -1931,13 +1934,106 @@ fi
...
+
+        boot_label = Some(uki::get_boot_label(&efi_bin).context("Getting UKI boot label")?);
+    }
+
+    // Write the UKI to ESP
+    let efi_linux_path = mounted_efi.join(EFI_LINUX);
+    create_dir_all(&efi_linux_path).context("Creating EFI/Linux")?;
+
+    let final_pe_path = if let Some(parent) = file_path.parent() {
+        let renamed_path = if parent.as_str()?.ends_with(EFI_ADDON_DIR_EXT) {
```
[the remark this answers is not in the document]
Johan-Liebert1: `Path::ends_with` checks per-component which is not what we want here

Johan-Liebert1:
Rebased and fix conflicts. @cgwalters could I have a review?

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -720,6 +711,118 @@ pub(crate) fn setup_composefs_uki_boot(
...
+
+    Task::new("Mounting ESP", "mount")
+        .args([&PathBuf::from(&esp_device), &mounted_efi.clone()])
+        .run()?;
+
+    let mut boot_label = String::new();
+
+    for entry in entries {
+        match entry {
+            ComposefsBootEntry::Type1(..) => tracing::debug!("Skipping Type1 Entry"),
```
cgwalters, line 783: FWIW I basically don't think bootc should support this. The main use case is kernel arguments which we already cover with bootc kargs.

So my preference going forward is a hard error, not just a debug.

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -720,6 +711,118 @@ pub(crate) fn setup_composefs_uki_boot(
...
+            ComposefsBootEntry::UsrLibModulesVmLinuz(..) => {
+                tracing::debug!("Skipping vmlinuz in /usr/lib/modules")
```
cgwalters, line 785: Shouldn't this one also be an error?

Johan-Liebert1: Currently, using existing images to create UKIs also has kernel+initrd in /usr/lib/modules. But, yes, makes sense for these to be errors as we only ever want one type of boot entry. But, let's say, if an image does have both type of entries, as users could be using the same image, should we still throw an error?

Issue #1575, opened 2025-09-02
RPM warning for absolute symlink


jeckersb:
Throwing this here before I forget about it since off the top of my head I think via packit we derive from the spec in the upstream repo?

```
RPM build warnings:
    absolute symlink: /usr/lib/bootc/storage -> /sysroot/ostree/bootc/storage
```

PR #1577, opened 2025-09-02
Makefile: Create storage symlink as relative

jeckersb:
Resolves: #1575

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,12 +13,13 @@ CARGO_FEATURES ?= $(shell . /usr/lib/os-release; if echo "$$ID_LIKE" |grep -qF r
 all:
 	cargo build --release --features "$(CARGO_FEATURES)"
 
+STORAGE_RELATIVE_PATH ?= $(shell realpath -m -s --relative-to="$(prefix)/lib/bootc/storage" /sysroot/ostree/bootc/storage)
 install:
 	install -D -m 0755 -t $(DESTDIR)$(prefix)/bin target/release/bootc
 	install -D -m 0755 -t $(DESTDIR)$(prefix)/bin target/release/system-reinstall-bootc
 	install -d -m 0755 $(DESTDIR)$(prefix)/lib/bootc/bound-images.d
 	install -d -m 0755 $(DESTDIR)$(prefix)/lib/bootc/kargs.d
-	ln -s /sysroot/ostree/bootc/storage $(DESTDIR)$(prefix)/lib/bootc/storage
+	ln -s "$(STORAGE_RELATIVE_PATH)" "$(DESTDIR)$(prefix)/lib/bootc/storage"
 	install -D -m 0755 crates/cli/bootc-generator-stub $(DESTDIR)$(prefix)/lib/systemd/system-generators/bootc-systemd-generator 
 	install -d $(DESTDIR)$(prefix)/lib/bootc/install
 	# Support installing pre-generated man pages shipped in source tarball, to avoid


Makefile
```diff
@@ -13,12 +13,13 @@ CARGO_FEATURES ?= $(shell . /usr/lib/os-release; if echo "$$ID_LIKE" |grep -qF r
 all:
 	cargo build --release --features "$(CARGO_FEATURES)"
 
+STORAGE_RELATIVE_PATH ?= $(shell realpath -m --relative-to="$(prefix)/lib/bootc/storage" /sysroot/ostree/bootc/storage)
```
cgwalters: Are you sure this is doing the right thing? Trying this out in a non-bootc environment i.e. without the path present it outputs `.` for me which is obviously wrong.

Honestly I think we can reasonably just hardcode `prefix = /usr` and when we do that this becomes a constant value.

jeckersb: Hm I tested it with `make bin-archive` which seems correct:

```
ln -s "../../../../sysroot/ostree/bootc/storage" "tmp-install/usr/lib/bootc/storage"
```

And the resulting archive looks right:

```
$ less target/bootc.tar.zst | grep usr/lib/bootc/storage
lrwxrwxrwx 0/0               0 2025-09-02 12:09 ./usr/lib/bootc/storage -> ../../../../sysroot/ostree/bootc/storage
```

edit: this is in a toolbox, so I don't have `/usr/lib/bootc` or `/sysroot` at all

cgwalters: Definitely doesn't work for me, and CI is also broken

```
$ make bin-archive; tar tvf target/bootc.tar.zst|grep usr/lib/bootc/storage
...
make[1]: Leaving directory '/var/home/walters/src/github/bootc-dev/bootc'
lrwxrwxrwx 0/0               0 2025-09-02 12:09 ./usr/lib/bootc/storage -> .
$
```

jeckersb: We tracked this down in a live session, the problem occurs if you already have the bootc rpm installed.  Then you already have the `/usr/lib/bootc/storage -> /sysroot/ostree/bootc/storage` link present.  Then `realpath` will chase the symlink by default, so it tries to create a relative symlink from `/sysroot/ostree/bootc/storage` to `/sysroot/ostree/bootc/storage`, which is just `.`

Easy fix though, `realpath` has the `-s` flag which causes it to not expand symlinks.  This works even when bootc is already present:

```
$ podman run --rm -it quay.io/fedora/fedora-bootc:42 realpath -m -s --relative-to=/usr/lib/bootc/storage /sysroot/ostree/bootc/storage
../../../../sysroot/ostree/bootc/storage
```

Compared to before without `-s`:

```
$ podman run --rm -it quay.io/fedora/fedora-bootc:42 realpath -m --relative-to=/usr/lib/bootc/storage /sysroot/ostree/bootc/storage
.
```

cgwalters: What's extra reassuring is when I look at `strace realpath -m -s` vs without the `-s`, in the first case I see a bunch of standard dynamic linker stuff and then a single `write` call which strongly implies this is just an in-memory computation on strings, which is what we want.

cgwalters (approved):

cgwalters:
Hm this seems to have tripped up a downstream rpm linter

image

Looking...

cgwalters:
Hum maybe what we should do here actually is make it a mount point...

Or maybe we teach container-libs how to support AIS aliases

cgwalters:
Nah I think this is all fine, it's just that it doesn't like dangling links, which I get...but hard to fix.

PR #1578, opened 2025-09-02
rechunker: Duplicate xattrs when more than 256 links

ckyrouac:
This is to avoid creating (10s of) thousands of links to a single file which will cause the rechunker to fail after passing the filesystem's link limit. This will create a duplicate xattrs file to link to for each chunk of 256 links.

Assisted-by: Claude Code

Fixes https://github.com/coreos/rpm-ostree/issues/5479


crates/ostree-ext/src/tar/export.rs
```diff
@@ -141,9 +143,13 @@ pub(crate) fn object_path(objtype: ostree::ObjectType, checksum: &str) -> Utf8Pa
...
 }
 
-fn v1_xattrs_object_path(checksum: &str) -> Utf8PathBuf {
+fn v1_xattrs_object_path_with_suffix(checksum: &str, suffix: u32) -> Utf8PathBuf {
     let (first, rest) = checksum.split_at(2);
-    format!("{OSTREEDIR}/repo/objects/{first}/{rest}.file-xattrs").into()
+    if suffix == 0 {
+        format!("{OSTREEDIR}/repo/objects/{first}/{rest}.file-xattrs").into()
+    } else {
+        format!("{OSTREEDIR}/repo/objects/{first}/{rest}.file-xattrs.{suffix}").into()
```
cgwalters: I'm worried that `import.rs` might reject this, in fact I'm pretty sure it will since it's matching on the plain file suffix, see https://github.com/bootc-dev/bootc/blob/df2da1adaf17ad0ec5309ca5dc22118b414676e8/crates/ostree-ext/src/tar/import.rs#L410

We definitely should have an "end-to-end" test (see `tests/it/main.rs`) that verifies we can both export and import such a container.

cgwalters: Honestly the simplest approach might be to just stop hardlinking the xattr files. It'd probably make the tarballs bigger, but OTOH compressors tend to do well with stuff like this.

But my suggestion over in the rpm-ostree issue was that instead of adding a suffix, we keep a counter of the number of links we emitted.

crates/ostree-ext/src/tar/export.rs
```diff
@@ -126,6 +126,8 @@ struct OstreeTarWriter<'a, W: std::io::Write> {
     wrote_dirmeta: HashSet<String>,
     wrote_content: HashSet<String>,
     wrote_xattrs: HashSet<String>,
+    /// Tracks the number of links to each xattrs object (checksum -> (link count, suffix))
+    xattrs_link_counts: std::collections::HashMap<String, (u32, u32)>,
```
cgwalters: I don't think we need both `wrote_xattrs` and `xattrs_link_counts` right? See below

crates/ostree-ext/src/tar/export.rs
ckyrouac: > Honestly the simplest approach might be to just stop hardlinking the xattr files. It'd probably make the tarballs bigger, but OTOH compressors tend to do well with stuff like this.

Makes sense to me

ckyrouac: I pushed a new commit that removes the xattrs link code from both import/export. This is mostly just generated code from claude to show the idea and needs to be cleaned up. It'll take me a bit to understand the import code more to properly clean it up, but I wanted to publish it to see if it's generally reasonable.

crates/ostree-ext/src/tar/import.rs
```diff
@@ -447,50 +439,36 @@ impl Importer {
...
     fn process_file_xattrs(
         &mut self,
-        entry: tar::Entry<impl std::io::Read>,
-        checksum: String,
-    ) -> Result<()> {
-        self.cache_xattrs_content(entry, Some(checksum))?;
-        Ok(())
-    }
-
-    /// Process a `.file-xattrs-link` object (v1).
```
cgwalters: We can't remove the import code as we still need to support *old* images.

To this point...hmm, we should probably add an e2e test that verifies we can pull a permanently saved old image (e.g. `skopeo copy quay.io/centos-bootc/centos-bootc:stream9 ghcr.io/bootc-dev/centos-bootc-fixture:20250903` or something)

crates/ostree-ext/src/tar/export.rs
```diff
@@ -125,7 +125,6 @@ struct OstreeTarWriter<'a, W: std::io::Write> {
     wrote_dirtree: HashSet<String>,
     wrote_dirmeta: HashSet<String>,
     wrote_content: HashSet<String>,
-    wrote_xattrs: HashSet<String>,
```
cgwalters, line 128: Very straightforward to just delete 90% of this code for sure.

crates/ostree-ext/src/tar/export.rs
cgwalters: Yep pushing WIP is all good by me! I think per below we simply can't change the import code at all, so that simplifies things a lot.

ckyrouac: I ripped out all the import changes. They were in place to make the tests pass. Instead, the latest commit creates the file with the file-xattrs-link suffix to preserve all the same import code for backwards compatibility.

ckyrouac:
> Can you do a before/after comparison on sizes just as a sanity check?

The `quay.io/fedora/fedora-bootc:42` image is actually a few MB smaller without links:

With links: 1917.763676 MB
Without links: 1899.812956 MB

cgwalters:
Thanks for confirming that, very helpful to know.

PR #1579, opened 2025-09-02
Deduplicate main error printing

cgwalters:
As part of the tracing-subscriber CVE I did a quick audit for usages of `tracing::error!` and I noticed when we updated the `main()` function in the primary
crate we missed also doing the same for system-reinstall-bootc.

Move the handling of that to utils.

xref: https://bugzilla.redhat.com/show_bug.cgi?id=2392017


jeckersb (approved):
:+1:

PR #1580, opened 2025-09-03
Un-hide soft-reboot

cgwalters:
We debugged https://github.com/ostreedev/ostree/issues/3503 and it's basically a systemd/Anaconda bug. So we can consider soft reboots an available feature from our end (as systemd does), but certainly one that will require integration testing on the OS/distro side.

If you're building a system and you don't want to enable soft reboots, just `rm /usr/lib/systemd/system/soft-reboot.target` or so should do the trick.


jmarrero (approved):
lgtm

bshephar:
This feature is great btw @jmarrero. Really cool:
```
❯ sudo target/release/bootc upgrade --soft-reboot required --apply
```

So fast, and I really think this will help me with my `bootc` sales pitch!

PR #1583, opened 2025-09-04
Update reboot-required file with information about the staged image

### LINKED ISSUE #1574: [FEATURE] Put a brief explanation in /run/reboot-required
Currently `[tracer](https://github.com/FrostyX/tracer/blob/master/tracer/controllers/default.py#L146)` puts a message into `/run/reboot-required` with a brief explanation of why the reboot is required.

Id would be nice if `bootc` also put a message into `/run/reboot-required`.  This way I could a something like this to `/etc/profile.d` as a notice to folks:
```
if [[ -e /run/reboot-required ]]; then
  cat /run/reboot-required
fi
```

I was thinking something like:
```
bootc says a reboot is required to load staged image: XYZ
```

bshephar:
This change updates the /run/reboot-required file to include information
about the staged bootc image. This will provide users with information
about what bootc image will be effective after the system has been
rebooted.

Fixes #1574

### DIFF
--- a/crates/lib/src/deploy.rs
+++ b/crates/lib/src/deploy.rs
@@ -794,9 +794,10 @@ pub(crate) async fn stage(
 
     // Unconditionally create or update /run/reboot-required to signal a reboot is needed.
     // This is monitored by kured (Kubernetes Reboot Daemon).
+    let reboot_message = format!("bootc: Reboot required for image: {}", &spec.image.image);
     let run_dir = Dir::open_ambient_dir("/run", cap_std::ambient_authority())?;
     run_dir
-        .atomic_write("reboot-required", b"")
+        .atomic_write("reboot-required", reboot_message.as_bytes())
         .context("Creating /run/reboot-required")?;
 
     Ok(())
--- a/tmt/tests/booted/test-image-pushpull-upgrade.nu
+++ b/tmt/tests/booted/test-image-pushpull-upgrade.nu
@@ -65,9 +65,9 @@ RUN echo test content > /usr/share/blah.txt
     let progress = open --raw $progress_json | from json -o
     sanity_check_switch_progress_json $progress
 
-    # Check that /run/reboot-required exists and is a zero-byte file
+    # Check that /run/reboot-required exists and is not empty
     let rr_meta = (ls /run/reboot-required | first)
-    assert equal $rr_meta.size 0b
+    assert ($rr_meta.size > 0b)
 
     # Also test that the mtime changes on modification
     let new_root_mtime = ls -Dl /ostree/bootc | get modified


crates/lib/src/store/mod.rs
```diff
@@ -49,6 +49,13 @@ pub const COMPOSEFS_MODE: Mode = Mode::from_raw_mode(0o700);
 /// system root
 pub(crate) const BOOTC_ROOT: &str = "ostree/bootc";
 
+/// The path for the reboot-required file, relative to the system
+/// root
+pub(crate) const REBOOT_REQUIRED: &str = "run/reboot-required";
+
+/// Reboot required message that will appear in /run/reboot-required
+pub(crate) const REBOOT_REQUIRED_MESSAGE: &[u8] = b"Reboot required to complete bootc operation";
```
cgwalters: This seems OK as is. But it'd likely be even nicer to have some kind of human-readable message like `Reboot required to update to bootc image: quay.io/exampleos/example@sha256:...` or something? And yeah it gets messy to try to "stringfiy" all operations, maybe we can just make it `Reboot required for quay.io/exampleos/example@sha256:...` even if we're rolling back etc.

crates/lib/src/store/mod.rs
```diff
@@ -175,4 +182,18 @@ impl Storage {
             .update_timestamps(std::path::Path::new(BOOTC_ROOT))
             .context("update_timestamps")
     }
+
+    #[context("Updating the reboot-required file")]
+    pub(crate) fn update_reboot_file(&self) -> Result<()> {
```
cgwalters: Conceptually `Storage` holds container images, I don't think its scope includes this conceptually. Offhand I'd say a free function in `deploy.rs` is probably best.

Either way...I don't quite get the idea of having a PR which adds an unused helper function, why wouldn't we just also change the code to use it in a single step?

crates/lib/src/store/mod.rs
```diff
@@ -49,6 +49,13 @@ pub const COMPOSEFS_MODE: Mode = Mode::from_raw_mode(0o700);
 /// system root
 pub(crate) const BOOTC_ROOT: &str = "ostree/bootc";
 
+/// The path for the reboot-required file, relative to the system
+/// root
+pub(crate) const REBOOT_REQUIRED: &str = "run/reboot-required";
```
cgwalters: One thing that's interesting here is I think we should only write this file if the system is actually *booted*. Which means (unlike most of my comments around file access!) we don't actually need to use fd-relative APIs. But see below

crates/lib/src/store/mod.rs
bshephar: Yeah, I'll add the rest of it in. I mainly wanted the feedback from the reporter on whether they thought the message was sufficient to just say, "Hey, you need to reboot." Or whether they were after, "Hey, you need to reboot to complete the switch operation to image: blah"

But I think both you and the reporter of the issue have conferred that we want to implement the latter.

I'll move it to deploy.rs while I'm implementing the rest of the functionality.

crates/lib/src/store/mod.rs
bshephar: Yeah, just using the image reference will be easier than tracking which operation is in progress. This is the middle ground between my initial suggestion, and my comment below. So, we can just say that  "Reboot is required for bootc image: `quay.io/blah/blah@`"

bshephar:
I didn't realize that we were already creating this file, so this change is much more simple than I had initially planned. I think it should always be updated in that `stage` function, so we don't need a helper and to handle it for each operation.

bshephar:
Result looks like:
```
❯ sudo target/release/bootc upgrade
layers already present: 67; layers needed: 1 (695.2 MB)
Fetched layers: 662.96 MiB in 34 seconds (19.51 MiB/s)
  Deploying: done (6 seconds)
Queued for next boot: ghcr.io/bshephar/fedora-asahi-bootc:42
  Version: 42.20250904.0
  Digest: sha256:8ea48cc32573500a367526e687cd61510859e097df82402bf352d98da755c6c9
Total new layers: 68    Size: 3.2 GB
Removed layers:   1     Size: 688.5 MB
Added layers:     1     Size: 695.2 MB
```

```
❯ cat /run/reboot-required
Reboot required for bootc image: ghcr.io/bshephar/fedora-asahi-bootc:42
```

I guess if this was a `bootc switch` instead of a `bootc upgrade` it would be more useful, because the image would be different to the one I'm currently running. But because I have used a tag, this is always just going to say the same thing. Do we feel like that is still valuable? Or, do we realistically want the context of which operation has staged the image?

Maybe there's something more informative than just `&spec.image.image` like I'm using here?

Edit: Maybe `image.manifest_digest.as_ref()` would be better here actually:
https://github.com/bootc-dev/bootc/blob/main/crates/lib/src/deploy.rs#L668

Using the digest looks like:
```
❯ cat /run/reboot-required
Reboot required for bootc image digest: sha256:c0aaf49e1df7ec2f4baa4ed8077b684cad36820381f1a7511247558fb35e2bea%
```

crates/lib/src/deploy.rs
bshephar, line 1: I noted that the `rollback` code path doesn't use this `stage` function, so it might still be desirable to split this out into it's own function that can be called from any operation that has changed the next boot target.

crates/lib/src/deploy.rs
```diff
@@ -794,9 +794,13 @@ pub(crate) async fn stage(
 
     // Unconditionally create or update /run/reboot-required to signal a reboot is needed.
     // This is monitored by kured (Kubernetes Reboot Daemon).
+    let reboot_message = format!(
+        "Reboot required for bootc image digest: {}",
```
cgwalters: Bikeshed time but how about:

```
bootc: Reboot required for image: quay.io/exampleos/exampleos@sha256:...
```

The leading `bootc: ` helps clarify that we're the *source* of the change. I want to get away from the idea there are "bootc images" in that sense, there are just container images that happen to have a kernel.

crates/lib/src/deploy.rs
cgwalters: Yes! Let's fix that too, though it should probably be a separate commit or even PR for clarity?

crates/lib/src/deploy.rs
bshephar: Yeah, that's a good plan. I've updated this now

crates/lib/src/deploy.rs
bshephar: Sounds good, I can send a follow up PR to create the dedicated function and find all of the places where it should be used.

Issue #1428, opened 2025-07-18
Site: Odd formatting (random `\`) throughout docs.


cdrage:
<img width="877" height="421" alt="Image" src="https://github.com/user-attachments/assets/90b9c952-12f4-4c07-a08a-4fb7ec02c7ef" />

Throughout the docs, I am noticing a lot of backslashes in some of the generated pages.

For example on: https://bootc-dev.github.io/bootc/man/bootc-install-to-disk.html

You get:

```
The implementation of this uses systemd \`tmpfiles.d\`, writing to a
file named \`/etc/tmpfiles.d/bootc-root-ssh.conf\`. This will have
the effect that by default, the SSH credentials will be set if not
present. The intention behind this is to allow mounting the whole
\`/root\` home directory as a \`tmpfs\`, while still getting the SSH
key replaced on boot.
```

jeckersb:
Maybe collateral damage from the quoting shenanigans in https://github.com/bootc-dev/bootc/pull/1385?  Although that was dealing single quotes / apostrophe, whereas this is for backticks.  Maybe not the same thing but related.

cgwalters:
Yeah the docs are just a mess, I am of the mind lately to just bite the bullet and manually duplicate the text and try to enforce with some automation (maybe AI assisted) that man pages are *man*ually kept in sync with the Rust/clap help text.

PR #1587, opened 2025-09-05
Switch to hand-written man pages with auto option sync

cgwalters:
See the updates to `Justfile` for how to use this.

Closes: #1428

Assisted-By: Claude Code (opus + sonnet)


cgwalters:
Besides closing the formatting bug, the big thing here is we can write a lot more text in the command man page. There's a lot of overlap between the currently hand-written `docs/src/bootc-install.md` and the formerly auto-generated `docs/src/bootc-install.md`. This PR doesn't rectify that, but what I want to do is basically fold the two together for the most part.

gursewak1997:
Changes look good. CI fails with
```
error: Packaging
Caused by:
    0: Converting docs/src/man/bootc.md to man page
    1: command not found: `pandoc`
```

cgwalters:
OK this is now architecturally closer to what's happening in e.g. https://github.com/containers/netavark/blob/main/docs/Makefile

A nice thing is we can replace pandoc with `go-md2man` which is way simpler of a project now that we don't need to have a crazy chain of Rust → man → markdown.

docs/src/man/bootc-usr-overlay.8.md
```diff
@@ -0,0 +1,42 @@
...
+# SYNOPSIS
+
+**bootc usr-overlay** \[**-h**\|**\--help**\]
+
+# DESCRIPTION
+
+Adds a transient writable overlayfs on `/usr` that will be discarded
+on reboot.
+
+\## Use cases
```
cgwalters: This looks wrong

docs/src/man/bootc-container-lint.8.md
```diff
@@ -5,20 +5,21 @@ checks as part of a container build
...
 container build.
 
-This is intended to be invoked via e.g. \`RUN bootc container lint\` as
+This is intended to be invoked via e.g. `RUN bootc container lint` as
 part of a build process; it will error if any problems are detected.
 
 # OPTIONS
 
+<!-- BEGIN GENERATED OPTIONS -->
+<!-- END GENERATED OPTIONS -->
```
cgwalters: Need to remove the hand written options below

cgwalters:
OK this one should be ready for review!

gursewak1997:
Is there a reason we cancelled the CI runs?
Errors on
`
Problem: cannot install both ostree-libs-2025.4-2.el9.aarch64 from appstream and ostree-libs-2025.6-1.el9.aarch64 from @System
`

cgwalters:
The GHA matrix there defaults to fail-fast; maybe we could change that. Anyways rebased on https://github.com/bootc-dev/bootc/pull/1609

cgwalters:
Fallout in https://github.com/bootc-dev/bootc/issues/1616

PR #1588, opened 2025-09-05
Initial factory reset code

ckyrouac:
This is a rebased version of #1389 with some fixes on top to make `bootc install reset --experimental` work with the latest soft reboot code. I'd like to get this in hidden behind `--experimental`, then iterate on it. I plan to add some e2e tests as a followup but I can work on including them as part of this PR if you want.


cgwalters:
I think we'll want tests at least before merging, and supporting that with tmt is probably going to be an excellent driver for the feature.

I haven't dug in but I think tmt is relying on injecting ssh keys w/cloud-init, so it might Just Work to do a reset there?

ckyrouac:
> I think we'll want tests at least before merging

Updated with a basic test.

crates/lib/src/deploy.rs
```diff
@@ -634,6 +634,18 @@ async fn deploy(
...
     let staged = ostree.staged_deployment().unwrap();
     assert_eq!(staged.index(), r);
+
+    // This file is used to signal to other CLI commands that this deployment is a factory reset
+    // (e.g. bootc status)
+    if matches!(from, MergeState::Reset { .. }) {
+        let deployment_dir = ostree.deployment_dirpath(&staged);
+        let deployment_dir = std::path::Path::new(deployment_dir.as_str());
+        if deployment_dir.exists() {
+            let marker_path = deployment_dir.join(".bootc-factory-reset");
```
cgwalters: Let's have clear `const` for things like this with an explanation there.

But backing up a second, in the ostree side we have "deployment backing" directories already where we can put metadata.

Though of course we should also be thinking about how this would work with the composefs backend.

Hmmm hmm. Here's an idea, how about we use the deployment's `etc` directory for state like this instead? I think we could save e.g. `system.bootc.merged_from` xattr or so?

Basically I lean towards extended attributes over "stamp files" for metadata.

BTW though do note the intersection here with `etc.transient` - I wonder if we should make that more visible in status output too. If that's enabled each upgrade is already like a reset of `/etc`.

crates/lib/src/status.rs
```diff
@@ -95,7 +95,20 @@ impl From<ImageReference> for OstreeImageReference {
 
 /// Check if a deployment has soft reboot capability
 fn has_soft_reboot_capability(sysroot: &SysrootLock, deployment: &ostree::Deployment) -> bool {
-    ostree_ext::systemd_has_soft_reboot() && sysroot.deployment_can_soft_reboot(deployment)
+    let deployment_dir = sysroot.deployment_dirpath(deployment);
+    let deployment_dir = std::path::Path::new(deployment_dir.as_str());
+    let is_factory_reset =
+        deployment_dir.exists() && deployment_dir.join(".bootc-factory-reset").exists();
```
cgwalters: This is where we want a `const` that is shared

crates/lib/src/status.rs
```diff
@@ -95,7 +95,20 @@ impl From<ImageReference> for OstreeImageReference {
...
+
+    if is_factory_reset {
+        // Factory reset deployments don't support soft reboots
+        // this is primarily because the kargs validation will fail when checking for soft reboot
+        // compatibility in the ostree code, which will cause bootc status and upgrade to fail.
```
cgwalters: I don't understand why we need to special case this. Why is bootc status failing exactly?

If we're changing (usually dropping) kernel arguments won't the existing karg check work?

cgwalters (changes requested):

jlebon:
Just saw the demo for this. A few random thoughts.

It'd be nice to be able to have different reset levels. One is just resetting `/etc`. This doesn't require a new stateroot. Another would be resetting both `/etc` and `/var`.

Should this use stateroots at all? This would be another thing to implement in the composefs backend (or maybe that code and concept already exist there?). But also it's quite common for `/var` or subdirectories of `/var` to be mounts. I.e. the stateroot var is not used in that case. We should probably detect this and error out.

One alternative though is basically to just move everything in `/var` into `/var/.previous`. That would work in both cases, is cheap, and makes access to the previous `/var` trivial.

(There's still the case of submounts in `/var` of course. We could apply that rename strategy recursively, or just warn and leave it up to the user -- or require them to be unmounted before proceeding?)

crates/lib/src/status.rs
cgwalters: As a followup - we talked about this in a live chat and there seems to be a libostree bug here where the staged deployment is missing `ostree=` and that trips an assertion in the soft reboot checks.

Ideally we root cause and fix that.

But we may be able to work around this by detecting when there's no kargs in the staged deployment?

ckyrouac: I think this will fix the factory resest issue: https://github.com/ostreedev/ostree/pull/3532. Still investigating why the deployment is missing the `ostree` arg.

Dockerfile
```diff
@@ -5,7 +5,7 @@
 # Use e.g. --build-arg=base=quay.io/fedora/fedora-bootc:42 to target
 # Fedora instead.
 
-ARG base=quay.io/centos-bootc/centos-bootc:stream10
+ARG base=quay.io/fedora/fedora-bootc:43
```
cgwalters: This is a pretty big change of default. I've been trying to ensure that what we do always works on C10S as it's a primary target.

Of course for testing out this PR pre merge that's fine!

Though I think in practice it doesn't really matter because CI is going to do the full matrix anyways...

Dockerfile
```diff
@@ -113,5 +113,10 @@ touch /usr/lib/.bootc-dev-stamp
 # And test our own linting
 ## Workaround for https://github.com/bootc-dev/bootc/issues/1546
 rm -rf /root/buildinfo
-bootc container lint --fatal-warnings
+
+dnf -y install 'dnf5-command(copr)'
```
cgwalters: This one needs detection for dnf4 vs dnf5

ckyrouac:
> It'd be a bit ugly but maybe we could detect the ostree version and avoid crashing if it's too old?

Just pushed up a commit to check the ostree version in the has_soft_reboot_capability function where the crash happens. I haven't had a chance to dig into the deeper root cause of why the ostree= karg is getting set but the version check should unblock this PR's CI. The result is on systems with an outdated ostree, soft reboot will be disabled for factory reset deployments. If it works for you, I'll just file an issue to track fixing the ostree= karg issue.

Dockerfile
ckyrouac: with the ostree version check I think we can avoid manually installing the latest ostree. The tests should pass with any ostree version.

Dockerfile
ckyrouac: yep I was just doing this for testing. Reverted in the latest push.

tmt/tests/booted/test-factory-reset.nu
```diff
@@ -0,0 +1,58 @@
...
+    $orig_stateroot | save /var/tmp/tmt/orig_stateroot
+
+    bootc install reset --experimental
+
+    # sanity check that bootc status shows a new deployment with a non default stateroot
+    let reset_status = bootc status --json | from json
+    assert not equal $reset_status.status.otherDeployments.0.ostree.stateroot "default"
+
+    # we need tmt in the new stateroot for second_boot
+    print "Copying tmt into new stateroot"
```
cgwalters, line 23: I guess this one is really asking for enhancing reset to support copying data in a nicer way.

ckyrouac: working on adding `bootc mount`

ckyrouac:
Ah, so the current CI failures are due to the outdated ostree in the tmt env which is causing the soft reboot tests to fail because softRebootCapable is defaulting to false. Not sure how we want to handle this. With the current code in this PR, soft reboot will be disabled for all systems with ostree < 2025.7. I can add back the code to the Dockerfile to install ostree from the continuous repo so the tests pass. Or, a more complicated workaround would be to add back the code to flag a deployment as a factory reset, then disable soft reboot only for ostree < 2025.7 and isFactoryReset. @cgwalters thoughts?

cgwalters:
Oh hmmmm. Yes a bit of a pickle. Can you see if (perhaps with agent assist) we can detect if the staged deployment is missing the `ostree=` karg instead and our conditional logic skips if that's the case?

or slightly more elaborated: `if (ostree is new enough || !staged_missing_karg() { we're good } else { skip } ` or so right?

ckyrouac:
@cgwalters added another commit to automatically preserve the /boot mount

cgwalters (approved):

cgwalters:
Nice! Hopefully we get some good feedback from this one.

PR #1589, opened 2025-09-05
log to journal if root

cgwalters:
I went to look at our journal logging and it didn't do what I expected.

The intention was to always log to the systemd journal,
even if we're spawned directly on the CLI outside of a unit.


cgwalters:
After a ton of head scratching I figured out this soft requires https://github.com/tokio-rs/tracing/pull/3372

cgwalters:
> After a ton of head scratching I figured out this soft requires https://github.com/tokio-rs/tracing/pull/3372

What we really want to be clear is for the syslog identifier to be `bootc`, and we just lose that if we have to do the "re-exec dance", which we shouldn't in default production paths.

i.e. we want `journalctl -t bootc` to give us the logs.

PR #1600, opened 2025-09-09
ostree: Set parent of merge commit to base commit if present

cgwalters:
Motivated by https://github.com/ostreedev/ostree/pull/3523

This is an obvious and trivially easy thing to do here, and makes dereferencing from "merge -> base" client side also trivial which is especially important in the initramfs.


alexlarsson:
I like this. It is useful, but it also fundamentally true. The real commit *is* the parent.

alexlarsson (approved):

jeckersb (approved):

Issue #20, opened 2022-12-06
Add alternative backends


cgwalters:
We should support one or both of

- composefs
- dm-verity

composefs maps pretty naturally to OCI; we could store the extra data it needs in a derived image or an OCI artifact.  dm-verity less naturally so, though we could do something [jigdo](https://wiki.debian.org/jigdo) (or more recently [osmet](https://coreos.github.io/coreos-installer/osmet/)) like where we reassemble a filesystem image using components - i.e. bootc fetches the stock tarball layers plus a lookaside map which tells it how to reconstruct a filesystem image.

jlebon:
Would containers-storage be another potential backend? Some considerations there of course (mixing images of different trust level and duplication) but it seems natural to have the OS container go there just like any other image so then you'd manage it with `podman` as usual.

cgwalters:
Potentially.  That said, it was an intentional decision so far to separate application container storage from the OS.  This ties into larger architectural decisions taken.  Specifically today, we tend to view containers as optional.  For example, we'd need to ensure things like `podman image prune -a` don't just blow away your OS too.  I could imagine the OS being represented as an always-running container or something internally?  But it needs careful design and thought.

cheese:
How about a [transactional-update](https://github.com/openSUSE/transactional-update)/BTRFS backend? It is straightforward and naturally supports `apply-live`.

vrothberg:
@cgwalters, do you want to track this separately from the podman backend?

cgwalters:
> @cgwalters, do you want to track this separately from the podman backend?


I've reworked this issue to now be about unified storage.

> How about a [transactional-update](https://github.com/openSUSE/transactional-update)/BTRFS backend? It is straightforward and naturally supports `apply-live`.

Long story short, the goal here is to make composefs the technological heart of things. Depending on BTRFS specifically doesn't give us a lot of value versus what we already have, augmented with composefs. "Live apply" type things are related but distinct.

cgwalters:
In the short term one thing we could do is basically add an opt-in flag (part of the origin? `bootc.conf`? something) that changes the flow for fetching the base image to:

- pull into bootc-owned c/storage (where LBIs go today)
- pull from *that* via ostree-container path into the ostree repo (this will suck, it will be a full physical copy for now) via the `containers-storage:` transport
- Then we optimize that

cgwalters:
[...]

PR #1601, opened 2025-09-09
WIP: Add --unified to use our container-storage for host images

jmarrero:
Compiles and at least does not fail a bootc switch... and can see the image in the container storage we own.

```
[core@cosa-devsh ~]$ sudo podman --storage-opt=additionalimagestore=/usr/lib/bootc/storage images
REPOSITORY                       TAG         IMAGE ID      CREATED       SIZE
quay.io/jmarrero_rh/soft-reboot  1           97f84fcb062e  25 hours ago  1.83 GB
```

This was drafted with the help of Claude Code.

### DIFF
--- a/crates/lib/src/cli.rs
+++ b/crates/lib/src/cli.rs
@@ -89,6 +89,11 @@ pub(crate) struct UpgradeOpts {
     #[clap(long = "soft-reboot", conflicts_with = "check")]
     pub(crate) soft_reboot: Option<SoftRebootMode>,
 
+    /// Use podman/skopeo to pull image to additionalimagestore, then read from container storage.
+    /// This provides a unified approach that leverages existing container tooling.
+    #[clap(long)]
+    pub(crate) unified: bool,
+
     #[clap(flatten)]
     pub(crate) progress: ProgressOptions,
 }
@@ -143,6 +148,11 @@ pub(crate) struct SwitchOpts {
 
     /// Target image to use for the next boot.
     pub(crate) target: String,
+
+    /// Use podman/skopeo to pull image to additionalimagestore, then read from container storage.
+    /// This provides a unified approach that leverages existing container tooling.
+    #[clap(long)]
+    pub(crate) unified: bool,
 
     #[clap(flatten)]
     pub(crate) progress: ProgressOptions,
@@ -975,7 +985,11 @@ async fn upgrade(opts: UpgradeOpts) -> Result<()> {
             }
         }
     } else {
-        let fetched = crate::deploy::pull(repo, imgref, None, opts.quiet, prog.clone()).await?;
+        let fetched = if opts.unified {
+            crate::deploy::pull_unified(repo, imgref, None, opts.quiet, prog.clone(), sysroot).await?
+        } else {
+            crate::deploy::pull(repo, imgref, None, opts.quiet, prog.clone()).await?
+        };
         let staged_digest = staged_image.map(|s| s.digest().expect("valid digest in status"));
         let fetched_digest = &fetched.manifest_digest;
         tracing::debug!("staged: {staged_digest:?}");
@@ -1096,7 +1110,11 @@ async fn switch(opts: SwitchOpts) -> Result<()> {
 
     let new_spec = RequiredHostSpec::from_spec(&new_spec)?;
 
-    let fetched = crate::deploy::pull(repo, &target, None, opts.quiet, prog.clone()).await?;
+    let fetched = if opts.unified {
+        crate::deploy::pull_unified(repo, &target, None, opts.quiet, prog.clone(), sysroot).await?
+    } else {
+        crate::deploy::pull(repo, &target, None, opts.quiet, prog.clone()).await?
+    };
 
     if !opts.retain {
         // By default, we prune the previous ostree ref so it will go away after later upgrades
--- a/crates/lib/src/deploy.rs
+++ b/crates/lib/src/deploy.rs
@@ -380,6 +380,98 @@ pub(crate) async fn prepare_for_pull(
     Ok(PreparedPullResult::Ready(Box::new(prepared_image)))
 }
 
+/// Unified approach: First pull with podman to containers-storage, then prepare from containers-storage
+pub(crate) async fn prepare_for_pull_unified(
+    repo: &ostree::Repo,
+    imgref: &ImageReference,
+    target_imgref: Option<&OstreeImageReference>,
+    store: &Storage,
+) -> Result<PreparedPullResult> {
+    // Ensure bootc storage is properly initialized before using unified storage
+    let _imgstore = store.get_ensure_imgstore()?;
+    
+    // First, pull the image using podman with unified storage
+    crate::podman::pull_image_unified(&format!("{imgref:#}")).await?;
+    
+    // Now create a containers-storage reference to the pulled image
+    let containers_storage_imgref = ImageReference {
+        transport: "containers-storage".to_string(),
+        image: imgref.image.clone(),
+        signature: imgref.signature.clone(),
+    };
+    let ostree_imgref = &OstreeImageReference::from(containers_storage_imgref);
+    
+    // Use the standard preparation flow but reading from containers-storage
+    let mut imp = new_importer(repo, ostree_imgref).await?;
+    if let Some(target) = target_imgref {
+        imp.set_target(target);
+    }
+    let prep = match imp.prepare().await? {
+        PrepareResult::AlreadyPresent(c) => {
+            println!("No changes in {imgref:#} => {}", c.manifest_digest);
+            return Ok(PreparedPullResult::AlreadyPresent(Box::new((*c).into())));
+        }
+        PrepareResult::Ready(p) => p,
+    };
+    check_bootc_label(&prep.config);
+    if let Some(warning) = prep.deprecated_warning() {
+        ostree_ext::cli::print_deprecated_warning(warning).await;
+    }
+    ostree_ext::cli::print_layer_status(&prep);
+    let layers_to_fetch = prep.layers_to_fetch().collect::<Result<Vec<_>>>()?;
+
+    let prepared_image = PreparedImportMeta {
+        imp,
+        n_layers_to_fetch: layers_to_fetch.len(),
+        layers_total: prep.all_layers().count(),
+        bytes_to_fetch: layers_to_fetch.iter().map(|(l, _)| l.layer.size()).sum(),
+        bytes_total: prep.all_layers().map(|l| l.layer.size()).sum(),
+        digest: prep.manifest_digest.clone(),
+        prep,
+    };
+
+    Ok(PreparedPullResult::Ready(Box::new(prepared_image)))
+}
+
+/// Unified pull: Use podman to pull to containers-storage, then read from there
+pub(crate) async fn pull_unified(
+    repo: &ostree::Repo,
+    imgref: &ImageReference,
+    target_imgref: Option<&OstreeImageReference>,
+    quiet: bool,
+    prog: ProgressWriter,
+    store: &Storage,
+) -> Result<Box<ImageState>> {
+    match prepare_for_pull_unified(repo, imgref, target_imgref, store).await? {
+        PreparedPullResult::AlreadyPresent(existing) => {
+            // Log that the image was already present (Debug level since it's not actionable)
+            const IMAGE_ALREADY_PRESENT_ID: &str = "5c4d3e2f1a0b9c8d7e6f5a4b3c2d1e0f9";
+            tracing::debug!(
+                message_id = IMAGE_ALREADY_PRESENT_ID,
+                bootc.image.reference = &imgref.image,
+                bootc.image.transport = &imgref.transport,
+                bootc.status = "already_present",
+                "Image already present: {}",
+                imgref
+            );
+            Ok(existing)
+        }
+        PreparedPullResult::Ready(prepared_image_meta) => {
+            // Log that we're pulling a new image
+            const PULLING_NEW_IMAGE_ID: &str = "6d5e4f3a2b1c0d9e8f7a6b5c4d3e2f1a0";
+            tracing::info!(
+                message_id = PULLING_NEW_IMAGE_ID,
+                bootc.image.reference = &imgref.image,
+                bootc.image.transport = &imgref.transport,
+                bootc.status = "pulling_new",
+                "Pulling new image: {}",
+                imgref
+            );
+            pull_from_prepared(imgref, quiet, prog, *prepared_image_meta).await
+        }
+    }
+}
+
 #[context("Pulling")]
 pub(crate) async fn pull_from_prepared(
     imgref: &ImageReference,
--- a/crates/lib/src/podman.rs
+++ b/crates/lib/src/podman.rs
@@ -51,3 +51,27 @@ pub(crate) fn storage_exists(root: &Dir, path: impl AsRef<Utf8Path>) -> Result<b
 pub(crate) fn storage_exists_default(root: &Dir) -> Result<bool> {
     storage_exists(root, CONTAINER_STORAGE.trim_start_matches('/'))
 }
+
+/// Pull an image using podman with additionalimagestore pointing to bootc storage.
+/// This allows the image to be available in both regular container storage and bootc storage.
+pub(crate) async fn pull_image_unified(imgref: &str) -> Result<()> {
+    use bootc_utils::CommandRunExt;
+    
+    // Use podman pull with additionalimagestore pointing to bootc storage
+    let bootc_storage_path = "/usr/lib/bootc/storage";
+    
+    tracing::info!("Pulling image via podman with unified storage: {}", imgref);
+    
+    std::process::Command::new("podman")
+        .args([
+            "pull",
+            "--storage-opt",
+            &format!("additionalimagestore={}", bootc_storage_path),
+            imgref,
+        ])
+        .run_capture_stderr()
+        .map_err(|e| anyhow::anyhow!("Failed to pull image via podman: {}", e))?;
+        
+    tracing::info!("Successfully pulled image to unified storage: {}", imgref);
+    Ok(())
+}


crates/lib/src/cli.rs
```diff
@@ -89,6 +89,11 @@ pub(crate) struct UpgradeOpts {
     #[clap(long = "soft-reboot", conflicts_with = "check")]
     pub(crate) soft_reboot: Option<SoftRebootMode>,
 
+    /// Use podman/skopeo to pull image to additionalimagestore, then read from container storage.
+    /// This provides a unified approach that leverages existing container tooling.
+    #[clap(long)]
+    pub(crate) unified: bool,
```
cgwalters: I don't think this should be a CLI flag that operates just once; it should be something like `bootc image --set-unified` or so and act persistently.

Also we should support setting this at install time so that it happens from the very start.

crates/lib/src/podman.rs
```diff
@@ -51,3 +51,27 @@ pub(crate) fn storage_exists(root: &Dir, path: impl AsRef<Utf8Path>) -> Result<b
...
     storage_exists(root, CONTAINER_STORAGE.trim_start_matches('/'))
 }
+
+/// Pull an image using podman with additionalimagestore pointing to bootc storage.
+/// This allows the image to be available in both regular container storage and bootc storage.
+pub(crate) async fn pull_image_unified(imgref: &str) -> Result<()> {
+    use bootc_utils::CommandRunExt;
+    
+    // Use podman pull with additionalimagestore pointing to bootc storage
+    let bootc_storage_path = "/usr/lib/bootc/storage";
```
cgwalters: Use podstorage.rs instead please

crates/lib/src/deploy.rs
```diff
@@ -380,6 +380,98 @@ pub(crate) async fn prepare_for_pull(
...
+    imgref: &ImageReference,
+    target_imgref: Option<&OstreeImageReference>,
+    quiet: bool,
+    prog: ProgressWriter,
+    store: &Storage,
+) -> Result<Box<ImageState>> {
+    match prepare_for_pull_unified(repo, imgref, target_imgref, store).await? {
+        PreparedPullResult::AlreadyPresent(existing) => {
+            // Log that the image was already present (Debug level since it's not actionable)
+            const IMAGE_ALREADY_PRESENT_ID: &str = "5c4d3e2f1a0b9c8d7e6f5a4b3c2d1e0f9";
```
cgwalters, line 530: Note that if we go this route, we should also *not* log to the journal in the ostree pull path because otherwise we're double logging.

crates/lib/src/cli.rs
```diff
@@ -975,7 +985,11 @@ async fn upgrade(opts: UpgradeOpts) -> Result<()> {
             }
         }
     } else {
-        let fetched = crate::deploy::pull(repo, imgref, None, opts.quiet, prog.clone()).await?;
+        let fetched = if opts.unified {
+            crate::deploy::pull_unified(repo, imgref, None, opts.quiet, prog.clone(), sysroot).await?
+        } else {
+            crate::deploy::pull(repo, imgref, None, opts.quiet, prog.clone()).await?
```
cgwalters, line 971: This relates to https://github.com/bootc-dev/bootc/pull/1599#discussion_r2333604990

Basically how about changing this code to take a `Storage` which would hold the persistent flag, and then `crate::deploy::pull` would itself query that flag and change its behavior.

That would also implicitly then fix the install path to behave the same.

cgwalters (commented):
Thanks for working on this!

crates/lib/src/cli.rs
cgwalters: (hmm my bad I thought we discussed this but I failed to update the issue https://github.com/bootc-dev/bootc/issues/20 or something maybe?)

jmarrero: Ahh man yeah, we talked about saving the config on the origin file IIRC, I forgot.

crates/lib/src/podman.rs
cgwalters: To elaborate on this, an important aspect here is currently the GC of the bootc/storage instance is rooted in the set of LBIs. That will need to be extended to include the host image.

cgwalters: Another big task related to this - we'll eventually really want to use the podman HTTP APIs so we can properly monitor progress. Right now we are very crudely relying on doing this synchronously to the existing tty and reusing the podman CLI tty output. But that can't really work if we ever need to render our own progress APIs.

https://lib.rs/crates/bollard might be a good choice for this?

jlebon:
Do we get reflinks when importing from the alternate storage into the ostree repo?

cgwalters:
> Do we get reflinks when importing from the alternate storage into the ostree repo?

https://github.com/bootc-dev/bootc/issues/20 links to https://github.com/containers/container-libs/issues/144

jlebon:
> > Do we get reflinks when importing from the alternate storage into the ostree repo?
> 
> #20 links to [containers/container-libs#144](https://github.com/containers/container-libs/issues/144)

Hmm, that's about copying between two container storages, but I was asking about the ostree import. But #20 does say:

> From there, we do a reflink/hardlink copy into the ostree store (i.e. This would duplicate metadata, but not data on disk. Reflinks should always work.

Which kinda answers the question.

Basically without reflinks into the ostree storage, IMO this makes it a lot less obvious that this new option is objectively better and it instead becomes a discussion of trade-offs.

crates/lib/src/image.rs
```diff
@@ -181,3 +181,122 @@ pub(crate) async fn imgcmd_entrypoint(
...
+    tracing::info!(
+        message_id = SET_UNIFIED_JOURNAL_ID,
+        bootc.image.reference = &imgref_display.image,
+        bootc.image.transport = &imgref_display.transport,
+        "Re-pulling booted image into bootc storage via unified path: {}",
+        imgref_display
+    );
+
+    // Check if this is a localhost image with registry transport - these images
+    // were built locally and don't exist in any remote registry. We need to
```
cgwalters: I don't understand why we're special casing this

cgwalters:
On the progress status side maybe we can use https://lib.rs/crates/bollard to talk to podman and I think it should help us get data out of the events?

crates/lib/src/image.rs
jmarrero: When running the tests and doing the install there is no image in the host's containers-storage. I guess that is what `bcvk libvirt run --update-from-host` was created for, I'll try adding that to the tmt tests here. We still would copy from container-storage to container-storage so it would still be a bit different than pulling from the network.

crates/lib/src/image.rs
```diff
@@ -181,3 +181,122 @@ pub(crate) async fn imgcmd_entrypoint(
...
+    const SET_UNIFIED_JOURNAL_ID: &str = "1a0b9c8d7e6f5a4b3c2d1e0f9a8b7c6d";
+    tracing::info!(
+        message_id = SET_UNIFIED_JOURNAL_ID,
+        bootc.image.reference = &imgref_display.image,
+        bootc.image.transport = &imgref_display.transport,
+        "Re-pulling booted image into bootc storage via unified path: {}",
+        imgref_display
+    );
+
+    // Check if this is a localhost image with registry transport - these images
```
cgwalters: We had a live chat about this, and the thing we actually want to special case is "source transport is containers-storage"

crates/lib/src/cli.rs
```diff
@@ -149,6 +149,14 @@ pub(crate) struct SwitchOpts {
     #[clap(long)]
     pub(crate) retain: bool,
 
+    /// Use unified storage path to pull images (experimental)
+    ///
+    /// When enabled, this uses bootc's container storage (/usr/lib/bootc/storage) to pull
+    /// the image first, then imports it from there. This is the same approach used for
+    /// logically bound images.
+    #[clap(long = "experimental-unified-storage")]
```
cgwalters: I think this should be `hide = true`

crates/lib/src/deploy.rs
```diff
@@ -381,6 +404,188 @@ pub(crate) async fn prepare_for_pull(
...
+/// If `explicit_flag` is Some(true), unified storage is always used.
+/// If `explicit_flag` is Some(false), unified storage is never used.
+/// If `explicit_flag` is None, auto-detect based on whether the image already exists
+/// in bootc's container storage.
+///
+/// Returns true if unified storage should be used.
+pub(crate) async fn should_use_unified_storage(
+    store: &Storage,
+    imgref: &ImageReference,
+    explicit_flag: Option<bool>,
```
cgwalters: This seems strange to me; nothing AFAICS passes `Some(false)`. And if `Some(true)` is passed the caller will just get back `true`. So why even call this?

crates/lib/src/image.rs
```diff
@@ -15,11 +15,20 @@ use crate::{
...
 };
 
 /// The name of the image we push to containers-storage if nothing is specified.
 const IMAGE_DEFAULT: &str = "localhost/bootc";
 
+/// Check if an image exists in the default containers-storage (podman storage).
+async fn image_exists_in_host_storage(image: &str) -> Result<bool> {
+    use tokio::process::Command as AsyncCommand;
+    let mut cmd = AsyncCommand::new("podman");
+    cmd.args(["image", "exists", image]);
```
cgwalters, line 33: I've never been a fan of using exit codes to communicate things like this. Not a blocker but we should definitely try to cut over to using bollard or so.

crates/lib/src/image.rs
```diff
@@ -181,3 +190,137 @@ pub(crate) async fn imgcmd_entrypoint(
...
+        imgref_display
+    );
+
+    // Determine the appropriate source for pulling the image into bootc storage.
+    //
+    // Case 1: If source transport is containers-storage, the image was installed from
+    //         local container storage. The image only exists in ostree now, so we need
+    //         to export from ostree to bootc storage.
+    // Case 2: Otherwise, pull from the specified transport (usually a remote registry).
+    let is_containers_storage = imgref.transport == "containers-storage";
```
cgwalters: We have a `Transport` enum for this

crates/lib/src/utils.rs
```diff
@@ -277,6 +277,18 @@ pub(crate) fn path_relative_to(from: &Path, to: &Path) -> Result<PathBuf> {
     return Ok(final_path);
 }
 
+/// Converts an ImageReference to a container reference string suitable for use with container storage APIs.
+/// For registry transport, returns just the image name. For other transports, prepends the transport.
+pub(crate) fn imageref_to_container_ref(imgref: &crate::spec::ImageReference) -> String {
```
cgwalters: Seems cleaner to have this as a method `impl ImageReference { fn to_short_container_ref()` ?

crates/lib/src/image.rs
```diff
@@ -181,3 +190,137 @@ pub(crate) async fn imgcmd_entrypoint(
...
+        bootc.image.transport = &imgref_display.transport,
+        "Re-pulling booted image into bootc storage via unified path: {}",
+        imgref_display
+    );
+
+    // Determine the appropriate source for pulling the image into bootc storage.
+    //
+    // Case 1: If source transport is containers-storage, the image was installed from
+    //         local container storage. The image only exists in ostree now, so we need
+    //         to export from ostree to bootc storage.
```
cgwalters: It *might* only exist in ostree - but it will also commonly exist in the "app" default container storage in /var/lib/containers/storage right?

Optional/for-followup: Detect when the exact image digest exists there, and then do a copy from that instead. This would in theory be faster/better (after https://github.com/containers/container-libs/issues/144 is fixed).

crates/lib/src/image.rs
```diff
@@ -181,3 +190,137 @@ pub(crate) async fn imgcmd_entrypoint(
...
+
+    if is_containers_storage {
+        tracing::info!(
+            "Source transport is containers-storage; checking if image exists in host storage"
+        );
+
+        // Check if the image already exists in the default containers-storage.
+        // This can happen if someone did a local build (e.g., podman build) and
+        // we don't want to overwrite it with an export from ostree.
+        ensure_floating_c_storage_initialized();
```
cgwalters: I think we can put this global init somewhere farther up?

crates/lib/src/image.rs
```diff
@@ -181,3 +190,137 @@ pub(crate) async fn imgcmd_entrypoint(
...
+            tracing::info!(
+                "Image {} already exists in containers-storage, skipping ostree export",
+                &imgref.image
+            );
+        } else {
+            // The image was installed from containers-storage and now only exists in ostree.
+            // We need to export from ostree to default containers-storage (/var/lib/containers)
+            tracing::info!("Image not found in containers-storage; exporting from ostree");
+            let booted = host.status.booted.as_ref().unwrap();
+            let booted_image = booted.image.as_ref().unwrap();
```
cgwalters: Hmm I think we can avoid these unwraps by using the first arg from `crate::status::get_status_require_booted(ostree)?;` above

crates/lib/src/image.rs
```diff
@@ -181,3 +190,137 @@ pub(crate) async fn imgcmd_entrypoint(
...
+                transport: Transport::try_from(booted_image.image.transport.as_str())?,
+                name: booted_image.image.image.clone(),
+            };
+            let target = ImageReference {
+                transport: Transport::ContainerStorage,
+                name: imgref.image.clone(),
+            };
+
+            let mut opts = ostree_ext::container::store::ExportToOCIOpts::default();
+            opts.progress_to_stdout = true;
```
cgwalters, line 293: Let's add a `// TODO: bridge to progress API` here

crates/system-reinstall-bootc/src/main.rs
```diff
@@ -30,6 +30,8 @@ pub(crate) struct ReinstallOpts {
     // Note if we ever add any other options here,
     #[arg(long)]
     pub(crate) composefs_backend: bool,
+    #[arg(long = "experimental-unified-storage")]
```
cgwalters: Ditto re hide = true

crates/lib/src/image.rs
```diff
@@ -181,3 +190,137 @@ pub(crate) async fn imgcmd_entrypoint(
...
 }
+
+/// Re-pull the currently booted image into the bootc-owned container storage.
+///
+/// This onboards the system to unified storage for host images so that
+/// upgrade/switch can use the unified path automatically when the image is present.
+#[context("Setting unified storage for booted image")]
+pub(crate) async fn set_unified_entrypoint() -> Result<()> {
+    let sysroot = crate::cli::get_storage().await?;
+    let ostree = sysroot.get_ostree()?;
```
cgwalters, line 217: I think this stuff should be passed down from the caller, that's how most of the other logic works

crates/lib/src/deploy.rs
```diff
@@ -381,6 +404,180 @@ pub(crate) async fn prepare_for_pull(
...
+        "Unified pull: pulling from transport '{}' to bootc storage",
+        &imgref.transport
+    );
+
+    // Pull the image to bootc storage using the same method as LBIs
+    // Show a spinner since podman pull can take a while and doesn't output progress
+    let pull_msg = format!("Pulling {} to bootc storage", &image_ref_str);
+    async_task_with_spinner(&pull_msg, async move {
+        imgstore
+            .pull(&image_ref_str, crate::podstorage::PullMode::Always)
```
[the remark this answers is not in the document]
jmarrero: LBI are pinned to a version, I think it makes more sense when we are dealing with the image we are going to boot to Always pull if there is anything new.

cgwalters:
Hmm the failure of test-install is weird; I'm not seeing the job fail in e.g. https://github.com/bootc-dev/bootc/pull/1846 or other PRs.

crates/lib/src/deploy.rs
```diff
@@ -381,6 +404,180 @@ pub(crate) async fn prepare_for_pull(
...
+    let ostree_imgref = OstreeImageReference::from(containers_storage_imgref);
+
+    // Configure the importer to use bootc storage as an additional image store
+    let mut config = ostree_ext::containers_image_proxy::ImageProxyConfig::default();
+    let mut cmd = Command::new("skopeo");
+    // Use the actual physical path to bootc storage
+    // During install, this is the target disk's mount point; otherwise default to /sysroot
+    let sysroot_base = sysroot_path
+        .map(|p| p.to_string())
+        .unwrap_or_else(|| "/sysroot".to_string());
```
cgwalters: Let's clean this up post merge - should get this stuff from the Storage

crates/lib/src/install.rs
```diff
@@ -1237,7 +1265,10 @@ fn require_host_userns() -> Result<()> {
         .uid();
     // We must really be in a rootless container, or in some way
     // we're not part of the host user namespace.
-    ensure!(pid1_uid == 0, "{proc1} is owned by {pid1_uid}, not zero; this command must be run in the root user namespace (e.g. not rootless podman)");
+    ensure!(
```
cgwalters: Ideally we wouldn't reformat these - I have a suspicion that the test install failure is because we changed message output and that might have broken the expect() statements

crates/lib/src/install.rs
```diff
@@ -1379,8 +1413,18 @@ async fn prepare_install(
         .target_imgref
         .as_deref()
         .unwrap_or(source.imageref.name.as_str());
+    // Determine the target transport:
+    // - If the source transport is containers-storage, use that as the target too
```
cgwalters: I think this an incompatible behavior change to do by default.

tmt/tests/booted/test-switch-to-unified.nu
```diff
@@ -0,0 +1,84 @@
+use std assert
+use tap.nu
+
```
cgwalters, line 8: You could opt in to 
```
# extra:
#   try_bind_storage: true
```

And when we do this, we'll have the container image accessible via the additionalimagestore (AIS) from the *host*.

crates/lib/src/install.rs
cgwalters: We chatted about this live and the result is that we have to special case our tests, not the install code.

tmt/tests/booted/test-switch-to-unified.nu
jmarrero: I was able to just use   bootc image copy-to-storage

crates/lib/src/install.rs
jmarrero: removed this change and now just deal with it on the test.

crates/lib/src/image.rs
jmarrero: TODO added for bollard or so.

jmarrero:
Failures on rawhide are the same we saw in another PR yesterday: https://github.com/bootc-dev/bootc/issues/1849

docs/src/man/bootc-install-to-disk.8.md
```diff
@@ -92,6 +92,10 @@ partition is auto-discovered.
 
     Verify the image can be fetched from the bootc image. Updates may fail when the installation host is authenticated with the registry but the pull secret is not in the bootc image
 
+**--experimental-unified-storage**
```
cgwalters: Hmm why did this show up in the man pages? Is our man page not handling `hide = true`? Or did you forget to regenerate? (We probably need a CI gate for this too)

crates/lib/src/spec.rs
```diff
@@ -132,6 +132,23 @@ impl ImageReference {
             }
         }
     }
+
+    /// Parse the transport string into a Transport enum.
+    pub fn transport(&self) -> Transport {
+        Transport::try_from(self.transport.as_str()).unwrap_or(Transport::Registry)
```
cgwalters: Under what scenarios would we want to silently fall back to registry if we fail to parse the input string? I think parsing should *usually* error out and not have silent fallbacks.

In the function just above this we do `        let transport = Transport::try_from(self.transport.as_str())?;` which feels more correct to me.

(This *shouldn't* fail of course and per that TODO it'd be better to represent with a proper enum in this code too)

jmarrero: I think there is no scenario this is needed. Updated in the last push.

docs/src/man/bootc-install-to-disk.8.md
jmarrero: I just missed this when I added the hidden to the removed system-reinstall-bootc code flags.

crates/lib/src/deploy.rs
jmarrero: Added a new commit doing this.

crates/lib/src/deploy.rs
```diff
@@ -467,14 +462,10 @@ pub(crate) async fn prepare_for_pull_unified(
     // Configure the importer to use bootc storage as an additional image store
     let mut config = ostree_ext::containers_image_proxy::ImageProxyConfig::default();
     let mut cmd = Command::new("skopeo");
-    // Use the actual physical path to bootc storage
-    // During install, this is the target disk's mount point; otherwise default to /sysroot
-    let sysroot_base = sysroot_path
-        .map(|p| p.to_string())
-        .unwrap_or_else(|| "/sysroot".to_string());
+    // Use the physical path to bootc storage from the Storage struct
```
cgwalters, line 465: This is OK but I've been trying to ensure that bootc uses file descriptors and not file paths as much as possible. The problem here is the case where we have a fd but need to pass that state down to a subprocess which expects a file path.

podstorage.rs goes to quite some effort to do this for the primary storage, see `new_podman_cmd_in` - I think we could use a similar technique here.

PR #1603, opened 2025-09-09
kernel_cmdline: Refactor into separate `bytes` and `utf8` modules

jeckersb:
Split the kernel command line parsing functionality into two focused
modules. The `bytes` module handles raw byte parsing without UTF-8
requirements, matching kernel behavior for arbitrary byte
sequences. The `utf8` module provides string-based parsing for cases
where UTF-8 validation is needed.  The `utf8` module reuses the
`bytes` module primitives where possible, and uses the fact that
`utf8::Cmdline` can only be constructed from valid UTF-8 to do
unchecked conversions between the two.

### DIFF
--- a/crates/kernel_cmdline/src/bytes.rs
+++ b/crates/kernel_cmdline/src/bytes.rs
@@ -5,6 +5,8 @@
 
 use std::borrow::Cow;
 
+use crate::utf8;
+
 use anyhow::Result;
 
 /// A parsed kernel command line.
@@ -64,6 +66,13 @@ impl<'a> Cmdline<'a> {
         CmdlineIter(&self.0)
     }
 
+    /// Returns an iterator over all parameters in the command line
+    /// which are valid UTF-8.
+    pub fn iter_utf8(&'a self) -> impl Iterator<Item = utf8::Parameter<'a>> {
+        self.iter()
+            .filter_map(|p| utf8::Parameter::try_from(p).ok())
+    }
+
     /// Locate a kernel argument with the given key name.
     ///
     /// Returns the first parameter matching the given key, or `None` if not found.
@@ -73,14 +82,32 @@ impl<'a> Cmdline<'a> {
         self.iter().find(|p| p.key == key)
     }
 
+    /// Locate a kernel argument with the given key name.
+    ///
+    /// Returns an error if a parameter with the given key name is
+    /// found, but the value is not valid UTF-8.
+    ///
+    /// Otherwise, returns the first parameter matching the given key,
+    /// or `None` if not found.  Key comparison treats dashes and
+    /// underscores as equivalent.
+    pub fn find_utf8(&'a self, key: impl AsRef<str>) -> Result<Option<utf8::Parameter<'a>>> {
+        let bytes = match self.find(key.as_ref()) {
+            Some(p) => p,
+            None => return Ok(None),
+        };
+
+        Ok(Some(utf8::Parameter::try_from(bytes)?))
+    }
+
     /// Find all kernel arguments starting with the given prefix.
     ///
     /// This is a variant of [`Self::find`].
-    pub fn find_all_starting_with(
+    pub fn find_all_starting_with<T: AsRef<[u8]> + ?Sized>(
         &'a self,
-        prefix: &'a [u8],
+        prefix: &'a T,
     ) -> impl Iterator<Item = Parameter<'a>> + 'a {
-        self.iter().filter(move |p| p.key.0.starts_with(prefix))
+        self.iter()
+            .filter(move |p| p.key.0.starts_with(prefix.as_ref()))
     }
 
     /// Locate the value of the kernel argument with the given key name.
@@ -199,11 +226,10 @@ impl<'a> Parameter<'a> {
                 value = &value[1..];
 
                 // *Only* the first and last double quotes are stripped
-                value = value
-                    .strip_prefix(b"\"")
-                    .unwrap_or(value)
-                    .strip_suffix(b"\"")
-                    .unwrap_or(value);
+                value = {
+                    let v = value.strip_prefix(b"\"").unwrap_or(value);
+                    v.strip_suffix(b"\"").unwrap_or(v)
+                };
 
                 Self {
                     key,
@@ -237,9 +263,13 @@ impl<'a> PartialEq for Parameter<'a> {
 mod tests {
     use super::*;
 
-    // convenience method for tests
+    // convenience methods for tests
     fn param(s: &str) -> Parameter<'_> {
         Parameter::parse(s.as_bytes()).0.unwrap()
+    }
+
+    fn param_utf8(s: &str) -> utf8::Parameter<'_> {
+        utf8::Parameter::parse(s).0.unwrap()
     }
 
     #[test]
@@ -281,6 +311,12 @@ mod tests {
     fn test_parameter_quoted() {
         let p = param("foo=\"quoted value\"");
         assert_eq!(p.value, Some(b"quoted value".as_slice()));
+
+        let p = param("foo=\"unclosed quotes");
+        assert_eq!(p.value, Some(b"unclosed quotes".as_slice()));
+
+        let p = param("foo=trailing_quotes\"");
+        assert_eq!(p.value, Some(b"trailing_quotes".as_slice()));
     }
 
     #[test]
@@ -367,6 +403,38 @@ mod tests {
         // Test the find API
         assert_eq!(kargs.find("foo").unwrap().value.unwrap(), b"bar,bar2");
         assert!(kargs.find("nothing").is_none());
+    }
+
+    #[test]
+    fn test_kargs_iter_utf8() {
+        let kargs = Cmdline::from(b"foo=bar,bar2 \xff baz=fuz bad=oh\xffno wiz");
+        let mut iter = kargs.iter_utf8();
+
+        assert_eq!(iter.next(), Some(param_utf8("foo=bar,bar2")));
+        assert_eq!(iter.next(), Some(param_utf8("baz=fuz")));
+        assert_eq!(iter.next(), Some(param_utf8("wiz")));
+        assert_eq!(iter.next(), None);
+    }
+
+    #[test]
+    fn test_kargs_find_utf8() {
+        let kargs = Cmdline::from(b"foo=bar,bar2 \xff baz=fuz bad=oh\xffno wiz");
+
+        // found it
+        assert_eq!(
+            kargs.find_utf8("foo").unwrap().unwrap().value().unwrap(),
+            "bar,bar2"
+        );
+
+        // didn't find it
+        assert!(kargs.find_utf8("nothing").unwrap().is_none());
+
+        // found it but key is invalid
+        let p = kargs.find_utf8("bad");
+        assert_eq!(
+            p.unwrap_err().to_string(),
+            "Parameter value is not valid UTF-8"
+        );
     }
 
     #[test]
--- a/crates/kernel_cmdline/src/utf8.rs
+++ b/crates/kernel_cmdline/src/utf8.rs
@@ -3,6 +3,8 @@
 //! This module provides functionality for parsing and working with kernel command line
 //! arguments, supporting both key-only switches and key-value pairs with proper quote handling.
 
+use std::ops::Deref;
+
 use crate::bytes;
 
 use anyhow::Result;
@@ -34,7 +36,7 @@ impl<'a> Iterator for CmdlineIter<'a> {
     type Item = Parameter<'a>;
 
     fn next(&mut self) -> Option<Self::Item> {
-        self.0.next().map(Into::into)
+        self.0.next().map(Parameter::from_bytes)
     }
 }
 
@@ -71,7 +73,8 @@ impl<'a> Cmdline<'a> {
     /// Returns the first parameter matching the given key, or `None` if not found.
     /// Key comparison treats dashes and underscores as equivalent.
     pub fn find(&'a self, key: impl AsRef<str>) -> Option<Parameter<'a>> {
-        self.iter().find(|p| p.key() == key.as_ref().into())
+        let key = ParameterKey::from(key.as_ref());
+        self.iter().find(|p| p.key() == key)
     }
 
     /// Find all kernel arguments starting with the given UTF-8 prefix.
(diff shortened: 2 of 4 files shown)


crates/kernel_cmdline/src/bytes.rs
```diff
@@ -0,0 +1,474 @@
...
+        Parameter::parse(s.as_bytes()).0.unwrap()
+    }
+
+    #[test]
+    fn test_parameter_parse() {
+        let (p, rest) = Parameter::parse(b"foo");
+        let p = p.unwrap();
+        let key = p.key();
+        println!("{key:?}");
+        //assert_eq!(*p.key(), b"foo");
```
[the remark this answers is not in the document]
jeckersb: good bot

crates/kernel_cmdline/src/utf8.rs
```diff
@@ -0,0 +1,479 @@
...
+
+    /// Locate the value of the kernel argument with the given key name.
+    ///
+    /// Returns the first value matching the given key, or `None` if not found.
+    /// Key comparison treats dashes and underscores as equivalent.
+    pub fn value_of(&'a self, key: impl AsRef<str>) -> Option<&'a str> {
+        self.0.value_of(key.as_ref().as_bytes()).map(|v| {
+            // SAFETY: We know this is valid UTF-8 since we only
+            // construct the underlying `bytes` from valid UTF-8
+            #[allow(unsafe_code)]
```
cgwalters: I am not a big fan of unsafe code when we're talking about string/buffer handling.

That said, this pattern is also how e.g. [camino works](https://github.com/camino-rs/camino/blob/2d6994cc68d298dcae922ec21bb5d5efb9aca0b0/src/lib.rs#L647). Or probably [parent() is a better example](https://github.com/camino-rs/camino/blob/2d6994cc68d298dcae922ec21bb5d5efb9aca0b0/src/lib.rs#L880) - basically calling into the std `Path` to parse and then re-casting.

---

Offhand...is it *really* that onerous to duplicate the parsing for bytes vs str? It seems like it shouldn't.

Honestly though the more I think about this the more I feel we probably when parsing `/proc/cmdline` just want to basically ignore non-UTF8 values right? Do we actually *need* any of this binary handling? I can't think of when we would...

I get that you already did all this work and the code looks generally sane, but since it has `unsafe` it just needs careful review and I think we need to weigh that vs other approaches.

cgwalters: When this whole karg thing came up I just wanted us to not panic or totally fail if `/proc/cmdline` has binary data.

I can see two potential behaviors for non-utf8 in the kargs:

- Ignore them
- Error out cleanly

I think *most* cases we want the first. But ostree does have some code that basically "propagates forward" `/proc/cmdline` back into the bootloader entries, and in that case we should cleanly error out I'd say.

jeckersb: Yeah I really try to avoid `unsafe` too, but it felt like this was one of the cases where it actually made sense.  I'm not tied to it, and at the very least I could easily change it to a checked conversion everywhere and `.expect()` it to be valid UTF-8, then at least we'd get a panic if something really was wrong instead of UB.

FWIW the only way I think this could possibly go wrong would be if you constructed a valid UTF-8 string that, when split on ascii whitespace, would result in one of the split halves somehow being invalid UTF-8.  I don't think that's possible, and claude at least seems to agree:

```
> is it possible to split a utf-8 string on ascii whitespace, and for one of
  the resulting halves to *not* be valid utf-8?

● No, it's not possible. When you split a valid UTF-8 string on ASCII
  whitespace, all resulting parts will be valid UTF-8.

  This is because:

  1. ASCII whitespace characters (space, tab, newline, etc.) are single
  bytes with values < 128
  2. UTF-8 is designed so that ASCII characters never appear as part of
  multi-byte sequences
  3. Multi-byte UTF-8 sequences use bytes with the high bit set (≥ 128)

  Since ASCII whitespace can never occur in the middle of a multi-byte UTF-8
   character, splitting on whitespace will never break apart a multi-byte
  sequence, ensuring all resulting substrings remain valid UTF-8.
```

jeckersb: re: actually needing the binary handling, no not currently in the way we're using it in bootc (and soon composefs-boot).  That was just trying to get it as close as possible to the way the kernel actually parses things.  But as a general-use cmdline library it's desirable because you can absolutely pass around non-utf8 data in the cmdline.

cgwalters: > and at the very least I could easily change it to a checked conversion everywhere and .expect() it to be valid UTF-8, then at least we'd get a panic if something really was wrong instead of UB.

Yes let's at least do that

jeckersb:
I replaced all of the `unsafe` usage with checked + `expect`.  It's not like this is performance-critical, the extra checks aren't going to hurt anything.  I'm as certain as I can be that it will never panic (but if it does... better than UB).

crates/kernel_cmdline/src/utf8.rs
```diff
@@ -0,0 +1,470 @@
...
+    }
+}
+
+impl<'a> Cmdline<'a> {
+    /// Reads the kernel command line from `/proc/cmdline`.
+    ///
+    /// Returns an error if:
+    ///   - The file cannot be read
+    ///   - There are I/O issues
+    ///   - The cmdline from proc is not valid UTF-8
```
cgwalters, line 49: You're saying in a different thread that it's valid to have non-utf8 in the kernel commandline - but if we we error out in this case it will just fundamentally break bootc and hence their ability to upgrade the OS right?

One of these things is incorrect. What I think is incorrect is the second - the `utf8::Cmdline` should have `from_bytes_skip_non_utf8` or so (and `from_proc` wraps that feeding it the contents of `/proc/cmdline`). This would skip parameters that have invalid UTF-8.

And that's what we should use across the board.

Or if you want what might be simpler is have `Cmdline::from_proc_cmdline_utf8_lossy()`, but what makes me uncomfortable about that it is probably *more* confusing if a parmeter we're looking for has its value silently changed into a UTF8 replacement char, vs just not seen at all?

cgwalters: If we don't error out (and I don't think we should) we should probably still *warn* at least to the systemd journal? (Bikeshed: In theory we could implement warn-on-once with a stamp file in `/run/bootc/karg-warning-emitted` or something but eh)

jeckersb: We [originally](https://github.com/bootc-dev/bootc/pull/1476) had the whole lossy thing and got rid of it.

jeckersb: I lean towards hard-requiring UTF-8 for bootc, because we're planning to follow the bls spec and it [explicitly requires](https://uapi-group.org/specifications/specs/boot_loader_specification/#type-1-boot-loader-specification-entries) UTF-8, and as you point out there's not really a "right" way to go from non-UTF-8 to UTF-8.

cgwalters: > We https://github.com/bootc-dev/bootc/pull/1476 had the whole lossy thing and got rid of it.

Yeah I don't like just blindly blasting `.to_utf8_lossy()` on things. It's fine in e.g. tests when we're parsing subprocess output or whatever, but in prod code paths it's different.

The 90% case is we are parsing the cmdline to find a specific key I think.

Then there are two sub-cases:

- We encounter a key with invalid UTF8; I think we should skip these
- We encounter a key with *valid* UTF8 but whose value is invalid UTF8 *and* it's a key we're looking for (e.g. `bootc.composefs.digest=` or whatever); I think this should probably be an error

crates/kernel_cmdline/src/utf8.rs
```diff
@@ -0,0 +1,470 @@
...
+    }
+}
+
+impl<'a> From<&'a str> for ParameterKey<'a> {
+    fn from(input: &'a str) -> Self {
+        Self(bytes::ParameterKey(input.as_bytes()))
+    }
+}
+
+impl<'a> From<bytes::ParameterKey<'a>> for ParameterKey<'a> {
```
jeckersb: This is wrong because implementing the trait makes it part of the public api, allowing users to create `utf8::ParameterKey` from a non-UTF-8 `bytes::ParameterKey`.  Should be `TryFrom` instead, and/or moved to private as `from_unchecked` (probably both).

crates/kernel_cmdline/src/utf8.rs
```diff
@@ -0,0 +1,470 @@
...
+    pub fn value(&'a self) -> Option<&'a str> {
+        self.0.value().map(|p| {
+            // SAFETY: We know this is valid UTF-8 since we only
+            // construct the underlying `bytes` from valid UTF-8
+            str::from_utf8(p).expect("We only construct the underlying bytes from valid UTF-8")
+        })
+    }
+}
+
+impl<'a> From<bytes::Parameter<'a>> for Parameter<'a> {
```
jeckersb: Ditto above re: `From/TryFrom`.

crates/kernel_cmdline/src/utf8.rs
jeckersb: So in that case, I guess we switch bootc back to the `bytes` variant (fortunately very easy!), and then implement `TryFrom<bytes::Parameter> for utf8::Parameter`.  Then in bootc we can (for example) `cmdline.find("root").try_into()?` and that will only return an error in the case that the value for "root" is invalid.  But it will ignore any invalid utf-8 elsewhere in the cmdline.

cgwalters: > I lean towards hard-requiring UTF-8 for bootc, because we're planning to follow the bls spec

We'll parse a kernel commandline parameter in the initramfs for composefs; are you saying we should *fail to boot* then if there's non-UTF8 in the commandline?

The other case where we parse the commandline is `install to-existing-root` where we auto-discover some kargs for the rootfs. I could imagine hard failing there, but OTOH again we are only explicitly taking some arguments ISTM we should just ignore non-UTF8 key/value pairs there. Unless someone has `root=<some non-UTF8 value>` in which case we'd error.

Of note of course since ostree is written in C which long predates Unicode, there isn't an attempt to enforce UTF8 in kargs and the parser is I believe lenient in that sense, but there's no tests that I know of.

I bet systemd also does not require UTF8.

jeckersb: No for the initramfs case clearly that would be wrong.  But I think this is a good exercise and shows why this refactor is ultimately useful - there's lots of little subtle landmines here and I think this makes it more obvious what's going on.  This gives the flexibility to mix and match both where it makes sense.  For the code dealing with reading/writing BLS files we'll definitely want to use only the `utf8` module.

jeckersb:
Ok I think this new commit works the way we want it to.  Gracefully handles non-UTF-8 data but errors out if we have non-UTF-8 data in the specific parameters we're interested in.

crates/lib/src/install.rs
```diff
@@ -1653,22 +1657,26 @@ struct RootMountInfo {
...
-        .value_of_utf8("root")
-        .context("Parsing root= karg")?;
+fn find_root_args_to_inherit(
+    cmdline: &bytes::Cmdline,
+    root_info: &Filesystem,
+) -> Result<RootMountInfo> {
     // If we have a root= karg, then use that
+    let root = cmdline
+        .find("root")
+        .map(utf8::Parameter::try_from)
```
cgwalters: How about `find_utf8` which handles this since that's going to be the 90% case - us looking for a utf8 key+value?

That would be best implemented in terms of `iter_utf8` on the bytes one which would basically be `.iter().filter(utf8::Parameter::try_from.ok())` right?

cgwalters (approved):

crates/kernel_cmdline/src/bytes.rs
```diff
@@ -0,0 +1,472 @@
...
+                // skip `=`, we know it's the first byte because we
+                // found it above
+                value = &value[1..];
+
+                // *Only* the first and last double quotes are stripped
+                value = value
+                    .strip_prefix(b"\"")
+                    .unwrap_or(value)
+                    .strip_suffix(b"\"")
+                    .unwrap_or(value);
```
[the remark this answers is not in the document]
jeckersb: Good catch bot, although this hasn't changed, how did it take you until now to realize this? :laughing:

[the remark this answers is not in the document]
jeckersb: Although I don't think this can actually happen, since due to the way we parse it we always match quotes.

Edit:  Well if we run up against EOS... I dunno.  Should write some tests I guess!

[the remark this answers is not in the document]
jeckersb: Yeah right now what gemini says is correct if we run up against EOS.  Referring back to the [kernel implementation](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/lib/cmdline.c?id=e8d780dcd957d80725ad5dd00bab53b856429bc0#n227) it will blindly strip from both ends even if it turns out it was unmatched.  So we should fix this to do the same.

crates/lib/src/install.rs
jeckersb: Hm I implemented this, but now we don't return an error in the case where kargs is `root=invalid\xffutf8`, and instead fall back to the case where `root` doesn't appear at all.  Desirable?  I'm not sure.

cgwalters: Right, to get the error semantic we'd need `find_utf8` to be directly coded to do that. Would make sense to me.

crates/kernel_cmdline/src/utf8.rs
jeckersb: This should all be resolved in the latest version

crates/lib/src/install.rs
jeckersb: Ok I've updated `find_utf8` to return `Result<Option<utf8::Parameter>>>` and instead of using `iter_utf8` it instead uses the underlying `find` so it will catch the case of `root=invalid\xffutf8` above and return an error.

crates/lib/src/install.rs
```diff
@@ -83,6 +83,10 @@ const SELINUXFS: &str = "/sys/fs/selinux";
 /// The mount path for uefi
 const EFIVARFS: &str = "/sys/firmware/efi/efivars";
 pub(crate) const ARCH_USES_EFI: bool = cfg!(any(target_arch = "x86_64", target_arch = "aarch64"));
+/// This is used by dracut.
+pub const INITRD_ARG_PREFIX: &[u8] = b"rd.";
+/// The kernel argument for configuring the rootfs flags.
+pub const ROOTFLAGS: &[u8] = b"rootflags";
```
cgwalters: Can't these remain as `&str`?

jeckersb: Yeah but requires updating `bytes::Cmdline::find_all_starting_with` to take `AsRef<[u8]>` instead of just `&'a [u8]` which is fine.

PR #1606, opened 2025-09-11
ci: Also validate the composefs-backend feature

cgwalters:
I was hitting a compliation error which *looks*
like it was actaully an incremental compilation bug? Or it might have been rust-analyzer and local builds fighting over enabled features.

Anyways, this ensures that we're gating on the composefs backend compiling.

### DIFF
--- a/crates/initramfs/src/lib.rs
+++ b/crates/initramfs/src/lib.rs
@@ -112,7 +112,7 @@ pub fn mount_at_wrapper(
         .with_context(|| format!("Mounting at path {path:?}"))
 }
 
-/// Wrapper around [`rustix::openat`]
+/// Wrapper around [`rustix::fs::openat`]
 #[context("Opening dir {name:?}")]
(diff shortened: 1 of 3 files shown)


crates/initramfs/src/lib.rs
```diff
@@ -112,7 +112,7 @@ pub fn mount_at_wrapper(
         .with_context(|| format!("Mounting at path {path:?}"))
 }
 
-/// Wrapper around [`rustix::openat`]
+/// Wrapper around [`rustix::fs::openat`]
```
jeckersb, line 115: accidentally included this?

cgwalters: Nope it's intentional, `make validate-rust` trips on it. See https://docs.rs/rustix/latest/rustix/fs/fn.openat.html

jeckersb:
Hm is something wrong with the repo config?  I have the "Merge pull request" button but there's not an approval on this yet.  I won't touch it for the moment just incase we need this state to troubleshoot...

cgwalters:
branch protection doesn't apply to anything other than main right now

Issue #1473, opened 2025-08-01
rework tmt to use pre-built container


cgwalters:
OK so we have a lot of CI sprawl, across GHA and Testing Farm among others. That's not going to go away, but we need to keep chipping away at it.

## Testing Farm/packit is really RPM oriented and less so containers, and especially less so bootc

We've been through a few iterations of bootc/tmt/testing farm iteration. A while ago we added the `bootc` provisioner plugin but that only works on host systems with virtualization available (if run in a cloud would require nested virt) and is specific to bootc-image-builder. (Of course we could teach it to make custom AMIs etc. but eww)

More recently we added https://github.com/bootc-dev/bootc/blob/main/tmt/tests/bootc-install-provision.sh which uses the `install to-existing-root` flow which simplifies things hugely from a testing perspective. And of course - it has the same "you don't need to build disk images" simplicity for users!

But the massive downside of the way we're doing things right now is we build from source for *each test*. (Also it's annoying to share code between this and the "main" Dockerfile, though we could figure out how to automate that)

What I really really really want is to be able to just test a bootc container image I have already built.

### Wedging containers somehow into packit + testing-farm + tmt

The trifecta of packit + testing-farm + tmt is mostly logical, but it's all quite complicated and one can clearly see the different "gaps" between them technically/organizationally.

By far the biggest thing here is again the use case of "I want to run my local build in the target environment" - xref https://github.com/teemtee/tmt/issues/1018

OK so here's my thought:

- Kill the `bootc-install-provision.sh` thing
- Change our tmt tests to require a container image reference in the environment
- Require use of https://github.com/cgwalters/cstor-dist for local testing

Anyone see issues with that?

### More on the "packit -> testing-farm bridge"

So right now AFAIK basically our CI is doing "build an RPM in copr driven by packit" and then passing that off to testing farm and we end up sticking that RPM into a local container build run in each test.

What I'm thinking here is for pull request CI we could probably try either:

- Patch packit to learn how to build and push containers too
- Figure out how to insert custom functionality like this into the CI flow (*maybe* packit has some kind of hook we could use?) Basically, wait for the COPR build and then we take over build+push container to e.g. `ghcr.io/bootc-dev/bootc-pr-12345-<digest>` and passing that into testing farm?

### Alternative 2: Drop packit+copr

Another alternative which actually has a fair amount of appeal is to just stop using packit - we just use the upstream `podman build` (which doesn't build an RPM today).

[...]

PR #1607, opened 2025-09-11
ci: Unify more of hack/ and tests/

cgwalters:
A key thing for me is that the `Justfile` should be a one-stop shop for development of the project. It can't have everything but it should answer the basic questions of "how do I build and test this project".

This aligns the recently added tmt-on-GHA flow a *bit* more closely with some of that. Biggest is to use the `just build-integration-test-image` as the canonical way to build a container image with our testing stuff in it; which uses our main Dockerfile

Other cleanups:
- Change the scripts to accept data via argv[1] and not environment
- Drop the hardcoded testing directory and use `target/` as a generic build artifact dir

Closes: https://github.com/bootc-dev/bootc/issues/1473


.github/workflows/integration.yml
```diff
@@ -1,14 +1,16 @@
...
-    branches: [main]
+    branches: [main]  
+  workflow_dispatch:
 
 jobs:
   build:
     strategy:
       matrix:
-        test_os: [fedora-41, fedora-42, fedora-43, centos-9]
-        test_runner: [ubuntu-latest, ubuntu-24.04-arm]
```
henrywang, line 11: We do not run arm build test in github action runner? No nested virt support in github arm runner. 

So only bootc build and bootc image build test can be run in github action in our case.

We can only keep `x86_64` test running in github action, and run arm test on `Packit`.

.github/workflows/integration.yml
```diff
@@ -18,67 +20,71 @@ jobs:
...
 
       - uses: actions/checkout@v4
 
-      - name: Build bootc and bootc image
-        env:
-          TEST_OS: ${{ matrix.test_os }}
-        run: sudo -E TEST_OS=$TEST_OS tests/build.sh
+      - name: Set architecture variable
+        id: set_arch
+        run: echo "ARCH=$(arch)" >> $GITHUB_ENV
```
henrywang, line 35: If we do not run arm test here, we don't need `ARCH` env.

tests/build.sh
```diff
@@ -4,115 +4,65 @@ set -exuo pipefail
...
+        *) echo "Unknown OS: ${OS_ID}" 1>&2; exit 1
+        ;;
+    esac
+    build_args+=("--build-arg=base=$BASE")
+fi
 
-rm -rf /var/cache /var/lib/dnf
-EORUN
-CONTAINERFILEOF
+just build ${build_args[@]}
```
henrywang, line 36: For local running, the `just` has to be installed locally, right?

.github/workflows/integration.yml
```diff
@@ -18,67 +20,71 @@ jobs:
...
+
       - name: Install dependence
         run: |
           sudo apt-get update
-          sudo apt install -y qemu-kvm qemu-system
-          pip install --user tmt
+          sudo apt install -y qemu-kvm qemu-system just
+          sudo pip install --user tmt
```
henrywang: Can't use `sudo` in our case. The later `tmt` is running on rootless.

tests/build.sh
cgwalters: Yeah I'm trying to centralize more of the dev/test entrypoint around `just`

.github/workflows/integration.yml
cgwalters: > No nested virt support in github arm runner.

:cry: 

> We can only keep x86_64 test running in github action, and run arm test on Packit.

Hmmm...maybe yeah that's one way to dice the split.

I mean now, in reality, very very very little of the code is architecture sensitive and it's not clear to me we need to run *all* tests across both architectures across *four operating system major versions*...but trimming down that matrix is a whole other topic.

.github/workflows/integration.yml
cgwalters: True, but it feels cleaner to include it right?

.github/workflows/integration.yml
cgwalters: OK back to running tmt as non-root.

cgwalters:
OK right, so moving the tmt stuff into a subdirectory just breaks packit. Argh! We're just really going to have to get that tmt+rsync bug fixed. I filed https://github.com/teemtee/tmt/issues/4062

For now I'll go back to undoing the move and seeing if we can work around the tmt-rsync bug another way...

cgwalters:
> For now I'll go back to undoing the move and seeing if we can work around the tmt-rsync bug another way...

Hacked this by wrapping our tmt bits with a copy to a tempdir of the tests and running tmt from there.

cgwalters:
Man, I lost way too much time to https://github.com/teemtee/testcloud/issues/17

tmt/tests/bootc-install-provision.sh
henrywang, line 1: I know it's bad, but we have to keep this script and tmt plan to work with gating test. Gating test does not have virt support TF runner.

henrywang: Oh, we can use `system-reinstall-bootc` for this part. We can land this PR first, then I'll send another PR to have gating test supported.

cgwalters:
OK so this now works, but in unifying the (3!) different container builds we had going on and now we're including LBIs in some of the `install to-existing-root` tests, that reveals what looks like a different bug where we're leaking the overlay mount for our containers-storage instance.

EDIT: filed as https://github.com/bootc-dev/bootc/issues/1618

tmt/tests/bootc-install-provision.sh
cgwalters: > I know it's bad, but we have to keep this script and tmt plan to work with gating test. Gating test does not have virt support TF runner.

Right. OK let's back up, in general our tests should work in *all* provisioning modes (Anaconda, to-existing-root, starting from an existing disk image (whether BIB or a plain `to-disk`). That's of course quite a matrix of things.

The biggest problem with the prior status quo was building bootc from source code in each test. That completely ruins iteration speed.

Now the more I think about this...at least for the Packit flow, it always injects a built RPM into the target environment right? So we should have always been able to use that instead of rebuilding.

Although digging in, I'm not seeing anything like that in our runs... Ah is it because we have `skip_build: true` in our packit config? It must be.

The annoying thing here AFAIK is https://github.com/teemtee/tmt/issues/1018 - in order to run *that* workflow locally we'll need bespoke scripts to build an RPM from current sources and pass it to tmt the same way copr+testing-farm is doing.

cgwalters: There's also a related but different thread here in that I want to continue to support a container-native CI flow that isn't deeply tied into the plethora of Fedora-derivative specific CI/RPM tools...i.e. I want `docker|podman build` to continue to do something useful directly. But that shouldn't be too hard.

henrywang: `container-native CI flow` +1

henrywang: > The biggest problem with the prior status quo was building bootc from source code in each test. That completely ruins iteration speed.

In Packit/Gating test, yes, we have to do that. The Packit and gating environment can't deploy image mode VM for running test.

Testing Farm can deploy a image mode VM, so we can build a new image and switch to it.

cgwalters: > In Packit/Gating test, yes, we have to do that.

I don't think we ever needed to build from source in packit/gating, we just need to be able to start from a built RPM and turn the existing system into image mode via a local build...I may try a spike on that.

> The Packit and gating environment can't deploy image mode VM for running test.

Surely they can be made to do so, right? Is this about missing compose bound images?

cgwalters:
YES! This one finally passes.

cgwalters:
Now in order to merge this, we need to change the merge gates to require different GHA contexts. I can do that once we agree to merge.

(What I actually want is to have the list of required CI contexts listed in the repo itself; at least Gemini suggests open coding this with a polling task which is kind of icky but maybe we can put something for this in https://github.com/bootc-dev/infra )

cgwalters:
Who is taking the review/merge of this?

jeckersb:
> Who is taking the review/merge of this?

I'll take it

jeckersb:
Initial thoughts are this looks great to me, I'm running through it again except this time trying the `bcvk` route.

jeckersb (approved):
Well, the non-bcvk case works fine for me, I still need to do some untangling of things on my end so I won't block on that.

cgwalters:
> Initial thoughts are this looks great to me, I'm running through it again except this time trying the bcvk route.

You need https://github.com/cgwalters/bcvk/pull/6

cgwalters:
> Now in order to merge this, we need to change the merge gates to require different GHA contexts. I can do that once we agree to merge.

This should now be done, the merge gates are now switched to the GHA runs of tmt for c9s, c10s and fedora-42.

PR #1610, opened 2025-09-12
    generator: Conditionally enable bootc-status units

cgwalters:
Right now this service fails in `bcvk run-ephemeral`, but
also likely fails in any non-bootc system that has `subscription-manager`
installed.
    
A problem is that dependencies of units are started even
if the dependee has a condition that disables it.
    
This basically the target and path depend on `/run/ostree-booted`
being present (which yes, won't work for composefs...)
    
Tests: Covered by extant `012-test-unit-status.nu`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -87,8 +87,15 @@ RUN --mount=type=cache,target=/build/target --mount=type=cache,target=/var/rooth
 
 # The final image that derives from the original base and adds the release binaries
 FROM base
-# First, create a layer that is our new binaries.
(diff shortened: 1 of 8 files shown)


cgwalters:
OK I reworked this more deeply.

cgwalters:
OK, reworked this one even more. It took me a bit of time to figure out of course the reason my changes weren't working was actually just the build leaking state from the old RPM. Per the comment I guess we probably need to build packages in the Dockerfile.

PR #1611, opened 2025-09-12
cmdline tweaks and improvements

jeckersb:
See individual commits, these are things that came up as I was switching over composefs-boot to use bootc-kernel-cmdline.  Leaving as draft for now since the composefs-boot side is still WIP so this may change.


jeckersb:
Gonna go ahead and merge this, if there's anything else that comes up on the composefs-boot side we can do a followup

PR #1614, opened 2025-09-15
examples: Add bootc UKI & BLS examples

jeckersb:
Rebirth of https://github.com/bootc-dev/bootc/pull/1533 targeting
`main` after the `composefs-backend` branch was merged.


tmt/plans/bootc-integration.fmf
```diff
@@ -44,3 +44,10 @@ execute:
     how: fmf
     test:
         - /tmt/tests/test-24-local-upgrade-reboot
+
+/test-26-examples-build:
```
jeckersb: I'm not totally sold on testing this way.  This was the result of asking claude to write a test to make sure the examples build.  Is it too heavyweight to go through the matrix of spawning VMs with tmt just to sanity check the examples work?

jeckersb: (At the very least if we do it via tmt we need to add it to the list of tests in the github workflow)

jeckersb: > (At the very least if we do it via tmt we need to add it to the list of tests in the github workflow)

I added this at least for the moment just to see it run and see what happens...

cgwalters: Yes, it's messy to build containers inside a tmt test, and we definitely want to make it easy to build it outside of such a test too.

But OTOH, I think it's totally fine to do it this way as is right now. BTW though this will just conflict with https://github.com/bootc-dev/bootc/pull/1607 - what we're going to need to do is at least have a separate tmt "plan" for tests which are testing booting via bootc vs testing the build side.

jeckersb: Well first order of business...

```
./tmt-test-wrapper-inner.sh-tmt-tests-test-26-examples-build-1-default-0-default-0: line 6: cd: examples/bootc-bls: No such file or directory
```

Guess it's time to learn what path tomfoolery is going on in tmt

jeckersb: Maybe also a good point to go on a side quest and finally learn what nushell is all about...

cgwalters: I'm kind of regretting nushell a bit because IME ai models are more likely to confabulate with it. But writing nontrivial amounts of bash sucks.

One issue is TMT really pushes for having tests be interpreted code, but maybe we could mitigate that. There's obviously a ton of interpreted options, from python to ansible.

Ideally though for me, tests are like production code:

- in a static compiled language ([nushell is at least partially static](https://www.nushell.sh/book/thinking_in_nu.html#think-of-nushell-as-a-compiled-language))
- actors operate autonomously instead of having logic pushed to them

But anyways for what's going on here, bash is probably OK...

examples/bootc-bls/extra/etc/dracut.conf.d/no-xattr.conf
```diff
@@ -0,0 +1 @@
+export DRACUT_NO_XATTR=1
```
travier: We can remove those now

travier: See https://gitlab.com/fedora/bootc/base-images/-/merge_requests/261

examples/bootc-bls/extra/usr/lib/dracut/dracut.conf.d/37composefs.conf
```diff
@@ -0,0 +1,6 @@
+# we want to make sure the virtio disk drivers get included
+hostonly=no
```
travier: Can be removed as well

tmt/plans/bootc-integration.fmf
jeckersb: Yeah it was less "rewrite this tiny bash snippet in nushell" and moreso "most everything else is in nushell so I should probably at least understand how everything else works"

tmt/tests/examples/bootc-uki/install-grub.sh
```diff
@@ -0,0 +1,29 @@
...
+    --rm --privileged \
+    --pid=host \
+    -v /dev:/dev \
+    -v /var/lib/containers:/var/lib/containers \
+    -v /srv/bootc:/usr/bin/bootc:ro,Z \
+    -v /var/tmp:/var/tmp \
+    --security-opt label=type:unconfined_t \
+    "${IMAGE}" \
+    bootc install to-disk \
+        --composefs-native \
```
cgwalters, line 22: Right now bcvk doesn't allow passing arbitrary flags; we can add that of course, but I do think we should be aiming to move this to an install flag or autodetection instead.

jeckersb:
Added some commits here, tried to break it up so everything was self-explanatory in its intentions.  Tests are finally passing for me locally so I suspect this should pass and be ready for review.

jeckersb:
Hm looks like `051-test-initramfs.nu` failed?

jeckersb:
> Hm looks like `051-test-initramfs.nu` failed?

```
19:42:30                 out: TAP version 14
19:42:30                 out: initramfs
19:42:30                 out: -- No entries --
19:42:30                 out: Connection to 127.0.0.1 closed.
19:42:30         Command returned '1' (failure).
```

:thinking:

cgwalters:
This needs a rebase

cgwalters:
We had a live chat about this. So this gives us test coverage of the build+install, but not run. We should land this, but then an important next step will be teasing things out so that basically add composefs-backend as an option to the existing matrix (at least getting through `test-readonly` and ideally `test-20-local-upgrade`).

I think as part of that we'll want at least `env FEATURES=composefs just build` (or maybe we make it "sticky" with `touch target/composefs`) or so? 

Then the GHA matrix expands.

tests/build.sh
```diff
@@ -45,7 +45,7 @@ DISK=target/bootc-integration-test.qcow2
 rm -vf "${DISK}"
 # testcloud barfs on .raw
 if test -n "${bcvk}"; then
-  bcvk to-disk --format=qcow2 --disk-size "${SIZE}" localhost/bootc-integration "${DISK}"
+  bcvk to-disk --format=qcow2 --disk-size "${SIZE}" --filesystem ext4 localhost/bootc-integration "${DISK}"
```
cgwalters, line 48: OK for now but I think the composefs base image build should override the filesystem type here?

Or hmmm...actually we should fail at `bootc container lint` time if the base image specifies a filesystem which is known not to support fsverity probably and it's a sealed image.

PR #1617, opened 2025-09-17
WIP: examples: Add bootc UKI & BLS examples

travier:
Work in progress for https://github.com/bootc-dev/bootc/pull/1614

### DIFF
--- a/examples/bootc-bls/Containerfile.uki-simplified
+++ b/examples/bootc-bls/Containerfile.uki-simplified
@@ -1,0 +1,33 @@
+FROM quay.io/fedora/fedora-bootc-bls:42 AS base
+
+FROM base as kernel
+
+ARG COMPOSEFS_FSVERITY
+
+RUN --mount=type=secret,id=key \
+    --mount=type=secret,id=cert <<EOF
+
+    echo "composefs=${COMPOSEFS_FSVERITY} rw" > /etc/kernel/cmdline
+
+    dnf install -y \
+        systemd-ukify \
+        sbsigntools \
+        systemd-boot-unsigned
+
+    kver=$(cd /usr/lib/modules && echo *)
+    ukify build \
+        --linux "/usr/lib/modules/$kver/vmlinuz" \
+        --initrd "/usr/lib/modules/$kver/initramfs.img" \
+        --uname="${kver}" \
+        --cmdline "@/etc/kernel/cmdline" \
+        --os-release "@/etc/os-release" \
+        --signtool sbsign \
+        --secureboot-private-key "/run/secrets/key" \
+        --secureboot-certificate "/run/secrets/cert" \
+        --measure \
+        --json pretty \
+        --output "/boot/EFI/Linux/$kver.efi"
+EOF
+
+FROM base as final
+COPY --from=final /boot /boot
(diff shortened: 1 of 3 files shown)


examples/to-disk.sh
```diff
@@ -0,0 +1,61 @@
...
+    exit 1
+fi
+
+if [[ ! -f systemd-bootx64.efi ]]; then
+    echo "Needs /srv/bootc/examples/systemd-bootx64.efi to exists for now"
+    exit 1
+fi
+
+rm -rf ./test.img
+rm -rf ./test.img
```
jeckersb: duplicated rm here

crates/lib/Cargo.toml
```diff
@@ -74,7 +74,7 @@ similar-asserts = { workspace = true }
 static_assertions = { workspace = true }
 
 [features]
-default = ["install-to-disk"]
+default = ["install-to-disk", "composefs-backend"]
```
jeckersb, line 76: Fine for WIP (I know for example this makes rust-analyzer annoying to work with without it), but we probably don't want to enable this by default quite yet, even though adding these tests goes a long way to increasing our confidence.  Maybe let it bake in CI and get early community feedback for a release or two before we turn it on?

crates/lib/src/bootc_composefs/repo.rs
```diff
@@ -49,7 +49,10 @@ pub(crate) async fn initialize_composefs_repository(
         &Arc::new(repo),
         &format!("{transport}{image_name}"),
         None,
-        None,
+        Some(ostree_ext::containers_image_proxy::ImageProxyConfig {
```
jeckersb, line 52: Why is this required?  Looking at `SourceInfo` the docs at least claim to always use `containers-storage`.  Either I'm misunderstanding the flow here (very possible!) or there's a mismatch between the docs and what is actually happening.

crates/lib/src/install.rs
```diff
@@ -1512,6 +1521,9 @@ async fn install_to_filesystem_impl(
         }
     }
 
+    println!("state: {state:#?}");
+    println!("root_setup: {rootfs:#?}");
```
jeckersb, line 1549: Fine for WIP just wanted to note here to remove these (plus one more below) before final merge

crates/lib/Cargo.toml
travier: Yes, definitely, this is not a branch to merge but more to share progress.

travier:
Agree that we do not want to merge this PR as is right now. This is more place to share progress and the developments scripts we are currently using.

crates/lib/src/store/mod.rs
```diff
@@ -35,8 +35,10 @@ use crate::spec::ImageStatus;
 use crate::utils::deployment_fd;
 
 /// See https://github.com/containers/composefs-rs/issues/159
+// pub type ComposefsRepository =
+//    composefs::repository::Repository<composefs::fsverity::Sha512HashValue>;
 pub type ComposefsRepository =
```
cgwalters, line 40: Per https://github.com/bootc-dev/bootc/issues/1190 I think we should change all of the code to use this alias instead

crates/lib/src/store/mod.rs
```diff
@@ -35,8 +35,10 @@ use crate::spec::ImageStatus;
...
-    composefs::repository::Repository<composefs::fsverity::Sha512HashValue>;
+    composefs::repository::Repository<composefs::fsverity::Sha256HashValue>;
```
cgwalters, line 41: And consensus seems to be keeping sha512

PR #1623, opened 2025-09-18
Cleanup systemd UKI support

Johan-Liebert1:
Changes in this PR

- Adds a field `cfg_type` to BLSConfig which will contain either of the
following key:

```rust
BLSConfig {
    cfg_type: NonEFI {
        linux: String,
        initrd: Vec<String>,
        options: Some(String)
    }
    |
    EFI {
        efi: String
    }
}
```

- We did not have config files for systemd-boot and were only using UKIs
which did not allow proper sorting of the UKIs. This adds .conf files
to `$ESP/loader/entries`

- Preserves UKI addons' names so we don't overwrite previously added addon

- Adds a cli option `--uki-addon` to select the addons to install

- Add logic for upgrading/switching to a deployment with systemd-boot as the bootloader

### DIFF
--- a/crates/lib/src/bootc_composefs/boot.rs
+++ b/crates/lib/src/bootc_composefs/boot.rs
@@ -1,7 +1,7 @@
+use std::ffi::OsStr;
 use std::fs::create_dir_all;
 use std::io::Write;
 use std::path::Path;
-use std::{ffi::OsStr, path::PathBuf};
 
 use anyhow::{anyhow, Context, Result};
 use bootc_blockdev::find_parent_devices;
@@ -64,7 +64,7 @@ const SYSTEMD_LOADER_CONF_PATH: &str = "loader/loader.conf";
 /// directory specified by the BLS spec. We do this because we want systemd-boot to only look at
 /// our config files and not show the actual UKIs in the bootloader menu
 /// This is relative to the ESP
-const SYSTEMD_UKI_DIR: &str = "EFI/Linux/uki";
+const SYSTEMD_UKI_DIR: &str = "EFI/Linux/bootc";
 
 pub(crate) enum BootSetupType<'a> {
     /// For initial setup, i.e. install to-disk
@@ -460,8 +460,10 @@ pub(crate) fn setup_composefs_bls_boot(
                 .with_sort_key(default_sort_key.into())
                 .with_version(version)
                 .with_cfg(BLSConfigType::NonEFI {
-                    linux: format!("/{}/{id_hex}/vmlinuz", entry_paths.abs_entries_path),
-                    initrd: vec![format!("/{}/{id_hex}/initrd", entry_paths.abs_entries_path)],
+                    linux: format!("/{}/{id_hex}/vmlinuz", entry_paths.abs_entries_path).into(),
+                    initrd: vec![
+                        format!("/{}/{id_hex}/initrd", entry_paths.abs_entries_path).into()
+                    ],
                     options: Some(cmdline_refs),
                 });
 
@@ -474,12 +476,14 @@ pub(crate) fn setup_composefs_bls_boot(
                             ..
                         } => {
                             *linux =
-                                format!("/{}/{symlink_to}/vmlinuz", entry_paths.abs_entries_path);
+                                format!("/{}/{symlink_to}/vmlinuz", entry_paths.abs_entries_path)
+                                    .into();
 
                             *initrd = vec![format!(
                                 "/{}/{symlink_to}/initrd",
                                 entry_paths.abs_entries_path
-                            )];
+                            )
+                            .into()];
                         }
 
                         _ => unreachable!(),
@@ -549,7 +553,7 @@ pub(crate) fn setup_composefs_bls_boot(
 fn write_pe_to_esp(
     repo: &ComposefsRepository<Sha256HashValue>,
     file: &RegularFile<Sha256HashValue>,
-    file_path: &PathBuf,
+    file_path: &Utf8Path,
     pe_type: PEType,
     uki_id: &String,
     is_insecure_from_opts: bool,
@@ -600,7 +604,7 @@ fn write_pe_to_esp(
 
     let final_pe_path = match file_path.parent() {
         Some(parent) => {
-            let renamed_path = match parent.as_str()?.ends_with(EFI_ADDON_DIR_EXT) {
+            let renamed_path = match parent.as_str().ends_with(EFI_ADDON_DIR_EXT) {
                 true => {
                     let dir_name = format!("{}{}", uki_id, EFI_ADDON_DIR_EXT);
 
@@ -626,13 +630,12 @@ fn write_pe_to_esp(
         .with_context(|| format!("Opening {final_pe_path:?}"))?;
 
     let pe_name = match pe_type {
-        PEType::Uki => format!("{}{}", uki_id, EFI_EXT),
+        PEType::Uki => &format!("{}{}", uki_id, EFI_EXT),
         PEType::UkiAddon => file_path
             .components()
             .last()
-            .unwrap()
-            .to_string_lossy()
-            .to_string(),
+            .ok_or_else(|| anyhow::anyhow!("Failed to get UKI Addon file name"))?
+            .as_str(),
     };
 
     pe_dir
@@ -750,7 +753,7 @@ fn write_systemd_uki_config(
     bls_conf
         .with_title(boot_label)
         .with_cfg(BLSConfigType::EFI {
-            efi: format!("/{SYSTEMD_UKI_DIR}/{}{}", id.to_hex(), EFI_EXT),
+            efi: format!("/{SYSTEMD_UKI_DIR}/{}{}", id.to_hex(), EFI_EXT).into(),
         })
         .with_sort_key(default_sort_key.into())
         // TODO (Johan-Liebert1): Get version from UKI like we get boot label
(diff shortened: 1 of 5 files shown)


crates/lib/src/install.rs
```diff
@@ -90,7 +90,7 @@ pub(crate) const ARCH_USES_EFI: bool = cfg!(any(target_arch = "x86_64", target_a
 #[cfg(any(feature = "composefs-backend", feature = "install-to-disk"))]
 pub(crate) const ESP_GUID: &str = "C12A7328-F81F-11D2-BA4B-00A0C93EC93B";
 #[cfg(any(feature = "composefs-backend", feature = "install-to-disk"))]
-pub(crate) const DPS_UUID: &str = "6523f8ae-3eb1-4e2a-a05a-18b695ae656f";
+pub(crate) const DPS_UUID: &str = "4f68bce3-e8cd-4db1-96e7-fbcaf984b709";
```
cgwalters: This is switching to hardcode x86_64 note, what we actually need is to do this per-arch properly

crates/lib/src/bootc_composefs/status.rs
```diff
@@ -162,8 +163,20 @@ fn get_bootloader() -> Result<Bootloader> {
 
     const EFI_LOADER_INFO: &str = "LoaderInfo-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f";
 
-    match efivarfs.read_to_string(EFI_LOADER_INFO) {
-        Ok(loader) => {
+    match efivarfs.read(EFI_LOADER_INFO) {
+        Ok(loader_bytes) => {
+            if loader_bytes.len() % 2 != 0 {
```
cgwalters: This would be cleaner in a helper method like `read_uefi_variable()`

cgwalters (approved):

crates/lib/src/install.rs
Johan-Liebert1: Done in https://github.com/bootc-dev/bootc/pull/1634

crates/lib/src/bootc_composefs/status.rs
Johan-Liebert1: Done in https://github.com/bootc-dev/bootc/pull/1633

Johan-Liebert1:
This was approved, but I think it needs a re-review

Johan-Liebert1:
@cgwalters This is also up for a re-review. Added a few more commits since the last one

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -55,6 +56,16 @@ pub(crate) const EFI_UUID_FILE: &str = "efiuuid.cfg";
 /// The EFI Linux directory
 const EFI_LINUX: &str = "EFI/Linux";
 
+/// Timeout for systemd-boot bootloader menu
+const SYSTEMD_TIMEOUT: &str = "timeout 5";
```
cgwalters, line 60: This relates to https://github.com/coreos/bootupd/pull/978 - I think the default for systems-boot should actually be set at install time.

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -603,7 +627,12 @@ fn write_pe_to_esp(
 
     let pe_name = match pe_type {
         PEType::Uki => format!("{}{}", uki_id, EFI_EXT),
-        PEType::UkiAddon => format!("{}{}", uki_id, EFI_ADDON_FILE_EXT),
+        PEType::UkiAddon => file_path
+            .components()
+            .last()
+            .unwrap()
+            .to_string_lossy()
```
cgwalters: Please no `to_string_lossy` in general except for display to the user. I think in this case we should be using `Utf8Path` here and propagate that constraint upwards.

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -55,6 +56,16 @@ pub(crate) const EFI_UUID_FILE: &str = "efiuuid.cfg";
...
 
+/// Timeout for systemd-boot bootloader menu
+const SYSTEMD_TIMEOUT: &str = "timeout 5";
+const SYSTEMD_LOADER_CONF_PATH: &str = "loader/loader.conf";
+
+/// We want to be able to control the ordering of UKIs so we put them in a directory that's not the
+/// directory specified by the BLS spec. We do this because we want systemd-boot to only look at
+/// our config files and not show the actual UKIs in the bootloader menu
+/// This is relative to the ESP
+const SYSTEMD_UKI_DIR: &str = "EFI/Linux/uki";
```
cgwalters: Hmm. If it's scoped to bootc maybe it should be `/EFI/Linux/bootc` or so?

crates/lib/src/bootc_composefs/repo.rs
```diff
@@ -54,6 +54,25 @@ pub(crate) async fn initialize_composefs_repository(
     .await
 }
 
+/// skopeo (in composefs-rs) doesn't understand "registry:"
```
cgwalters, line 57: Yeah...that was something I invented in ostree-ext and probably should have argued to add to the containers-transports so that podman/skopeo/etc understand.

crates/lib/src/bootc_composefs/repo.rs
```diff
@@ -54,6 +54,25 @@ pub(crate) async fn initialize_composefs_repository(
...
+/// This function will convert it to "docker://" and return the image ref
+///
+/// Ex
+/// docker://quay.io/some-image
+/// containers-storage:some-image
+pub(crate) fn get_imgref(transport: &String, image: &String) -> String {
```
cgwalters: Basically always use `&str` not `&String` for parameters so that it works with https://lib.rs/crates/smol_str etc.

cgwalters: This one looks very unit testable

I haven't followed the flow here but ideally we do https://github.com/youki-dev/oci-spec-rs/issues/205 and then we're always using a parsed representation for these things.

crates/lib/src/bootc_composefs/repo.rs
```diff
@@ -54,6 +54,25 @@ pub(crate) async fn initialize_composefs_repository(
...
+    let img = image.strip_prefix(":").unwrap_or(&image);
+    let transport = transport.strip_suffix(":").unwrap_or(&transport);
+
+    let final_imgref = if transport == "registry" {
```
cgwalters: Nit: No need for this `let` we can just return the value of the `if`

crates/lib/src/bootc_composefs/state.rs
```diff
@@ -17,40 +18,54 @@ use rustix::{
...
-        let entry = entry?;
+    let sorted_entries = get_sorted_type1_boot_entries(boot_dir, true)?;
 
-        if !entry.file_name().as_str()?.ends_with(".conf") {
-            continue;
-        }
+    for entry in sorted_entries {
+        match &entry.cfg_type {
+            BLSConfigType::EFI { efi } => {
+                let composfs_param_value = booted.value().ok_or(anyhow::anyhow!(
```
cgwalters: Typo here `composfs` also should use `ok_or_else`

crates/lib/src/parsers/bls_config.rs
```diff
@@ -5,10 +5,29 @@
...
+use core::fmt;
 use std::collections::HashMap;
 use std::fmt::Display;
 use uapi_version::Version;
 
+#[derive(Debug, PartialEq, PartialOrd, Eq, Default)]
+pub enum BLSConfigType {
+    EFI {
+        /// The path to the EFI binary, usually a UKI
+        efi: String,
```
cgwalters: If it's a file path let's make it `Utf8PathBuf`

cgwalters (approved):
A lot here, I didn't do an exhaustive review. What would probably help here is for us to ensure we're refining something like `docs/internals/` for how we're doing some of this.

crates/lib/src/bootc_composefs/boot.rs
Johan-Liebert1: Agreed. But, for now at least we keep it in bootc until we have support for systemd-boot in bootupd

crates/lib/src/parsers/bls_config.rs
Johan-Liebert1: Done. Also updated `linux` and `initrd` paths to be `Utf8PathBuf`

crates/lib/src/bootc_composefs/boot.rs
Johan-Liebert1: This turned into a `Utf8PathBuf` so now we just have

```rust
        PEType::UkiAddon => file_path
            .components()
            .last()
            .ok_or_else(|| anyhow::anyhow!("Failed to get UKI Addon file name"))?
            .as_str(),
```

crates/lib/src/bootc_composefs/boot.rs
Johan-Liebert1: Yeah, makes sense. Updated

crates/lib/src/bootc_composefs/repo.rs
Johan-Liebert1: There are two `ImageReference` structs `ostree_ext::container::ImageReference` and `lib::spec::ImageReference` and they I think `ostree_ext::container::ImageReference` is being converted to `lib::spec::ImageReference`, but `Display` for `Transport` is just "registry"

We should probably combine these throughout the codebase?

PR #1631, opened 2025-09-19
build-sys: Various cleanups, doc clarification

cgwalters:
First change `.dockerignore` to be an allowlist.
This avoids spurious rebuilds when touching e.g. `tmt/`, and also crucially we don't leak `.git/` into the sources which can easily change as one makes commits/branches.
This also requires touching the `hack/` directory which is now fully self contained.

While we're here, make clear the roles of Justfile vs Makefile. Remove the `make test`.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,11 +1,19 @@
-# The big one - this can get HUGE and we don't want
-# to copy it around.
-target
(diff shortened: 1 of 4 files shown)


jeckersb (approved):
:+1:

PR #1632, opened 2025-09-19
 tests: add custom selinux policy test

jmarrero:
Ensure that if a custom policy is added in a
Containerfile, the resulting deployment has the expected
labels as well.

### DIFF
--- a/tmt/plans/integration.fmf
+++ b/tmt/plans/integration.fmf
@@ -60,3 +60,11 @@ execute:
     how: fmf
     test:
         - /tmt/tests/test-26-examples-build
+
+/test-27-custom-selinux-policy:
+  summary: Execute restorecon test on system with custom selinux policy
+  discover:
+    how: fmf
+    test:
+        - /tmt/tests/bootc-install-provision
+        - /tmt/tests/test-27-custom-selinux-policy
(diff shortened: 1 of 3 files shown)


tmt/tests/booted/readonly/025-test-restorecon.nu
```diff
@@ -0,0 +1,12 @@
+use std assert
+use tap.nu
+
+tap begin "Run restorecon"
+
+# Run restorecon and capture trimmed output
+let out = (restorecon -vnr /var/ /etc/ /usr/ /boot/ | str trim)
```
cgwalters: I think I'd say we should test all of these separately; we'll need to have `/boot` marked as a known issue at least right?

I am a little wary of adding `/var` into this in the general case as labels might be different for at least semi-valid reasons. `/var` is conceptually a big bag of state not owned by us.

cgwalters (commented):
Thanks, looks like a good start!

tmt/tests/booted/readonly/025-test-restorecon.nu
jmarrero: I have removed this test for now and just added the custom policy one. Will do a new PR once we merge this one and open an issue with it's output as it looks like we do have some selinux labels issues on the base images.

PR #1634, opened 2025-09-22
Bootc status fix

Johan-Liebert1:
For BLS complaint bootloaders, just systemd for now, we were still
looking for entries in `boot/loader/entries` which was failing as
entries for these bootloaders will be stored in the ESP.

Also, refactor to use `TempMount::mount_dev` for mounting ESP

### DIFF
--- a/crates/lib/src/bootc_composefs/boot.rs
+++ b/crates/lib/src/bootc_composefs/boot.rs
@@ -44,7 +44,7 @@ use crate::{
         BOOT_LOADER_ENTRIES, COMPOSEFS_CMDLINE, ORIGIN_KEY_BOOT, ORIGIN_KEY_BOOT_DIGEST,
         STAGED_BOOT_LOADER_ENTRIES, STATE_DIR_ABS, USER_CFG, USER_CFG_STAGED,
     },
-    install::{DPS_UUID, ESP_GUID, RW_KARG},
+    install::{dps_uuid::DPS_UUID, ESP_GUID, RW_KARG},
     spec::{Bootloader, Host},
 };
 
--- a/crates/lib/src/install.rs
+++ b/crates/lib/src/install.rs
@@ -89,9 +89,26 @@ pub(crate) const EFIVARFS: &str = "/sys/firmware/efi/efivars";
 pub(crate) const ARCH_USES_EFI: bool = cfg!(any(target_arch = "x86_64", target_arch = "aarch64"));
 #[cfg(any(feature = "composefs-backend", feature = "install-to-disk"))]
 pub(crate) const ESP_GUID: &str = "C12A7328-F81F-11D2-BA4B-00A0C93EC93B";
+
 #[cfg(any(feature = "composefs-backend", feature = "install-to-disk"))]
-// TODO (Johan-Liebert1): We'd want this for all archs
-pub(crate) const DPS_UUID: &str = "4f68bce3-e8cd-4db1-96e7-fbcaf984b709";
+// Architecture-specific DPS UUIDs for install-to-disk flow
+// See: https://uapi-group.org/specifications/specs/discoverable_partitions_specification/#defined-partition-type-uuids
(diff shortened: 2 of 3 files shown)


crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -352,85 +399,55 @@ pub(crate) fn setup_composefs_bls_boot(
...
+            let osrel_res = osrel_title_and_version(fs, &repo)?;
+
+            let (title, version) = match osrel_res {
+                Some((t, v)) => (
+                    t.unwrap_or(default_title_version.0),
+                    v.unwrap_or(default_title_version.1),
+                ),
+
+                None => default_title_version,
+            };
```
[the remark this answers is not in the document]
Johan-Liebert1: The control flows are distinct so this isn't an error

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -142,6 +142,10 @@ pub fn get_sysroot_parent_dev() -> Result<String> {
     return Ok(parent);
 }
 
+pub fn type1_entry_conf_file_name(sort_key: impl AsRef<str> + std::fmt::Display) -> String {
```
jeckersb: You can remove the `AsRef<str>` here; `impl std::fmt::Display` should be sufficient.

crates/lib/src/install.rs
```diff
@@ -90,7 +90,8 @@ pub(crate) const ARCH_USES_EFI: bool = cfg!(any(target_arch = "x86_64", target_a
 #[cfg(any(feature = "composefs-backend", feature = "install-to-disk"))]
 pub(crate) const ESP_GUID: &str = "C12A7328-F81F-11D2-BA4B-00A0C93EC93B";
 #[cfg(any(feature = "composefs-backend", feature = "install-to-disk"))]
-pub(crate) const DPS_UUID: &str = "6523f8ae-3eb1-4e2a-a05a-18b695ae656f";
+// TODO (Johan-Liebert1): We'd want this for all archs
```
jeckersb: Do we want to go ahead and fix this here?  Looks like there's only [5 of them](https://www.freedesktop.org/wiki/Specifications/DiscoverablePartitionsSpec/#definedpartitiontypeguids), and we can ignore IA-64; at a cursory glance rust itself [doesn't support IA-64](https://doc.rust-lang.org/stable/rustc/platform-support.html) whatsoever.

Johan-Liebert1: Yes, it's a low hanging fruit. Went ahead and added UUID all archs

crates/lib/src/bootc_composefs/boot.rs
Johan-Liebert1: Ah, right. Updated

Johan-Liebert1:
Tests for centos9 seem to be failing with 

```
cargo_build: invalid option -- '-'
error: /builddir/build/originals/bootc.spec: line 94: Unknown option - in cargo_build(naf:)
```

I can't figure out what the macro actually expands to, but the problem is probably in the following control flow

```
# Build the main bootc binary
%if 0%{?fedora} || 0%{?rhel} >= 10
    %cargo_build %{?with_rhsm:-f rhsm}
%else
    %cargo_build %{?with_rhsm:--features rhsm}
%endif
```

jeckersb:
Hm the macros must have been updated in c9s?  I'll look into that, I had to figure out that mess the first time around.

jeckersb:
> Hm the macros must have been updated in c9s? I'll look into that, I had to figure out that mess the first time around.

Yep, that's indeed the case - https://gitlab.com/redhat/centos-stream/rpms/rust/-/commit/54e4b48d95a9fe0d459f464b2a672325673254a8

I'll fix it in a separate PR and then we can just rebase this one.

jeckersb:
Actually nm, those jobs aren't required, we can just merge this and then fix the macros issue.

Issue #1635, opened 2025-09-22
update CONTRIBUTING.md for recent changes


cgwalters:
Our dev docs are a bit out of date since https://github.com/bootc-dev/bootc/issues/1473

PR #1638, opened 2025-09-22
Updates to build sys and CONTRIBUTING.md

cgwalters:
The emphasis here is on trying to have
the `Justfile` be the default entrypoint,
wrapping other tools.

- Replace mentions of podman-bootc with bcvk
  since I hope the latter supercedes the former
- Unify the unit test entrypoint
- Set up /var/tmp as a tmpdir to fix the etc merge
  test (otherwise, selinux failures w/tmp)
- Run the unit+container tests in integration.yml
- Have `just validate` run in a container

Closes: https://github.com/bootc-dev/bootc/issues/1635

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -30,6 +30,9 @@ jobs:
 
       - uses: actions/checkout@v4
 
+      - name: Free up disk space on runner
+        run: sudo ./ci/clean-gha-runner.sh
+
       - name: Set architecture variable
         id: set_arch
         run: echo "ARCH=$(arch)" >> $GITHUB_ENV
@@ -37,6 +40,10 @@ jobs:
       - name: Build container and disk image
         run: |
           sudo tests/build.sh ${{ matrix.test_os }}
+
+      - name: Run container tests
+        run:
+          sudo just test-container
 
       - name: Archive disk image
         uses: actions/upload-artifact@v4
@@ -52,10 +59,13 @@ jobs:
       matrix:
         test_os: [fedora-42, fedora-43, centos-9, centos-10]
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
 
     steps:
       - uses: actions/checkout@v4
+
+      - name: Free up disk space on runner
+        run: sudo ./ci/clean-gha-runner.sh
 
       - name: Set architecture variable
         id: set_arch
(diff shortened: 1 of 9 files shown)


.github/workflows/integration.yml
```diff
@@ -38,6 +38,10 @@ jobs:
         run: |
           sudo tests/build.sh ${{ matrix.test_os }}
 
+      - name: Run container tests
+        run:
+          sudo just test-container
```
p5, line 46: Perhaps it may be nicer to invoke `sudo` within the justfile?  

I know it goes against best practice, though if someone must remember which commands require `sudo` and which ones don't, I know I'd probably just default to sudoing more than needed.

p5 (commented):
All we need next is a bootc image with all dependencies preinstalled for devs! 😆

This is much nicer though.  Making it easier to run tests locally through a single interface is a huge improvement.  

Reminder: `just --choose` exists and is awesome - especially after documenting each entrypoint :)

.github/workflows/integration.yml
p5: Ah, so this `test-container` entrypoint is expected to be ran both rootful and rootless depending on the executing workflow.  

Yeah, there's not really anything that can be done here.

---

I no longer seem to be able to resolve my own threads, so feel free to resolve/ignore this one.

CONTRIBUTING.md
```diff
@@ -34,47 +34,68 @@ accepted!
...
 
-```
-$ podman-bootc run localhost/bootc
-```
+You can of course `podman run --rm -ti localhost/bootc bash` to get a shell,
+and try running `bootc`.
+
+### Running container-oriented integration tests
+
+`just run-container-integration`
```
p5: This command seems to have been removed from the Justfile in this PR

.github/workflows/integration.yml
cgwalters: Thanks for looking at this btw!

> Perhaps it may be nicer to invoke sudo within the justfile?

With https://github.com/bootc-dev/bcvk and `bcvk to-disk` in particular, everything fully supports rootless (and rootful).

cgwalters: (But I still need to wire up bcvk into the CI on this repo and drop all the `sudo`)

Makefile
```diff
@@ -98,23 +112,19 @@ test-bin-archive: all
 # We intentionally don't gate on this for local builds in cargo.toml
 # because it impedes iteration speed.
 CLIPPY_CONFIG = -A clippy::all -D clippy::correctness -D clippy::suspicious -D clippy::disallowed-methods -Dunused_imports -Ddead_code
-validate-rust:
```
jeckersb, line 101: I am going to type `make validate-rust` probably 50 times before this sinks in :laughing:

cgwalters: Yeah fair point, though it's probably good since I think ideally we all switch to `just validate`.

Though, a big tension for me now is using a toolbox and not devcontainer, one ends up with needing e.g. rust-analyzer on the host system, which means you're building all the Rust code there anyways...

PR #1642, opened 2025-09-24
test: Add Packit and gating test

henrywang:
Use 'system-reinstall-bootc' to re-install TF runner from package mode to image mode

### DIFF
--- a/tmt/plans/integration.fmf
+++ b/tmt/plans/integration.fmf
@@ -1,7 +1,17 @@
 provision:
   how: virtual
-  # Build via `./tests/build.sh`
   image: $@{test_disk_image}
+prepare:
+  # Replace package mode with image mode
+  - how: shell
+    script:
+      - pwd && ls -al
+      - if [[ -d hack ]]; then cd hack && ./provision-packit.sh; fi
+  # tmt-reboot and reboot do not work in this case
+  # reboot in ansible is the only way to reboot in tmt prepare
+  - how: ansible
+    playbook:
+      - https://github.com/henrywang/bootc/raw/refs/heads/gating/hack/packit-reboot.yml
 execute:
   how: tmt
 
@@ -60,11 +70,18 @@ execute:
     how: fmf
     test:
         - /tmt/tests/test-26-examples-build
+  adjust:
+    - when: running_env == packit
+      enabled: false
+      because: packit tests use RPM bootc and does not install /usr/lib/bootc/initramfs-setup
 
 /test-27-custom-selinux-policy:
   summary: Execute restorecon test on system with custom selinux policy
   discover:
     how: fmf
     test:
-        - /tmt/tests/bootc-install-provision
         - /tmt/tests/test-27-custom-selinux-policy
+  adjust:
+    - when: running_env == packit
+      enabled: false
+      because: tmt-reboot does not work with systemd reboot in testing farm environment
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -64,21 +64,23 @@ jobs:
       failure_comment:
         message: "bootc Copr build failed for {commit_sha}. @admin check logs {logs_url} and packit dashboard {packit_dashboard_url}"
 
-  # TODO: Readd some tmt tests that install the built RPM and e.g. test out system-reinstall-bootc
-  # - job: tests
-  #   trigger: pull_request
-  #   targets:
-  #     - centos-stream-9-x86_64
-  #     - centos-stream-9-aarch64
-  #     - centos-stream-10-x86_64
-  #     - centos-stream-10-aarch64
-  #     - fedora-42-x86_64
-  #     - fedora-42-aarch64
-  #     - fedora-rawhide-x86_64
-  #     - fedora-rawhide-aarch64
-  #   tmt_plan: /integration
-  #   skip_build: true
-  #   identifier: integration-test
+  - job: tests
+    trigger: pull_request
+    targets:
+      - centos-stream-9-x86_64
+      - centos-stream-9-aarch64
+      - centos-stream-10-x86_64
+      - centos-stream-10-aarch64
+      - fedora-42-x86_64
+      - fedora-42-aarch64
+      - fedora-rawhide-x86_64
+      - fedora-rawhide-aarch64
+    tmt_plan: /tmt/plans/integration
+    tf_extra_params:
+      environments:
+        - tmt:
+            context:
+              running_env: "packit"
 
   - job: propose_downstream
     trigger: release
--- a/hack/Containerfile.packit
+++ b/hack/Containerfile.packit
@@ -1,0 +1,44 @@
+# Build image for system-reinstall-bootc test
+
+# Use centos-bootc:stream10 as default
+FROM quay.io/centos-bootc/centos-bootc:stream10
+
+WORKDIR /bootc-test
+
+# Some rhts-*, rstrnt-* and tmt-* commands are in /usr/local/bin
+COPY bin /usr/local/bin
+# Save testing farm run files
+COPY ARTIFACTS /var/ARTIFACTS
+# Copy bootc repo
+COPY test-artifacts /var/share/test-artifacts
+
+RUN <<EORUN
+set -xeuo pipefail
+. /usr/lib/os-release
+if [[ $ID == "rhel" ]]; then
+    cp rhel.repo /etc/yum.repos.d/
+fi
+cp test-artifacts.repo /etc/yum.repos.d/
+dnf -y update bootc
+./provision-derived.sh
+
+# For test-22-logically-bound-install
+cp -a lbi/usr/. /usr
+for x in curl.container curl-base.image podman.image; do
+    ln -s /usr/share/containers/systemd/$x /usr/lib/bootc/bound-images.d/$x
+done
+
+# Add some testing kargs into our dev builds
+install -D -t /usr/lib/bootc/kargs.d test-kargs/*
+# Also copy in some default install configs we use for testing
+install -D -t /usr/lib/bootc/install/ install-test-configs/*
+
+# Remove bootc repo, bootc updated already
+rm -rf /var/share/test-artifacts /etc/yum.repos.d/test-artifacts.repo
+# Clean up dnf
+dnf -y clean all
+rm -rf /var/cache /var/lib/dnf
+
+# Finally, test our own linting
+# bootc container lint --fatal-warnings
+EORUN
--- a/hack/os-image-map.json
+++ b/hack/os-image-map.json
@@ -1,0 +1,9 @@
+{
+  "rhel-9.8": "images.paas.redhat.com/bootc/rhel-bootc:latest-9.8",
+  "rhel-10.2": "images.paas.redhat.com/bootc/rhel-bootc:latest-10.2",
+  "centos-9": "quay.io/centos-bootc/centos-bootc:stream9",
+  "centos-10": "quay.io/centos-bootc/centos-bootc:stream10",
+  "fedora-42": "quay.io/fedora/fedora-bootc:42",
+  "fedora-43": "quay.io/fedora/fedora-bootc:43",
+  "fedora-44": "quay.io/fedora/fedora-bootc:rawhide"
+}
--- a/hack/packit-reboot.yml
+++ b/hack/packit-reboot.yml
@@ -1,0 +1,19 @@
+---
+
+# tmt-reboot and reboot do not work in this case
+# reboot in ansible is the only way to reboot in tmt prepare
+- name: Reboot after system-reinstall-bootc
+  hosts: all
+  tasks:
+    - name: Check system mode
+      shell: bootc status --json | jq -r '.status.type'
+      register: os_mode
+
+    # null type means package mode
+    - name: Reboot system to image mode
+      reboot:
+      when: os_mode.stdout == 'null'
+
+    - name: Wait for connection to become reachable/usable
+      wait_for_connection:
+        delay: 30
--- a/hack/provision-packit.sh
+++ b/hack/provision-packit.sh
@@ -1,0 +1,110 @@
+#!/bin/bash
+set -exuo pipefail
+
+# Check environment
+printenv
+
+# This script only runs on Packit and gating environment
+# Do not run this script for image mode system
+if command -v bootc >/dev/null && bootc status --json | grep '"type":"bootcHost"'; then
+    echo "This system is Image Mode."
+    exit 0
+fi
+
+# Install required packages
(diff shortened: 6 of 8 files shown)


tmt/plans/integration.fmf
```diff
@@ -4,51 +4,66 @@ provision:
   image: $@{test_disk_image}
 execute:
   how: tmt
+  exit-first: true
 
-/readonly-tests:
+/test-00-system-reinstall-bootc:
```
cgwalters: Can't we make this a separate plan?

tmt/tests/system-reinstall/test-system-reinstall-bootc.sh
```diff
@@ -0,0 +1,140 @@
...
+    pwd
+    # change to hack directory
+    cd ../../hack
+
+    ARCH=$(uname -m)
+    # Get OS info
+    source /etc/os-release
+    case "$ID" in
+        "centos")
+            BASE="quay.io/centos-bootc/centos-bootc:stream${VERSION_ID}"
```
cgwalters: We keep adding new copies of this...I'd like to consolidate that. here's an random idea. How about we add a field `BOOTC_IMAGE` into `os-release`?

But in the intermediate term, I think we could probably have a `lib.sh` or something that's shared with `build.sh` too that we use here?

A different idea: Move all the tmt stuff into tests/ and have this "image mapping helper" live in tests?

tmt/tests/system-reinstall/test-system-reinstall-bootc.sh
```diff
@@ -0,0 +1,140 @@
...
+            if [[ "$VERSION_ID" == 44 ]]; then
+                TEST_OS="${ID}-rawhide"
+            fi
+            # Some rhts-*, rstrnt-* and tmt-* commands are in /usr/local/bin
+            cp -r /usr/local/bin .
+            ;;
+        "rhel")
+            # OSCI gating only
+            BASE="images.paas.redhat.com/bootc/rhel-bootc:latest-${VERSION_ID}"
+            # Some rhts-*, rstrnt-* and tmt-* commands are in /var/lib/tmt/scripts
```
cgwalters: Why duplicate this code in all three cases instead of having it be a second step?

tmt/tests/system-reinstall/test-system-reinstall-bootc.sh
```diff
@@ -0,0 +1,140 @@
...
+            cp -r /var/lib/tmt/scripts ./bin
+
+            CURRENT_COMPOSE_ID=$(skopeo inspect --no-tags --retry-times=5 --tls-verify=false "docker://${BASE}" | jq -r '.Labels."redhat.compose-id"')
```
cgwalters: Bigger picture I want to do https://gitlab.com/redhat/centos-stream/containers/bootc/-/issues/1174#note_2775383139

tmt/tests/system-reinstall/test-system-reinstall-bootc.sh
```diff
@@ -0,0 +1,140 @@
...
+        ls -al "$TMT_SOURCE_DIR"
+        ls -al "$TMT_SOURCE_DIR/SRPMS"
+        ls -al /etc/yum.repos.d
+        cat /etc/yum.repos.d/test-artifacts.repo
+        ls -al /var/share/test-artifacts
+        # copy bootc rpm repo into image building root
+        cp /etc/yum.repos.d/test-artifacts.repo ../../hack/
+    # Packit
+    else
+        tee "test-artifacts.repo" <<PACKITEOF
```
cgwalters: Hmmm won't tmt+testing farm inject something like this by default into `/etc/yum.repos.d`?

tmt/tests/system-reinstall/test-system-reinstall-bootc.sh
```diff
@@ -0,0 +1,140 @@
...
+enabled=1
+gpgcheck=0
+repo_gpgcheck=0
+PACKITEOF
+    fi
+
+    # Let's check things in hack folder
+    ls -al . bin
+
+    # Do not use just because it's only available on Fedora, not on CS and RHEL
```
cgwalters: True, though we can pull in EPEL easily enough?

But we don't *really* need to use the Justfile here either from my PoV.

cgwalters (commented):
Thanks for picking this up!

tmt/tests/system-reinstall/test-system-reinstall-bootc.sh
henrywang: That might not work in RHEL. I can build repo file in /etc/yum.repos.d in nightly compose rhel-bootc image.

tmt/tests/system-reinstall/test-system-reinstall-bootc.sh
henrywang: That's wired thing. The Packit, Fedora CI and  testing farm keep useing `/usr/local/bin` to save `tmt-` files, but OSCI changed to save `tmt-` files into `/var/lib/tmt/scripts`. So I have to make them separated. `rhel` means OSCI in this case.

tmt/plans/integration.fmf
henrywang: Right. It's already in a separate plan. Will be in next push.

tmt/tests/system-reinstall/test-system-reinstall-bootc.sh
henrywang: Thinking... Create a file with content like `BASE_CENOTS_10=quay.io/centos-bootc/centos-bootc:stream10` and source it?

hack/bin/aws
```diff
@@ -0,0 +1 @@
+/usr/local/aws-cli/v2/current/bin/aws
```
cgwalters: Can you explain what will use these binaries (and tomlq, and xq, and yq) ?

tmt/plans/packit.fmf
```diff
@@ -0,0 +1,64 @@
...
+execute:
+  how: tmt
+  exit-first: true
+
+/test-01-readonly:
+  summary: Execute booted readonly/nondestructive tests
+  discover+:
+    how: fmf
+    test:
+        - /tmt/tests/test-00-system-reinstall-bootc
```
cgwalters: Can't we do this step not as a test but as a "provision phase" step?

tmt/plans/packit.fmf
```diff
@@ -0,0 +1,64 @@
+discover:
+  how: fmf
+provision:
+  how: virtual
+  image: $@{test_disk_image}
```
cgwalters: This would be unused here right?

tmt/plans/packit.fmf
```diff
@@ -0,0 +1,64 @@
...
+  - how: install
+    package:
+      - podman
+      - skopeo
+      - jq
+execute:
+  how: tmt
+  exit-first: true
+
+/test-01-readonly:
```
cgwalters: We had an existing problem where "add a new test" requires a lot of ceremony (write the code as a file, write the fmf file, then add it to the plan) - it'd be really great to fix this (can't we use the tmt discover stuff).

But this takes that problem and makes it worse as now all the tests are listed *twice* more in this file.

tmt/tests/system-reinstall/test-system-reinstall-bootc.sh
```diff
@@ -0,0 +1,130 @@
...
+ls -al /etc/yum.repos.d
+
+# This script only runs on Packit and gating environment
+if [ "$TMT_REBOOT_COUNT" -eq 0 ]; then
+    # temp folder to save building files and folders
+    BOOTC_TEMPDIR=$(mktemp -d)
+    trap 'rm -rf -- "$BOOTC_TEMPDIR"' EXIT
+
+    # Copy files and folders in hack to TEMPDIR
+    cp -a ../../hack/* "$BOOTC_TEMPDIR"
```
cgwalters: I think it's a lot cleaner to do `cd $BOOTC_TMPDIR` and then everything can just reference `.` below

tmt/tests/system-reinstall/test-system-reinstall-bootc.sh
```diff
@@ -0,0 +1,130 @@
...
+
+    # Copy bootc repo
+    cp -r /var/share/test-artifacts "$BOOTC_TEMPDIR"
+
+    ARCH=$(uname -m)
+    # Get OS info
+    source /etc/os-release
+
+    # Some rhts-*, rstrnt-* and tmt-* commands are in /usr/local/bin
+    if [[ $ID == rhel ]]; then
```
cgwalters: How about just `if test -d /var/lib/tmt/scripts` instead? Then we're not detecting the OS, and if how we do RHEL/testing farm changes it just transparently starts working

tmt/tests/system-reinstall/test-system-reinstall-bootc.sh
```diff
@@ -0,0 +1,130 @@
...
+    case "$ID" in
+        "centos")
+            BASE="quay.io/centos-bootc/centos-bootc:stream${VERSION_ID}"
+            TEST_OS="${ID}-stream-${VERSION_ID}"
+            ;;
+        "fedora")
+            BASE="quay.io/fedora/fedora-bootc:${VERSION_ID}"
+            TEST_OS="${ID}-${VERSION_ID}"
+            if [[ "$VERSION_ID" == 44 ]]; then
+                TEST_OS="${ID}-rawhide"
```
cgwalters: This one is obviously going to be a maintenance trap over time...how about at least having `FEDORA_RAWHIDE_VERSION=44` at the top?

But if we can avoid this mapping that'd be even better

tmt/tests/system-reinstall/test-system-reinstall-bootc.sh
```diff
@@ -0,0 +1,130 @@
...
+    podman build --jobs=4 --from "$BASE" -v "$BOOTC_TEMPDIR":/bootc-test:z -t localhost/bootc-integration -f "${BOOTC_TEMPDIR}/Containerfile.packit" "$BOOTC_TEMPDIR"
+
+    # Keep these in sync with what's used in hack/lbi
+    podman pull -q --retry 5 --retry-delay 5s quay.io/curl/curl:latest quay.io/curl/curl-base:latest registry.access.redhat.com/ubi9/podman:latest
+
+    # Install system-reinstall-bootc
+    dnf install -y system-reinstall-bootc expect
+
+    # Run system-reinstall-bootc
+    system-reinstall/system-reinstall-bootc.exp
```
cgwalters: Yeah we definitely want system-reinstall-bootc to be more scriptable, let's track that as an issue or even just TODO in this file

tmt/tests/system-reinstall/test-system-reinstall-bootc.sh
```diff
@@ -0,0 +1,130 @@
...
+    # Some simple and fast checkings
+    bootc status
+    echo "$PATH"
+    printenv
+    ls -al /var/ARTIFACTS
+    ls -al /usr/local/bin
+    echo "Bootc system on TMT/TF runner"
+
+    exit 0
+fi
```
cgwalters, line 90: Should fatally error out if `TMT_REBOOT_COUNT` is not 0 or 1? e.g. use `case` to match it?

tmt/tests/system-reinstall/test-system-reinstall-bootc.sh
henrywang: The tmt test running folder is at tmt test fmf file folder. The folder id not identical every tmt/testing farm run. So I can't leave this folder before copy all files into `$BOOTC_TMPDIR`.

tmt/plans/packit.fmf
henrywang: I use this to debug locally with `tmt --context "test_disk_image=centos-stream-10" run --all -e PACKIT_COPR_PROJECT=packit/bootc-dev-bootc-1642 plans -n /tmt/plans/packit/test-01-readonly`.

tmt/tests/system-reinstall/test-system-reinstall-bootc.sh
henrywang: I can't find `just` in EPEL9, but it's available in EPEL10.

cgwalters: Let's make it a general principle that `just` is only used as a top-level entrypoint for two cases:

- human user operating on a git repo
- toplevel CI flow like GHA

This case I think it's fine to skip `just` and directly run a `podman build` type flow.

tmt/plans/packit.fmf
henrywang: I'm trying to run this in tmt `prepare` https://github.com/bootc-dev/bootc/pull/1642/files#diff-bbbfb45ca8f2b20a0985a7efd1f2d17bda0102fb40dcee7f43b6db2ef32a4433R16, and it'll not be a tmt `test` any more. And it's possible to share same tmt plan (only `package mode` VM, Packit and gating in this case, run this provision script, `github and local` run is running on image mode VM) to avoid duplicate plan issue.

tmt/plans/packit.fmf
henrywang: Same as above comment.

tmt/tests/system-reinstall/test-system-reinstall-bootc.sh
henrywang: @cgwalters , What about this https://github.com/bootc-dev/bootc/pull/1642/files#diff-85c232ec82f12105613d6c349d87940c2b6dc21e6d77ad85964adc1c4a7b9183R38?

tmt/plans/packit.fmf
henrywang: Yes! That works!

henrywang:
CS9 failure: issue https://github.com/bootc-dev/bootc/issues/1659. Can we skip CS9 packit test until issue fixed?

tmt/tests/system-reinstall/test-system-reinstall-bootc.sh
cgwalters: I like the JSON a lot better than doing it in shell for sure

cgwalters (approved):

cgwalters:
Thanks! So now we're running all the tests in two different ways, via a pregenerated disk image and via s-r-b.

I'm not sure we need to do that on *every* PR, but we can continue to fine tune this. Offhand for example I'd probably say that we could do the full matrix as an opt-in via label again or pre-release? Anyways not a blocker for now so let's keep moving forward!

> CS9 failure: issue https://github.com/bootc-dev/bootc/issues/1659. Can we skip CS9 packit test until issue fixed?

Sure, though they aren't gating right now, and hopefully the fix will be pretty easy, and then we can prove the PR to fix it works.

PR #1643, opened 2025-09-24
composefs-backend: Implement `bootc usr-overlay`

Johan-Liebert1:
Similar to ostree, mount a transient overlayfs on /usr

### DIFF
--- a/crates/lib/src/bootc_composefs/state.rs
+++ b/crates/lib/src/bootc_composefs/state.rs
@@ -2,10 +2,13 @@ use std::os::unix::fs::symlink;
 use std::{fs::create_dir_all, process::Command};
 
 use anyhow::{Context, Result};
+use bootc_initramfs_setup::overlay_transient;
 use bootc_kernel_cmdline::utf8::Cmdline;
 use bootc_mount::tempmount::TempMount;
 use bootc_utils::CommandRunExt;
 use camino::Utf8PathBuf;
+use cap_std_ext::cap_std::ambient_authority;
+use cap_std_ext::cap_std::fs::Dir;
 use cap_std_ext::{cap_std, dirext::CapStdExtDirExt};
 use composefs::fsverity::{FsVerityHashValue, Sha256HashValue};
 use fn_error_context::context;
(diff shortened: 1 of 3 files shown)


crates/lib/src/bootc_composefs/state.rs
```diff
@@ -163,3 +167,29 @@ pub(crate) fn write_composefs_state(
...
     Ok(())
 }
+
+pub(crate) fn composefs_usr_overlay() -> Result<()> {
+    let findmnt_rs = run_findmnt(&[], None)?;
+    let root_mnt = findmnt_rs
+        .filesystems
+        .iter()
+        .next()
+        .ok_or(anyhow::anyhow!("Failed to get root mount"))?;
```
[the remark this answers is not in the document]
Johan-Liebert1: finmnt with no args does always have root as the first element in the array. I don't think the output is going to change from under us

crates/lib/src/bootc_composefs/state.rs
```diff
@@ -163,3 +167,29 @@ pub(crate) fn write_composefs_state(
...
+
+    let usr_mnt = mnt_point_details("/usr", root_mnt);
+
+    if let Some(usr_mnt) = usr_mnt {
```
cgwalters: I think we can replace all of this with basically https://docs.rs/cap-std-ext/latest/cap_std_ext/dirext/trait.CapStdExtDirExt.html#tymethod.is_mountpoint

crates/lib/src/bootc_composefs/state.rs
```diff
@@ -163,3 +167,29 @@ pub(crate) fn write_composefs_state(
...
+        if usr_mnt.fstype == "overlay" {
```
cgwalters: I think we can just drop the checking for overlay, assume that if it's a mountpoint then we made the overlay.

That said, in the end what we really want here is metadata on mountpoints.

cgwalters (changes requested):
Thanks for doing this!

crates/lib/src/bootc_composefs/state.rs
Johan-Liebert1: > I think we can just drop the checking for overlay

Yeah, I think that's okay

> That said, in the end what we really want here is metadata on mountpoints

We do get metadata using `mnt_point_details`, which I didn't remove from this PR even though it's not used yet

crates/lib/src/bootc_composefs/state.rs
Johan-Liebert1: Updated to use `is_mountpoint`. The only thing is it uses `statx` which seems to be linux specific, so right now we fail if `statx` isn't a syscall

crates/lib/src/bootc_composefs/state.rs
```diff
@@ -163,3 +166,30 @@ pub(crate) fn write_composefs_state(
...
+        .context("Failed to get mount details for /usr")?;
+
+    match is_usr_mounted {
+        Some(x) if x == true => {
+            println!("A writeable overlayfs is already mounted on /usr");
+            return Ok(());
+        }
+
+        None => {
+            anyhow::bail!("Falied to get mountinfo");
```
cgwalters: Typo here, also I think it'd be cleaner to do it above like:
```let is_usr_mounted = is_mountpoint().ok_or_else(||)```

File: crates/mount/src/mount.rs [hunk 5]
```diff
@@ -115,6 +115,27 @@ pub fn is_source_mounted(path: &str, mounted_fs: &Filesystem) -> bool {
     false
 }
 
+// Recursively check a given filesystem to see if it contains an already mounted target
+// Returns the mount details
+pub fn mnt_point_details<'a>(
```
cgwalters: We don't need this anymore right?

Johan-Liebert1: Yeah, I'd kept it in case we wanted to get mount info in the future. Guess we can always get it back; removed it

Issue #806, opened 2024-09-27
UKI/systemd-boot tracker


cgwalters:
systemd-boot has a lot of uptake and is very simple for the UEFI path. We need to support it.

One thing this deeply intersects with is https://github.com/ostreedev/ostree/issues/2753 and https://github.com/ostreedev/ostree/issues/1951 as well as https://github.com/containers/bootc/issues/20

I think with the new https://github.com/containers/composefs/issues/332 we could try doing a "big bang" where we:

- Teach bootloader.rs how to detect and install systemd-boot (trivial)
- Have a mode where we basically drop all the ostree stuff out of the container image...we make a merged composefs client side from the container image content and for good measure just do all the selinux labeling *client* side to start (ref https://github.com/ostreedev/ostree-rs-ext/issues/510 - though we could also start honoring `security.selinux` in the tar stream)
- Deploying an image also copies the UKI out into the ESP (and we use the UKIs as a garbage collection root for the objects)
- We could hence then potentially start by dropping `ostree-prepare-root.service` from the initramfs and have a spike on what it'd look like to move the mount logic in the initramfs maybe in this project.

(A challenge in this is we document configuring things in `ostree-prepare-root.conf` today but we could start parsing those options in bootc instead)

cgwalters:
One thing that came up is it could be really helpful for us to support the "pure UKI" path first with no bootloader. That's likely a good way to try severing some of the ostree dependency in that path - we'd just need to handle UKI copying and garbage collection on the bootc side alone.

cgwalters:
cc @allisonkarlitskaya

prydom:
A note for when you do start integrating with systemd-boot is that if you want the ostree slot number to appear in the systemd-boot menu (and the order of the menu to be correct) you need to embed a custom os-release file into the .osrel section of the binary. In particular you need to set or add a suffix to `PRETTY_NAME=`. You also need to determine how discovery of the correct `ostree` deployment to boot during initramfs would be done, which is covered in some of the issues linked in this issue's description.

[...]

PR #1644, opened 2025-09-24
New changes / features for composefs-backend

Johan-Liebert1:
Finish up implementation for `install-to-filesystem` for `composefs-backend`

We were using composefs-native and composefs-backend interchangeably.
Replace all instances of `composefs-native` with `composefs-backend`

Move all composefs-backend options to a single struct so that we can
test for boolean instead of testing for Some/None for composefs-backend
options

The source imgref was being stored in the .origin file, even if target imgref was 
provided. Now we store the target imgref

Start finalize-staged service on update/switch

Add composefs opts to install-to-existing-root

### DIFF
--- a/crates/lib/src/install.rs
+++ b/crates/lib/src/install.rs
@@ -1322,7 +1322,10 @@ async fn prepare_install(
     } else {
         false
     };
+
     tracing::debug!("Composefs required: {composefs_required}");
+
+    #[cfg(feature = "composefs-backend")]
     if composefs_required {
         composefs_options.composefs_backend = true;
     }
--- a/crates/lib/src/bootloader.rs
+++ b/crates/lib/src/bootloader.rs
@@ -7,6 +7,8 @@ use fn_error_context::context;
 
 use bootc_blockdev::{Partition, PartitionTable};
 use bootc_mount as mount;
+
+#[cfg(any(feature = "composefs-backend", feature = "install-to-disk"))]
 use bootc_mount::tempmount::TempMount;
 
 use crate::utils;


Johan-Liebert1:
Currently with the option `--bootloader=systemd` we do everything according to the BLS spec, but the bootloader is still grub, as we don't have systemd-boot support in bootupd. Until https://github.com/coreos/bootupd/pull/978 lands, I was thinking of a very crude hack of simply replacing `grubx64.efi` with `systemd-boot.efi` (renamed to `grubx64.efi` so shim picks it up); this is basically how I've been testing `systemd-boot` until now.

Does it make sense to do the above until we have systemd-boot support in bootupd?

cgwalters:
Yes I'll try to take a look at https://github.com/coreos/bootupd/pull/978 soon.

Otherwise see https://github.com/bootc-dev/bootc/issues/806#issuecomment-3145079372 - I think the "no bootupd = systemd-boot" path also makes sense and should be relatively straightforward to implement here.

cgwalters:
Ah my bad, I think we have some textual (partial semantic) conflicts in https://github.com/bootc-dev/bootc/pull/1662

crates/lib/src/composefs_consts.rs
```diff
@@ -36,3 +36,5 @@ pub(crate) const USER_CFG_STAGED: &str = "user.cfg.staged";
 /// This is relative to the boot/efi directory
 pub(crate) const TYPE1_ENT_PATH: &str = "loader/entries";
 pub(crate) const TYPE1_ENT_PATH_STAGED: &str = "loader/entries.staged";
+
+pub(crate) const COMPOSEFS_FINALIZE_STAGED_SERVICE: &str = "composefs-finalize-staged.service";
```
cgwalters: Unrelated to this PR but I think we should call it `bootc-finalize-staged.service`

crates/lib/src/install.rs
```diff
@@ -249,10 +249,21 @@ pub(crate) struct InstallConfigOpts {
...
+    /// Make fs-verity validation optional in case the filesystem doesn't support it
     #[clap(long, default_value_t)]
     #[serde(default)]
     pub(crate) insecure: bool,
 
+    /// The bootloader to use.
+    /// Currently supported:
+    ///
+    /// - systemd-boot
+    /// - grub
```
cgwalters: Note that clap should emit this automatically

crates/lib/src/install.rs
```diff
@@ -372,6 +378,10 @@ pub(crate) struct InstallToFilesystemOpts {
 
     #[clap(flatten)]
     pub(crate) config_opts: InstallConfigOpts,
+
+    #[cfg(feature = "composefs-backend")]
+    #[clap(flatten)]
+    pub(crate) compoesfs_opts: InstallComposefsOpts,
```
cgwalters: Typo

Johan-Liebert1:
That's okay. I can rebase on main afterwards

crates/lib/src/install.rs
Johan-Liebert1: What's your opinion on adding something like [typos](https://github.com/crate-ci/typos) to the CI? There are quite a few typos across the repo

crates/lib/src/composefs_consts.rs
Johan-Liebert1: Updated

crates/lib/src/install.rs
Johan-Liebert1: Updated

cgwalters:
I rebased this one

cgwalters:
OK, the new composefs test fails on this one; it's possible I broke something when rebasing, or it's a legit regression.

Johan-Liebert1:
So, `cd crates/lib && cargo check --no-default-features` failed. Running it on my local branch succeeds

PR #1646, opened 2025-09-24
ci: Consolidate and cleanup workflows

cgwalters:
- Having separate workflows made sharing build artifacts not really possible, and it just makes sense to consolidate
- Refactor shared parts into a little internal action


cgwalters:
To merge this we'll need to force it to override the expected CI contexts.

.github/workflows/ci.yml
henrywang, line 1: Hi @cgwalters, Do we still need to run all those test when `push` to `main`?

cgwalters: Probably not? I am not aware of anyone monitoring the actions that run post-merge (though of course we could start doing so).

What it would catch is semantic merge conflicts, but the more I think about it the more I feel we should go to [merge queues](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue).

henrywang: Yeah, we can move integration test into `merge_queue` trigger and keep the rest test in `pull_request` trigger?

cgwalters: Do you think that should block this one?

henrywang: We do not need to run same test twice (`pull_request` and `merge_queue`). 
We can land this PR first and make changes when we enable `merge_queue`.

cgwalters: OK can you approve the PR please?

cgwalters:
I've adjusted the required CI context names to match this PR.

cgwalters:
Note: This means all outstanding PRs will need to be rebased

PR #1660, opened 2025-10-02
Update ostree binding to 0.20.5

alexlarsson:
This will be needed by https://github.com/coreos/rpm-ostree/pull/5497 as it picks up ostree-rs via ostree-ext.


alexlarsson:
I tried to do that, but it failed because only one cargo module can link to the ostree library.

PR #1664, opened 2025-10-03
test: Fix packit-reboot.yml download URL issue

henrywang:
PR #1642 merged, so the `packit-reboot.yml` ansible playbook download URL should be moved to `bootc-dev/bootc` now.

### DIFF
--- a/tmt/plans/integration.fmf
+++ b/tmt/plans/integration.fmf
@@ -11,7 +11,7 @@ prepare:
   # reboot in ansible is the only way to reboot in tmt prepare
   - how: ansible
     playbook:
-      - https://github.com/henrywang/bootc/raw/refs/heads/gating/hack/packit-reboot.yml
+      - https://github.com/bootc-dev/bootc/raw/refs/heads/main/hack/packit-reboot.yml
 execute:
   how: tmt
 


tmt/plans/integration.fmf
```diff
@@ -11,7 +11,7 @@ prepare:
   # reboot in ansible is the only way to reboot in tmt prepare
   - how: ansible
     playbook:
-      - https://github.com/henrywang/bootc/raw/refs/heads/gating/hack/packit-reboot.yml
+      - https://github.com/bootc-dev/bootc/raw/refs/heads/main/hack/packit-reboot.yml
```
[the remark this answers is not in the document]
henrywang: This is a really simple playbook for reboot, no additional code included. So no need to pin to `commit`.

cgwalters, line 14: Wait but can't we just specify running the playbook from the local git checkout?

henrywang: Two reasons:

1. According to https://tmt.readthedocs.io/en/stable/plugins/prepare.html#ansible, I can't trace the git checkout folder. Very sad thing for me.
2. The Fedora CI and OSCI gating does not have local git. I need extract `.src.rpm`. And it does not work with `ansible` prepare.

henrywang: And I'm working on another PR to use `.src.rpm` instead of local git checkout. Just for working with dist-git gating test.

cgwalters: >  I can't trace the git checkout folder. Very sad thing for me.

What do you mean "trace"? Oh...are you getting hit by https://github.com/bootc-dev/bootc/blob/3fd43a747f148af0f55a048a9081faef433dbf6d/tests/run-tmt.sh#L17 ?
We can add `hack/` there too.

> The Fedora CI and OSCI gating does not have local git. I need extract .src.rpm. And it does not work with ansible prepare.

Hmm. I often go back to referencing https://cockpit-project.org/blog/fmf-unified-testing.html

Though it looks like things have changed there for cockpit nowadays....

I also looked at https://src.fedoraproject.org/rpms/dnf5/blob/rawhide/f/.packit.yaml#_85 which is interesting.

Anyways I think the idea here is we should definitely be able to store this in a git repo directly.

henrywang: The dnf5 fedora dist-git does not have gating test for release PR, like https://src.fedoraproject.org/rpms/dnf5/pull-request/102.

PR #1665, opened 2025-10-03
build-sys: Use stream10 by default (+ 2 misc)

cgwalters:


cgwalters:
> Error: configure storage: 'overlay' is not supported over overlayfs, a mount_program is required: backing file system is unsupported for this graph driver

~Looking at this~ :arrow_right: https://github.com/bootc-dev/bootc/pull/1666

PR #1666, opened 2025-10-03
install: Add a tmpfs for /var/lib/containers

cgwalters:
Recent podman versions got stricter about not doing overlay-on-overlay by default.


jmarrero (approved):
/lgtm

PR #1670, opened 2025-10-04
test: Use SRPM as test code source to work with gating test

henrywang:
In Fedora or OSCI dist-git gating, SRPM is the only way to get test code. Change to extracting SRPM to get source code. That works with both Packit and dist-git gating.

This PR also includes a `when: running_env != unify` in prepare to make it only runs on Packit or dist-git gating test. In our case, `unify` means running on Github CI or locally.

### DIFF
--- a/hack/Containerfile.packit
+++ b/hack/Containerfile.packit
@@ -12,6 +12,7 @@ COPY ARTIFACTS /var/ARTIFACTS
 # Copy bootc repo
 COPY test-artifacts /var/share/test-artifacts
 
+ARG GATING
 RUN <<EORUN
 set -xeuo pipefail
 . /usr/lib/os-release
@@ -21,6 +22,12 @@ fi
 cp test-artifacts.repo /etc/yum.repos.d/
 dnf -y update bootc
 ./provision-derived.sh
+
+if [[ "$GATING" == true ]]; then
+    # Install Fedora CI/OSCI required packages for "Prepare dist-git sources" task
+    dnf install -y  rpm-build @buildsys-build 'dnf-command(builddep)'
+    dnf builddep -y /var/share/test-artifacts/*.src.rpm
+fi
 
 # For test-22-logically-bound-install
 cp -a lbi/usr/. /usr
--- a/tmt/plans/integration.fmf
+++ b/tmt/plans/integration.fmf
@@ -2,16 +2,32 @@ provision:
   how: virtual
   image: $@{test_disk_image}
 prepare:
-  # Replace package mode with image mode
+  # Install image mode system on package mode system
+  # Do not run on image mode VM running on Github CI and Locally
+  # Run on package mode VM running on Packit and Gating
+  - how: install
+    package:
+      - podman
+      - skopeo
+      - jq
+      - bootc
+      - system-reinstall-bootc
+      - expect
+      - ansible-core
+      - zstd
+    when: running_env != image_mode
   - how: shell
     script:
-      - pwd && ls -al
-      - if [[ -d hack ]]; then cd hack && ./provision-packit.sh; fi
+      - mkdir -p bootc && cp /var/share/test-artifacts/*.src.rpm bootc
+      - cd bootc && rpm2cpio *.src.rpm | cpio -idmv && rm -f *-vendor.tar.zstd && zstd -d *.tar.zstd && tar -xvf *.tar -C . --strip-components=1 && ls -al
+      - pwd && ls -al && cd bootc/hack && ./provision-packit.sh
+    when: running_env != image_mode
   # tmt-reboot and reboot do not work in this case
   # reboot in ansible is the only way to reboot in tmt prepare
   - how: ansible
     playbook:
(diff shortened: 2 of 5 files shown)


henrywang:
`ansible` in `prepare` summary:

`TMT_TREE=/var/ARTIFACTS/work-readonly-testso2z42hbo/tmt/plans/integration/readonly-tests/tree`

`prepare` ansible playbook search path:

1. `bootc/hack/packit-reboot.yml`: `fail: Playbook '/var/ARTIFACTS/git-18806550f6eec4bdd5b829fbdac8e3a17cc25194x66xpqn2/bootc/hack/packit-reboot.yml' does not exist.`
2. `file:///bootc/hack/packit-reboot.yml`: `fail: Playbook '/var/ARTIFACTS/git-18806550f6eec4bdd5b829fbdac8e3a17cc25194g61y9ec8/bootc/hack/packit-reboot.yml' does not exist.`
3. `${TMT_TREE}/bootc/hack/packit-reboot.yml`: `fail: Playbook '/var/ARTIFACTS/git-18806550f6eec4bdd5b829fbdac8e3a17cc25194wiir1lrw/var/artifacts/work-readonly-testsilntmhhw/tmt/plans/integration/readonly-tests/tree/bootc/hack/packit-reboot.yml' does not exist.`
4. `file://../../../${TMT_TREE}/bootc/hack/packit-reboot.yml`: `fail: '/var/ARTIFACTS/git-18806550f6eec4bdd5b829fbdac8e3a17cc251949fqj3f3q/../../../var/artifacts/work-readonly-tests4que9qig/tmt/plans/integration/readonly-tests/tree/bootc/hack/packit-reboot.yml' is not relative to the expected root '/var/ARTIFACTS/git-18806550f6eec4bdd5b829fbdac8e3a17cc251949fqj3f3q'`

Okey, I tried all different configurations, no one works in this case. I still prefer `https://github.com/bootc-dev/bootc/raw/refs/heads/main/hack/packit-reboot.yml` because it's clear and straightforward.

henrywang:
CS9 failure needs fix https://github.com/bootc-dev/bootc/pull/1663.

hack/Containerfile.packit
```diff
@@ -22,6 +23,12 @@ cp test-artifacts.repo /etc/yum.repos.d/
 dnf -y update bootc
 ./provision-derived.sh
 
+if [[ "$GATING" == true ]]; then
+    # Install Fedora CI/OSCI required packages for "Prepare dist-git sources" task
```
cgwalters, line 27: To be clear, this is to get the tests? Ah I see, we're shipping them in the `.tar.zstd` today. But in theory we shouldn't need the full build dependencies just to get those...

Is extracting the .src.rpm really what others do here? I guess we could make a `-tests` subpackage?

tmt/plans/integration.fmf
```diff
@@ -3,15 +3,31 @@ provision:
   image: $@{test_disk_image}
 prepare:
   # Replace package mode with image mode
+  # DO NOT RUN ON Github or Locally
```
cgwalters: It's a bit unclear what "DO NOT RUN ON Github" means. The term "unify" also isn't obvious to me.

How about calling this "init_package_mode" or something - making clear that in this flow we're starting from package mode.

tmt/plans/integration.fmf
```diff
@@ -3,15 +3,31 @@ provision:
...
+      - expect
+      - ansible-core
+      - zstd
+    when: running_env != unify
   - how: shell
     script:
-      - pwd && ls -al
-      - if [[ -d hack ]]; then cd hack && ./provision-packit.sh; fi
+      - mkdir -p bootc && cp /var/share/test-artifacts/*.src.rpm bootc
+      - cd bootc && rpm2cpio *.src.rpm | cpio -idmv && rm -f *-vendor.tar.zstd && zstd -d *.tar.zstd && tar -xvf *.tar -C . --strip-components=1 && ls -al
```
cgwalters, line 22: Ah. But if we do this, do we need the `dnf builddep` per above at all? I don't think so...

hack/Containerfile.packit
henrywang: Because in Fedora/OSCI gating, two tasks will be running after our `prepare`:
```
queued prepare task #1: essential-requires on default-0
queued prepare task #2: default-0 on default-0 --> our install prepare
queued prepare task #3: default-1 on default-0 --> our provision prepare
queued prepare task #4: Prepare dist-git sources (buildrequires, patches, discovery...) on default-0 --> install packages with `rpm-build @buildsys-build 'dnf-command(builddep)`
queued prepare task #5: requires (dist-git) on default-0 --> install builddeps packages
queued prepare task #6: recommends (dist-git) on default-0
```
Because those package installations prepares are under our provision prepare, so we need pre-build those packages, which will be used by task 3 and task 4, into our image.

tmt/plans/integration.fmf
henrywang: Right. We don't need buiddep packages, but FedoraCI/OSCI needs those packages.

PR #1674, opened 2025-10-07
Makefile: targeted workaround for s390x builds on c10s

jeckersb:

### DIFF
--- a/contrib/packaging/bootc.spec
+++ b/contrib/packaging/bootc.spec
@@ -95,6 +95,13 @@ cat vendor-config.toml >> .cargo/config.toml
 rm vendor-config.toml
 
 %build
+# https://issues.redhat.com/browse/RHEL-116881
+%if 0%{?rhel} >= 10
+if test $(arch) = s390x; then
+  export MALLOC_MMAP_MAX_=0
+fi
+%endif
+
 # Build the main bootc binary
 %if %new_cargo_macros
     %cargo_build %{?with_rhsm:-f rhsm}


jeckersb:
Actually I don't think this is going to help things in CI because IIUC packit is using the spec under contrib/packaging and that calls cargo directly instead of building via the Makefile.

Makefile
```diff
@@ -27,10 +27,18 @@ TAR_REPRODUCIBLE = tar --mtime="@${SOURCE_DATE_EPOCH}" --sort=name --owner=0 --g
 # (Note we should also make installation of the units conditional on the rhsm feature)
 CARGO_FEATURES ?= $(shell . /usr/lib/os-release; if echo "$$ID_LIKE" |grep -qF rhel; then echo rhsm; fi)
 
+# https://issues.redhat.com/browse/RHEL-116881
+ARCH ?= $(shell uname -m)
+ID ?= $(shell [ -f /etc/os-release ] && . /etc/os-release && echo $$ID)
+VERSION_ID ?= $(shell [ -f /etc/os-release ] && . /etc/os-release && echo $$VERSION_ID)
```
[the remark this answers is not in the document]
jeckersb: silly bot you just told me to do it the other way, which i prefer despite the repeating since it's explicit

contrib/packaging/bootc.spec
```diff
@@ -97,9 +104,9 @@ rm vendor-config.toml
 %build
 # Build the main bootc binary
 %if %new_cargo_macros
-    %cargo_build %{?with_rhsm:-f rhsm}
+    %bootc_cargo_build %{?with_rhsm:-f rhsm}
```
cgwalters: It seems somehow simpler to do this in shell script just above like
```
if test $(arch) = s390x; then
  export MALLOC_MMAP_MAX_=0
fi
```
or so?

Us having our own cargo macros feels like overkill.

cgwalters:
That said though, since other things are broken by this today, if that's a valid workaround, why not just ask the toolchain maintainers to ship it by default?

contrib/packaging/bootc.spec
jeckersb: Yeah that's better.  I really hate bash scripting, and I really hate rpm macros, and any time I have to combine the two (and/or Makefiles...) it usually ends poorly :laughing:

jeckersb:
> That said though, since other things are broken by this today, if that's a valid workaround, why not just ask the toolchain maintainers to ship it by default?

From the Jira it seems like the confidence is not particularly high on the workaround, but if we could provide early feedback on whether it helps in our case it would be quite useful to them.

jeckersb:
I'm going to let this round of copr builds run before I make any more changes just to see how things turn out.

contrib/packaging/bootc.spec
```diff
@@ -95,6 +95,13 @@ cat vendor-config.toml >> .cargo/config.toml
 rm vendor-config.toml
 
 %build
+# https://issues.redhat.com/browse/RHEL-116881
+%if 0%{?rhel} >= 10
```
cgwalters, line 99: Wasn't this happening on c9s too? I suspect it has to do with the rust version, not the host OS version.

It seems simpler to just unconditionally do it on s390x anyways.

cgwalters (approved):

contrib/packaging/bootc.spec
jeckersb: @henrywang mentioned it happening on c9s but I haven't seen it happen yet in our CI, I figured I would be more conservative at first and if we start to see it showing up on 9 we can add it.

If it's not broken on Fedora for example I'd rather not enable the workaround there.

henrywang: OH, sorry, my bad @jeckersb. It's CS10, not CS9. CS9 is empty my brain. Sorry about that!

jeckersb: Less broken things = better!

jeckersb:
Hm looks like the change worked, but it still segfaulted even with the environment variable added

https://download.copr.fedorainfracloud.org/results/packit/bootc-dev-bootc-1674/centos-stream-10-s390x/09666012-bootc/builder-live.log.gz

Relevant (trimmed) bits:

```
Executing(%build): /bin/sh -e /var/tmp/rpm-tmp.75FzzG
...
++ arch
+ test s390x = s390x
+ export MALLOC_MMAP_MAX_=0
+ MALLOC_MMAP_MAX_=0
...
+ make manpages
cargo run --package xtask -- manpages
...
error: rustc interrupted by SIGSEGV, printing backtrace
...
error: Generating manpages: Syncing man pages: Extracting CLI: Running CLI JSON dump command: command exited with non-zero code `cargo run --release --features=docgen -- internals dump-cli-json`: 101
make: *** [Makefile:45: manpages] Error 1
error: Bad exit status from /var/tmp/rpm-tmp.75FzzG (%build)
```

cgwalters:
Marking draft since this isn't yet doing what we want and it's not clear to me that it's really a good idea to work around it only here.

henrywang:
Hi @jeckersb, the latest cs10 s390x build is passed. https://dashboard.packit.dev/jobs/copr/2855430

jeckersb:
Closing this since it doesn't seem to help (but it was good to at least verify that!)

PR #1676, opened 2025-10-07
Adopt Swatinem/rust-cache for faster CI

sideeffffect:


sideeffffect:
I've noticed that compiling Rust takes a lot of time in CI. Hopefully this will help :crossed_fingers:

.github/workflows/ci.yml
```diff
@@ -32,6 +32,11 @@ jobs:
       - uses: actions/checkout@v4
       - name: Bootc Ubuntu Setup
         uses: ./.github/actions/bootc-ubuntu-setup
+      - name: Setup Rust cache
+        uses: Swatinem/rust-cache@v2
```
p5, line 36: (not a maintainer)
It may be good to, if the maintainers are okay with introducing additional dependencies, to use a fork of the action provided by step-security.  

They selectively cherry pick from upstream into a hardened repo to protect the supply chain.
They have forked well-used (and therefore potentially targets) GitHub Actions, and are generally trusted within the space.

https://github.com/step-security/rust-cache

sideeffffect: Thanks for the suggestion :+1:

sideeffffect: It doesn't work :(
```
Error: Subscription is not valid. Reach out to support@stepsecurity.io
```
https://github.com/bootc-dev/bootc/actions/runs/18330132425/job/52203562664?pr=1676#step:4:15

p5: Oh.  I guess they release _some_ actions to the public and not others.  I couldn't see any docs that suggested it was paywalled.

cgwalters:
Hi, thanks for submitting this! In general actually I'm trying to move most things into a container build, and there's some more nontrivial work to enable using this action with that.

We're relying today on intermediate cache layers in the container build. Now docker does have support for caching layers to/from GHA but podman doesn't directly. It might not be had to add though.

My inclination on compliation speedup here was to look at https://github.com/mozilla/sccache/

.github/workflows/ci.yml
```diff
@@ -32,6 +32,12 @@ jobs:
...
+        with:
+          cache-all-crates: true
+          save-if: ${{ github.ref == 'refs/heads/main' }}
+          lookup-only: ${{ github.ref == 'refs/heads/main' }}
```
cgwalters, line 40: Why are you setting `lookup-only`?

.github/workflows/ci.yml
```diff
@@ -32,6 +32,12 @@ jobs:
       - uses: actions/checkout@v4
       - name: Bootc Ubuntu Setup
         uses: ./.github/actions/bootc-ubuntu-setup
+      - name: Setup Rust cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          cache-all-crates: true
+          save-if: ${{ github.ref == 'refs/heads/main' }}
```
cgwalters, line 39: Why? My understanding is that due to how github handles caching we're safe against e.g. malicious PRs.

.github/workflows/ci.yml
sideeffffect: This will ensure that on `main`, no cache will be used and everything will be built from scratch.
But if you don't like it, it doesn't have to be there.

.github/workflows/ci.yml
sideeffffect: This will ensure that caches are saved only on `main`, not on any other PRs. I think that's enough caching. It may also have some vague security benefits.
But if you don't like it, it doesn't have to be there.

.github/workflows/ci.yml
cgwalters: > This will ensure that on main, no cache will be used and everything will be built from scratch.

Yes, that's the *effect* the change has. I'm asking about *why* we want that effect - it's not the default, so you went to some effort to turn it on. 

A [github-wide search](https://github.com/search?q=%22save-if%3A+%24%7B%7B+github.ref+%3D%3D+%27refs%2Fheads%2Fmain%27+%7D%7D%22&type=code) shows it's definitely used elsewhere but...why?

.github/workflows/ci.yml
cgwalters: I think anything that's not the default should have a comment with a rationale. When you say "enough caching" - is that to ensure that we're not evicting caches from git main with pull requests?

sideeffffect:
Regarding why change the defaults: if the cache is always created in a pristine environment, there's much higher chance of working around any potential compiler bugs.

If the compiler and it's incremental compilation were 100% correct, we wouldn't need to worry about it.

But the potential for bugs is always there. This would give us most of the benefits of caching while avoiding potentially nasty problems related to caching.

But if you'd rather keep things simple, without these opt-in features, no worries, I'd be happy to remove it.

------

Regarding cashing via containers/layers/images, I don't have a good experience with it. You're in all likelihood much bigger expert then me. But I had trouble with determinism, the layers will have always different hashes, even for the same data; this is caused by timestamps which will continue to change. Thus the only reproducible container tool that I use now is JIB from Google (sbt-jib), it fixes the timestamps to 0 (or 2010-01-01 or something like that), ensuring the whole layer(s) have bit per bit identical data and thus hashes.

I hope this helps :)

.github/workflows/ci.yml
cgwalters: Followup for this in https://github.com/bootc-dev/bootc/pull/1678/commits/bb0693c32699378a093b92e5fee88718a93713bd which adds comments, let me know if I got that wrong

PR #1680, opened 2025-10-10
ci: Use bcvk

cgwalters:
Make builds unprivileged by default and start using bcvk in our own CI, the same way we expect people to do so locally now.

### DIFF
--- a/.github/actions/bootc-ubuntu-setup/action.yml
+++ b/.github/actions/bootc-ubuntu-setup/action.yml
@@ -65,5 +65,20 @@ runs:
       if: ${{ inputs.libvirt == 'true' }}
       shell: bash
       run: |
-        set -eux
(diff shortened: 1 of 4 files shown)


cgwalters:
>  Error: reading blob sha256:70bccdb16279fab744edcf78bf4ef42ca06cddc350ad437cfd1d0a0f1702d3b8: open /run/virtiofs-mnt-hoststorage/overlay/70bccdb16279fab744edcf78bf4ef42ca06cddc350ad437cfd1d0a0f1702d3b8/diff/usr/lib64/python3.9/profile.py: too many open files

I think this is https://gitlab.com/virtio-fs/virtiofsd/-/issues/210 - investigating.

cgwalters:
OK yep https://github.com/bootc-dev/bcvk/pull/60 will fix this

cgwalters:
@henrywang can you review?

Issue #1276, opened 2025-04-15
DOC: does `bootc` prune obsolete container layers from previous deployments


jcpunk:
`rpm-ostree cleanup` provides a way to remove old data. Is there a `bootc` integration/automation/script/etc?

cgwalters:
bootc automatically prunes data not referenced by a deployment, the same as rpm-ostree. However, we don't yet offer a CLI or API to explicitly remove the pending or rollback state, but we should.

(We do interoperate though today, using `rpm-ostree cleanup -b` will remove the rollback and bootc handles that, as does the lower level `ostree admin undeploy`, but that won't automatically GC container layers)

This relates a lot to https://github.com/bootc-dev/bootc/issues/904 on the opposite end.

I am retitling this issue under the assumption that your RFE is a way to remove the rollback, is that correct?

jcpunk:
I'm primarily interested in getting the current behavior documented.

RFEs take a while and may require some design work...

> that won't automatically GC container layers

Eventually I'd love that too!  My projected "nodes" look like they may have a 25 year field lifetime, so garbage collection will eventually be very important along side historical records.

PR #1685, opened 2025-10-13
Composefs deleting deployment

Johan-Liebert1:
Add a command to delete a composefs native deployment

Deleting a deployment would mean, deleting the EROFS image, the
bootloader entries for that deployment and deleting any objects in the
composefs repository that are only referenced by said deployment.

Also refactor some functions and add error contexts in some places

Draft PR for now as it requires some more adjustments

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -589,7 +589,7 @@ checksum = "55b672471b4e9f9e95499ea597ff64941a309b2cdbffcc46f2cc5e2d971fd335"
 [[package]]
 name = "composefs"
 version = "0.3.0"
-source = "git+https://github.com/containers/composefs-rs?rev=82e211ee3e5c8e2b7b3bc00b40b828133633c41b#82e211ee3e5c8e2b7b3bc00b40b828133633c41b"
+source = "git+https://github.com/containers/composefs-rs?rev=0f636031a1ec81cdd9e7f674909ef6b75c2642cb#0f636031a1ec81cdd9e7f674909ef6b75c2642cb"
 dependencies = [
  "anyhow",
  "hex",
@@ -609,7 +609,7 @@ dependencies = [
 [[package]]
 name = "composefs-boot"
 version = "0.3.0"
-source = "git+https://github.com/containers/composefs-rs?rev=82e211ee3e5c8e2b7b3bc00b40b828133633c41b#82e211ee3e5c8e2b7b3bc00b40b828133633c41b"
+source = "git+https://github.com/containers/composefs-rs?rev=0f636031a1ec81cdd9e7f674909ef6b75c2642cb#0f636031a1ec81cdd9e7f674909ef6b75c2642cb"
 dependencies = [
  "anyhow",
  "composefs",
@@ -622,7 +622,7 @@ dependencies = [
 [[package]]
 name = "composefs-oci"
 version = "0.3.0"
-source = "git+https://github.com/containers/composefs-rs?rev=82e211ee3e5c8e2b7b3bc00b40b828133633c41b#82e211ee3e5c8e2b7b3bc00b40b828133633c41b"
+source = "git+https://github.com/containers/composefs-rs?rev=0f636031a1ec81cdd9e7f674909ef6b75c2642cb#0f636031a1ec81cdd9e7f674909ef6b75c2642cb"
 dependencies = [
  "anyhow",
  "async-compression",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,9 +41,9 @@ chrono = { version = "0.4.38", default-features = false }
 clap = "4.5.4"
 clap_mangen = { version = "0.2.20" }
 # Reviewers (including AI tools): This hash is duplicated as there's not a much better way to handle it right now
-composefs = { git = "https://github.com/containers/composefs-rs", rev = "82e211ee3e5c8e2b7b3bc00b40b828133633c41b", package = "composefs", features = ["rhel9"] }
-composefs-boot = { git = "https://github.com/containers/composefs-rs", rev = "82e211ee3e5c8e2b7b3bc00b40b828133633c41b", package = "composefs-boot" }
-composefs-oci = { git = "https://github.com/containers/composefs-rs", rev = "82e211ee3e5c8e2b7b3bc00b40b828133633c41b", package = "composefs-oci" }
+composefs = { git = "https://github.com/containers/composefs-rs", rev = "0f636031a1ec81cdd9e7f674909ef6b75c2642cb", package = "composefs", features = ["rhel9"] }
+composefs-boot = { git = "https://github.com/containers/composefs-rs", rev = "0f636031a1ec81cdd9e7f674909ef6b75c2642cb", package = "composefs-boot" }
+composefs-oci = { git = "https://github.com/containers/composefs-rs", rev = "0f636031a1ec81cdd9e7f674909ef6b75c2642cb", package = "composefs-oci" }
 fn-error-context = "0.2.1"
 hex = "0.4.3"
 indicatif = "0.18.0"


crates/lib/src/cli.rs
```diff
@@ -1548,6 +1556,11 @@ async fn run_from_opt(opt: Opt) -> Result<()> {
 
         #[cfg(feature = "composefs-backend")]
         Opt::ConfigDiff => get_etc_diff().await,
+
+        #[cfg(feature = "composefs-backend")]
+        Opt::DeleteDeployment { depl_id, delete } => {
```
cgwalters: Note this is also an RFE here https://github.com/bootc-dev/bootc/issues/1276

crates/lib/src/spec.rs
```diff
@@ -207,6 +207,9 @@ pub struct BootEntryComposefs {
     pub boot_type: BootType,
     /// Whether we boot using systemd or grub
     pub bootloader: Bootloader,
+    /// The sha256sum of vmlinuz + initrd
+    /// Only `Some` for Type1 boot entries
+    pub boot_digest: Option<String>,
```
cgwalters, line 214: Instead of String let's use say https://docs.rs/oci-spec/latest/oci_spec/image/struct.Sha256Digest.html for stuff like this?

Though that might be a bit harder with the jsonschema

crates/lib/src/bootc_composefs/delete.rs
```diff
@@ -0,0 +1,370 @@
...
+    let all_depls = host.all_composefs_deployments()?;
+
+    let depl_to_del = all_depls
+        .iter()
+        .find(|d| d.deployment.verity == deployment_id);
+
+    let Some(depl_to_del) = depl_to_del else {
+        anyhow::bail!("Deployment {deployment_id} not found");
+    };
+
```
cgwalters: I think there's a missing check here that refuses to delete the booted one?

crates/lib/src/bootc_composefs/delete.rs
```diff
@@ -0,0 +1,370 @@
...
+    // Unqueue rollback. This makes it easier to delete boot entries later on
+    if matches!(depl_to_del.ty, Some(Slot::Rollback)) && host.status.rollback_queued {
+        composefs_rollback().await?;
+    }
+
+    let sysroot =
+        Dir::open_ambient_dir("/sysroot", ambient_authority()).context("Opening sysroot")?;
+
+    let repo = open_composefs_repo(&sysroot)?;
```
cgwalters, line 257: I'd still like this stuff to key off Storage but maybe we can just do a mass conversion later

crates/lib/src/bootc_composefs/delete.rs
```diff
@@ -0,0 +1,370 @@
...
+                    )
+                }
+            }
+        }
+
+        Bootloader::Systemd => {
+            let device = get_sysroot_parent_dev()?;
+            let (esp_part, ..) = get_esp_partition(&device)?;
+
+            let esp_mount = TempMount::mount_dev(&esp_part)?;
```
cgwalters: This will need updating for https://github.com/bootc-dev/bootc/pull/1691

crates/lib/src/bootc_composefs/delete.rs
```diff
@@ -0,0 +1,370 @@
...
+    grub_dir
+        .atomic_write(USER_CFG_STAGED, buffer)
+        .with_context(|| format!("Writing to {USER_CFG_STAGED}"))?;
+
+    rustix::fs::fsync(grub_dir.reopen_as_ownedfd().context("Reopening")?).context("fsync")?;
+
+    rename_exchange_user_cfg(&grub_dir)
+}
+
+fn delete_depl_boot_entries(deployment: &DeploymentEntry, deleting_staged: bool) -> Result<()> {
```
cgwalters, line 218: Hmm entrie*s* plural? There's a 1:1 mapping with the UKI case, right?

I guess with type1 we're trying to handle multiple?

crates/lib/src/bootc_composefs/delete.rs
```diff
@@ -0,0 +1,370 @@
...
+        }
+
+        buffer.write_all(entry.to_string().as_bytes())?;
+    }
+
+    grub_dir
+        .atomic_write(USER_CFG_STAGED, buffer)
+        .with_context(|| format!("Writing to {USER_CFG_STAGED}"))?;
+
+    rustix::fs::fsync(grub_dir.reopen_as_ownedfd().context("Reopening")?).context("fsync")?;
```
cgwalters: We could make an `atomic_write_fsync` helper

cgwalters: Also btw there's https://docs.rs/cap-std-ext/latest/cap_std_ext/dirext/trait.CapStdExtDirExt.html#tymethod.atomic_replace_with that supports writing-as-you-go instead of building up the in memory buffer.

crates/lib/src/bootc_composefs/delete.rs
```diff
@@ -0,0 +1,370 @@
...
+        tracing::debug!("Deleting staged file {file:?}");
+        std::fs::remove_file(file).context("Removing staged file")?;
+    }
+
+    let state_dir = Path::new(STATE_DIR_RELATIVE).join(deployment_id);
+    sysroot
+        .remove_dir_all(&state_dir)
+        .with_context(|| format!("Removing dir {state_dir:?}"))?;
+
+    for sha in diff {
```
cgwalters: This should really be an api in composefs-rs right?

Also bigger picture...I think we want to *very clearly* separate two things (as libostree does): managing the "GC roots" vs a GC operation.

Deleting a deployment starts with unlinking its GC root: the bootloader entry.

But thereafter we should just invoke a generic "GC" operation which traverses all active roots.

The idea here is we *must* support being interrupted - and we want to be idempotent.

cgwalters (changes requested):

crates/lib/src/spec.rs
Johan-Liebert1: Yeah, I think that's a good idea

crates/lib/src/bootc_composefs/delete.rs
Johan-Liebert1: It's at https://github.com/bootc-dev/bootc/pull/1685/files/883110d50b9245951ba693b87aad02d132c73e8f#diff-12f8773f52c03d119c8cc2cb67374f160f918ff0baae4bd10306f1ce84120dd6R326

I have a `delete` param passed in to this function for debugging purposes. Might be helpful to just see what all will be deleted

crates/lib/src/bootc_composefs/delete.rs
Johan-Liebert1: Yes, I would like to pick it up after this

crates/lib/src/bootc_composefs/delete.rs
Johan-Liebert1: Agreed. There are a few failure cases that need to be handled

> This should really be an api in composefs-rs right?

The deletion of unreferenced objects? It does make sense for it to be in composefs-rs. There's even a command for it, but right now it's basically a no-op.

crates/lib/src/bootc_composefs/delete.rs
Johan-Liebert1: It was more to do with deleting the .conf file and the UKI, and for Type1 deleting the .conf file and the kernel + initrd. This could definitely use some documentation

crates/lib/src/bootc_composefs/delete.rs
Johan-Liebert1: I removed the param, as it doesn't really make much sense

crates/lib/src/spec.rs
Johan-Liebert1: With the jsonschema it's a bit tricky

crates/lib/src/bootc_composefs/delete.rs
Johan-Liebert1: Updated to use `atomic_replace_with`

> We could make an atomic_write_fsync helper

looking at it again, we don't have any one show write + fsync. It's mostly a couple of writes then a final fsync

Johan-Liebert1:
Had some discussions with @allisonkarlitskaya on Matrix on how we'd want to handle gc for a composefs repo. The `gc_objects` function should not be in bootc and will be moved to composefs-rs

crates/lib/src/bootc_composefs/delete.rs
```diff
@@ -0,0 +1,365 @@
...
+
+    let kind = if depl_to_del.pinned {
+        "pinned "
+    } else if deleting_staged {
+        "staged "
+    } else {
+        ""
+    };
+
+    tracing::info!("Deleting {kind}deployment '{deployment_id}'");
```
cgwalters, line 358: Minor but let's use structured logging for this instead like `tracing::info!("Deleting deployment: {deployment_id}", kind = kind)` or so

crates/lib/src/bootc_composefs/delete.rs
```diff
@@ -0,0 +1,365 @@
...
+
+    delete_depl_boot_entries(&depl_to_del, deleting_staged)?;
+
+    composefs_gc().await?;
```
cgwalters, line 362: I think the GC should be a clearly separate step in general so we can amortize its cost (e.g. "delete 3 deployments, then gc") and also skip the gc ("lazy prune storage later")

crates/lib/src/bootc_composefs/gc.rs
```diff
@@ -0,0 +1,226 @@
...
+        },
+    },
+    composefs_consts::{STATE_DIR_RELATIVE, USER_CFG},
+    spec::Bootloader,
+};
+
+#[fn_error_context::context("Listing EROFS images")]
+fn list_erofs_images(sysroot: &Dir) -> Result<Vec<String>> {
+    let images_dir = sysroot
+        .open_dir("composefs/images")
```
cgwalters, line 30: Hmm there isn't an API for this in `Repository`?

I think what would help here is if this API actually took a `Repository` instance for now at least - we can reference its fd and read things directly. Same elsewehere

cgwalters:
Hmm it looks like we had a timeout during install in this run only for c9s. https://github.com/bootc-dev/bootc/actions/runs/18873507377/job/53857419924?pr=1685

PR #1687, opened 2025-10-16
ci/setup: Parallelize again, add timing logs

cgwalters:
Motivation is faster provisioning and easier to debug when something goes wrong.
Also cleanup the codeql cache.


cgwalters:
@henrywang you're on default PR review rotation this cycle, can you look at this one and the others? (Or feel free to reassign reviews)

PR #1691, opened 2025-10-19
bootloader: Mount the ESP with restricted fmask+dmask

cgwalters:
This avoids warnings from `bootctl install` for good reasons. Visible from `bootc install` using systemd-boot.


jmarrero (approved):
lgtm

PR #1693, opened 2025-10-19
work around mock (COPR) regression

cgwalters:
This suddenly started in our RPM (COPR/mock) builds, my suspicion is that seccomp got turned on inadvertently, but let's add some error context here on general principle.

```
thread 'boundimage::tests::test_parse_spec_dir' panicked at crates/lib/src/boundimage.rs:290:49:
called `Result::unwrap()` on an `Err` value: Querying bound images

Caused by:
    Function not implemented (os error 38)
```

### DIFF
--- a/contrib/packaging/bootc.spec
+++ b/contrib/packaging/bootc.spec
@@ -143,7 +143,11 @@ find %{?buildroot}/%{_docdir} ! -type d -printf '%{_docdir}/%%P\n' > bootcdoclis
 
 %if %{with check}
 %check
-%cargo_test
+if grep -qEe 'Seccomp:.*0$' /proc/self/status; then
+    %cargo_test
+else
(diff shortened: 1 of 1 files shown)


cgwalters:
Man I really should have looked carefully at the full list of failing tests. There were a ton from the importer failing, but that wasn't all of it. This seccomp profile turns of *all* use of openat2 which is going to be painful to work around consistently; we also use it in the linting code, etc.

cgwalters:
OK I just turned off all the unit tests in the RPM build if we detect seccomp for now. We already do run the unit tests inside the container build, *and* this only affects COPR not production Koji so there's no real coverage loss.

cgwalters:
@henrywang you're assigned reviewer this cycle, can you take this?

PR #1694, opened 2025-10-20
crates: Add discoverable partition module

cgwalters:
This fleshes out what we had with a more rigorous
binding to the spec.

Assisted-by: Claude Code


cgwalters:
ooo. Hooray for the new composefs integration tests catching a bug.

PR #1695, opened 2025-10-20
ostree-ext: Replace config loading with a stub if we detect mock

cgwalters:
Workaround for https://github.com/rpm-software-management/mock/pull/1613#issuecomment-3421908652


cgwalters:
Just going to fold this into https://github.com/bootc-dev/bootc/pull/1693 - no point having two PRs

PR #1697, opened 2025-10-21
Fix a few instances of oddball baud rate

jeckersb:
Claude pointed this out while I was working on something else
completely unrelated.  But it looks like somewhere along the way
114800 snuck its way in and probably got copy/pasted into a few
places.  The maximum baud rate is 115200[1], and that is the more
typical usage.

[1] https://www.kernel.org/doc/html/latest/admin-guide/serial-console.html


jeckersb:
(If there's some reason it was purposefully like this, please chime in!)

cgwalters (approved):

PR #1708, opened 2025-10-28
lib: Remove composefs-backend feature gate

cgwalters:
While composefs is still experimental, after looking at this I think the feature gating we're doing has a pretty high "pain:gain" ratio - in other words, the risk we're mitigating by having it off is very low.

Since composefs is a focus of development, let's just remove the feature gate. We have good CI coverage for the non-composefs case.


cgwalters:
Hmm interesting the c9s install is timing out here too, this may be a generic bug

PR #1711, opened 2025-10-29
tmt: Try to describe how to reproduce tmt+packages locally

cgwalters:
The whole chain of packit -> tmt+testing-farm is very complicated and I struggle to wrap my head around it even *without* involving bootc.

With a bit of AI assistance we came up with this, which I didn't try yet truly running and clearly if we were to do so we'd need to automate it.

And doing *that* really is probably going to pull us to split out "build an rpm" from "build a container" because the former is what the packit flow wants today.

xref https://github.com/teemtee/tmt/issues/1018

### DIFF
--- a/tmt/plans/integration.fmf
+++ b/tmt/plans/integration.fmf
@@ -1,3 +1,40 @@
+# Note that this provisioning defaults to using a locally built bootc
+# disk image. The default runs in Github Actions use nested KVM
+# to run most of these tests.
+#
+# However, Packit+Testing Farm provisions a stock package-based
+# cloud VM. When it does that, it ignores the `provision:` tag
+# here, sets running_env=packit, and installs the build artifacts to
+# /var/share/test-artifacts/. The prepare phase then builds a bootc
+# image and reinstalls the system.
+#
+# The idea here is all of our tests should work when started from
+# a package-based host *or* an image-based one.
+#
+# For local testing:
+# - Standard workflow: Use tests/run-tmt.sh (builds bootc image first, sets running_env=image_mode)
+#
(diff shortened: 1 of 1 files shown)


jeckersb:
I will look at this in a bit, but note that this is relevant to my interests because the [functional tests in bodhi](https://bodhi.fedoraproject.org/updates/FEDORA-2025-c3ef769947) are failing because apparently [nu isn't installed](https://artifacts.dev.testing-farm.io/53e71b90-a4d0-4422-a7a7-f66bdff72bed/) and I figured it had something to do with the goofy way we provision things in tmt/testing-farm but I haven't dug much more into it.

cgwalters:
Yes...I think that may be due to packit != dist-git CI...https://src.fedoraproject.org/rpms/bootc/pull-request/59 may be aiming to help this?

cgwalters:
cc @henrywang

henrywang:
What about put RPM building commands into justfile? I can work on it.

cgwalters:
OK sounds good; I looked at this briefly and it was a bit complicated at least to satisfy the constraints I wanted to achieve.

My initial goal was having the default `just` (i.e. a container build) implicitly do an RPM build but what is an absolute requirement in this is we preserve the fast incremental build flow. One thing I got hung up on here is ensuring that the Rust build still uses the Rust caching. 

A constraint I would also like to retain is that the RPM (in general, package) build infra is always itself in a container and isn't required in the outer build env (e.g. GHA runner, dev container).

henrywang:
If we can build bootc RPM in Containerfile, we can drop Packit build and make localy test easy.

PR #1715, opened 2025-10-30
Various composefs changes

Johan-Liebert1:
Increase ESP size to 1GB for composefs-backend

Handle the case when an update is already staged. Handle the case when
the staged image is not the same as the update image. Implement `--check`
option

Get OS version from UKI for display

### DIFF
--- a/crates/lib/src/bootc_composefs/boot.rs
+++ b/crates/lib/src/bootc_composefs/boot.rs
@@ -15,8 +15,11 @@ use cap_std_ext::{
 use clap::ValueEnum;
 use composefs::fs::read_file;
 use composefs::tree::RegularFile;
-use composefs_boot::bootloader::{PEType, EFI_ADDON_DIR_EXT, EFI_ADDON_FILE_EXT, EFI_EXT};
 use composefs_boot::BootOps;
+use composefs_boot::{
+    bootloader::{PEType, EFI_ADDON_DIR_EXT, EFI_ADDON_FILE_EXT, EFI_EXT},
+    uki::UkiError,
+};
 use fn_error_context::context;
 use ostree_ext::composefs::fsverity::{FsVerityHashValue, Sha512HashValue};
 use ostree_ext::composefs_boot::bootloader::UsrLibModulesVmlinuz;
@@ -365,7 +368,6 @@ struct BLSEntryPath<'a> {
 #[context("Setting up BLS boot")]
 pub(crate) fn setup_composefs_bls_boot(
     setup_type: BootSetupType,
-    // TODO: Make this generic
     repo: crate::store::ComposefsRepository,
     id: &Sha512HashValue,
     entry: &ComposefsBootEntry<Sha512HashValue>,
@@ -563,6 +565,11 @@ pub(crate) fn setup_composefs_bls_boot(
     Ok(boot_digest)
 }
 
+struct UKILabels {
+    boot_label: String,
+    version: Option<String>,
+}
+
 /// Writes a PortableExecutable to ESP along with any PE specific or Global addons
 #[context("Writing {file_path} to ESP")]
 fn write_pe_to_esp(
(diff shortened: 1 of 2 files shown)


crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -607,7 +614,15 @@ fn write_pe_to_esp(
             );
         }
 
-        boot_label = Some(uki::get_boot_label(&efi_bin).context("Getting UKI boot label")?);
+        let osrel = uki::get_text_section(&efi_bin, ".osrel")
```
cgwalters, line 617: Yes...though we should also have exactly this same data in `/usr/lib/os-release` in the target image right? My inclination would be to use that over parsing the UKI, but in the end it doesn't really matter.

crates/lib/src/bootc_composefs/update.rs
```diff
@@ -1,33 +1,153 @@
...
     cli::UpgradeOpts,
+    spec::ImageReference,
+    store::ComposefsRepository,
 };
 
+use cap_std_ext::cap_std::{ambient_authority, fs::Dir};
+
+#[context("Getting SHA256 Digest for {id}")]
+pub fn str_to_sha256digest(id: &str) -> Result<Sha256Digest> {
+    let id = id.strip_prefix("sha256:").unwrap_or(id);
```
cgwalters, line 24: Hmm it feels like being "loose" in our inputs is not a good idea; we should either strictly require a digest prefix or not.

crates/lib/src/bootc_composefs/update.rs
```diff
@@ -1,33 +1,153 @@
...
+        Dir::open_ambient_dir("/sysroot", ambient_authority()).context("Opening sysroot")?;
+    let repo = open_composefs_repo(&sysroot)?;
+
+    let (img_pulled, mut manifest, mut config) = is_image_pulled(&repo, imgref).await?;
+    let booted_img_digest = manifest.config().digest().digest();
+
+    // We already have this container config. No update available
+    if img_pulled {
+        println!("No changes in: {imgref:#}");
+        // TODO(Johan-Liebert1): What if we have the config but we failed the previous update in the middle?
```
cgwalters, line 85: Yes, I think we need to only write the config object after all of its dependent layers.

crates/lib/src/bootc_composefs/update.rs
```diff
@@ -1,33 +1,153 @@
...
+
+        // We already have this container config. No update available
+        if img_pulled {
+            println!("No changes in staged image: {imgref:#}");
+            return Ok(());
+        }
+    }
+
+    if opts.check {
+        // TODO(Johan-Liebert1): If we have the previous, i.e. the current manifest with us then we can replace the
```
cgwalters, line 122: And yes we need to store the manifest.

crates/lib/src/install/baseline.rs
```diff
@@ -36,6 +36,10 @@ use bootc_mount::is_mounted_in_pid1_mountns;
 // This ensures we end up under 512 to be small-sized.
 pub(crate) const BOOTPN_SIZE_MB: u32 = 510;
 pub(crate) const EFIPN_SIZE_MB: u32 = 512;
+/// EFI Partition size for composefs installations
+/// We need more space than ostree as we have UKIs and UKI addons
+/// We might also need to store UKIs for pinned deployments
+pub(crate) const CFS_EFIPN_SIZE_MB: u32 = 1024;
```
cgwalters, line 42: (Ultimately of course we want this to be more OS controllable, and that thread is adding support for repart.d in to-disk)

crates/lib/src/bootc_composefs/boot.rs
Johan-Liebert1: Yes, I think we could've gone either way as the data from `/usr/lib/os-release` will be embedded in the UKI anyway

crates/lib/src/bootc_composefs/update.rs
Johan-Liebert1: I see. This was just me being careful. I think the layer ids always start with `sha256:`

crates/lib/src/bootc_composefs/update.rs
Johan-Liebert1: That's what we do. This comment was for the cases where we pull the entire image and create an erofs image, but fail to create some files in the state dir, or maybe something else. The bootloader entries will be overwritten, but we will fail the update if the state directory already exists.

PR #1716, opened 2025-10-30
ci: Just use one job for build + test

cgwalters:
In the future, we may want to "shard" tests across multiple runners, but as is right now it's basically just adding overhead to copy the disk image as an artifact across the job.

While we're here, clean things up further to match the general principle that GHA flows should mostly just be running `just`. The logic from build.sh to map from strings -> containers though moves into the GHA for now.


cgwalters:
@gursewak1997 this one should be a straightforward review when you have a sec

PR #1717, opened 2025-10-30
docs: Add docs for install reset to experimental section

ckyrouac:
Assisted-by: Claude Code

### DIFF
--- a/docs/src/experimental-install-reset.md
+++ b/docs/src/experimental-install-reset.md
@@ -1,0 +1,118 @@
+# Factory reset with `bootc install reset`
+
+This is an experimental feature; use `--experimental` flag to acknowledge.
+
+## Overview
+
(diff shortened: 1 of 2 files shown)


docs/src/experimental-install-reset.md
```diff
@@ -0,0 +1,118 @@
...
+
+1. First, remove any remaining deployments from the old stateroot:
+
+```bash
+# List all deployments to find the old stateroot's deployment index
+ostree admin status
+
+# Remove the old deployment(s) by index
+# The index is shown in the output (e.g., "1" for the second deployment)
+ostree admin undeploy <index>
```
cgwalters, line 96: I think this is something we need to wrap in bootc more nicely...perhaps something like `bootc stateroot rm <name>` or so that would also remove all deployments associated with it?

gursewak1997 (approved):

PR #1718, opened 2025-10-31
composefs: Refactoring to use storage

cgwalters:
A key goal here is to avoid global state and have as much as possible be fd-relative as `Storage` encourages, so that things can work consistently in the install and update paths.

### DIFF
--- a/crates/lib/src/cli.rs
+++ b/crates/lib/src/cli.rs
@@ -8,7 +8,7 @@ use std::os::unix::process::CommandExt;
 use std::process::Command;
 use std::sync::Arc;
 
-use anyhow::{ensure, Context, Result};
+use anyhow::{anyhow, ensure, Context, Result};
 use camino::{Utf8Path, Utf8PathBuf};
 use cap_std_ext::cap_std;
 use cap_std_ext::cap_std::fs::Dir;
@@ -24,7 +24,6 @@ use ostree_ext::composefs::fsverity;
 use ostree_ext::composefs::fsverity::FsVerityHashValue;
 use ostree_ext::composefs::splitstream::SplitStreamWriter;
 use ostree_ext::container as ostree_container;
-use ostree_ext::container_utils::ostree_booted;
 use ostree_ext::containers_image_proxy::ImageProxyConfig;
 use ostree_ext::keyfileext::KeyFileExt;
 use ostree_ext::ostree;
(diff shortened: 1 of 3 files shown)


crates/lib/src/status.rs
```diff
@@ -365,13 +369,16 @@ pub(crate) fn get_status(
 
 async fn get_host() -> Result<Host> {
     let host = if ostree_booted()? {
```
Johan-Liebert1: This skips the status check for composefs-backend. We shouldn't need this anyway with the `super::cli:;get_storage` right?

crates/lib/src/cli.rs
```diff
@@ -712,24 +712,11 @@ pub(crate) fn ensure_self_unshared_mount_namespace() -> Result<()> {
...
-}
-
 /// Load global storage state, expecting that we're booted into a bootc system.
 #[context("Initializing storage")]
-pub(crate) async fn get_storage() -> Result<crate::store::Storage> {
-    let global_run = Dir::open_ambient_dir("/run", cap_std::ambient_authority())?;
-    let sysroot = get_locked_sysroot().await?;
-    crate::store::Storage::new(sysroot, &global_run)
+pub(crate) async fn get_storage() -> Result<crate::store::BootedStorage> {
+    prepare_for_write()?;
```
Johan-Liebert1, line 718: There's a check in `prepare_for_write` which exits the program if the system is not ostree booted

Johan-Liebert1: We can probably move this check inside the ostree `BootedStorage` branch. Of course we still need to handle repository locks for composefs, which I don't think all operations are doing

crates/lib/src/cli.rs
```diff
@@ -1109,36 +1084,85 @@ async fn switch(opts: SwitchOpts) -> Result<()> {
...
+                    })
+                    .await??
+                };
+                println!("Updated {deployid} to pull from {target}");
+                return Ok(());
+            }
+            switch_ostree(opts, storage, &booted_ostree).await
+        }
+        BootedStorageKind::Composefs(booted_cfs) => {
+            if opts.mutate_in_place {
```
Johan-Liebert1, line 1106: For composefs we would require a hard reboot for any switch right? I'm not sure if we can switch without rebooting. Maybe we could for Type1 booted systems, but we will definitely need to reboot for UKIs

cgwalters: No I think the same principle applies - yes a new UKI would have a new rootfs digest always and hence be different, but I think what we want here is to check the kernel version, and if that's the same, then we can do a soft reboot.

Johan-Liebert1:
So
https://github.com/bootc-dev/bootc/blob/27b6189f2e3d2662c8548650679326ff619c048f/crates/tests-integration/src/container.rs#L12

fails. This test is supposed to be run inside a container and should be expected to fail. I'm not sure how it was passing until now

Johan-Liebert1:
My bad, the issue was not the test failing, but the fact that previously we didn't error out if the system wasn't deployed via bootc, but now we do.

Johan-Liebert1:
@cgwalters Could you review my changes?

cgwalters:
Hmmm. Yes this is complicated. I am not quite happy with having storage call back into the CLI code, we should clean that up.

I think we need to have a higher level wrapper here that returns `Result<Option<BootedStorage>>`

cgwalters:
Also this shows another bug, we definitely need to do the "re-exec self" dance even in the composefs path.

cgwalters:
OK i tried to clean this up even more. There's a few prep commits here and then I added a final one which uses `Result<Option<>>`.

Johan-Liebert1:
LGTM

PR #1721, opened 2025-10-31
tests: Use systemd-i128 over uuidgen

cgwalters:
`just test-composefs` fails in the new default devcontainer ref <https://github.com/bootc-dev/infra/blob/main/common/.devcontainer> without this.


cgwalters:
@gursewak1997 ping you're on first call review cycle this week, can you look through some PRs? This one is easy

gursewak1997 (approved):
lgtm

PR #1723, opened 2025-10-31
prep work / enhancements to kernel_cmdline

jeckersb:
See individual commits, this is mostly stuff to prepare for bootc_kargs to switch over to using kernel_cmdline for all of its parsing.

### DIFF
--- a/crates/kernel_cmdline/src/bytes.rs
+++ b/crates/kernel_cmdline/src/bytes.rs
@@ -156,7 +156,11 @@ impl<'a> Cmdline<'a> {
 
         // The exact parameter was not found, so we append it.
         let self_mut = self.0.to_mut();
-        if !self_mut.is_empty() && !self_mut.last().unwrap().is_ascii_whitespace() {
+        if self_mut
+            .last()
+            .filter(|v| !v.is_ascii_whitespace())
+            .is_some()
+        {
             self_mut.push(b' ');
         }
         self_mut.extend_from_slice(param.parameter);
@@ -203,7 +207,11 @@ impl<'a> Cmdline<'a> {
         if !seen_key {
             // The parameter was not found, so we append it.
             let self_mut = self.0.to_mut();
-            if !self_mut.is_empty() && !self_mut.last().unwrap().is_ascii_whitespace() {
+            if self_mut
+                .last()
+                .filter(|v| !v.is_ascii_whitespace())
+                .is_some()
+            {
                 self_mut.push(b' ');
             }
             self_mut.extend_from_slice(param.parameter);


crates/kernel_cmdline/src/bytes.rs
```diff
@@ -132,6 +132,36 @@ impl<'a> Cmdline<'a> {
...
+        for p in self.iter() {
+            if p == *param {
+                // Exact match found, don't add duplicate
+                return Action::Existed;
+            }
+        }
+
+        // The exact parameter was not found, so we append it.
+        let self_mut = self.0.to_mut();
+        if !self_mut.is_empty() && !self_mut.last().unwrap().is_ascii_whitespace() {
```
cgwalters: I suspect there's a way to avoid this unwrap

crates/kernel_cmdline/src/bytes.rs
```diff
@@ -678,6 +708,49 @@ mod tests {
...
+    fn test_add() {
+        let mut kargs = Cmdline::from(b"console=tty0 console=ttyS1");
+
+        // add new parameter with duplicate key but different value
+        assert!(matches!(
+            kargs.add(&param("console=ttyS2")),
+            Action::Added
+        ));
+        let mut iter = kargs.iter();
+        assert_eq!(iter.next(), Some(param("console=tty0")));
```
cgwalters, line 752: Minor but I think we could use https://docs.rs/itertools/latest/itertools/fn.equal.html

cgwalters (approved):
Looks sane to me

crates/kernel_cmdline/src/bytes.rs
jeckersb: Good catch, that was all claude throwing that in there :grin:

jeckersb: Oh it's because we already did that in `add_or_modify` and it just copy/pasted it from there.  It's totally inert though because we check to make sure it's not empty immediately before.

cgwalters: This works for me to avoid the unwrap

```
diff --git i/crates/kernel_cmdline/src/bytes.rs w/crates/kernel_cmdline/src/bytes.rs
index 8ff518e8..7c45f54d 100644
--- i/crates/kernel_cmdline/src/bytes.rs
+++ w/crates/kernel_cmdline/src/bytes.rs
@@ -166,7 +166,7 @@ impl<'a> Cmdline<'a> {
         if !seen_key {
             // The parameter was not found, so we append it.
             let self_mut = self.0.to_mut();
-            if !self_mut.is_empty() && !self_mut.last().unwrap().is_ascii_whitespace() {
+            if self_mut.last().filter(|v| !v.is_ascii_whitespace()).is_some() {
                 self_mut.push(b' ');
             }
             self_mut.extend_from_slice(param.parameter);
```

cgwalters: And yes I'm aware that some people think in the age of AI reviewing code is like reviewing assembly but really, this project is managing my root filesystem and I'm going to look at the code.

crates/kernel_cmdline/src/bytes.rs
```diff
@@ -210,20 +257,47 @@ impl<'a> AsRef<[u8]> for Cmdline<'a> {
...
+        self.iter()
+    }
+}
+
+impl<'a, 'other> Extend<Parameter<'other>> for Cmdline<'a> {
+    fn extend<T: IntoIterator<Item = Parameter<'other>>>(&mut self, iter: T) {
+        for param in iter {
+            self.add(&param);
+        }
+    }
```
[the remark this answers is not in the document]
cgwalters: I don't care about `O(N*M)` for this code, kernel commandlines will never be anywhere near a size where it could possibly matter. But well spotted and perhaps worth a `// NOTE this is O(N*M) but we don't care because the sizes will never be large`

crates/kernel_cmdline/src/bytes.rs
jeckersb: Shockingly it looks like we don't (directly) use itertools anywhere in bootc right now.

cgwalters: Yep totally fine, we can do that kind of stuff as async followup

cgwalters (approved):

PR #1724, opened 2025-10-31
Switch bootc_kargs to use kernel_cmdline for parsing

jeckersb:
Had this working earlier this afternoon but broke it in rebase somehow.  Time for halloween festivities but wanted to throw it up incase my laptop is hit by a meteor over the weekend.

### DIFF
--- a/tmt/tests/booted/test-image-pushpull-upgrade.nu
+++ b/tmt/tests/booted/test-image-pushpull-upgrade.nu
@@ -10,12 +10,14 @@ use tap.nu
 const kargsv0 = ["testarg=foo", "othertestkarg", "thirdkarg=bar"]
 const kargsv1 = ["testarg=foo", "thirdkarg=baz"]
 let removed = ($kargsv0 | filter { not ($in in $kargsv1) })
+const quoted_karg = '"thisarg=quoted with spaces"'
 
 # This code runs on *each* boot.
 # Here we just capture information.
 bootc status
 let st = bootc status --json | from json
 let booted = $st.status.booted.image
+let is_composefs = ($st.status.booted.composefs? != null)
 
 # Parse the kernel commandline into a list.
 # This is not a proper parser, but good enough
@@ -76,6 +78,13 @@ RUN echo test content > /usr/share/blah.txt
     let new_root_mtime = ls -Dl /ostree/bootc | get modified
     assert ($new_root_mtime > $orig_root_mtime)
 
+    # Test for https://github.com/ostreedev/ostree/issues/3544
+    # Add a quoted karg using rpm-ostree if available
+    if not $is_composefs {
+        print "Adding quoted karg via rpm-ostree to test ostree issue #3544"
+        rpm-ostree kargs --append=($quoted_karg)
+    }
+
     # And reboot into it
     tmt-reboot
 }
@@ -125,6 +134,14 @@ def second_boot [] {
         assert ($x in $cmdline)
     }
 
+    # Test for https://github.com/ostreedev/ostree/issues/3544
+    # Verify the quoted karg added via rpm-ostree is still present
+    if not $is_composefs {
+        print "Verifying quoted karg persistence (ostree issue #3544)"
+	let cmdline = open /proc/cmdline
+        assert ($quoted_karg in $cmdline) $"Expected quoted karg ($quoted_karg) not found in cmdline"
+    }
+
     # Now do another build where we drop one of the kargs
     let td = mktemp -d
     cd $td
@@ -167,6 +184,14 @@ def third_boot [] {
     # And the kargs that should be removed are gone
     for x in $removed {
         assert not ($removed in $cmdline)
+    }
+
+    # Test for https://github.com/ostreedev/ostree/issues/3544
+    # Verify the quoted karg added via rpm-ostree is still present after upgrade
+    if not $is_composefs {
+        print "Verifying quoted karg persistence after upgrade (ostree issue #3544)"
+	let cmdline = open /proc/cmdline
+        assert ($quoted_karg in $cmdline) $"Expected quoted karg ($quoted_karg) not found in cmdline after upgrade"
     }
 
     tap ok


jeckersb:
Rebased on https://github.com/bootc-dev/bootc/pull/1729, so this is pending getting that sorted out and merged first.

jeckersb:
Hooray for tests

```
verifying third boot
cmdline=[BOOT_IMAGE=(hd0,gpt3)/boot/ostree/default-7e0ca7b10ef8102e3e459b6a8de1b8d09f8c700e82e97d2af2c1d940d967c9ab/vmlinuz-6.17.6-200.fc42.x86_64, root=UUID=d83586c7-8a15-41d1-a383-aca4d4def032, rootflags=subvol=root, rw, boot=UUID=72a5cd34-0b37-4dc9-9907-0524161248c3, localtestkarg=somevalue, otherlocalkarg=42, kargsd-test=1, kargsd-othertest=2, console=ttyS0,115200n8, testing-kargsd=3, ostree=/ostree/boot.1/default/7e0ca7b10ef8102e3e459b6a8de1b8d09f8c700e82e97d2af2c1d940d967c9ab/0, testarg=foo, thirdkarg=bar]
Verifying karg testarg=foo
Verifying karg thirdkarg=baz
Error:   × Assertion failed.
     ╭─[/var/ARTIFACTS/work-test-20-local-upgradem4wulh6_/tmt/plans/integration/test-20-local-upgrade/discover/default-0/tests/tmt/tests/booted/test-image-pushpull-upgrade.nu:165:16]
 164 │         print $"Verifying karg ($x)"
 165 │         assert ($x in $cmdline)
     ·                ────────┬───────
     ·                        ╰── It is not true.
 166 │     }
     ╰────
```

crates/lib/src/bootc_kargs.rs
```diff
@@ -46,39 +47,42 @@ impl Config {
...
-                || arg.starts_with(INITRD_ARG_PREFIX)
-        })
-        .copied()
-        .collect()
+pub(crate) fn root_args_from_cmdline(cmdline: &Cmdline) -> Cmdline<'static> {
+    let mut result = Cmdline::new();
+    for param in cmdline {
+        let key = param.key();
+        if key.starts_with(ROOT) || key.starts_with(ROOTFLAGS) || key.starts_with(INITRD_ARG_PREFIX)
+        {
```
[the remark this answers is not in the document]
jeckersb: Good catch, will update the constants though and use those instead of hard-coding.

crates/lib/src/bootc_kargs.rs
```diff
@@ -216,17 +227,24 @@ mod tests {
...
     fn test_arch() {
         // no arch specified, kargs ensure that kargs are applied unconditionally
         let sys_arch = "x86_64";
         let file_content = r##"kargs = ["console=tty0", "nosmt"]"##.to_string();
-        let parsed_kargs = parse_kargs_toml(&file_content, sys_arch).unwrap();
-        assert_eq!(parsed_kargs, ["console=tty0", "nosmt"]);
+        let parsed_kargs = parse_kargs_toml(&file_content, sys_arch).unwrap().unwrap();
+        let mut iter = parsed_kargs.iter();
+        assert_eq!(iter.next(), Some(Parameter::parse("console=tty0").unwrap()));
+        assert_eq!(iter.next(), Some(Parameter::parse("nosmt").unwrap()));
```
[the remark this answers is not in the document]
jeckersb: Sure

jeckersb:
Also we should modify the upgrade integration test to include quoted kargs as in https://github.com/ostreedev/ostree/issues/3544 which originally prompted all of this

jeckersb:
Ok good news is that cherry-picking the patch that modifies the upgrade test onto main reproduces the original problem where ostree aborts:

```
16:32:23                         content:  INFO Staging image for deployment: ostree-unverified-image:containers-storage:localhost/bootc-derived (digest: sha256:84c3e696405c96ad1a4ed6f122f5588a14dbe4bdf78e70999a1fb75e0599ce70) message_id="8f7a2b1c3d4e5f6a7b8c9d0e1f2a3b4c" bootc.image.reference="localhost/bootc-derived" bootc.image.transport="containers-storage" bootc.manifest_digest="sha256:84c3e696405c96ad1a4ed6f122f5588a14dbe4bdf78e70999a1fb75e0599ce70"
16:32:23                         content: DEBUG kargs: added=["thirdkarg=baz"] removed=["othertestkarg", "thirdkarg=bar"]
16:32:23                         content: Deploying...**
16:32:23                         content: OSTree:ERROR:src/libostree/ostree-kernel-args.c:197:split_kernel_args: 'quoted' should be FALSE
16:32:23                         content: Bail out! OSTree:ERROR:src/libostree/ostree-kernel-args.c:197:split_kernel_args: 'quoted' should be FALSE
16:32:23                         content: Error: nu::shell::core_dumped
16:32:23                         content:
16:32:23                         content:   x External command core dumped
16:32:23                         content:      ,-[/var/tmp/tmt/run-019/tmt/plans/integration/test-20-local-upgrade/discover/default-0/tests/tmt/tests/booted/test-image-pushpull-upgrade.nu:160:20]
16:32:23                         content:  159 |     # We should already be fetching updates from container storage
16:32:23                         content:  160 |     RUST_LOG=debug bootc upgrade
16:32:23                         content:      :                    ^^|^^
16:32:23                         content:      :                      `-- core dumped with SIGABRT (6)
16:32:23                         content:  161 |     # Verify we staged an update
16:32:23                         content:      `----
```

crates/lib/src/bootc_kargs.rs
```diff
@@ -161,50 +168,56 @@ pub(crate) fn get_kargs(
...
+    }
 
     Ok(kargs)
 }
 
 /// This parses a bootc kargs.d toml file, returning the resulting
 /// vector of kernel arguments. Architecture matching is performed using
 /// `sys_arch`.
-fn parse_kargs_toml(contents: &str, sys_arch: &str) -> Result<Vec<String>> {
+fn parse_kargs_toml(contents: &str, sys_arch: &str) -> Result<Option<Cmdline<'static>>> {
```
cgwalters: It confused me at first that this is using `'static` but I guess it means we're using owned values. Maybe worth a `CmdlineOwned` alias?

crates/lib/src/install.rs
```diff
@@ -2318,29 +2319,28 @@ pub(crate) async fn install_reset(opts: InstallResetOpts) -> Result<()> {
...
-    let kargs = crate::bootc_kargs::get_kargs_in_root(rootfs, std::env::consts::ARCH)?
-        .into_iter()
-        .chain(root_kargs.into_iter())
-        .chain(opts.karg.unwrap_or_default())
-        .collect::<Vec<_>>();
+    // Extend with user-provided kargs
+    if let Some(user_kargs) = opts.karg {
+        if !user_kargs.is_empty() {
+            let user_kargs_cmdline = Cmdline::from(user_kargs.join(" "));
+            kargs.extend(&user_kargs_cmdline);
```
cgwalters: This looks wrong offhand to serialize the user kargs back to a single string

cgwalters (changes requested):
One thing I'm questioning but otherwise looks sane

crates/lib/src/bootc_kargs.rs
jeckersb: :+1: Done

crates/lib/src/install.rs
jeckersb: I changed this to map directly to `Parameter` instead.

Edit: or more precisely, collect into `Result<Vec<Parameter>>`

crates/lib/src/install.rs
```diff
@@ -2318,29 +2319,34 @@ pub(crate) async fn install_reset(opts: InstallResetOpts) -> Result<()> {
...
+    }
 
-    let kargs = crate::bootc_kargs::get_kargs_in_root(rootfs, std::env::consts::ARCH)?
-        .into_iter()
-        .chain(root_kargs.into_iter())
-        .chain(opts.karg.unwrap_or_default())
-        .collect::<Vec<_>>();
+    // Extend with user-provided kargs
+    if let Some(user_kargs) = opts.karg {
+        if !user_kargs.is_empty() {
```
cgwalters: I don't think we need this condition anymore since we're no longer incorrectly mapping an empty vector to an empty string.

crates/lib/src/install.rs
```diff
@@ -2318,29 +2319,34 @@ pub(crate) async fn install_reset(opts: InstallResetOpts) -> Result<()> {
...
+            let user_kargs = user_kargs
+                .iter()
+                .map(|arg| {
+                    Parameter::parse(arg).ok_or_else(|| anyhow!("Unable to parse parameter: {arg}"))
```
cgwalters: Looks good though one thing to bear in mind is that we can push string parsing all the way up into clap structs where applicable, and I think this is one of those cases

jeckersb:
Hmm

```
thread 'main' (60193) panicked at crates/tests-integration/src/system_reinstall.rs:104:17:
File not found: /ostree/deploy/default/deploy/fee15478094fcc50471aad464eb2c51f8fb180c05b659ec22f11b1b270c831a1.0/usr/lib/bootc/fedora-bootc-destructive-cleanup
/home/runner/work/_temp/1fcac6ae-bc62-420f-bb5d-9c76858f2d28.sh: line 32: 60192 Aborted                 sudo bootc-integration-tests system-reinstall localhost/bootc-integration --test-threads=1
Error: Process completed with exit code 134.
```

That looks unrelated to this but worth noting.

jeckersb:
> Hmm
> 
> ```
> thread 'main' (60193) panicked at crates/tests-integration/src/system_reinstall.rs:104:17:
> File not found: /ostree/deploy/default/deploy/fee15478094fcc50471aad464eb2c51f8fb180c05b659ec22f11b1b270c831a1.0/usr/lib/bootc/fedora-bootc-destructive-cleanup
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> /home/runner/work/_temp/1fcac6ae-bc62-420f-bb5d-9c76858f2d28.sh: line 32: 60192 Aborted                 sudo bootc-integration-tests system-reinstall localhost/bootc-integration --test-threads=1
> Error: Process completed with exit code 134.
> ```
> 
> That looks unrelated to this but worth noting.

Ah I see https://github.com/bootc-dev/bootc/issues/1748 now :+1:

crates/lib/src/install.rs
jeckersb: Yeah that would be ideal.  It's a bit awkward the way things are right now though, because `Parameter` is always borrowed from elsewhere (typically from its parent `Cmdline`).  I guess we could refactor it to work exactly like `Cmdline` so it uses a `Cow` internally and then we could build owned values from the clap options.

crates/lib/src/install.rs
jeckersb: :+1: Removed the `is_empty()` check.

crates/lib/src/install.rs
jeckersb: > Yeah that would be ideal. It's a bit awkward the way things are right now though, because `Parameter` is always borrowed from elsewhere (typically from its parent `Cmdline`). I guess we could refactor it to work exactly like `Cmdline` so it uses a `Cow` internally and then we could build owned values from the clap options.

Actually after sleeping on it, it occurs to me that we can just use `Option<Vec<CmdlineOwned>>` for the `karg` flag.  As a side-effect of that you would be able to specify multiple parameters in one option while still supporting multiple options.  So for example: `--karg "foo bar baz" --karg "wuz"` would correctly parse that as three and one arg respectively, and add all four.

PR #1727, opened 2025-11-04
install: Handle mount points in rootfs operations

### LINKED ISSUE #1728: Separate /var part required for ISO based installation but forbidden for 'bootc install to-filesystem'
What were you trying to do that didn't work?
[bootc] Separate /var part required for ISO based installation but forbidden for 'bootc install to-filesystem'

What is the impact of this issue to you?
impossible to create a separate logvols for directories under `/var/` if /var itself is not a seprate logvol as well.

Please provide the package NVR for which the bug is seen:
Package - bootc-1.5.1-1.el9.x86_64.rpm

How reproducible is this bug?:
Steps to reproduce:
Successfully installed  bootc container and ISO using instructions below:

Step 1 :

```
[root@test-rhel9-bootc-builder bootc-build]# cat Containerfile
FROM registry.redhat.io/rhel9/rhel-bootc:latest

RUN dnf -y install iproute rhc unzip tmux && dnf clean all
RUN systemctl mask bootc-fetch-apply-updates.timer
RUN mkdir -p /var/home
RUN useradd -d /home/admin -m -u 1001 -G wheel -p 'password' admin
RUN usermod -p 'password' root
RUN echo "test file" > /home/admin/test_file
RUN chown 1001:1001 /home/admin/test_file
RUN mkdir /foo
RUN mkdir /bar
```

Step 2:

```
RUN bootc container lint
```

Step 3:

```
podman build -t  localhost/bootc .
```

Step 4:

```
[root@test-rhel9-bootc-builder bootc-build]# cat config.toml
[customizations.installer.kickstart]
contents = """

lang en_US.UTF-8
keyboard us
timezone UTC
network --device=link --bootproto=dhcp --onboot=on --activate

zerombr
clearpart --all --initlabel
part /boot --fstype ext4 --size=1024 --fsoptions='defaults,sync,nosuid'
part /boot/efi --fstype vfat --size 200 --fsoptions='defaults,sync,uid=0,gid=0,umask=0077,shortname=winnt,nosuid'
part pv.10 --size=1024 --grow
volgroup BL pv.10
logvol swap --fstype swap --name=swapLV --vgname=BL --size=4096
#logvol /var --fstype ext4 --name=var01 --vgname=BL --size=4096
logvol /var/log --fstype ext4 --name=log --vgname=BL --size=4096
logvol /var/log/audit --fstype ext4 --name=audit --vgname=BL --size=1024
logvol /var/home --fstype ext4 --name=home --vgname=BL --size=512
logvol /foo --fstype ext4 --name=root02 --vgname=BL --size=4096
part /foo/boot --fstype ext4 --size=1024 --fsoptions='defaults,sync,nosuid'
part /foo/boot/efi --fstype vfat --size 200 --fsoptions='defaults,sync,uid=0,gid=0,umask=0077,shortname=winnt,nosuid'
#logvol /foo/var --fstype ext4 --name=var02 --vgname=BL --size=4096
logvol /tmp --fstype ext4 --name=tmp --vgname=BL --size=1656
logvol /opt --fstype ext4 --name=opt --vgname=BL --size=4096
logvol /var/crash --fstype ext4 --name=crash --vgname=BL --size=1024
logvol /var/crypted --fstype ext4 --name=crypted --vgname=BL --size=512
logvol /var/reserved_space --fstype ext4 --name=reserved_space --vgname=BL --size=34096
logvol /var/lib/containers --fstype xfs --name=containers --vgname=BL --size=24096
logvol / --fstype ext4 --name=root01 --vgname=BL --fsoptions='defaults' --size=4096
logvol /ssp_disks --fstype ext4 --name=mep_storage --vgname=BL --percent=100

reboot --eject

"""
```

Step 5: 

```
I have build installation ISO using command:
[root@test-rhel9-bootc-builder bootc-build]# podman run --rm -it --privileged --pull=newer --security-opt label=type:unconfined_t -v /var/lib/containers/storage:/var/lib/containers/storage -v $(pwd)/config.toml:/config.toml -v $(pwd)/output:/output registry.redhat.io/rhel9/bootc-image-builder:latest --local --type iso --config /config.toml localhost/bootc
```

Step 6:

When I try to install machine using build above ISO I'm getting error:

```
`The command 'mount --bind /mnt/sysimage/var/crash /mnt/sysroot/var/crash' exited with the code 32`
```

See attached screenshoot for the anaconda with the error. 

Step 7:

The issue dissapears is if in config.toml I uncomment line creating separte logvol under /var

```
#logvol /var --fstype ext4 --name=var01 --vgname=BL --size=4096
```

In other words, it is impossible to create a separate logvols for directories under `/var/` if /var itself is not a seprate logvol as well.

Expected results:
According documentation linked above:

"There is just one /var directory. If it is not a distinct partition, then physically the /var directory is a bind mount into /ostree/deploy/$stateroot/var and is shared across the available boot loader entries deployments."

So it suggest that using separate lvm/part for /var is not required.

Could you elaborate on this?

1. Should it be possible to have separate LVMs crated under var (especially /var/lib/containers) but not /var itself?

2. At the same time is not not possible to use separate /var mount point if I'm trying to deploy build in the same way image using option 

"bootc install to-filesystem"
Error returned in such case is:
ERROR Installing to filesystem: Verifying empty rootfs: Non-empty root filesystem; found "var"

To be more precise, structure to which I woujld like to run "bootc install to-filesystem" (only relevant part is included)

````
├─sda3                    1G  0 part /foo/boot                          26d962b2-1111-4453-9614-bf3f91234567
├─sda4                  200M  0 part /foo/boot/efi    

  ├─BL-var02              4G  0 lvm  /foo/var                           c026ae7e-1111-4842-9f90-0f5fc3456789
  ├─BL-root02             5G  0 lvm  /foo     
````

Command used:

````
podman run --rm --privileged -v /dev:/dev -v /var/lib/containers:/var/lib/containers -v /foo:/target -v /run/user/0/containers/auth.json:/etc/ostree/auth.json --pid=host --security-opt label=type:unconfined_t localhost/bootc bootc install to-filesystem --karg=root=UUID=52d280c9-1234-49cc-9a38-4818a1234567 --root-mount-spec=UUID=52d280c9-1234-49cc-9a38-4818a1234567 --boot-mount-spec=UUID=26d962b2-1111-4453-9614-bf3f91234567 /target
Note option:  -v /foo:/target used.
````

### PR DESCRIPTION
**Pull request by ckyrouac**
Allow mount points (separate filesystems) to exist during to-filesystem installations by detecting and skipping them in the empty rootdir check.

When cleaning directories during install, recursively clean the contents of mount point directories while preserving the mount point itself. This is necessary for cases where subdirectories like /var/log or /var/home are separate logical volumes.

Assisted-by: Claude Code

Fixes #1728

---

I need to test this more and add automated tests. Looking for feedback on possible side effects I'm not aware of.

### DIFF
--- a/crates/lib/src/install.rs
+++ b/crates/lib/src/install.rs
@@ -1803,6 +1803,46 @@ pub(crate) async fn install_to_disk(mut opts: InstallToDiskOpts) -> Result<()> {
     Ok(())
 }
 
+/// Require that a directory contains only mount points recursively.
+/// Returns Ok(()) if all entries in the directory tree are either:
+/// - Mount points (on different filesystems)
+/// - Directories that themselves contain only mount points (recursively)
+/// - The lost+found directory
+///
+/// Returns an error if any non-mount entry is found.
+///
+/// This handles cases like /var containing /var/lib (not a mount) which contains
+/// /var/lib/containers (a mount point).
+#[context("Requiring directory contains only mount points")]
+fn require_dir_contains_only_mounts(parent_fd: &Dir, dir_name: &str) -> Result<()> {
+    let Some(dir_fd) = parent_fd.open_dir_noxdev(dir_name)? else {
+        // The directory itself is a mount point
+        return Ok(());
+    };
+
+    if dir_fd.entries()?.next().is_none() {
+        anyhow::bail!("Found empty directory: {dir_name}");
+    }
+
+    for entry in dir_fd.entries()? {
+        let entry = DirEntryUtf8::from_cap_std(entry?);
+        let entry_name = entry.file_name()?;
+
+        if entry_name == LOST_AND_FOUND {
+            continue;
+        }
+
+        let etype = entry.file_type()?;
+        if etype == FileType::dir() && dir_fd.open_dir_noxdev(&entry_name)?.is_some() {
+            require_dir_contains_only_mounts(&dir_fd, &entry_name)?;
+        } else {
+            anyhow::bail!("Found entry in {dir_name}: {entry_name}");
+        }
+    }
+
+    Ok(())
+}
+
 #[context("Verifying empty rootfs")]
 fn require_empty_rootdir(rootfs_fd: &Dir) -> Result<()> {
     for e in rootfs_fd.entries()? {
@@ -1811,17 +1851,11 @@ fn require_empty_rootdir(rootfs_fd: &Dir) -> Result<()> {
         if name == LOST_AND_FOUND {
             continue;
         }
-        // There must be a boot directory (that is empty)
-        if name == BOOT {
-            let mut entries = rootfs_fd.read_dir(BOOT)?;
-            if let Some(e) = entries.next() {
-                let e = DirEntryUtf8::from_cap_std(e?);
-                let name = e.file_name()?;
-                if matches!(name.as_str(), LOST_AND_FOUND | crate::bootloader::EFI_DIR) {
-                    continue;
-                }
-                anyhow::bail!("Non-empty boot directory, found {name}");
-            }
+
+        // Check if this entry is a directory
+        let etype = e.file_type()?;
+        if etype == FileType::dir() {
+            require_dir_contains_only_mounts(rootfs_fd, &name)?;
         } else {
             anyhow::bail!("Non-empty root filesystem; found {name:?}");
         }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: crates/lib/src/install.rs [hunk 1]
```diff
@@ -1771,6 +1771,19 @@ fn require_empty_rootdir(rootfs_fd: &Dir) -> Result<()> {
         if name == LOST_AND_FOUND {
             continue;
         }
+
+        // Check if this entry is a directory
+        let etype = e.file_type()?;
+        if etype == FileType::dir() {
+            // Check if this directory is a mount point (separate filesystem)
```
cgwalters: Offhand this seems sane to me

crates/ostree-ext/src/commit.rs
```diff
@@ -57,10 +57,23 @@ fn clean_subdir(root: &Dir, rootdev: u64) -> Result<()> {
             tracing::trace!("Skipping entry in foreign dev {path:?}");
             continue;
         }
-        // Also ignore bind mounts, if we have a new enough kernel with statx()
-        // that will tell us.
+        // For mount points (e.g., separate LVM volumes under /var), we need to clean
```
cgwalters: For the record I think this code is "almost" dead, it should only be reachable from the `ostree container commit` CLI verb which we have backed away from.

While it is related, I think this would be better as a separate PR with its own rationale, because it's really not related at all to the install flow.

cgwalters:
I'm not quite sure this is going to actively fix https://github.com/bootc-dev/bootc/issues/1728 though. This case is more like "system-reinstall-bootc on a system where e.g. `/foo` is a mount point. Except `to-existing-root` already defaults to `--wipe=alongside` which should skip this logic anyways?

crates/lib/src/install.rs
```diff
@@ -1803,6 +1803,54 @@ pub(crate) async fn install_to_disk(mut opts: InstallToDiskOpts) -> Result<()> {
...
+/// Check if a directory contains only mount points (or is empty), recursively.
+/// Returns true if all entries in the directory tree are either:
+/// - Mount points (on different filesystems)
+/// - Directories that themselves contain only mount points (recursively)
+/// - The lost+found directory
+///
+/// This handles cases like /var containing /var/lib (not a mount) which contains
+/// /var/lib/containers (a mount point).
+#[context("Checking if directory contains only mount points")]
+fn dir_contains_only_mounts(parent_fd: &Dir, dir_name: &str) -> Result<bool> {
```
cgwalters: Could have a unit test for this for the trivial cases at least (no inner mounts). Actually testing the mount case needs an integration test.

crates/lib/src/install.rs
```diff
@@ -1811,6 +1859,27 @@ fn require_empty_rootdir(rootfs_fd: &Dir) -> Result<()> {
...
+            // This handles the case where subdirectories like /var/log, /var/home are separate
+            // partitions, creating a /var directory that exists only to hold mount points.
+            if dir_contains_only_mounts(rootfs_fd, &name)? {
+                tracing::debug!("Skipping directory containing only mount points: {name}");
+                continue;
+            }
+        }
+
         // There must be a boot directory (that is empty)
         if name == BOOT {
```
cgwalters: I think we might be able to consolidate this logic with the above?

crates/lib/src/install.rs
```diff
@@ -1803,6 +1803,54 @@ pub(crate) async fn install_to_disk(mut opts: InstallToDiskOpts) -> Result<()> {
...
+
+            // Not a mount point itself, but check recursively if it contains only mounts
+            if dir_contains_only_mounts(&dir_fd, &entry_name)? {
+                tracing::debug!("Directory {dir_name}/{entry_name} contains only mount points");
+                continue;
+            }
+        }
+
+        // Found a non-mount, non-directory-of-mounts entry
+        tracing::debug!("Found non-mount entry in {dir_name}: {entry_name}");
```
cgwalters: The logs here feel a bit noisy potentially, how about making this

```fn require_dir_contains_only_mounts(...) -> Result<()>``` and the error message is the first non-mount entry?

Then I think we probably don't need the logs at all, the error will include the info.

File: crates/lib/src/install.rs [hunk 6]
```diff
@@ -1811,6 +1859,27 @@ fn require_empty_rootdir(rootfs_fd: &Dir) -> Result<()> {
...
+                continue;
+            }
+
+            // If the directory itself is not a mount point, check if it contains only mount points.
+            // This handles the case where subdirectories like /var/log, /var/home are separate
+            // partitions, creating a /var directory that exists only to hold mount points.
+            if dir_contains_only_mounts(rootfs_fd, &name)? {
+                tracing::debug!("Skipping directory containing only mount points: {name}");
+                continue;
+            }
```
cgwalters: Feels a bit weird to me to fall through here on error instead of just erroring out, see above

crates/lib/src/install.rs
```diff
@@ -1803,6 +1803,50 @@ pub(crate) async fn install_to_disk(mut opts: InstallToDiskOpts) -> Result<()> {
...
+/// /var/lib/containers (a mount point).
+#[context("Requiring directory contains only mount points")]
+fn require_dir_contains_only_mounts(parent_fd: &Dir, dir_name: &str) -> Result<()> {
+    let Some(dir_fd) = parent_fd.open_dir_noxdev(dir_name)? else {
+        // The directory itself is a mount point
+        return Ok(());
+    };
+
+    if dir_fd.entries()?.next().is_none() {
+        anyhow::bail!("Found empty directory: {dir_name}");
```
cgwalters, line 1824: Wait why is this an error? It seems to conflict with the doc string.

crates/lib/src/install.rs
```diff
@@ -1803,6 +1803,50 @@ pub(crate) async fn install_to_disk(mut opts: InstallToDiskOpts) -> Result<()> {
...
+
+    for entry in dir_fd.entries()? {
+        let entry = DirEntryUtf8::from_cap_std(entry?);
+        let entry_name = entry.file_name()?;
+
+        if entry_name == LOST_AND_FOUND {
+            continue;
+        }
+
+        if dir_name == BOOT && entry_name == crate::bootloader::EFI_DIR {
```
cgwalters: This is a bit of ugly special casing, I wonder if we still need it? If we do...it may be better to keep the special casing in `require_empty_rootdir` and *this* function is otherwise has very simple logic.

crates/lib/src/install.rs
ckyrouac: this is how the check previously behaved so I wanted to preserve the same behavior. The doc string is wrong

crates/lib/src/install.rs
ckyrouac: ah right, for some reason I was thinking /boot/EFI would be an empty dir but we just need to allow it to be a mount point. Removed this check.

Issue #1624, opened 2025-09-18
tmt/testing followups phase 2: flexible provisioning and build


cgwalters:
Splitting out of https://github.com/bootc-dev/bootc/issues/1473 where https://github.com/bootc-dev/bootc/pull/1607 changes a lot of things but as always there's lots of followups.

This also relates to https://gitlab.com/fedora/bootc/tests/bootc-workflow-test/-/issues/361

Basically we support multiple provisioning systems (Anaconda, bib/to-disk, s-r-b). Our tests should support being run from *any* of these.

In this I think we can change our TMT logic to execute different [prepare phases](https://tmt.readthedocs.io/en/stable/plugins/prepare.html)

[ ] Support a packit-based flow where packit builds an RPM and injects it into the target environment and then we just do `FROM <base> RUN dnf -y upgrade bootc.rpm` and then do a system-reinstall-bootc to that as a preparatory step
[ ] Support Anaconda

In the Anaconda flow we require a container image as input, and I'd really like to support that locally with libvirt without pushing to a registry, but also in the cloud.

---

Also, I need to publish "bcvk" (formerly bootc-kit) which already streamlines the local unprivileged `to-disk` flow and can be extended for local Anaconda and bib.

PR #1733, opened 2025-11-05
Switch over qemu tests to use bcvk

cgwalters:
Closes: https://github.com/bootc-dev/bootc/issues/1624

### DIFF
--- a/tmt/tests/examples/bootc-uki/build.final
+++ b/tmt/tests/examples/bootc-uki/build.final
@@ -9,9 +9,10 @@ cp /usr/bin/bootc .
 rm -rf tmp/sysroot
 mkdir -p tmp/sysroot/composefs
 
+# TODO port this over to container compute-composefs-digest
 IMAGE_ID="$(sed s/sha256:// tmp/iid)"
-./bootc internals cfs --repo tmp/sysroot/composefs oci pull containers-storage:"${IMAGE_ID}"
-COMPOSEFS_FSVERITY="$(./bootc internals cfs --repo tmp/sysroot/composefs oci compute-id --bootable "${IMAGE_ID}")"
+./bootc internals cfs --repo tmp/sysroot/composefs --insecure oci pull containers-storage:"${IMAGE_ID}"
+COMPOSEFS_FSVERITY="$(./bootc internals cfs --repo tmp/sysroot/composefs --insecure oci compute-id --bootable "${IMAGE_ID}")"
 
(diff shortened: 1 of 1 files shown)


cgwalters:
OK hum that's interesting, all tests passed *except* ` /tmt/plans/integration/test-27-custom-selinux-policy`....because of a reboot timeout? Weird...

(Also man digging through tmt logs to find the actual failure is really just terrible...it looks like https://gitlab.com/testing-farm/oculus is the thing that fixes all this up for testing-farm?)

In debugging it seems like something is going wrong with `tmt-reboot`, if I ssh into the guest and just run `reboot` then the test passes.

tmt/bug-soft-reboot.md
```diff
@@ -0,0 +1,35 @@
+# TMT soft-reboot limitation
+
+TMT does not currently support systemd soft-reboots. It detects reboots by checking
```
cgwalters, line 3: Man did I get burned by this. It's very subtle because basically tmt has a *another* bug where `tmt-reboot` with testcloud for some reason goes out to the hypervisor and "physically" shuts down and restarts...bypassing the in-guest reboot logic entirely.

But of course, we're not doing that with bcvk because 1) there's no magic backchannel to do so 2) why would we it doesn't make sense

cgwalters:
>  + test centos-9 = centos-10

Oh man am I so glad I had that last minute thought while refactoring to cross-check that we're actually building for the specified target OS...

cgwalters:
Oh man so close, the last bug seems to be that the two jobs for centos-10 (one ostree, one composefs) were trying to save the same artifact ID, that should be fixed...

cgwalters:
:tada: it's passing!
@gursewak1997 can you review?

Justfile
```diff
@@ -59,27 +68,27 @@ validate-local:
...
+# bcvk.
 #
-# This task runs *all* of the tmt-based tests targeting the disk image generated
-# in the previous step.
-test-tmt *ARGS: build-disk
-    ./tests/run-tmt.sh {{ARGS}}
+# To run an individual test, pass it as an argument like:
+# `just test-tmt readonly`
+test-tmt *ARGS: build-integration-test-image
+    cargo xtask run-tmt --env=BOOTC_variant={{variant}} localhost/bootc-integration {{ARGS}}
```
gursewak1997, line 77: This kinda confused me, as in we are using cargo xtask run-tmt (Rust implementation using bcvk)

PR #1736, opened 2025-11-06
blockdev: add `bootable` property to find ESP partition

HuijingHei:
As Fedora IoT on the Raspberry Pi 3 is using `MBR` by default, we need to find ESP partition via `bootable`.
See https://github.com/coreos/bootupd/issues/1019

### DIFF
--- a/crates/blockdev/src/blockdev.rs
+++ b/crates/blockdev/src/blockdev.rs
@@ -117,6 +117,7 @@ pub struct Partition {
     pub parttype: String,
     pub uuid: Option<String>,
     pub name: Option<String>,
+    pub bootable: Option<bool>,
 }
 
 #[derive(Debug, Deserialize, PartialEq, Eq)]
@@ -169,6 +170,11 @@ impl PartitionTable {
     pub fn find_partition_of_type(&self, uuid: &str) -> Option<&Partition> {
         self.partitions.iter().find(|p| p.parttype_matches(uuid))
     }
+
+    /// Find the partition with bootable is 'true'.
+    pub fn find_partition_of_bootable(&self) -> Option<&Partition> {
+        self.partitions.iter().find(|p| p.is_bootable())
+    }
 }
 
 impl Partition {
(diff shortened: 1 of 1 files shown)


cgwalters:
Nothing wrong with adding a bootable property, but I don't think that's how we should find the ESP - per [Wikipedia](https://en.wikipedia.org/wiki/EFI_system_partition)

> The [globally unique identifier](https://en.wikipedia.org/wiki/Globally_unique_identifier) (GUID) for the EFI system partition in the [GUID Partition Table](https://en.wikipedia.org/wiki/GUID_Partition_Table) (GPT) scheme is C12A7328-F81F-11D2-BA4B-00A0C93EC93B, while its ID in the [master boot record](https://en.wikipedia.org/wiki/Master_boot_record) (MBR) partition-table scheme is [0xEF](https://en.wikipedia.org/wiki/Partition_type#PID_EFh).

So we should search for 0xEF instead.

cgwalters (approved):

HuijingHei:
> Nothing wrong with adding a bootable property, but I don't think that's how we should find the ESP - per [Wikipedia](https://en.wikipedia.org/wiki/EFI_system_partition)
> 
> > The [globally unique identifier](https://en.wikipedia.org/wiki/Globally_unique_identifier) (GUID) for the EFI system partition in the [GUID Partition Table](https://en.wikipedia.org/wiki/GUID_Partition_Table) (GPT) scheme is C12A7328-F81F-11D2-BA4B-00A0C93EC93B, while its ID in the [master boot record](https://en.wikipedia.org/wiki/Master_boot_record) (MBR) partition-table scheme is [0xEF](https://en.wikipedia.org/wiki/Partition_type#PID_EFh).
> 
> So we should search for 0xEF instead.

Agree, find `bootable` looks more easier, and `0xEF` is more accurate, will add it later, thank you for the pointer!

cgwalters:
So we should define a `const ESP_ID_MBR` and use that both in bootupd and bootc, and move the `esp_in()` code down to blockdev.rs and have it detect MBR vs GPT.

Issue #1737, opened 2025-11-06
kernel_cmdline quoting equivalence


jeckersb:
Side quest spawned from investigating https://github.com/ostreedev/ostree/issues/3544 ...

Quoted kargs end up getting transformed between grub and the kernel, here's an example:

```
# Original kargs does not have a "foo" value
[root@ibm-p8-kvm-03-guest-02 ~]# rpm-ostree kargs
root=UUID=4f68bce3-e8cd-4db1-96e7-fbcaf984b709 rw ostree=/ostree/boot.1/default/944761db707ac4bce2b4e1362b79cfea486bcd764ac56131a30780b4927ee26b/0
# Append a quoted arg
[root@ibm-p8-kvm-03-guest-02 ~]# rpm-ostree kargs --append='foo="quoted value"'
Staging deployment... done
Changes queued for next boot. Run "systemctl reboot" to start a reboot
# Sanity check that the quoted arg was added, and in the format we expect
[root@ibm-p8-kvm-03-guest-02 ~]# rpm-ostree kargs
root=UUID=4f68bce3-e8cd-4db1-96e7-fbcaf984b709 rw ostree=/ostree/boot.1/default/944761db707ac4bce2b4e1362b79cfea486bcd764ac56131a30780b4927ee26b/0 foo="quoted value"
[root@ibm-p8-kvm-03-guest-02 ~]# systemctl reboot
[...]
# After reboot, boot loader entry looks correct
[root@ibm-p8-kvm-03-guest-02 ~]# grep options /boot/loader/entries/ostree-2.conf
options root=UUID=4f68bce3-e8cd-4db1-96e7-fbcaf984b709 rw ostree=/ostree/boot.0/default/944761db707ac4bce2b4e1362b79cfea486bcd764ac56131a30780b4927ee26b/0 foo="quoted value"
# However, the kernel sees the value as "externally" quoted in early boot
[root@ibm-p8-kvm-03-guest-02 ~]# journalctl -b --grep 'Command line:' --no-pager
[...]

PR #1739, opened 2025-11-06
kernel_cmdline: Fix parsing/equivalence of "outside" quoted args

jeckersb:
Primary motivation here is that these two should be equivalent:

foo="quoted value"
"foo=quoted value"

This also adds tests for a few more oddball cases that weren't covered
before but clarify the expected kernel behavior.

Closes: #1737

### DIFF
--- a/crates/kernel_cmdline/src/bytes.rs
+++ b/crates/kernel_cmdline/src/bytes.rs
@@ -421,27 +421,31 @@ impl<'a> Parameter<'a> {
     /// This is an internal method that assumes the input has already been
     /// split into a single parameter (e.g., by CmdlineIterBytes).
     fn parse_internal(input: &'a [u8]) -> Option<Self> {
-        let equals = input.iter().position(|b| *b == b'=');
+        // *Only* the first and last double quotes are stripped
+        let dequoted_input = input.strip_prefix(b"\"").unwrap_or(input);
+        let dequoted_input = dequoted_input.strip_suffix(b"\"").unwrap_or(dequoted_input);
+
+        let equals = dequoted_input.iter().position(|b| *b == b'=');
 
         match equals {
             None => Some(Self {
                 parameter: input,
-                key: ParameterKey(input),
+                key: ParameterKey(dequoted_input),
                 value: None,
             }),
             Some(i) => {
-                let (key, mut value) = input.split_at(i);
+                let (key, mut value) = dequoted_input.split_at(i);
                 let key = ParameterKey(key);
 
                 // skip `=`, we know it's the first byte because we
                 // found it above
                 value = &value[1..];
 
-                // *Only* the first and last double quotes are stripped
-                value = {
-                    let v = value.strip_prefix(b"\"").unwrap_or(value);
-                    v.strip_suffix(b"\"").unwrap_or(v)
-                };
+                // If there is a quote after the equals, skip it.  If
+                // there was a closing quote at the end of the value,
+                // we would have already removed it in
+                // `dequoted_input` above
+                value = value.strip_prefix(b"\"").unwrap_or(value);
 
                 Some(Self {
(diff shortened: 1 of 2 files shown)


crates/kernel_cmdline/src/bytes.rs
```diff
@@ -549,9 +557,22 @@ mod tests {
     fn test_parameter_pathological() {
         // valid things that certified insane people would do
 
-        // quotes don't get removed from keys
```
jeckersb, line 552: This was just incorrect before.

crates/kernel_cmdline/src/bytes.rs
```diff
@@ -421,27 +421,31 @@ impl<'a> Parameter<'a> {
...
-                // *Only* the first and last double quotes are stripped
-                value = {
-                    let v = value.strip_prefix(b"\"").unwrap_or(value);
-                    v.strip_suffix(b"\"").unwrap_or(v)
-                };
+                // If there is a quote after the equals, skip it.  If
+                // there was a closing quote at the end of the value,
+                // we would have already removed it in
+                // `dequoted_input` above
+                value = value.strip_prefix(b"\"").unwrap_or(value);
```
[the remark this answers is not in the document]
jeckersb: I don't believe you but it's time to go home so I will look at it tomorrow :laughing:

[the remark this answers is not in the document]
jeckersb: Ok not quite tomorrow but there's a passing test for that case...

```rust
    #[test]
    fn test_parameter_quoted() {
        let p = param("foo=\"quoted value\"");
        assert_eq!(p.value, Some(b"quoted value".as_slice()));
```

Plus this whole bit just doesn't make any sense:

> For an argument like `foo="bar"`, the closing quote is at the end of the value but not at the end of the whole parameter, so it isn't removed by the initial dequoted_input step.

:-1:

crates/kernel_cmdline/src/bytes.rs
```diff
@@ -421,27 +421,31 @@ impl<'a> Parameter<'a> {
     /// This is an internal method that assumes the input has already been
     /// split into a single parameter (e.g., by CmdlineIterBytes).
     fn parse_internal(input: &'a [u8]) -> Option<Self> {
-        let equals = input.iter().position(|b| *b == b'=');
+        // *Only* the first and last double quotes are stripped
+        let dequoted_input = input.strip_prefix(b"\"").unwrap_or(input);
+        let dequoted_input = dequoted_input.strip_suffix(b"\"").unwrap_or(dequoted_input);
```
cgwalters, line 426: What happens if there's only one quote at the start? Shouldn't we *not* strip the trailing one in that case? What does the Linux kernel do in this corner case?

crates/kernel_cmdline/src/utf8.rs
```diff
@@ -426,6 +426,16 @@ mod tests {
     fn test_parameter_quoted() {
         let p = param("foo=\"quoted value\"");
         assert_eq!(p.value(), Some("quoted value"));
+
+        let p = param("foo=\"unclosed quotes");
+        assert_eq!(p.value(), Some("unclosed quotes"));
```
cgwalters, line 431: I guess this is an answer to my question above, but is it the correct answer?

cgwalters (approved):

crates/kernel_cmdline/src/utf8.rs
jeckersb: Yeah that's this line - https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/lib/cmdline.c#n256.  We see '=', replace it with '\0', advance one, if it's '"' we advance again past it, effectively stripping it.

PR #1743, opened 2025-11-07
rust: cargo update

cgwalters:
Renovate is currently failing because of the composefs-rs git submodule (trying to debug that) so we're behind. At least this one gets rid of multiple copies of rustix and an old toml version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,15 +3,6 @@
 version = 3
 
 [[package]]
-name = "addr2line"
-version = "0.24.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
-dependencies = [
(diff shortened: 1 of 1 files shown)


jeckersb:
> 👍
> 
> Only thing I see is that we get a new deprecation warning:
> 
> ```
> warning: use of deprecated unit variant `rustix::thread::Capability::SystemAdmin`: Use CapabilitySet with a single bit instead
>    --> crates/lib/src/cli.rs:743:83
>     |
> 743 |         rustix::thread::capability_is_in_bounding_set(rustix::thread::Capability::SystemAdmin)?,
>     |                                                                                   ^^^^^^^^^^^
>     |
>     = note: `#[warn(deprecated)]` on by default
> 
> warning: `bootc-lib` (lib) generated 1 warning
> ```
> 
> But that can be fixed in a followup.

:arrow_right: https://github.com/bootc-dev/bootc/pull/1750

PR #1746, opened 2025-11-09
Justfile: Two minor tweaks

cgwalters:


henrywang:
Just make a rebase for `test install` failure.

PR #1749, opened 2025-11-10
kernel_cmdline: Simplify Display impl for utf8::Parameter

jeckersb:
Really we want to just return the original slice that the Parameter
was created from.  This also aligns with how Cmdline and ParameterKey
impl Display.

Where this really matters the most is to ensure we retain the quoting
that the parameter was created with, so I added a test just to sanity
check that.  Before this change the test would fail because "foo"
would be stripped of its quotes and just rendered as foo unquoted.

### DIFF
--- a/crates/kernel_cmdline/src/utf8.rs
+++ b/crates/kernel_cmdline/src/utf8.rs
@@ -232,8 +232,7 @@ impl<'a> AsRef<str> for Cmdline<'a> {
 
 impl<'a> std::fmt::Display for Cmdline<'a> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
-        let as_str: &str = self.as_ref();
-        write!(f, "{as_str}")
+        f.write_str(self)
     }
 }
 
@@ -267,7 +266,7 @@ impl<'a, 'other> Extend<Parameter<'other>> for Cmdline<'a> {
 #[derive(Clone, Debug, Eq)]
 pub struct ParameterKey<'a>(bytes::ParameterKey<'a>);
 
-impl<'a> std::ops::Deref for ParameterKey<'a> {
+impl<'a> Deref for ParameterKey<'a> {
     type Target = str;
 
     fn deref(&self) -> &Self::Target {
@@ -305,8 +304,7 @@ impl<'a, T: AsRef<str> + ?Sized> From<&'a T> for ParameterKey<'a> {
 
 impl<'a> std::fmt::Display for ParameterKey<'a> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
-        let as_str: &str = self;
-        write!(f, "{as_str}")
+        f.write_str(self)
     }
 }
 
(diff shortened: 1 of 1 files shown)


crates/kernel_cmdline/src/utf8.rs
```diff
@@ -368,16 +368,8 @@ impl<'a> TryFrom<bytes::Parameter<'a>> for Parameter<'a> {
...
-                if v.contains(|ch: char| ch.is_ascii_whitespace()) {
-                    write!(f, "{}=\"{}\"", self.key(), v)
-                } else {
-                    write!(f, "{}={}", self.key(), v)
-                }
-            }
-            None => write!(f, "{}", self.key()),
-        }
+        let as_str: &str = self.as_ref();
+        write!(f, "{as_str}")
```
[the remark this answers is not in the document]
jeckersb: Oh nice, I guess I'll update all three Display impls to work that way...

crates/kernel_cmdline/src/utf8.rs
```diff
@@ -213,6 +213,16 @@ impl<'a> Cmdline<'a> {
     }
 }
 
+impl Deref for Cmdline<'_> {
+    type Target = str;
+
+    fn deref(&self) -> &Self::Target {
+        // SAFETY: We know this is valid UTF-8 since we only
+        // construct the underlying `bytes` from valid UTF-8
+        str::from_utf8(&self.0).expect("We only construct the underlying bytes from valid UTF-8")
```
bshephar, line 222: `from_utf8` is a safe operation right? Isn't it `from_utf8_unchecked` that is the unsafe method?
Safe:
https://doc.rust-lang.org/std/primitive.str.html#method.from_utf8

> If you are sure that the byte slice is valid UTF-8, and you don’t want to incur the overhead of the validity check, there is an unsafe version of this function, [from_utf8_unchecked](https://doc.rust-lang.org/std/str/fn.from_utf8_unchecked.html), which has the same behavior but skips the check.

Unsafe:
https://doc.rust-lang.org/std/str/fn.from_utf8_unchecked.html

Just wondering if the safety comment is necessary here. Seems like it would only be necessary if using the `unchecked` version.

jeckersb: Yeah it's safe and probably not correct to the letter of the law of rust conventions to label it like that.  It's like that because in the very earliest version of this code, it did actually use `from_utf8_unchecked`.  But really it's not so performance critical that we need to squeeze out a few cycles and avoid the safe code.  So we switched it to do the checked version and take the performance "penalty".  In theory however, those comments still hold true and we should *never* hit the case where the underlying bytes are invalid UTF-8 and if we really really wanted to we could switch back to using the unchecked version.

Or put another way... it should be 100% impossible for this to panic.  But if something *does* goes wrong, at least it will panic instead of undefined behavior.

henrywang:
Just make a rebase for `test install` failure.

jeckersb:
Hm not sure what the testing farm failures on stream 10 are about:

```
        Command 'reboot' returned 1.

        # stderr (1 lines)
        Call to Reboot failed: Access denied
```

Unrelated to this change but worth noting.

Issue #1388, opened 2025-06-29
install fstab for `/boot` conflicts with transient etc


stephan-tolksdorf:
After `install to-filesystem` the generated `fstab` is present in the sysroot `etc` folder, but not the `usr/etc` folder, which then with a transient etc leads to the file missing after boot in the `etc` overlay, because the lower is `usr/etc`.

cgwalters:
Hi, thanks for the report.

As of the current code, we only write a fstab entry for `/boot` if we detect one in use. One solution is to not use a separate `/boot` partition. Another option is to align with https://uapi-group.org/specifications/specs/discoverable_partitions_specification/ - though it doesn't quite cover the "classic linux /boot" partition I think. (Though that needs some digging in and testing, I may be wrong and we could actually just use that even if it's not VFAT)

I think we can unconditionally switch `to-filesystem` to use the (somewhat new) `systemd.mount-extra` kernel argument instead of us writing a fstab. That functionality isn't in RHEL9 today is the only trap, but we could probably detect the systemd version in use.

If systemd is new enough we should unconditionally use the karg I think. If systemd is too old *and* transient etc is in use, then we error out by default.

stephan-tolksdorf:
Thanks! `systemd.mount-extra` seems interesting, I'll try that.

Currently I'm helping myself by copying the fstab to `/usr/etc` before the first boot. (In my workflow I build images for the partitions which are then copied to the target disk by an installer script. root is luks encrypted, while boot isn't.)

cgwalters:
> Currently I'm helping myself by copying the fstab to /usr/etc before the first boot. 

That will go away on the first upgrade, so it can't really work in general.

stephan-tolksdorf:
The idea is to do upgrades like first installations, i.e. overwrite the root and boot partitions with the new disk images from a script.

cgwalters:
> The idea is to do upgrades like first installations, i.e. overwrite the root and boot partitions with the new disk images from a script.

OK, though at that point you're only using 10% of bootc; in particular it will be hard to implement transactional updates. If the goal is to absolutely minimize state drift, see https://github.com/bootc-dev/bootc/issues/404 (which has some recent WIP).

The first solution though to state drift in `/etc` is just to limit which tools write to it. For `/var`, it's possible to clean things on shutdown/reboot via e.g. `tmpfiles.d`.

cgwalters:
[...]

Issue #820, opened 2024-10-11
install to-existing-root: don't blow away /boot until later


cgwalters:
In `install to-existing-root`, we can delay blowing away `/boot` until very late in the process - specifically until we've pulled the container image for example, which may hit problems.

It'd make us overall more robust to being `ctrl-c`'d in general and we could even have a last minute extra warning + delay there as it's really that process that's the destructive thing.

PR #1752, opened 2025-11-11
install: add `target_root_path` for `RootSetup`

HuijingHei:
- install: empty `/boot` & `/boot/efi`
- When install to-filesystem on ostree OS, will pass `target_root_path` to `bootupctl` to install bootloader.
- install: should remove the empty dir under `/boot`
- install: mount esp part before `clean_boot_directories()`

### DIFF
--- a/tmt/plans/tests-install.fmf
+++ b/tmt/plans/tests-install.fmf
@@ -14,6 +14,8 @@ provision:
   user: root
   memory: 4096
   disk: 20
+  ssh-option:
+    - "PubkeyAcceptedAlgorithms=+ssh-rsa"
 
 /plan-bootc-install-on-coreos:
   summary: Execute bootc install on ostree OS


HuijingHei:
Build the patch based on `quay.io/centos-bootc/centos-bootc:stream9`, run `bootc install to-existing-root` on rhel9.6-edge. 

Check the `/boot/efi/EFI/centos` directory is created, `bootuuid.cfg` and `grub.cfg` is updated. Is this expected?

If I reboot, seems it still booted into old rhel9.6, how can I check if the installation is successful?

- Run bootc install
```
# podman run --rm --privileged -v /dev:/dev -v /var/lib/containers:/var/lib/containers -v /:/target \
--pid=host --security-opt label=type:unconfined_t \
localhost/bootc:latest \
env BOOTC_BOOTLOADER_DEBUG=1 bootc install to-existing-root --skip-fetch-check --acknowledge-destructive --root-ssh-authorized-keys /target/root/.ssh/id_rsa.pub --karg=console=ttyS0,115200n8
...
Installing image: docker://localhost/bootc:latest
Digest: sha256:f23c996f7ef414d2a9fbfef366b1dbc513c1f564064fc687da39e793365a5ac9
Bootloader: grub
Reusing extant ostree layout
layers already present: 0; layers needed: 70 (2.2 GB)
Deploying container image...done (26 seconds)
Injected: etc/tmpfiles.d/bootc-root-ssh.conf
[TRACE bootupd] executing cli
[INFO  bootupd::bootupd] System boot method: EFI
[DEBUG bootupd::efi] Found metadata grub2-efi-x64-1:2.06-118.el9.x86_64,shim-x64-15.8-2.el9.x86_64
[DEBUG bootupd::efi] Reusing existing mount point "/target/boot/efi"
[DEBUG bootupd::efi] Get product name: 'CentOS Stream'
Executing: "efibootmgr" "--create" "--disk" "/dev/vda" "--part" "1" "--loader" "\\EFI\\centos\\shimx64.efi" "--label" "CentOS Stream"
[DEBUG bootupd::efi] Creating new EFI boot entry using 'CentOS Stream'
BootCurrent: 0002
Timeout: 0 seconds
BootOrder: 0005,0002,0003,0000,0001,0004
Boot0000* BootManagerMenuApp
Boot0001* EFI Firmware Setup
Boot0002* Red Hat Enterprise Linux
Boot0003* UEFI Misc Device
Boot0004* EFI Internal Shell
Boot0005* CentOS Stream
[INFO  bootupd::bootupd] Installed EFI grub2-efi-x64-1:2.06-118.el9.x86_64,shim-x64-15.8-2.el9.x86_64
[DEBUG bootupd::grubconfigs] root_dev=64768 boot_dev=64514
Added 01_users.cfg
Added 10_blscfg.cfg
Added 14_menu_show_once.cfg
Added 30_uefi-firmware.cfg
Added 41_custom.cfg
Installed: grub.cfg
Installed: bootuuid.cfg
[DEBUG bootupd::grubconfigs] vendordir="centos"
Installed: "centos/grub.cfg"
Installed: "centos/bootuuid.cfg"
[DEBUG bootupd::efi] Unmounting
[TRACE bootupd::efi] Unmounted
Installation complete!

```

- Check the `/boot/efi/EFI/centos` dir exists, `bootuuid.cfg` and `grub.cfg` is updated.
```
[root@localhost ~]# rpm-ostree status
State: idle
Deployments:
● edge:rhel/9/x86_64/edge
                  Version: 9.6 (2025-11-11T07:44:55Z)
                   Commit: e1cfc9ab4666fece360f19620d7a4afdb10905454c40c3258d5215949f2838c2

[root@localhost ~]# ls /boot/efi/EFI/*
/boot/efi/EFI/BOOT:
BOOTX64.EFI  fbx64.efi

/boot/efi/EFI/centos:
BOOTX64.CSV   grub.cfg	   mmx64.efi  shimx64-centos.efi
bootuuid.cfg  grubx64.efi  shim.efi   shimx64.efi

/boot/efi/EFI/redhat:
BOOTX64.CSV  grubx64.efi  shim.efi	      shimx64.efi
grub.cfg     mmx64.efi	  shimx64-redhat.efi

[root@localhost ~]# cat /boot/efi/EFI/centos/bootuuid.cfg 
set BOOT_UUID="9098a193-ff0f-4f34-aa2f-fcff9bb043c0"
[root@localhost ~]# blkid /dev/vda2
/dev/vda2: UUID="9098a193-ff0f-4f34-aa2f-fcff9bb043c0" TYPE="xfs" PARTUUID="1d7009d0-6e6e-4093-b35c-137e3c61531b"

[root@localhost ~]# cat /boot/efi/EFI/centos/grub.cfg 
if [ -e (md/md-boot) ]; then
  # The search command might pick a RAID component rather than the RAID,
  # since the /boot RAID currently uses superblock 1.0.  See the comment in
  # the main grub.cfg.
  set prefix=md/md-boot
else
  if [ -f ${config_directory}/bootuuid.cfg ]; then
    source ${config_directory}/bootuuid.cfg
  fi
  if [ -n "${BOOT_UUID}" ]; then
    search --fs-uuid "${BOOT_UUID}" --set prefix --no-floppy
  else
    search --label boot --set prefix --no-floppy
  fi
fi
if [ -d ($prefix)/grub2 ]; then
  set prefix=($prefix)/grub2
  configfile $prefix/grub.cfg
else
  set prefix=($prefix)/boot/grub2
  configfile $prefix/grub.cfg
fi
boot
```
- Reboot and check rpm-ostreed.service failed
```
[root@localhost ~]# journalctl -u rpm-ostreed.service --no-page
Nov 12 12:19:11 localhost.localdomain systemd[1]: Starting rpm-ostree System Management Daemon...
Nov 12 12:19:11 localhost.localdomain rpm-ostree[951]: Reading config file '/etc/rpm-ostreed.conf'
Nov 12 12:19:11 localhost.localdomain rpm-ostree[951]: error: Couldn't start daemon: Error setting up sysroot: loading sysroot: Unexpected state: /run/ostree-booted found and in / sysroot, but bootloader entry not found
Nov 12 12:19:11 localhost.localdomain systemd[1]: rpm-ostreed.service: Main process exited, code=exited, status=1/FAILURE
Nov 12 12:19:11 localhost.localdomain systemd[1]: rpm-ostreed.service: Failed with result 'exit-code'.
Nov 12 12:19:11 localhost.localdomain systemd[1]: Failed to start rpm-ostree System Management Daemon.
```

cgwalters:
This is a complex topic. A first thing here is that `bootc install` is always destructive with respect to `/boot` and the ESP. We definitely need to avoid that in the case where the system is already managed via bootupd. I think that's causing part of the problem here.

This issue also relates to https://github.com/bootc-dev/bootc/issues/820

I think at the current time it's expected that `to-existing-root` in this scenario does wipe things, but I think we're also missing cleanup of the prior deployments by default.

> If I reboot, seems it still booted into old rhel9.6, how can I check if the installation is successful?

That is strange; what does `bootc status` say in this setup before reboot?

cgwalters:
One issue also related to this is that some of the install integration tests are still mostly orchestrated via the GHA `test-install` and haven't been lowered into a more rigorous/reproducible setup.

I think we should likely switch them over to tmt. Although it gets subtle as reproducing the Edge starting state via tmt will require some work, but we can do that as a followup.

HuijingHei:
> This is a complex topic. A first thing here is that `bootc install` is always destructive with respect to `/boot` and the ESP. We definitely need to avoid that in the case where the system is already managed via bootupd. I think that's causing part of the problem here.

Actually on ostree OS, it only removes `bootupd-state.json` (see https://github.com/bootc-dev/bootc/blob/main/crates/lib/src/install.rs#L1845), and keep all the other files, but maybe cleanup some files later?
If I boot rhel-edge and bootc install centos, then there are 2 vendors under `/boot/efi`, maybe useful to keep the dual OS?

> That is strange; what does `bootc status` say in this setup before reboot?

```
[root@localhost ~]# bootc status
● Booted ostree
           Commit: e1cfc9ab4666fece360f19620d7a4afdb10905454c40c3258d5215949f2838c2
```
Run `bootc install`, before boot, not found new `/boot/loader/entries/ostree-2.conf` and the new default kernel file under `/boot/ostree`, are the files created during boot?
```
[root@localhost ~]# ls /boot/loader/entries/
ostree-1.conf
[root@localhost ~]# ls /boot/ostree/
rhel-73943340a8bb027bb3783557749af43d3b71ed9c720166c7431b08d8845d392c
```

After reboot, boot into the old rhel9.6-edge, find directory `/boot/grub2` and `/boot/bootupd-state.json` are missing, the new entry config and kernel files are created.
The workaround is to install bootloader again to bootinto the the bootc image.
```
# ls /boot/
boot  efi  loader  loader.1  ostree

# ls /boot/ostree/
default-abe244070c34287211ecad478d1ae50bc25a8cdfbefb55218ef19f8c3b3d4735
# cat /boot/loader/entries/ostree-1.conf 
title CentOS Stream 9 (ostree:0)
version 1
options root=/dev/mapper/rootvg-lv_root rd.lvm.lv=rootvg/lv_root rd.lvm.lv=rootvg/lv_swap rw boot=UUID=715f41aa-99c5-42b1-8072-2f73a6600c27 console=ttyS0,115200n8 ostree=/ostree/boot.1/default/abe244070c34287211ecad478d1ae50bc25a8cdfbefb55218ef19f8c3b3d4735/0
linux /boot/ostree/default-abe244070c34287211ecad478d1ae50bc25a8cdfbefb55218ef19f8c3b3d4735/vmlinuz-5.14.0-635.el9.x86_64
initrd /boot/ostree/default-abe244070c34287211ecad478d1ae50bc25a8cdfbefb55218ef19f8c3b3d4735/initramfs-5.14.0-635.el9.x86_64.img
```

cgwalters:
> Actually on ostree OS, it only removes bootupd-state.json 

You're right! And that's one of the issues here - this system is ostree but not bootupd, and that's one of the problem roots - I think in this case we'll just not change the bootloader state at all which is not expected.

I think basically if we detect ostree but not bootupd, then we should proceed by wiping out the complete ESP as before alongside everything in `/boot` *except* `/boot/loader` or so.

---

Now...there is a related but distinct thing here which is that if we detect bootupd, we should also forcibly do a `bootupctl update` *instead* of an install.

HuijingHei:
> I think basically if we detect ostree but not bootupd, then we should proceed by wiping out the complete ESP as before alongside everything in `/boot` _except_ `/boot/loader` or so.

Sorry that this is out of my knowledge, what I see is from [clean_boot_directories()](https://github.com/bootc-dev/bootc/blob/main/crates/lib/src/install.rs#L2048) before install_to_filesystem_impl(), any pointer for this? Thank you!

> Now...there is a related but distinct thing here which is that if we detect bootupd, we should also forcibly do a `bootupctl update` _instead_ of an install.

LGTM, have no idea where I should add this.

> Run bootc install, before boot, not found new /boot/loader/entries/ostree-2.conf and the new default kernel file under `/boot/ostree`

Should the new entries config and kernel file are created after running `bootc install` before reboot?

cgwalters:
> Sorry that this is out of my knowledge, what I see is from [clean_boot_directories()](https://github.com/bootc-dev/bootc/blob/main/crates/lib/src/install.rs#L2048) before install_to_filesystem_impl(), any pointer for this? Thank you!

Yep that's the right place

> LGTM, have no idea where I should add this.

In the logic where we have `install_via_bootupd` I think we'd want to instead do a `bootupctl update` right?

> Should the new entries config and kernel file are created after running bootc install before reboot?

Yes the boot loader entries are written by the install (and upgrade) process; those should stay the same as is today. What's special cased right now is ostree-vs-non-ostree in the `install to-existing-root` flow, but we could change to preserve `/boot/loader` by default in *all* cases, and only prune the non-ostree boot loader entries after an install by default or so.

HuijingHei:
> In the logic where we have install_via_bootupd I think we'd want to instead do a bootupctl update right?

The bootupctl update is a little different as it currently reads path form `/` (see [link](https://github.com/coreos/bootupd/blob/main/src/bootupd.rs#L500)), and it is running as systemd service which might not suitable in container.

> Yes the boot loader entries are written by the install (and upgrade) process; those should stay the same as is today. What's special cased right now is ostree-vs-non-ostree in the install to-existing-root flow, but we could change to preserve `/boot/loader` by default in all cases, and only prune the non-ostree boot loader entries after an install by default or so.

With @jbtrystram 's help, do testing with [change](https://github.com/bootc-dev/bootc/pull/1752/commits/13f3c2252594bfa7950578c5a118495ee3c3d29b), cleanup ESP files and preserve like `/boot/{loader, loader.0, ostree}`, run bootc and before reboot, run `ostree admin finalize-staged`, check new entries are not created, and no new kernel files are installed.

```
[root@localhost ~]# ostree admin finalize-staged -v
OT: root_is_ostree_booted: 1
OT: remounted /sysroot writable
OT: using fuse: 0
OT: Target rootdev key backing-root-device-inode found
OT: Deployment e1cfc9ab4666fece360f19620d7a4afdb10905454c40c3258d5215949f2838c2.0 unlocked=0
```
After reboot, boot into the old rhel9.6-edge, directory `/boot/grub2` and `/boot/bootupd-state.json` are missing, the new entry and kernel files are created, also get rpm-ostree error:

```
rpm-ostree[1273]: error: Couldn't start daemon: Error setting up sysroot: loading sysroot: Unexpected state: /run/ostree-booted found and in / sysroot, but bootloader entry not found
```

I think what you said for the [comment](https://cloud-native.slack.com/archives/C08SKSQKG1L/p1763051027964929?thread_ts=1763014559.071159&cid=C08SKSQKG1L) makes sense, but I have no idea how to fix this, any pointer is much appreciated. Copy the comment here:
```
The latter is likely the /boot partition (/etc/fstab ) content being lost as it's local state by default (this also relates to what we did recently by default in the still-experimental https://github.com/bootc-dev/bootc/pull/1588
```

HuijingHei:
Check more and find that the new `ostree-1.conf` is created after `bootc install`, but put under `/sysroot/boot` and also the kernel files, if I copy the new entry and kernel files to physical `/boot`, reboot failed with:
`error: ../../grub-core/fs/fshelp.c:257:file /boot/ostree/default-abe244070c34287211ecad478d1ae50bc25a8cdfbefb55218ef19f8c3b3d4735/vmlinuz-5.14.0-635.el9.x86_64' not found.` 

```
[root@localhost ~]# cd /sysroot/
[root@localhost sysroot]# cat boot/loader.1/entries/ostree-1.conf 
title CentOS Stream 9 (ostree:0)
version 1
options root=/dev/mapper/rootvg-lv_root rd.lvm.lv=rootvg/lv_root rd.lvm.lv=rootvg/lv_swap rw boot=UUID=6b3a964e-7790-4d9d-a1cf-4a7675a1fc89 console=ttyS0,115200n8 ostree=/ostree/boot.1/default/abe244070c34287211ecad478d1ae50bc25a8cdfbefb55218ef19f8c3b3d4735/0
linux /boot/ostree/default-abe244070c34287211ecad478d1ae50bc25a8cdfbefb55218ef19f8c3b3d4735/vmlinuz-5.14.0-635.el9.x86_64
initrd /boot/ostree/default-abe244070c34287211ecad478d1ae50bc25a8cdfbefb55218ef19f8c3b3d4735/initramfs-5.14.0-635.el9.x86_64.img

[root@localhost sysroot]# ls boot/ostree/default-abe244070c34287211ecad478d1ae50bc25a8cdfbefb55218ef19f8c3b3d4735/
initramfs-5.14.0-635.el9.x86_64.img  vmlinuz-5.14.0-635.el9.x86_64
```

The workaround is remove `/boot` in the entry file like following, can boot into the new bootc image.
```
title CentOS Stream 9 (ostree:0)
version 1
options root=/dev/mapper/rootvg-lv_root rd.lvm.lv=rootvg/lv_root rd.lvm.lv=rootvg/lv_swap rw boot=UUID=6b3a964e-7790-4d9d-a1cf-4a7675a1fc89 console=ttyS0,115200n8 ostree=/ostree/boot.1/default/abe244070c34287211ecad478d1ae50bc25a8cdfbefb55218ef19f8c3b3d4735/0
linux /ostree/default-abe244070c34287211ecad478d1ae50bc25a8cdfbefb55218ef19f8c3b3d4735/vmlinuz-5.14.0-635.el9.x86_64
initrd /ostree/default-abe244070c34287211ecad478d1ae50bc25a8cdfbefb55218ef19f8c3b3d4735/initramfs-5.14.0-635.el9.x86_64.img
```
After booted, check that `/boot` is bindmount with `/sysroot/boot`, not physical boot.
```
[root@localhost ~]# bootc status
● Booted image: localhost/bootc:latest
        Digest: sha256:b66719ea7fb1c4d520eb6e3af22b2e6e72c9d4c0437bd4c692aff1cfd0c25c2e (amd64)
       Version: 9 (2025-11-21T03:31:40Z)


[root@localhost ~]# findmnt /boot 
TARGET SOURCE                            FSTYPE OPTIONS
/boot  /dev/mapper/rootvg-lv_root[/boot] xfs    rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota
```
Edit:
Copy the claude summary here, let me know if I misunderstood this.
The key points we clarified:

1. bootc correctly detects the ostree environment and uses `/sysroot` as the physical root
2. Bootloader entries ARE created during installation to `/sysroot/boot/loader.*/entries/`
3. initramfs creates bind mounts to expose them
4. /etc/fstab boot entry is ignored due to ostree's bind mount happening first in ostree-prepare-root

This explains why the mount shows as `/dev/mapper/rootvg-lv_root[/boot]` instead of the separate boot partition

HuijingHei:
Based on the above [comment](https://github.com/bootc-dev/bootc/pull/1752#issuecomment-3561893321), one workaround is
- clean all files under `/boot & /boot/efi` in `clean_boot_directories()`
- copy `/sysroot/boot/loader*` to `/boot` if `/boot` is separate partition (this need do some change like remove `/boot` in linux & initrd commands)
- copy `/sysroot/boot/ostree` to `/boot`
But not sure if this would make other side effects. Any other concerns for this?

Edit:
A better workaround is Injecting kernel arguments using `systemd.mount-extra`, see https://bootc-dev.github.io/bootc/man/bootc-install-to-existing-root.8.html#injecting-kernel-arguments-for-local-state

HuijingHei:
Do some testing based on PR,

Before reboot:
```
[root@localhost ~]# bootc status
[ 4705.559201] SELinux:  Context unconfined_u:object_r:invalid_bootcinstall_testlabel_t:s0 is not valid (left unmapped).
● Booted ostree
           Commit: 524dddce375eb37d24ef040decf165e475f395e9a32399345bf9200ae836d268

   Other image: localhost/bootc:latest
        Digest: sha256:e30a9a716f4ac09cc9eda2eaf03016d3a1d93a5898521997c707be1e980c0512 (amd64)
       Version: 9 (2025-11-24T05:11:22Z)
```

After reboot and check
```
[systemd]
Failed Units: 1
  bootc-publish-rhsm-facts.service
[root@localhost ~]# bootc status
● Booted image: localhost/bootc:latest
        Digest: sha256:e30a9a716f4ac09cc9eda2eaf03016d3a1d93a5898521997c707be1e980c0512 (amd64)
       Version: 9 (2025-11-24T05:11:22Z)

 Other ostree
       Commit: 524dddce375eb37d24ef040decf165e475f395e9a32399345bf9200ae836d268
```

crates/lib/src/install.rs
```diff
@@ -970,6 +986,22 @@ async fn install_container(
         }
     }
 
+    // For seperate /boot filesystem, the better workaround is
+    // to inject kernel arguments during installation.
+    if let Some(boot) = root_setup.boot.as_ref() {
+        if !boot.source.is_empty() {
+            let mount_extra = format!(
+                "systemd.mount-extra={}:{}:{}:{}",
```
cgwalters: Let's link to  https://github.com/bootc-dev/bootc/issues/1388 here - and also note the problem is that functionality isn't in C9S today, and there were people who wanted this fix on that version.

There's more discussion in that issue, including the big wrinkle that for FIPS we need a `BOOT=` karg today.

crates/lib/src/install.rs
```diff
@@ -758,6 +758,22 @@ async fn initialize_ostree_root(state: &State, root_setup: &RootSetup) -> Result
...
 
+    // If we're installing alongside existing ostree and there's a separate boot partition,
+    // we need to bind mount it to the sysroot's /boot so ostree can write bootloader entries there
+    if has_ostree {
+        if let Some(target_root) = root_setup.target_root_path.as_ref() {
+            if root_setup.boot.is_some() {
+                let source_boot = target_root.join(BOOT);
+                let target_boot = root_setup.physical_root_path.join(BOOT);
+                tracing::debug!("Bind mounting {source_boot} to {target_boot}");
+                Command::new("mount")
```
cgwalters: We have a mount module it'd make sense to use

crates/lib/src/install.rs
```diff
@@ -758,6 +758,22 @@ async fn initialize_ostree_root(state: &State, root_setup: &RootSetup) -> Result
     // Another implementation: https://github.com/coreos/coreos-assembler/blob/3cd3307904593b3a131b81567b13a4d0b6fe7c90/src/create_disk.sh#L295
     crate::lsm::ensure_dir_labeled(rootfs_dir, "", Some("/".into()), 0o755.into(), sepolicy)?;
 
+    // If we're installing alongside existing ostree and there's a separate boot partition,
+    // we need to bind mount it to the sysroot's /boot so ostree can write bootloader entries there
+    if has_ostree {
+        if let Some(target_root) = root_setup.target_root_path.as_ref() {
+            if root_setup.boot.is_some() {
```
cgwalters: Could avoid one level of nesting by moving this to the top `if has_ostree && root_setup.boot.is_some()`

crates/lib/src/install.rs
```diff
@@ -2011,6 +2079,20 @@ pub(crate) async fn install_to_filesystem(
         .context("Mounting host / to {ALONGSIDE_ROOT_MOUNT}")?;
     }
 
+    let target_root_path = &fsopts.root_path.clone();
```
cgwalters: I don't think we need to clone here

crates/lib/src/install.rs
```diff
@@ -2011,6 +2079,20 @@ pub(crate) async fn install_to_filesystem(
...
+    // Get a file descriptor for the root path /target
+    let target_rootfs_fd = {
+        let rootfs_fd = Dir::open_ambient_dir(&target_root_path, cap_std::ambient_authority())
```
cgwalters: I think we can just pass `&fsopts.root_path` here

crates/lib/src/install.rs
```diff
@@ -2011,6 +2079,20 @@ pub(crate) async fn install_to_filesystem(
...
+            .with_context(|| format!("Opening target root directory {target_root_path}"))?;
+
+        tracing::debug!("Root filesystem: {target_root_path}");
```
cgwalters: This debug could go earlier if we need it, we're just a copy of `fsopts.root_path`

crates/lib/src/install.rs
```diff
@@ -2011,6 +2079,20 @@ pub(crate) async fn install_to_filesystem(
...
+
+        if let Some(false) = rootfs_fd.is_mountpoint(".")? {
```
cgwalters: Also I think we should verify this earlier.

To summarize: `target_root_path = fsopts.root_path` and I think it's a lot clearer if we don't have another alias for it.

(Though of course, even better if we *only* hold it as a `Dir` but that's going to be harder)

HuijingHei:
CI error:
```
podman run --privileged --pid=host --user=root:root -v /var/lib/containers:/var/lib/containers -v /dev:/dev --security-opt label=type:unconfined_t -v /:/target -v /tmp/.tmplt1aWT:/bootc_authorized_ssh_keys/root localhost/bootc-integration bootc install to-existing-root --acknowledge-destructive --skip-fetch-check --cleanup --root-ssh-authorized-keys /bootc_authorized_ssh_keys/root
        
        After reboot, the current root will be available in the /sysroot directory. Existing mounts will not be automatically mounted by the bootc system unless they are defined in the bootc image. Some automatic cleanup of the previous root will be performed.
        
        NOTICE: This will replace the installed operating system and reboot. Are you sure you want to continue? [y/N] 
        NOTICE: This will replace the installed operating system and reboot. Are you sure you want to continue? [Y/n] 
        NOTICE: This will replace the installed operating system and reboot. Are you sure you want to continue? yes
        Installing image: docker://localhost/bootc-integration:latest
        Digest: sha256:b441fc960f2cbe1ec4acef0485156f419fb775c89f480717568e56f4fbc2614a
        Initializing ostree layout
        error: Installing to filesystem: Creating ostree deployment: loading sysroot: Not a symbolic link: boot/loader
        error: run: running reinstall command: Failed to run command: Command {
            program: "podman",
            args: [
                "podman",
                "run",
                "--privileged",
                "--pid=host",
                "--user=root:root",
                "-v",
                "/var/lib/containers:/var/lib/containers",
                "-v",
                "/dev:/dev",
                "--security-opt",
                "label=type:unconfined_t",
                "-v",
                "/:/target",
                "-v",
                "/tmp/.tmplt1aWT:/bootc_authorized_ssh_keys/root",
                "localhost/bootc-integration",
                "bootc",
                "install",
                "to-existing-root",
                "--acknowledge-destructive",
                "--skip-fetch-check",
                "--cleanup",
                "--root-ssh-authorized-keys",
                "/bootc_authorized_ssh_keys/root",
            ],
            create_pidfd: false,
        }
```

crates/lib/src/install.rs
HuijingHei: ACK, let me try, thank you!

crates/lib/src/install.rs
HuijingHei: So we need both `target_root_path` and `root_path`, is this right?

- On ostree OS, `target_root_path` will be `/target` (used by bootupd), and `root_path` will be `/target/sysroot`, `physical_root` should be the directory file descriptor of `/target/sysroot`.

- On package mode, both `target_root_path` and `root_path` will be `/target`, `physical_root` should be the directory file descriptor of `/target`.

Edit:
I did testing and am sure that `physical_root` should be the dir of `/target/sysroot` on ostree.

HuijingHei: Also tried with `target_root_path = fsopts.root_path`, but make failed. So I do not change this, but maybe I misunderstood this.

crates/lib/src/install.rs
HuijingHei: With `bootc_mount::mount(source_boot.as_str(), &target_boot)?;`, so change to use boot.source, WDYT?
```
Mount /target/boot to /target/sysroot/boot on ostree
[19568.268855] /target/boot: Can't lookup blockdev
mount: /target/sysroot/boot: /target/boot is not a block device.
error: Installing to filesystem: Creating ostree deployment: Failed to run command: Command {
    program: "mount",
    args: [
        "mount",
        "/target/boot",
        "/target/sysroot/boot",
    ],
    create_pidfd: false,
}
```

crates/lib/src/install.rs
HuijingHei: Done, thank you!

HuijingHei:
Rebase the main branch and waiting for the CI result.

HuijingHei:
Maybe the CI failed is not related to the PR, could you help to review again? Thank you @cgwalters !

crates/lib/src/install.rs
```diff
@@ -1851,30 +1883,59 @@ fn remove_all_in_dir_no_xdev(d: &Dir, mount_err: bool) -> Result<()> {
...
-        if ARCH_USES_EFI {
-            if let Some(efidir) = bootdir
-                .open_dir_optional(crate::bootloader::EFI_DIR)
-                .context("Opening /boot/efi")?
-            {
-                remove_all_in_dir_no_xdev(&efidir, false)
-                    .context("Emptying EFI system partition")?;
-            }
+    // This should not remove /boot/efi note.
+    remove_all_except_loader_dirs(&bootdir, is_ostree).context("Emptying /boot")?;
```
cgwalters, line 1992: Hmm looks like this is going to conflict with https://github.com/bootc-dev/bootc/pull/1727

crates/lib/src/install.rs
cgwalters: This one will need to do something like run `systemctl --version` and detect if the karg is supported, and if not fall back to writing to `/etc/fstab` or so.

crates/lib/src/install.rs
```diff
@@ -758,6 +758,16 @@ async fn initialize_ostree_root(state: &State, root_setup: &RootSetup) -> Result
     // Another implementation: https://github.com/coreos/coreos-assembler/blob/3cd3307904593b3a131b81567b13a4d0b6fe7c90/src/create_disk.sh#L295
     crate::lsm::ensure_dir_labeled(rootfs_dir, "", Some("/".into()), 0o755.into(), sepolicy)?;
 
+    // If we're installing alongside existing ostree and there's a separate boot partition,
+    // we need to mount it to the sysroot's /boot so ostree can write bootloader entries there
```
cgwalters, line 783: One thing I don't quite understand is why other cases don't hit this. I feel like there might be a different bug somewhere.

HuijingHei: On ostree env, the bootloader entries are written under `/sysroot/boot`, this is why mount `/boot` to `/sysroot/boot`.

cgwalters:
On test cases: Hmm, so this case starts from an ostree base, maybe with some custom partitioning; there's a few options here.

- Use Fedora/RHEL CoreOS and set up partitions w/Ignition.
- bootc-image-builder might be able to emulate this (and yes we have a TODO do default to some tests using it here)
- Try replicating the full original scenario w/image builder service

HuijingHei:
> This one will need to do something like run systemctl --version and detect if the karg is supported

From [doc](https://www.freedesktop.org/software/systemd/man/latest/systemd-fstab-generator.html#systemd.mount-extra=WHAT:WHERE%5B:FSTYPE%5B:OPTIONS%5D%5D) get that `systemd.mount-extra` was added in `version 254`, but check 9.6.z find the [patch](https://gitlab.com/redhat/rhel/rpms/systemd/-/blob/rhel-9.6.0/0337-fstab-generator-add-rd.systemd.mount-extra-and-frien.patch?ref_type=heads) is included in 252-18, maybe I should check if on rhel, the version >= 252-18, else >=254.
```
$ systemctl --version  (on F43)
systemd 258 (258.2-1.fc43)

# systemctl --version (on RHEL 9.6)
systemd 252 (252-51.el9_6.2)
```
> and if not fall back to writing to /etc/fstab or so.

Check this already added in https://github.com/bootc-dev/bootc/blob/main/crates/lib/src/install.rs#L1028-L1034

cgwalters:
> From [doc](https://www.freedesktop.org/software/systemd/man/latest/systemd-fstab-generator.html#systemd.mount-extra=WHAT:WHERE%5B:FSTYPE%5B:OPTIONS%5D%5D) get that systemd.mount-extra was added in version 254, but check 9.6.z find the [patch](https://gitlab.com/redhat/rhel/rpms/systemd/-/blob/rhel-9.6.0/0337-fstab-generator-add-rd.systemd.mount-extra-and-frien.patch?ref_type=heads) is included in 252-18, maybe I should check if on rhel, the version >= 252-18, else >=254.

Ah! That's a really useful find! So...hum...special casing rhel version numbers here is a little ugly. We could add a Rust feature like `systemd-supports-mount-extra` and then set that to on in the spec file if we detect rhel >= 9.6 ?

HuijingHei:
Rebase main branch and doing testing, see this when booting the bootc image. Maybe we do not need the kernel argument for boot mount?

```
[    6.692668] systemd-fstab-generator[754]: Failed to create unit file '/run/systemd/generator/boot.mount', as it already exists. Duplicate entry in '/proc/cmdline'?
[    6.830908] systemd[743]: /usr/lib/systemd/system-generators/systemd-fstab-generator failed with exit status 1.
```

HuijingHei:
Hold on and will do more testing for this.

crates/lib/src/install.rs
```diff
@@ -1862,30 +1878,58 @@ fn remove_all_in_dir_no_xdev(d: &Dir, mount_err: bool) -> Result<()> {
...
+        let entry = entry.context("Reading directory entry")?;
+        let file_name = entry.file_name();
+        let file_name = if let Some(n) = file_name.to_str() {
+            n
+        } else {
+            anyhow::bail!("Invalid non-UTF8 filename: {file_name:?} in /boot");
+        };
+
+        // Only preserve loader on ostree
+        if is_ostree && file_name.starts_with("loader") {
```
cgwalters, line 1959: This also relates to an issue (TODO link) where we should support preserving basically everything (including the bootloader entries on non-ostree) by default until the very end of the install. And ideally make the "commit" phase an optional step after.

This one also textually conflicts with https://github.com/bootc-dev/bootc/pull/1727

HuijingHei: > This also relates to an issue (TODO link) where we should support preserving basically everything (including the bootloader entries on non-ostree) by default until the very end of the install. And ideally make the "commit" phase an optional step after.

I think you meant preserve everything under /boot & /boot/efi, then clean them before install bootloader, is this right?

> This one also textually conflicts with #1727

A little confused, this requires empty rootfs when `--replace` is not set, but when using `to-existing-root` it is using `ReplaceMode::Alongside` by default, maybe I misunderstood this?

HuijingHei:
Test locally using `tmt -vvvvv run --environment running_env=existing-ostree plans -n  /tmt/plans/tests-install/plan-bootc-install-on-ostree --debug`, after reboot, get error, will look more about this.
```
plan failed.

The exception was caused by 1 earlier exceptions

Cause number 1:

    execute step failed

    The exception was caused by 1 earlier exceptions

    Cause number 1:

        Failed to push workdir to the guest. This usually means that login as 'root' to the guest does not work.

        The exception was caused by 1 earlier exceptions

        Cause number 1:

            Command 'rsync --chmod=755 -s -p -e 'ssh -oForwardX11=no -oStrictHostKeyChecking=no -oUserKnownHostsFile=/dev/null -oConnectionAttempts=5 -oConnectTimeout=60 -oServerAliveInterval=5 -oServerAliveCountMax=60 -oIdentitiesOnly=yes -p10022 -i /var/tmp/tmt/run-003/tmt/plans/tests-install/plan-bootc-install-on-ostree/provision/default-0/id_ecdsa -oPasswordAuthentication=no -S/var/tmp/tmt/run-003/ssh-sockets/127.0.0.1-10022-root.socket' /var/tmp/tmt/run-003/tmt/plans/tests-install/plan-bootc-install-on-ostree/discover/default-0/tests/tmt/tests/tmt-test-wrapper-inner.sh-tmt-tests-install-test-bootc-install-on-ostree-1-default-0-default-0 root@127.0.0.1:/var/tmp/tmt/run-003/tmt/plans/tests-install/plan-bootc-install-on-ostree/discover/default-0/tests/tmt/tests/tmt-test-wrapper-inner.sh-tmt-tests-install-test-bootc-install-on-ostree-1-default-0-default-0' returned 3.

            # stderr (3 lines)
            Warning: Permanently added '[127.0.0.1]:10022' (ED25519) to the list of known hosts.
            rsync: [Receiver] change_dir#3 "/var/tmp/tmt/run-003/tmt/plans/tests-install/plan-bootc-install-on-ostree/discover/default-0/tests/tmt/tests" failed: No such file or directory (2)
            rsync error: errors selecting input/output files, dirs (code 3) at main.c(818) [Receiver=3.2.5]
```
Edit:
Had a short talk with @shi2wei3 and @henrywang , when create new deployment will use the same `stateroot`, this would use the shared `/var` instead of new `/var`.

cgwalters:
> Test locally using tmt -vvvvv run --environment running_env=existing-ostree plans -n  /tmt/plans/tests-install/plan-bootc-install-on-ostree --debug, after reboot, get error, will look more about this.

Yeah it doesn't really work to run `tmt` directly because of https://github.com/teemtee/tmt/issues/4062

We have a whole `cargo xtask run-tmt` for this reason, among others

HuijingHei:
> > Test locally using tmt -vvvvv run --environment running_env=existing-ostree plans -n  /tmt/plans/tests-install/plan-bootc-install-on-ostree --debug, after reboot, get error, will look more about this.
> 
> Yeah it doesn't really work to run `tmt` directly because of [teemtee/tmt#4062](https://github.com/teemtee/tmt/issues/4062)
> 
> We have a whole `cargo xtask run-tmt` for this reason, among others

As testing on existing FCOS/RHCOS, one easy workaround is using the same `stateroot` when install bootc, then `/var` will be shared.

Another issue is can not test uefi because of https://github.com/teemtee/tmt/issues/4203, set boot.method: uefi in the plan in `tests-install.fmf ` does not work.

Maybe I can try to use bcvk libvirt with `--bind-storage-ro` to start fedora-coreos VM, and then run boot-install, but failed, I post the log in https://cloud-native.slack.com/archives/C08SKSQKG1L/p1765525873281029, any pointer is much appreciated, thank you!

HuijingHei:
@cgwalters could you help to review again? Thank you!
The changes are:
- remove the kernel argument mount (looks no needed) for systemd
- mount esp before `clean_boot` as esp is not mounted after booted on FCOS
- add tmt tests on FCOS

Edit:
tmt test on FCOS get error, this is wired, can not reproduce locally. Not sure if this because of bindmount `-v /run/host-container-storage:/run/host-container-storage:ro`
```
podman run --rm --privileged -v /dev:/dev -v /run/host-container-storage:/run/host-container-storage:ro -v /:/target --pid=host --security-opt label=type:unconfined_t localhost/bootc-integration:latest env BOOTC_BOOTLOADER_DEBUG=1 STORAGE_OPTS=additionalimagestore=/run/host-container-storage bootc install to-existing-root --stateroot=default --skip-fetch-check --acknowledge-destructive --karg=console=ttyS0,115200n8
                        content: error: Installing to filesystem: Gathering source info from container env: Subprocess failed: ExitStatus(unix_wait_status(256))
                        content: error: readdir: Too many open files
```

Add `--ulimit nofile=524288:524288` to podman run, and hit the same error.

cgwalters:
> readdir: Too many open files

I think this is a generic virtiofs flake. There's also https://github.com/bootc-dev/bcvk/issues/157 which is a different error - but might be related, I haven't dug enough into it. You can try to restart to see if it is a flake.

HuijingHei:
Finally the test is passed (see https://github.com/bootc-dev/bootc/actions/runs/20326393875/job/58393662760), ready for reviewing now, thank you so much @cgwalters for the pointer.

Justfile
```diff
@@ -192,6 +197,19 @@ _build-upgrade-image:
 test-tmt-nobuild *ARGS:
     cargo xtask run-tmt --env=BOOTC_variant={{variant}} --upgrade-image={{integration_upgrade_img}} {{integration_img}} {{ARGS}}
 
+# Build test container image for testing on coreos with SKIP_CONFIGS=1,
+# without configs and no curl container image
+build-testimage-coreos PATH:
+    @just build-from-package {{PATH}}
```
cgwalters, line 203: Ah it looks like this is going to conflict with https://github.com/bootc-dev/bootc/pull/1864

But it's ok if this goes in first I can rebase

tmt/tests/install/test-on-ostree/test-install-on-ostree.sh
```diff
@@ -0,0 +1,57 @@
+# number: 50
+# tmt:
+#   summary: Test bootc install on ostree OS
+#   duration: 30m
+#   adjust:
+#     - when: VARIANT_ID != coreos
+#       enabled: false
+#       because: this needs to start an ostree OS firstly
```
cgwalters, line 8: Hum. But what does "ostree OS" mean here that's different from the other tests which start via `bcvk libvirt run`?

Doesn't this relate more to just the `/boot` partition setup?

If it's about having a separate `/boot` on an ostree based OS then note we should also be able to easily replicate this with Anaconda-based installs too (or bootc-image-builder).

crates/xtask/src/tmt.rs
```diff
@@ -137,6 +137,25 @@ fn build_firmware_args(sh: &Shell, image: &str) -> Result<Vec<String>> {
     Ok(r)
 }
 
+/// Detect VARIANT_ID from container image by reading os-release
+/// Returns string like "coreos" or empty
+#[context("Detecting distro from image")]
+fn detect_variantid_from_image(sh: &Shell, image: &str) -> Result<Option<String>> {
+    let variant_id = cmd!(
+        sh,
+        "podman run --rm {image} bash -c '. /usr/lib/os-release && echo $VARIANT_ID'"
```
cgwalters: Super minor nit but any time we do stuff like this we should remember to pass `--net=none` - it shaves off a bit of time on container startup.

crates/lib/src/bootloader.rs
```diff
@@ -30,6 +31,36 @@ pub(crate) fn esp_in(device: &PartitionTable) -> Result<&Partition> {
...
 
+/// Get esp partition node based on the root dir
+pub(crate) fn get_esp_partition_node(root: &Dir) -> Result<Option<String>> {
+    let device = get_sysroot_parent_dev(&root)?;
+    let base_partitions = bootc_blockdev::partitions_of(Utf8Path::new(&device))?;
+    let esp = base_partitions.find_partition_of_esp()?;
+    if let Some(esp) = esp {
+        return Ok(Some(esp.node.clone()));
+    };
+    Ok(None)
```
cgwalters: Minor but slightly more elegant perhaps as
`Ok(esp.map(|v| v.node.clone()))`

crates/lib/src/bootloader.rs
```diff
@@ -30,6 +31,36 @@ pub(crate) fn esp_in(device: &PartitionTable) -> Result<&Partition> {
...
+pub(crate) fn mount_esp_part(root: &Dir, root_path: &Utf8Path, physical_root: &Dir) -> Result<()> {
+    let efi_path = Utf8Path::new("boot").join(crate::bootloader::EFI_DIR);
+    if let Some(esp_fd) = root
+        .open_dir_optional(&efi_path)
+        .context("Opening /boot/efi")?
+    {
+        if let Some(false) = esp_fd.is_mountpoint(".")? {
+            tracing::debug!("Not a mountpoint: /boot/efi");
+            // On ostree env, should use /target/sysroot because of composefs
+            if let Some(esp_part) = get_esp_partition_node(&physical_root)? {
```
cgwalters: I prefer "let ... else { return Ok(()) };` style to avoid rightward drift for things like this (here and above)

crates/xtask/src/tmt.rs
```diff
@@ -387,7 +421,12 @@ pub(crate) fn run_tmt(sh: &Shell, args: &RunTmtArgs) -> Result<()> {
                     distro
                 );
             }
-
+            // Add --filesystem=xfs by default on fedora-coreos
```
cgwalters, line 424: OK as is but I think we should just put that in the default FCOS container image. As is done in https://gitlab.com/redhat/centos-stream/containers/bootc/-/commit/cfafc9e78a0c9cbdb016113edd6a5dd35d87ab8b
Then we can drop it here.

crates/lib/src/install.rs
```diff
@@ -2136,7 +2195,18 @@ pub(crate) async fn install_to_filesystem(
             tokio::task::spawn_blocking(move || remove_all_in_dir_no_xdev(&rootfs_fd, true))
                 .await??;
         }
-        Some(ReplaceMode::Alongside) => clean_boot_directories(&rootfs_fd, is_already_ostree)?,
+        Some(ReplaceMode::Alongside) => {
+            // On existing ostree OS like FCOS, esp is not mounted after booted,
```
cgwalters: That's not strictly true, Anaconda writes the ESP into `/boot/efi` and it is mounted by default I think.

I'd probably just drop the comment

crates/lib/src/install.rs
```diff
@@ -1924,30 +1940,61 @@ fn remove_all_in_dir_no_xdev(d: &Dir, mount_err: bool) -> Result<()> {
...
-                .open_dir_optional(crate::bootloader::EFI_DIR)
-                .context("Opening /boot/efi")?
-            {
-                remove_all_in_dir_no_xdev(&efidir, false)
-                    .context("Emptying EFI system partition")?;
-            }
+    // This should not remove /boot/efi note.
+    remove_all_except_loader_dirs(&bootdir, is_ostree).context("Emptying /boot")?;
+
+    // TODO: Discover the ESP the same way bootupd does it; we should also
```
cgwalters: I know you're just moving this existing code, but now in this same PR we are using `crate::bootloader::mount_esp_part` just below...so we could start doing the same here?

cgwalters (approved):
Thanks for the hard work on this! There's so many subtleties here. I think this looks right, but I would like to spend some time looking more carefully after break too.

tmt/tests/install/test-on-ostree/test-install-on-ostree.sh
HuijingHei: I think this is more related to esp partition and a separate `/boot` on existing ostree OS, like edge and FCOS/RHCOS.

Maybe I should use coreos here, as it is testing on FCOS. In this case, just use `bcvk libvirt run` to start fedora-coreos:testing-devel and then do testing.
I can try later if can use Anaconda, thank you for your pointer!

crates/lib/src/install.rs
HuijingHei: For Anaconda yes, but on booted FCOS, the ESP part is not mounted, and `/boot/efi` will be removed in `clean_boot_directories()`, so need to mount ESP before `clean_boot_directories()`.

crates/lib/src/install.rs
HuijingHei: Sure, that would be better.

crates/lib/src/install.rs
HuijingHei: Move `crate::bootloader::mount_esp_part()` to `clean_boot_directories()`

cgwalters:
Restarted some CI jobs here

HuijingHei:
Just trigger the tests.

HuijingHei:
Rebase the main branch to trigger the test.

I have no idea why test `test-23-install-outside-container` and `test-30-install-unified-flag` failed on F42
```
                        content: + test -d /sysroot/ostree
                        content: + mount --bind /usr/share/empty /sysroot/ostree
                        content: + mkdir -p /tmp/ovl/upper /tmp/ovl/work
                        content: + mount -t overlay -olowerdir=/usr,workdir=/tmp/ovl/work,upperdir=/tmp/ovl/upper overlay /usr
                        content: mount: /usr: wrong fs type, bad option, bad superblock on overlay, missing codepage or helper program, or other error.
```
Edit:
Looks the issue was fixed by https://github.com/bootc-dev/bootc/pull/1882, thank you @Johan-Liebert1 !

HuijingHei:
Hold on, try to find out what is actually block test install bootc image on fedora-coreos.

Edit: From logs, after reboot, ssh failed to connect the guest. But can not reproduce locally, add `ssh-option: PubkeyAcceptedAlgorithms=+ssh-rsa` to plan as workaround.

tmt/plans/tests-install.fmf
```diff
@@ -14,6 +14,8 @@ provision:
   user: root
   memory: 4096
   disk: 20
+  ssh-option:
+    - "PubkeyAcceptedAlgorithms=+ssh-rsa"
```
cgwalters, line 18: Interesting; this seems most likely to relate to changes in Fedora 43 though I don't understand why it'd only appear with FCOS.

crates/lib/src/install.rs
```diff
@@ -1122,6 +1122,8 @@ pub(crate) struct RootSetup {
     pub(crate) physical_root_path: Utf8PathBuf,
     /// Directory file descriptor for the above physical root.
     pub(crate) physical_root: Dir,
+    /// Target root path /target.
+    pub(crate) target_root_path: Option<Utf8PathBuf>,
```
cgwalters, line 1126: In general, I would like if we could go to having most things be file descriptor relative; it's unfortunate to need to add a new absolute path here.

But I understand that the bootupd install is using it today.

What I would find helpful is to understand a bit more strictly the meaing of "target" here. I think we're effectively using it to be "what the user provided as an install target" right?

And for ostree/bootc systems, that's always the mounted deployment root conventionally today when they say `-v /:/target`.

But that's just conventional - what would happen if they passed `-v /sysroot:/target` say?

The more I think about this the more I feel what we should do is detect when the passed root filesystem refers to *either* a mounted deployment root or the physical root, and automatically canonicalize internally to "physical root" and "deployment root" (which for non-ostree systems are the same thing), and then discard the user provided input path.

crates/lib/src/install.rs
```diff
@@ -2142,7 +2156,8 @@ pub(crate) async fn install_to_filesystem(
     };
 
     // Get a file descriptor for the root path
-    let rootfs_fd = {
+    // It will be /target/sysroot on ostree OS, or will be /target
+    let rootfs_fd = if is_already_ostree {
```
cgwalters, line 2160: While I know you're just changing existing code, I think it'd be cleaner to fold this conditional with the one above.

I had to read this whole logic a few times to notice that a key bit here is we're *mutating* the user input `fsopts.root_path = possible_physical_root` just above, and that's an important aspect.

crates/lib/src/install.rs
```diff
@@ -2218,9 +2237,10 @@ pub(crate) async fn install_to_filesystem(
     };
     // Find the UUID of /boot because we need it for GRUB.
     let boot_uuid = if boot_is_mount {
-        let boot_path = fsopts.root_path.join(BOOT);
+        let boot_path = target_root_path.join(BOOT);
```
cgwalters, line 2240: This seems to be a key change here. I guess we needed this because in the ostree case we'd mutated `fsopts.root_path` to refer to the physical root, and when `/boot` is actually a separate mount point that was broken on ostree setups?

But this relates to what I mean above - we could instead just know to look at the deployment's `/boot` explicitly.

crates/lib/src/install.rs
HuijingHei: > What I would find helpful is to understand a bit more strictly the meaing of "target" here. I think we're effectively using it to be "what the user provided as an install target" right?
> 
> And for ostree/bootc systems, that's always the mounted deployment root conventionally today when they say -v /:/target.
> 
> But that's just conventional - what would happen if they passed -v /sysroot:/target say?
> 
> The more I think about this the more I feel what we should do is detect when the passed root filesystem refers to either a mounted deployment root or the physical root, and automatically canonicalize internally to "physical root" and "deployment root" (which for non-ostree systems are the same thing), and then discard the user provided input path.

This sounds better and more sane, will try this later.

HuijingHei: > what would happen if they passed -v /sysroot:/target say?

We still need `/boot/efi` (or separate `/boot`) to update the bootloader, if we pass `-v /sysroot:/target`, then how could we get the "physical root" like `/` (to get `/boot/efi`)?

crates/lib/src/install.rs
HuijingHei: > This seems to be a key change here. I guess we needed this because in the ostree case we'd mutated `fsopts.root_path` to refer to the physical root, and when `/boot` is actually a separate mount point that was broken on ostree setups?

I am a little confused, do you mean "physical root" is `/target`, and "deployment root" is `/target/sysroot`?

From following log (copy from bug), we passed `/target/sysroot`, but we should pass `/target`.
```
> bootupctl backend install --write-uuid -vvvv --update-firmware --auto --device /dev/vda /target/sysroot
...
[DEBUG bootupd::efi] Mounted at "/target/sysroot/boot"
...
Could not find /boot/efi/EFI when installing "redhat/grub.cfg"
```

And I do not understand `/sysroot/boot` and `/boot` correctly, are they the same?

cgwalters: For non-bootc/ostree setups, "physical root == root", so the user always passes `-v /:/target` - in other words `/target` inside there is always the filesystem root.

For bootc/ostree setups, they are not the same. What I am saying though is we should support being passed *either* thing (`/sysroot` or `/`) for `/target` and automatically do the right thing, and resolve the right roots.

> And I do not understand /sysroot/boot and /boot correctly, are they the same?

Typically we only mount the separate boot partition in the target root, not the physical root.

HuijingHei: Understood, thank you for the detail info.

> For bootc/ostree setups, they are not the same. What I am saying though is we should support being passed either thing (/sysroot or /) for /target and automatically do the right thing, and resolve the right roots.

Try to catch your meaning, for bootc/ostree setups, if passed `-v /sysroot:/target`, my concern is how to get ESP part, but looks can get `/target` part, then get its parent disk, then get separate boot part if there is (maybe need to mount it to `/target/boot`?), and ESP.
```
podman run --rm -it --privileged -v /dev:/dev -v /var/lib/containers:/var/lib/containers -v /sysroot:/target \
--pid=host --security-opt label=type:unconfined_t \
quay.io/fedora/fedora-bootc:rawhide bash
bash-5.3# cd /target
bash-5.3# findmnt .
TARGET  SOURCE    FSTYPE OPTIONS
/target /dev/vda4 xfs    rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,prjquota
```

HuijingHei: Will try this later.

PR #1761, opened 2025-11-13
More kargs improvements and hoisting parsing into clap

jeckersb:
- **kernel_cmdline: Add some more derives for Cmdline**
- **install: Hoist InstallResetOpts kargs parsing up into clap**
- **kernel_cmdline: Standardize Deref/AsRef impls across types**
- **Use Cmdline for kargs fields in InstallConfigOpts and RootSetup**
- **kernel_cmdline: impl PartialOrd/Ord/PartialEq/Eq for Parameter/ParameterKey**
- **kernel_cmdline: Manually impl semantically-correct equivalence for Cmdline**

### DIFF
--- a/crates/kernel_cmdline/src/bytes.rs
+++ b/crates/kernel_cmdline/src/bytes.rs
@@ -17,7 +17,7 @@ use serde::{Deserialize, Serialize};
 /// Wraps the raw command line bytes and provides methods for parsing and iterating
 /// over individual parameters. Uses copy-on-write semantics to avoid unnecessary
 /// allocations when working with borrowed data.
-#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
 pub struct Cmdline<'a>(Cow<'a, [u8]>);
 
 /// An owned Cmdline.  Alias for `Cmdline<'static>`.
@@ -374,6 +374,19 @@ impl<'a, 'other> Extend<Parameter<'other>> for Cmdline<'a> {
         }
     }
 }
+
+impl PartialEq for Cmdline<'_> {
+    fn eq(&self, other: &Self) -> bool {
+        let mut our_params = self.iter().collect::<Vec<_>>();
+        our_params.sort();
+        let mut their_params = other.iter().collect::<Vec<_>>();
+        their_params.sort();
+
+        our_params == their_params
+    }
+}
+
+impl Eq for Cmdline<'_> {}
 
 /// A single kernel command line parameter key
 ///
@@ -1090,5 +1103,25 @@ mod tests {
         let params: Vec<_> = kargs.iter_bytes().collect();
 
         assert_eq!(params.len(), 0);
+    }
+
+    #[test]
+    fn test_cmdline_eq() {
+        // Ordering, quoting, and the whole dash-underscore
+        // equivalence thing shouldn't affect whether these are
+        // semantically equal
+        assert_eq!(
+            Cmdline::from("foo bar-with-delim=\"with spaces\""),
+            Cmdline::from("\"bar_with_delim=with spaces\" foo")
+        );
+
+        // Uneven lengths are not equal even if the parameters are. Or
+        // to put it another way, duplicate parameters break equality.
+        // Check with both orderings.
+        assert_ne!(Cmdline::from("foo"), Cmdline::from("foo foo"));
+        assert_ne!(Cmdline::from("foo foo"), Cmdline::from("foo"));
+
+        // Equal lengths but differing duplicates are also not equal
+        assert_ne!(Cmdline::from("a a b"), Cmdline::from("a b b"));
     }
 }
--- a/crates/kernel_cmdline/src/utf8.rs
+++ b/crates/kernel_cmdline/src/utf8.rs
@@ -928,4 +928,24 @@ mod tests {
         assert_eq!(params[1], param("baz=qux"));
         assert_eq!(params[2], param("wiz"));
     }
+
+    #[test]
+    fn test_cmdline_eq() {
+        // Ordering, quoting, and the whole dash-underscore
+        // equivalence thing shouldn't affect whether these are
+        // semantically equal
+        assert_eq!(
+            Cmdline::from("foo bar-with-delim=\"with spaces\""),
+            Cmdline::from("\"bar_with_delim=with spaces\" foo")
+        );
+
+        // Uneven lengths are not equal even if the parameters are. Or
+        // to put it another way, duplicate parameters break equality.
+        // Check with both orderings.
+        assert_ne!(Cmdline::from("foo"), Cmdline::from("foo foo"));
+        assert_ne!(Cmdline::from("foo foo"), Cmdline::from("foo"));
+
+        // Equal lengths but differing duplicates are also not equal
+        assert_ne!(Cmdline::from("a a b"), Cmdline::from("a b b"));
+    }
 }


crates/kernel_cmdline/src/bytes.rs
```diff
@@ -9,13 +9,14 @@ use std::ops::Deref;
...
 use anyhow::Result;
+use serde::{Deserialize, Serialize};
 
 /// A parsed kernel command line.
 ///
 /// Wraps the raw command line bytes and provides methods for parsing and iterating
 /// over individual parameters. Uses copy-on-write semantics to avoid unnecessary
 /// allocations when working with borrowed data.
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
```
jeckersb: TODO here, deriving `Eq` like this means the two `Cmdline`s have to be byte-for-byte identical but really what we should do is iter over `Parameter` and just make sure there's a match in the other.  Then two that are semantically identical but textually different would still be equal.  So these should be the same:

`foo bar-with-delim="with spaces"` == `"bar_with_delim=with spaces" foo`

crates/kernel_cmdline/src/utf8.rs
```diff
@@ -8,13 +8,14 @@ use std::ops::Deref;
...
 use anyhow::Result;
+use serde::{Deserialize, Serialize};
 
 /// A parsed UTF-8 kernel command line.
 ///
 /// Wraps the raw command line bytes and provides methods for parsing and iterating
 /// over individual parameters. Uses copy-on-write semantics to avoid unnecessary
 /// allocations when working with borrowed data.
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
```
jeckersb, line 18: Same TODO here as above.

crates/lib/src/parsers/bls_config.rs
```diff
@@ -15,7 +15,7 @@ use uapi_version::Version;
 
 use crate::composefs_consts::COMPOSEFS_CMDLINE;
 
-#[derive(Debug, PartialEq, PartialOrd, Eq, Default)]
```
jeckersb, line 18: I want to specifically call out that I removed this `PartialOrd` derive, I'm not sure why it was there and I don't think it makes sense to apply an ordering to this type.  But it was blocking me from switching to `Cmdline` here since `Cmdline` does not `impl PartialOrd`.  I *could* add it but I don't think it makes sense.

crates/kernel_cmdline/src/utf8.rs
jeckersb: Actually scratch that, this doesn't need any changes since it just wraps the `bytes` version so fixing it there is enough to make the derived `PartialEq`/`Eq` work just fine here.

crates/lib/src/install.rs
```diff
@@ -942,28 +941,35 @@ async fn install_container(
...
-        .chain(kargsd)
-        .chain(state.config_opts.karg.iter().flatten().map(|v| v.as_str()))
-        .collect::<Vec<_>>();
+    let mut kargs = Cmdline::new();
+
+    kargs.extend(&root_setup.kargs);
+
+    if let Some(install_config_kargs) = install_config_kargs {
+        let install_config_kargs = Cmdline::from(install_config_kargs.join(" "));
+        kargs.extend(&install_config_kargs);
```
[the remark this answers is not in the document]
jeckersb: Sure although these both probably perform about the same.

crates/kernel_cmdline/src/bytes.rs
```diff
@@ -369,16 +374,31 @@ impl<'a, 'other> Extend<Parameter<'other>> for Cmdline<'a> {
...
+    fn eq(&self, other: &Self) -> bool {
+        let our_params = self.iter().collect::<Vec<_>>();
+        let their_params = other.iter().collect::<Vec<_>>();
+
+        if our_params.len() != their_params.len() {
+            return false;
+        }
+
+        our_params.iter().all(|p| their_params.contains(p))
+    }
```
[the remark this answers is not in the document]
jeckersb: Good bot :robot:

crates/lib/src/parsers/bls_config.rs
jeckersb: Actually maybe it would make sense.  I could see adding a `canonicalize()` method to `Cmdline` which would sort the params and render them out with some opinionated quoting :thinking:

Edit: Or at least make sense for `Cmdline`... still not sure it makes sense here for `BLSConfigType` but if I end up adding it for Cmdline I'll just put it back for now.

crates/kernel_cmdline/src/bytes.rs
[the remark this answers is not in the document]
cgwalters: Yeah, impressive for sure.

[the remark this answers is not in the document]
jeckersb: Ok I went ahead and made `Parameter`/`ParameterKey` impl `Ord` so we could just sort them and compare that way.  A bit more efficient.

crates/kernel_cmdline/src/bytes.rs
```diff
@@ -399,32 +418,42 @@ impl<'a, T: AsRef<[u8]> + ?Sized> From<&'a T> for ParameterKey<'a> {
     }
 }
 
+impl ParameterKey<'_> {
+    /// Returns an iterator over the canonicalized bytes of the
+    /// parameter, with dashes turned into underscores.
+    fn iter(&self) -> impl Iterator<Item = u8> + use<'_> {
```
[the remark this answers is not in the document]
jeckersb: Nah the compiler straight told me to use that syntax (which I don't think I've seen before, so that was a bit of a learning experience).  See https://doc.rust-lang.org/std/keyword.use.html#precise-capturing.  I think if we mandated 2024 edition we could remove that but we need it for 2021-compat.

crates/lib/src/install.rs
```diff
@@ -942,28 +941,36 @@ async fn install_container(
...
-        .chain(kargsd)
-        .chain(state.config_opts.karg.iter().flatten().map(|v| v.as_str()))
-        .collect::<Vec<_>>();
+    let mut kargs = Cmdline::new();
+
+    kargs.extend(&root_setup.kargs);
+
+    if let Some(install_config_kargs) = install_config_kargs {
+        for karg in install_config_kargs {
+            kargs.extend(&Cmdline::from(karg.as_str()));
```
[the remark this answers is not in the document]
jeckersb: `karg` is *not* already a `Cmdline`, `install_config_kargs` is a `Vec<String>` so `karg` is `&String` here.

:-1:

crates/lib/src/parsers/bls_config.rs
jeckersb: Well down below I added `PartialOrd`/`Ord` for `Parameter` which makes sense in the context of comparing two `Cmdline`s by sorting their `Parameter`s.  But I still don't think it makes sense for `Cmdline` itself to have ordering.

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -407,12 +412,10 @@ pub(crate) fn setup_composefs_bls_boot(
...
-                [
-                    format!("root=UUID={}", this_arch_root()),
-                    RW_KARG.to_string(),
-                    format!("{COMPOSEFS_CMDLINE}={id_hex}"),
-                ]
-                .join(" "),
+                Cmdline::from(format!(
+                    "root=UUID={} {RW_KARG} {COMPOSEFS_CMDLINE}={id_hex}",
+                    this_arch_root()
+                )),
```
cgwalters, line 418: I think it'd be cleaner for this to remain split up kargs instead of being parsed from a string again.

jeckersb: It's all just `Vec<u8>` under the covers, it doesn't really get parsed again.  The only "parsing" is when it hands out slices via the various iter methods.

Issue #1760, opened 2025-11-13
Soft reboot: Also detect SELinux policy deltas


cgwalters:
SELinux policy is not reloaded across soft reboots, so we need to detect this and error out if a soft reboot is requested.

PR #1768, opened 2025-11-14
Soft reboot: Detect SELinux policy deltas

gursewak1997:
Add check to prevent soft reboot when SELinux policies differ between booted and target deployments, since policy is not reloaded across soft reboots.

Fixes: https://github.com/bootc-dev/bootc/issues/1760

### DIFF
--- a/crates/lib/src/status.rs
+++ b/crates/lib/src/status.rs
@@ -93,7 +93,45 @@ impl From<ImageReference> for OstreeImageReference {
     }
 }
 
+/// Check if SELinux policies are compatible between booted and target deployments.
+/// Returns false if SELinux is enabled and the policies differ or have mismatched presence.
+fn check_selinux_policy_compatible(
+    sysroot: &SysrootLock,
+    booted_deployment: &ostree::Deployment,
+    target_deployment: &ostree::Deployment,
+) -> Result<bool> {
+    // Only check if SELinux is enabled
+    if !crate::lsm::selinux_enabled()? {
+        return Ok(true);
+    }
+
+    let booted_fd = crate::utils::deployment_fd(sysroot, booted_deployment)
+        .context("Failed to get file descriptor for booted deployment")?;
+    let booted_policy = crate::lsm::new_sepolicy_at(&booted_fd)
+        .context("Failed to load SELinux policy from booted deployment")?;
+    let target_fd = crate::utils::deployment_fd(sysroot, target_deployment)
+        .context("Failed to get file descriptor for target deployment")?;
+    let target_policy = crate::lsm::new_sepolicy_at(&target_fd)
+        .context("Failed to load SELinux policy from target deployment")?;
+
+    let booted_csum = booted_policy.and_then(|p| p.csum());
+    let target_csum = target_policy.and_then(|p| p.csum());
+
+    match (booted_csum, target_csum) {
+        (None, None) => Ok(true), // Both absent, compatible
+        (Some(_), None) | (None, Some(_)) => {
+            // Incompatible: one has policy, other doesn't
+            Ok(false)
+        }
+        (Some(booted_csum), Some(target_csum)) => {
+            // Both have policies, checksums must match
+            Ok(booted_csum == target_csum)
+        }
+    }
+}
+
 /// Check if a deployment has soft reboot capability
+// TODO: Lower SELinux policy check into ostree's deployment_can_soft_reboot API
 fn has_soft_reboot_capability(sysroot: &SysrootLock, deployment: &ostree::Deployment) -> bool {
     if !ostree_ext::systemd_has_soft_reboot() {
         return false;
@@ -113,7 +151,22 @@ fn has_soft_reboot_capability(sysroot: &SysrootLock, deployment: &ostree::Deploy
         return false;
     }
 
-    sysroot.deployment_can_soft_reboot(deployment)
+    if !sysroot.deployment_can_soft_reboot(deployment) {
+        return false;
+    }
+
+    // Check SELinux policy compatibility with booted deployment
+    // Block soft reboot if SELinux policies differ, as policy is not reloaded across soft reboots
+    if let Some(booted_deployment) = sysroot.booted_deployment() {
+        // deployment_fd should not fail for valid deployments
+        if !check_selinux_policy_compatible(sysroot, &booted_deployment, deployment)
+            .expect("deployment_fd should not fail for valid deployments")
+        {
+            return false;
+        }
+    }
+
+    true
 }
 
 /// Parse an ostree origin file (a keyfile) and extract the targeted
--- a/tmt/tests/booted/test-soft-reboot-selinux-policy.nu
+++ b/tmt/tests/booted/test-soft-reboot-selinux-policy.nu
@@ -1,0 +1,104 @@
+# Verify that soft reboot is blocked when SELinux policies differ
+use std assert
+use tap.nu
+
+let soft_reboot_capable = "/usr/lib/systemd/system/soft-reboot.target" | path exists
+if not $soft_reboot_capable {
+    echo "Skipping, system is not soft reboot capable"
+    return
+}
+
+# Check if SELinux is enabled
+let selinux_enabled = "/sys/fs/selinux/enforce" | path exists
+if not $selinux_enabled {
+    echo "Skipping, SELinux is not enabled"
+    return
+}
+
+# This code runs on *each* boot.
+bootc status
+
+# Run on the first boot
+def initial_build [] {
+    tap begin "Build base image and test soft reboot with SELinux policy change"
+
+    let td = mktemp -d
+    cd $td
+
+    bootc image copy-to-storage
+
+    # Create a derived container that installs a custom SELinux policy module
+    # Installing a policy module will change the compiled policy checksum
+    # Following Colin's suggestion and the composefs-rs example
+    # We create a minimal policy module and install it
+    "FROM localhost/bootc
+# Install tools needed to build and install SELinux policy modules
+RUN dnf install -y selinux-policy-devel checkpolicy policycoreutils
+
+# Create a minimal SELinux policy module that will change the policy checksum
+# We install it to ensure it's part of the deployment filesystem
+RUN <<EORUN
+    set -eux
+    mkdir -p /tmp/bootc-test-policy
+    cd /tmp/bootc-test-policy
+    echo 'module bootc_test_policy 1.0;' > bootc_test_policy.te
+    echo 'require {' >> bootc_test_policy.te
+    echo '    type unconfined_t;' >> bootc_test_policy.te
+    echo '    class file { read write };' >> bootc_test_policy.te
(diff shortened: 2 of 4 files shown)


crates/lib/src/cli.rs
```diff
@@ -763,6 +763,51 @@ fn has_soft_reboot_capability(deployment: Option<&crate::spec::BootEntry>) -> bo
...
+    // Load SELinux policy from booted deployment
+    let booted_fd = crate::utils::deployment_fd(sysroot, booted_deployment)?;
+    let booted_policy = crate::lsm::new_sepolicy_at(&booted_fd)?;
+
+    // Load SELinux policy from target deployment
+    let target_fd = crate::utils::deployment_fd(sysroot, target_deployment)?;
+    let target_policy = crate::lsm::new_sepolicy_at(&target_fd)?;
+
+    // If either deployment doesn't have a policy, we can't compare
+    // In this case, we'll allow soft reboot (policy will be loaded on boot)
```
cgwalters: I don't think that's true, we should not allow soft reboot in the degenerate case that selinux is enabled in one but not the other.

crates/lib/src/cli.rs
```diff
@@ -763,6 +763,51 @@ fn has_soft_reboot_capability(deployment: Option<&crate::spec::BootEntry>) -> bo
     deployment.map(|d| d.soft_reboot_capable).unwrap_or(false)
 }
 
+/// Check if SELinux policy differs between booted and target deployments.
+/// Returns an error if SELinux is enabled and the policies differ.
+#[context("Checking SELinux policy compatibility with soft reboot")]
+fn check_selinux_policy_for_soft_reboot(
```
cgwalters: This helps, but I think what we really want is to change `has_soft_reboot_capability`.
 
But then that raises the next thing in that in theory this change should go in libostree which has a `deployment_can_soft_reboot` API which already does checks like this.

OTOH, I don't mind filling it in here either, we can just have a `// TODO lower into ostree` to start.

The other thing of course related to this is that when we get around to soft reboots + composefs, we'll need similar logic there.

crates/lib/src/cli.rs
```diff
@@ -763,6 +763,51 @@ fn has_soft_reboot_capability(deployment: Option<&crate::spec::BootEntry>) -> bo
...
+
+    if booted_csum != target_csum {
+        anyhow::bail!(
+            "SELinux policy differs between booted and target deployments (booted: {:?}, target: {:?}). \
+             Soft reboot cannot be used because SELinux policy is not reloaded across soft reboots. \
+             Please use a full reboot instead.",
+            booted_csum,
+            target_csum
+        );
+    }
```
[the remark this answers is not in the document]
gursewak1997: fixed

crates/lib/src/cli.rs
gursewak1997: Made some changes pertaining to this.

crates/lib/src/status.rs
```diff
@@ -93,7 +93,56 @@ impl From<ImageReference> for OstreeImageReference {
...
+) -> bool {
+    // Only check if SELinux is enabled
+    let Ok(selinux_enabled) = crate::lsm::selinux_enabled() else {
+        return true; // If we can't determine, assume compatible
+    };
+    if !selinux_enabled {
+        return true;
+    }
+
+    let Ok(booted_fd) = crate::utils::deployment_fd(sysroot, booted_deployment) else {
```
cgwalters: I don't expect this operation to ever fail, I don't think we should "swallow errors" in a case like this.

crates/lib/src/status.rs
```diff
@@ -93,7 +93,56 @@ impl From<ImageReference> for OstreeImageReference {
...
+        return false; // Can't check, be conservative
+    };
+    let Ok(booted_policy) = crate::lsm::new_sepolicy_at(&booted_fd) else {
```
cgwalters: This one...hmmm...it'd be quite surprising to me if we failed to init the selinux policy - and if we keep the code as is, failure to do so should at least use our `log_err_or_else` helper (which hmm I guess could nowadays e.g. at least always log to the systemd journal?).

I would say though we should just keep using the `?` here too.

crates/lib/src/status.rs
```diff
@@ -93,7 +93,56 @@ impl From<ImageReference> for OstreeImageReference {
...
+
+    match (booted_policy, target_policy) {
+        (None, None) => true, // Both absent, compatible
+        (Some(_), None) | (None, Some(_)) => {
+            // Incompatible: one has policy, other doesn't
+            false
+        }
+        (Some(booted), Some(target)) => {
+            // Both have policies, checksums must match
+            // SAFETY: new_sepolicy_at filters out policies without checksums
```
cgwalters: Yes though I think we could do
```
let booted_policy = booted_policy.map(|p| p.csum());
let target_policy = target_policy(|p| p.csum());
```

And then we just compare those two

tmt/tests/booted/test-soft-reboot-selinux-policy.nu
```diff
@@ -0,0 +1,115 @@
...
+    bootc image copy-to-storage
+
+    # Create a derived container that injects a local SELinux policy module
+    # This modifies the policy in a way that changes the policy checksum
+    # Following Colin's suggestion: inject a local selinux policy module
+    "FROM localhost/bootc
+# Inject a local SELinux policy change by modifying file_contexts
+# This will change the policy checksum between deployments
+RUN mkdir -p /opt/bootc-test-selinux-policy && \
+    echo '/opt/bootc-test-selinux-policy /opt/bootc-test-selinux-policy' >> /etc/selinux/targeted/contexts/files/file_contexts.subs_dist || true
```
cgwalters: Why the `|| true`? In general I don't think we should ignore errors

tmt/tests/booted/test-soft-reboot-selinux-policy.nu
```diff
@@ -0,0 +1,115 @@
...
+    let st = bootc status --json | from json
+    assert ($st.status.booted.image.name | str contains "bootc-derived-policy")
+    
+    # Now create another derived image with the SAME policy (no changes)
+    let td = mktemp -d
+    cd $td
+    
+    bootc image copy-to-storage
+    
+    # Create a derived container that doesn't change the policy
```
cgwalters: We're already covering the same-policy scenario in most other tests, I don't think we need to do this.

gursewak1997:
The test fails because modifying `file_contexts.local` doesn't change the compiled policy checksum that ostree uses. The checksum comes from the compiled policy binary, not source files. We need a different approach to change the policy checksum (e.g., installing a policy module or using different base images). 
The implementation in `status.rs` is correct afaik

cgwalters:
> The test fails because modifying file_contexts.local doesn't change the compiled policy checksum that ostree uses. The checksum comes from the compiled policy binary, not source files. We need a different approach to change the policy checksum (e.g., installing a policy module or using different base images).

Right let's install a policy module then. 

https://github.com/containers/composefs-rs/blob/0f636031a1ec81cdd9e7f674909ef6b75c2642cb/examples/uki/Containerfile#L49 
is one example of this

tmt/tests/booted/test-soft-reboot-selinux-policy.nu
```diff
@@ -0,0 +1,102 @@
...
+    # Installing a policy module will change the compiled policy checksum
+    # Following Colin's suggestion and the composefs-rs example
+    # We create a minimal policy module and install it
+    "FROM localhost/bootc
+# Install tools needed to build and install SELinux policy modules
+RUN dnf install -y selinux-policy-devel checkpolicy policycoreutils
+
+# Create a minimal SELinux policy module that will change the policy checksum
+# We install it to ensure it's part of the deployment filesystem
+RUN mkdir -p /tmp/bootc-test-policy && \\
```
cgwalters: This one cries out to be a heredoc via `RUN <<EORUN` like we do in other places and we can drop all the `&& \\`

tmt/tests/booted/test-soft-reboot-selinux-policy.nu
```diff
@@ -0,0 +1,102 @@
...
+    echo '    class file { read write };' >> bootc_test_policy.te && \\
+    echo '}' >> bootc_test_policy.te && \\
+    echo 'type bootc_test_t;' >> bootc_test_policy.te && \\
+    checkmodule -M -m -o bootc_test_policy.mod bootc_test_policy.te && \\
+    semodule_package -o bootc_test_policy.pp -m bootc_test_policy.mod && \\
+    semodule -i bootc_test_policy.pp && \\
+    # Ensure the policy module is in /usr/etc/selinux for the deployment
+    mkdir -p /usr/etc/selinux/targeted/modules/active/modules && \\
+    if [ -d /etc/selinux/targeted/modules/active/modules ]; then \\
+        cp -a /etc/selinux/targeted/modules/active/modules/* /usr/etc/selinux/targeted/modules/active/modules/ 2>/dev/null || true; \\
```
cgwalters: I think "error suppression and masking" via things like `2>/dev/null || true` need justification.

But all of this stuff related to `/usr/etc` just looks wrong/unnecessary to me; I think the `semodule -i` should be sufficient.

BTW a `bootc container lint --fatal-warnings` at the end of this build should have errored out due to having both `/etc` and `/usr/etc`

tmt/tests/booted/test-soft-reboot-selinux-policy.nu
```diff
@@ -0,0 +1,102 @@
...
+    assert ($st.status.staged != null) "Expected staged deployment to exist"
+    
+    # The staged deployment should NOT be soft-reboot capable because policies differ
+    assert (not $st.status.staged.softRebootCapable) "Expected soft reboot to be blocked due to SELinux policy difference, but softRebootCapable is true"
+    
+    # Verify soft reboot preparation didn't happen
+    assert (not ("/run/nextroot" | path exists)) "Soft reboot should not be prepared when policies differ"
+    
+    # Reset and do a full reboot instead
+    ostree admin prepare-soft-reboot --reset
```
cgwalters: I think this one should be unnecessary as we shouldn't have queued any soft reboot state in the first place

cgwalters:
I think this one needs a rebase on main to fix CI

tmt/tests/booted/test-soft-reboot-selinux-policy.nu
```diff
@@ -0,0 +1,104 @@
...
+EORUN
+" | save Dockerfile
+    
+    # Build the derived image
+    podman build --quiet -t localhost/bootc-derived-policy .
+    
+    # Verify soft reboot preparation hasn't happened yet
+    assert (not ("/run/nextroot" | path exists))
+    
+    # Try to soft reboot - this should fail because policies differ
```
cgwalters, line 68: I would say "Ask for a soft reboot, but we won't get one because policies differ"

gursewak1997:
`testing-farm:centos-stream-10-aarch64` consistently fails to run the tests.

PR #1769, opened 2025-11-14
initramfs: Mount /sysroot readonly for composefs by default

cgwalters:
This implements readonly mounting of /sysroot for composefs systems, matching the behavior that ostree systems already have. Previously, composefs left /sysroot mounted read-write, which was inconsistent and meant the readonly tests had to be skipped for composefs.

The implementation uses a direct `libc::syscall` wrapper for `mount_setattr` since rustix doesn't yet provide this API. The `MOUNT_ATTR_RDONLY` flag is applied recursively to three mount points during initramfs setup:
- The composefs rootfs image mount (becomes `/` after switch-root)
- The test root filesystem mount (used in testing scenarios)
- The sysroot clone mount (becomes `/sysroot` in the booted system)

With this change, the readonly /sysroot tests in test-status.nu now run for both ostree and composefs systems without conditional checks.

### DIFF
--- a/crates/lib/src/store/mod.rs
+++ b/crates/lib/src/store/mod.rs
@@ -35,7 +35,7 @@ use crate::bootc_composefs::status::{composefs_booted, ComposefsCmdline};
 use crate::lsm;
 use crate::podstorage::CStorage;
 use crate::spec::ImageStatus;
-use crate::utils::deployment_fd;
+use crate::utils::{deployment_fd, open_dir_remount_rw};
 
 /// See https://github.com/containers/composefs-rs/issues/159
 pub type ComposefsRepository =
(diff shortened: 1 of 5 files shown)


Johan-Liebert1:
We would need to update this for `bootc update` and `bootc switch` as they do write to `/sysroot`

cgwalters:
> We would need to update this for bootc update and bootc switch as they do write to /sysroot

We should now be consistently remounting `/sysroot` writable in a private mountns for both ostree and composefs paths. I'll double check this though.

cgwalters:
> We should now be consistently remounting /sysroot writable in a private mountns for both ostree and composefs paths. I'll double check this though.

Yep! This is covered by https://github.com/bootc-dev/bootc/commit/c0b9cde666b0b4ea8ba55668fabd044a8f393f4d

Johan-Liebert1:
Hmm... I still get a `Read-only file system (os error 30)` while switching

```
error: Switching: Composefs Switching: Pulling composefs repository: Pulling composefs repo: Unable to pull container image docker://192.168.122.1:5000/bootc-bls-upgrade: Failed to pull confi
g Descriptor { media_type: ImageConfig, digest: Digest { algorithm: Sha256, value: "sha256:47a7ba74f0fb105fc620cff9f2fc3ff3f823f5dca93d056c311751a0414e645a", split: 6 }, size: 13403, urls: No
ne, annotations: None, platform: None, artifact_type: None, data: None }: Read-only file system (os error 30)

```

Johan-Liebert1:
We don't really remount `/sysroot` as `rw` in the composefs case. Only in the ostree case we call `ostree_sysroot_lock` which remounts `/sysroot` and `/boot` as `rw`. In composefs case we probably don't need a lock as composefs-rs does have an advisory lock via `flock` for all repository related operations, so I think just remounting as RW should be enough

crates/lib/src/store/mod.rs
```diff
@@ -147,20 +147,25 @@ impl BootedStorage {
...
                 let mut composefs = ComposefsRepository::open_path(&physical_root, COMPOSEFS)?;
                 if cmdline.insecure {
                     composefs.set_insecure(true);
                 }
                 let composefs = Arc::new(composefs);
 
+                // Remount /sysroot rw only if we are in a new mount ns
+                if env.needs_mount_namespace() {
+                    physical_root = open_dir_remount_rw(&physical_root, ".".into())?;
+                }
```
cgwalters: I think we can move this up, like
```
let physical_root = {
  let d =  Dir::open_ambient_dir("/sysroot", cap_std::ambient_authority())
            .context("Opening /sysroot")?;
  open_dir_remount_rw(&d ".".into())?
};
```

above or so?

Issue #1771, opened 2025-11-15
Discoverable partition UUIDs misused


gucci-on-fleek:
There are (at least) 3 different types of UUIDs for Linux filesystems/partitions:

1. The filesystem UUID. This is intrinsic to the filesystem, so it can exist with both MBR and GPT, and its format depends on the specific filesystem (8 hex chars on FAT, a GUID for most other filesystems). You can mount a disk by its filesystem UUID by running `mount UUID=<uuid>`, and you can display a filesystem's UUID by running `lsblk --output=name,fstype,uuid`. Theoretically, no 2 disks should ever have the same filesystem UUIDs.

2. The partition UUID. This is a part of GPT, so it has the same format for every filesystem, but it doesn't work with MBR. You can mount a disk by its partition UUID by running `mount PARTUUID=<uuid>`, and you can display a partitions's UUID by running `lsblk --output=name,fstype,partuuid`. Theoretically, no 2 disks should ever have the same partition UUIDs.

3. The partition type UUID. This is a part of GPT, so it has the same format for every filesystem, but it doesn't work with MBR. You can mount a disk by its partition type UUID with [`systemd-gpt-auto-generator`](https://www.freedesktop.org/software/systemd/man/latest/systemd-gpt-auto-generator.html), and you can display a partitions's type UUID by running `lsblk --output=name,fstype,parttype` (or `…,parttypename` for its friendly name). Many disks will have the same partition type UUIDs, and common values are defined by the [Discoverable Partitions Specification](https://uapi-group.org/specifications/specs/discoverable_partitions_specification/#defined-partition-type-uuids).

Right now, `bootc` is setting the filesystem UUIDs to one of the values defined by the DPS:

https://github.com/bootc-dev/bootc/blob/71dc8e52d3697155c1f0b668313a00b555ce78f1/crates/lib/src/bootc_composefs/boot.rs#L415-L418

https://github.com/bootc-dev/bootc/blob/71dc8e52d3697155c1f0b668313a00b555ce78f1/crates/lib/src/install/baseline.rs#L116-L134

This is bad because 

1. This isn't actually following the DPS, so tools like `systemd-gpt-auto-generator` won't recognize this. 

2. Some software assumes that filesystem/partition UUIDs are truly unique, so bad things will happen if you try to mount 2 disks formatted this way at the same time.

Here's an example of the disk generated by `bootc install to-disk` (using v1.10):

```console
[...]

PR #1772, opened 2025-11-15
install: Fix DPS support

cgwalters:
This fixes bootc's use of the Discoverable Partition Specification (DPS) to properly support systemd-gpt-auto-generator. Previously, bootc was incorrectly setting filesystem UUIDs to the DPS partition type UUID value, which caused UUID collisions and prevented proper DPS functionality.

It's still a TODO on our side to support systemd-repart in this flow.

Note we go back to using random filesystem UUIDs with this, but per above we should likely reinitialize them on boot via repart.

Note we remove root= parameter from kernel cmdline for composefs sealed images, allowing systemd-gpt-auto-generator to auto-discover the root partition and we test this.

Fixes: #1771

### DIFF
--- a/crates/lib/src/bootc_composefs/boot.rs
+++ b/crates/lib/src/bootc_composefs/boot.rs
@@ -51,7 +51,6 @@ use crate::{
         BOOT_LOADER_ENTRIES, COMPOSEFS_CMDLINE, ORIGIN_KEY_BOOT, ORIGIN_KEY_BOOT_DIGEST,
         STAGED_BOOT_LOADER_ENTRIES, STATE_DIR_ABS, USER_CFG, USER_CFG_STAGED,
     },
-    discoverable_partition_specification::this_arch_root,
     install::RW_KARG,
     spec::{Bootloader, Host},
 };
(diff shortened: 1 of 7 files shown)


Johan-Liebert1 (approved):

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -412,10 +411,7 @@ pub(crate) fn setup_composefs_bls_boot(
             (
                 Utf8PathBuf::from("/sysroot"),
                 get_esp_partition(&sysroot_parent)?.0,
-                Cmdline::from(format!(
-                    "root=UUID={} {RW_KARG} {COMPOSEFS_CMDLINE}={id_hex}",
```
Johan-Liebert1, line 416: We would still require this for grub

crates/lib/src/bootc_composefs/boot.rs
Johan-Liebert1, line 382: On first install, we still get `root=` karg. Again, would require it for grub

crates/lib/src/bootc_composefs/boot.rs
cgwalters: FTR we had a live chat about this topic this morning and I think at least recent versions of GRUB do set the required UEFI variable for the DPS.

cgwalters:
OK right we have a mechanical issue in that @Johan-Liebert1 is not technically a maintainer and can't approve PRs. @Johan-Liebert1 are you OK to become a maintainer?

PR #1773, opened 2025-11-17
blockdev: find `esp` based on `MBR` or `GPT`

HuijingHei:
See https://github.com/bootc-dev/bootc/pull/1736#issuecomment-3497579510

### DIFF
--- a/crates/blockdev/src/blockdev.rs
+++ b/crates/blockdev/src/blockdev.rs
@@ -12,6 +12,13 @@ use serde::Deserialize;
 
 use bootc_utils::CommandRunExt;
 
+/// EFI System Partition (ESP) on MBR
+/// Refer to <https://en.wikipedia.org/wiki/Partition_type>
+pub const ESP_ID_MBR: &[u8] = &[0x06, 0xEF];
+
+/// EFI System Partition (ESP) for UEFI boot on GPT
+pub const ESP: &str = "c12a7328-f81f-11d2-ba4b-00a0c93ec93b";
+
 #[derive(Debug, Deserialize)]
 struct DevicesOutput {
     blockdevices: Vec<Device>,
@@ -174,6 +181,19 @@ impl PartitionTable {
     /// Find the partition with bootable is 'true'.
     pub fn find_partition_of_bootable(&self) -> Option<&Partition> {
         self.partitions.iter().find(|p| p.is_bootable())
+    }
+
+    /// Find the esp partition.
+    pub fn find_partition_of_esp(&self) -> Result<Option<&Partition>> {
+        match &self.label {
+            PartitionType::Dos => Ok(self.partitions.iter().find(|b| {
+                u8::from_str_radix(&b.parttype, 16)
+                    .map(|pt| ESP_ID_MBR.contains(&pt))
+                    .unwrap_or(false)
+            })),
+            PartitionType::Gpt => Ok(self.find_partition_of_type(ESP)),
+            _ => Err(anyhow::anyhow!("Unsupported partition table type")),
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


crates/blockdev/src/blockdev.rs
```diff
@@ -175,6 +181,20 @@ impl PartitionTable {
...
+    /// Find the esp partition.
+    pub fn find_partition_of_esp(&self) -> Result<Option<&Partition>> {
+        match &self.label {
+            PartitionType::Dos => Ok(ESP_ID_MBR
+                .iter()
+                .find_map(|&t| self.partitions.iter().find(|p| p.parttype_matches(t)))),
+            PartitionType::Gpt => Ok(self
+                .partitions
+                .iter()
+                .find(|p| p.parttype_matches(ESP_ID_GPT))),
```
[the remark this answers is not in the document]
HuijingHei: SGTM.

crates/blockdev/src/blockdev.rs
```diff
@@ -12,6 +12,12 @@ use serde::Deserialize;
 
 use bootc_utils::CommandRunExt;
 
+/// EFI System Partition (ESP) on MBR
+pub const ESP_ID_MBR: &[&str] = &["6", "ef"];
+
+/// EFI System Partition (ESP) for UEFI boot on GPT
+pub const ESP_ID_GPT: &str = "c12a7328-f81f-11d2-ba4b-00a0c93ec93b";
```
cgwalters: We already have this one the DPS spec too as `pub const ESP`

crates/blockdev/src/blockdev.rs
```diff
@@ -12,6 +12,12 @@ use serde::Deserialize;
 
 use bootc_utils::CommandRunExt;
 
+/// EFI System Partition (ESP) on MBR
+pub const ESP_ID_MBR: &[&str] = &["6", "ef"];
```
cgwalters: I wonder if we should actually be parsing these as hexadecimal *numbers* since it seems quite possible that tooling could output e.g. `06` instead of `6`.

So this would be `ESP_ID_MBR: &[u8] = [0x06, 0xEF]` or so and we detect if the partition is MBR and parse the `u8` out of the string in the JSON.

BTW for MBR it looks like https://en.wikipedia.org/wiki/Partition_type is a good resource to link to here.

crates/blockdev/src/blockdev.rs
```diff
@@ -636,11 +657,6 @@ mod test {
                         "start": 1028096,
                         "size": 2097152,
                         "type": "83"
-                    },{
```
cgwalters, line 640: Why this code change?

crates/blockdev/src/blockdev.rs
```diff
@@ -175,6 +181,17 @@ impl PartitionTable {
     pub fn find_partition_of_bootable(&self) -> Option<&Partition> {
         self.partitions.iter().find(|p| p.is_bootable())
     }
+
+    /// Find the esp partition.
+    pub fn find_partition_of_esp(&self) -> Result<Option<&Partition>> {
+        match &self.label {
+            PartitionType::Dos => Ok(ESP_ID_MBR
+                .iter()
+                .find_map(|&t| self.find_partition_of_type(t))),
```
cgwalters: This one feels like it's "backwards", I expected to iterate over all partitions and find the first which matches either type or so.

crates/blockdev/src/blockdev.rs
HuijingHei: Yes, simply do not want this to include more dependency, so I copy it from DPS.

crates/blockdev/src/blockdev.rs
HuijingHei: Just want to make it shorter, will revert this.

crates/blockdev/src/blockdev.rs
HuijingHei: Update this to `ESP` too.

crates/blockdev/src/blockdev.rs
HuijingHei: Is it necessary to also check if there is `bootable`? The first matches might do not have bootable.

cgwalters:
Very minor `make validate` failure

HuijingHei:
> Very minor `make validate` failure

Updated, thank you for the kind review!

PR #1775, opened 2025-11-17
Dockerfile: Use rpmbuild

cgwalters:
See individual commits.


cgwalters:
I was getting quite confused until I finally realized that the failure here is specific to fedora and things work on the default c10s...digging into what the difference is in the build env there.

cgwalters:
OK I think this one is now passing tests!

Issue #1707, opened 2025-10-28
Enhanced docker support for "bootc switch"


ericcurtin:
On my personal machine, been trying to use docker with bootc, we need to enhance to use docker's container storage somehow:

sudo bootc switch --transport containers-storage "$id"

Any opinions on how this should be implemented would be appreciated.

cgwalters:
container-libs/skopeo has the `docker-daemon:` transport but we don't wrap it, probably wouldn't be hard.

ericcurtin:
@cgwalters I am trying to use skopeo, but if that doesn't work out, maybe bootc <-> docker can interact directly. WDYT?

ericcurtin:
We may want to bypass skopeo docker-daemon anyway, performance seems shocking, takes a long time and eats up CPU hard...

ericcurtin:
Closing

PR #1776, opened 2025-11-18
--transport docker-daemon support

ericcurtin:
So we can do things like:

sudo bootc switch --transport docker-daemon localhost/bootc:latest

### DIFF
--- a/crates/ostree-ext/src/generic_decompress.rs
+++ b/crates/ostree-ext/src/generic_decompress.rs
@@ -22,11 +22,6 @@ use crate::oci_spec::image as oci_image;
 /// TODO: change the skopeo code to shield us from this correctly
 const DOCKER_TYPE_LAYER_TAR: &str = "application/vnd.docker.image.rootfs.diff.tar";
 
-/// The Docker MIME type for gzipped layers when stored in docker-daemon.
-/// Even though this indicates gzip compression, docker-daemon actually stores
-/// the layers uncompressed, so we need to treat this as uncompressed.
-const DOCKER_TYPE_LAYER_TAR_GZIP: &str = "application/vnd.docker.image.rootfs.diff.tar.gzip";
-
 /// Extends the `Read` trait with another method to get mutable access to the inner reader
 trait ReadWithGetInnerMut: Read + Send + 'static {
     fn get_inner_mut(&mut self) -> &mut dyn Read;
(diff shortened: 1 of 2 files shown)


ericcurtin:
Some updates made, now we are hitting this issue:

$ sudo bootc switch --transport docker-daemon localhost/bootc:latest
layers already present: 2; layers needed: 75 (31.1 GB)
Fetched layers: 0 B in 0 seconds (0 B/s)                                                                            error: Switching: Switching (ostree): Pulling: Importing: Unencapsulating base: Layer sha256:cda9fba25fac4004a9ea567d3bc2ba605df47467fc302613e3cddd85e8b5d2ee: Unhandled layer type: application/vnd.docker.image.rootfs.diff.tar.gzip

ericcurtin:
It works now:

https://github.com/bootc-dev/bootc/issues/1707

ericcurtin:
@cgwalters @jmarrero PTAL

crates/ostree-ext/src/generic_decompress.rs
```diff
@@ -22,6 +22,11 @@ use crate::oci_spec::image as oci_image;
 /// TODO: change the skopeo code to shield us from this correctly
 const DOCKER_TYPE_LAYER_TAR: &str = "application/vnd.docker.image.rootfs.diff.tar";
 
+/// The Docker MIME type for gzipped layers when stored in docker-daemon.
+/// Even though this indicates gzip compression, docker-daemon actually stores
+/// the layers uncompressed, so we need to treat this as uncompressed.
```
cgwalters: This seems surprising to me on two levels.

First, in theory the current skopeo proxy *should* be translating Docker media types to OCI unconditionally.

(Though honestly, I want to get away from supporting docker v2s2 at all, we should emit a deprecation warning or somethign)

Second, yeah the compression thing. Although that may just be really fallout from us not translating from v2s2 to OCI correctly.

---

Can we instead special case this workaround in the `layer_info` processing we do *only* if the input transport is `Transport::DockerDaemon`?

---

Also worth at least trying to spawn an agent to cross check all of this.

ericcurtin: I added another commit. Do you prefer the approach in the new commit?

cgwalters (approved):

crates/ostree-ext/src/generic_decompress.rs
cgwalters: Yes though I suspect there's still an underlying bug in the skopeo or containers-image-proxy logic here and note that because composefs goes through a different path it'd likely need a similar workaround right now.

cgwalters:
Also while I'm OK letting this slide really in general changes like this should come with a test. I know doing that would be slightly involved at the moment though.

Issue #1777, opened 2025-11-18
Missing default systemd-boot load entry (wrong boot order)


vaskozl:
When using `bootc install to-disk --composefs-backend --via-loopback bootable.img --filesystem ext4 --wipe --bootloader systemd` first boot works fine since only the following loader entry exist.

```
/boot/loader/entries/bootc-composefs-1.conf
```

However following `bootc switch` or `bootc update` the system fails to update properly due to there being no default in `/boot/loader/loader.conf` and `sd-boot` preferring smaller numbers/strings. 

The `bootc update` process creates:
```
/boot/loader/entries/bootc-composefs-0.conf (current/old)
/boot/loader/entries/bootc-composefs-1.conf (new/staged)
```

However systemd-boot always boots `0.conf` first (it also has `sort-key 0` in the file.)

This can be reproduced and fixed on the amazing bootcrew images linked in https://github.com/bootc-dev/bootc/pull/1735 via the following command:

```
mount /dev/vda1 /boot && echo 'default bootc-composefs-1.conf' > /boot/loader/loader.conf && umount /boot
```

PR #1779, opened 2025-11-18
composefs/boot: Fix sd-boot order on update

Johan-Liebert1:
Grub sorts its BLS config in descending order, while sd-boot sorts the configs in ascending order. While upgrading we were always setting the new sort key to be `1` which would work for Grub but not for sd-boot.

Fixes: #1777

### DIFF
--- a/crates/lib/src/bootc_composefs/boot.rs
+++ b/crates/lib/src/bootc_composefs/boot.rs
@@ -1,3 +1,66 @@
+//! Composefs boot setup and configuration.
+//!
+//! This module handles setting up boot entries for composefs-based deployments,
+//! including generating BLS (Boot Loader Specification) entries, copying kernel/initrd
+//! files, managing UKI (Unified Kernel Images), and configuring the ESP (EFI System
+//! Partition).
+//!
+//! ## Boot Ordering
+//!
+//! A critical aspect of this module is boot entry ordering, which must work correctly
+//! across both Grub and systemd-boot bootloaders despite their fundamentally different
+//! sorting behaviors.
+//!
+//! ## Critical Context: Grub's Filename Parsing
+//!
+//! **Grub does NOT read BLS fields** - it parses the filename as an RPM package name!
+//! See: <https://github.com/ostreedev/ostree/issues/2961>
+//!
+//! Grub's `split_package_string()` parsing algorithm:
+//! 1. Strip `.conf` suffix
+//! 2. Find LAST `-` → extract **release** field
+//! 3. Find SECOND-TO-LAST `-` → extract **version** field
+//! 4. Remainder → **name** field
+//!
+//! Example: `kernel-5.14.0-362.fc38.conf`
+//! - name: `kernel`
+//! - version: `5.14.0`
+//! - release: `362.fc38`
+//!
+//! **Critical:** Grub sorts by (name, version, release) in DESCENDING order.
+//!
+//! ## Bootloader Differences
+//!
+//! ### Grub
+//! - Ignores BLS sort-key field completely
+//! - Parses filename to extract name-version-release
+//! - Sorts by (name, version, release) DESCENDING
+//! - Any `-` in name/version gets incorrectly split
+//!
+//! ### Systemd-boot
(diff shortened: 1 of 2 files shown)


Johan-Liebert1:
I'm also working on adding tests for these cases

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -474,7 +474,8 @@ pub(crate) fn setup_composefs_bls_boot(
             let boot_digest = compute_boot_digest(usr_lib_modules_vmlinuz, &repo)
                 .context("Computing boot digest")?;
 
-            let default_sort_key = "1";
+            let default_sort_key = bootloader.default_sort_key();
```
cgwalters: Hmm I think it'd be cleaner if computed all the BLS entries without writing them, and then walk over all of them and reverse the order if we detect grub or so?

crates/lib/src/spec.rs
```diff
@@ -197,6 +197,31 @@ impl FromStr for Bootloader {
...
+impl Bootloader {
+    pub(crate) const SYSTEMD_PRIMARY_SORT_KEY: &str = "0";
+    // Grub entries are sorted by their filename in reverse order
+    pub(crate) const GRUB_PRIMARY_SORT_KEY: &str = "1";
+
+    pub(crate) const SYSTEMD_SECONDARY_SORT_KEY: &str = "1";
+    pub(crate) const GRUB_SECONDARY_SORT_KEY: &str = "0";
+
+    /// Sort key for the primary BLS entry
+    pub(crate) fn default_sort_key(&self) -> &str {
```
cgwalters: How about 

```
sorting_keys(&self) -> impl Iterator<Item=&'static str> {
  let r = ["0", "1"]
  match self {
    Systemd => r
    Grub => r.rev()
  }
```

cgwalters:
OK first when talking about grub, there's a giant mess there because it has a wildly varying patch set per distro/OS. In particular https://src.fedoraproject.org/rpms/grub2 now has 390 (!) patches.

Anyways I think the *real* problem here is:

```
~/src/src.fedoraproject.org/rpms/grub2/grub2-2.12-build/grub-2.12> grep -r sort-key .
./docs/grub.texi:sort-key     fedora
~/src/src.fedoraproject.org/rpms/grub2/grub2-2.12-build/grub-2.12>
```

i.e. at least Fedora's grub2 isn't implementing `sort-key` at all AFAICS. I suspect we're getting sorting by filenames instead.

We should make this at least build or runtime configurable probably in the future, but for now until we gather more data on grub on other OSes it's likely simplest to just not emit `sort-key` at all with grub and ensure that we always do filename ordering.

benjamin-pilgrim:
From what I can see, this patch doesn't change the current behaviour for Grub, but fixes the currently broken behaviour for systemd-boot. Is it worth getting this in first and then dealing with Grub as a separate PR if more info is needed before figuring out the best solution for Grub?

Johan-Liebert1:
> I suspect we're getting sorting by filenames instead

That's what's happening, at least in fedora based images. The `sort-key` is present in the BLS Config even for grub is because we only have one `write` function for the `BLSConfig` struct which writes every field in the struct regardless of which bootloader is used

crates/lib/src/spec.rs
Johan-Liebert1: We use the secondary keys when upgrading/switching, so this doesn't quite work for that case

crates/lib/src/bootc_composefs/boot.rs
Johan-Liebert1: That is kind of what we're doing, but in a bit of a weird way. This might be good for a cleanup. I'll pick it up

cgwalters:
> From what I can see, this patch doesn't change the current behaviour for Grub, but fixes the currently broken behaviour for systemd-boot. Is it worth getting this in first and then dealing with Grub as a separate PR if more info is needed before figuring out the best solution for Grub?

Yes I think that's true, however I think it's well worth some further cleanups here. We'll get this fixed soon.

PR #1783, opened 2025-11-19
Get kargs from usr/lib/bootc/kargs.d

Johan-Liebert1:
Parse toml files in usr/lib/bootc/kargs.d and append them to kernel
cmdline on install and upgrade/switch. Also, copy over current
deployment's cmdline args on upgrade/switch to another deployment

storage: Add `boot_dir` and `esp` fields

We have a lot of places where we mount the ESP temporarily and a lot of
switch cases for Grub's vs SystemdBoot's 'boot' directory.

We add a `boot_dir` field in Storage which points to `/sysroot/boot` for
systems with Grub as the bootloader and points to the ESP for systems
with SystemdBoot as the bootloader.

Also we mount the ESP temporarily while creating the storage struct,
which cleans up the code quite a bit.

### DIFF
--- a/crates/lib/src/bootc_composefs/boot.rs
+++ b/crates/lib/src/bootc_composefs/boot.rs
@@ -381,7 +381,7 @@ pub(crate) fn setup_composefs_bls_boot(
     repo: crate::store::ComposefsRepository,
     id: &Sha512HashValue,
     entry: &ComposefsBootEntry<Sha512HashValue>,
-    mounted_erofs: &Dir
+    mounted_erofs: &Dir,
 ) -> Result<String> {
     let id_hex = id.to_hex();
 
@@ -446,9 +446,15 @@ pub(crate) fn setup_composefs_bls_boot(
         }
     };
 
-    kargs_from_composefs_filesystem(mounted_erofs, &mut cmdline_refs)?;
-
     let is_upgrade = matches!(setup_type, BootSetupType::Upgrade(..));
+
+    let current_root = if is_upgrade {
+        Some(&Dir::open_ambient_dir("/", ambient_authority()).context("Opening root")?)
+    } else {
+        None
+    };
+
+    kargs_from_composefs_filesystem(mounted_erofs, current_root, &mut cmdline_refs)?;
 
     let (entry_paths, _tmpdir_guard) = match bootloader {
         Bootloader::Grub => {
@@ -1076,7 +1082,7 @@ pub(crate) fn setup_composefs_boot(
                 repo,
                 &id,
                 entry,
-                &mounted_fs
+                &mounted_fs,
             )?;
 
             boot_digest = Some(digest);
--- a/crates/lib/src/bootc_kargs.rs
+++ b/crates/lib/src/bootc_kargs.rs
@@ -45,13 +45,49 @@ impl Config {
     }
 }
 
+/// Compute the diff between existing and remote kargs
+/// Apply the diff to the new kargs
+fn compute_apply_kargs_diff(
+    existing_kargs: &Cmdline,
+    remote_kargs: &Cmdline,
+    new_kargs: &mut Cmdline,
+) {
+    // Calculate the diff between the existing and remote kargs
+    let added_kargs: Vec<_> = remote_kargs
+        .iter()
+        .filter(|item| !existing_kargs.iter().any(|existing| *item == existing))
+        .collect();
+    let removed_kargs: Vec<_> = existing_kargs
+        .iter()
+        .filter(|item| !remote_kargs.iter().any(|remote| *item == remote))
+        .collect();
+
+    tracing::debug!("kargs: added={:?} removed={:?}", added_kargs, removed_kargs);
+
+    // Apply the diff to the system kargs
+    for arg in &removed_kargs {
+        new_kargs.remove_exact(arg);
+    }
+    for arg in &added_kargs {
+        new_kargs.add(arg);
+    }
+}
+
 /// Looks for files in usr/lib/bootc/kargs.d and parses cmdline agruments
 pub(crate) fn kargs_from_composefs_filesystem(
     new_fs: &Dir,
-    cmdline: &mut Cmdline,
+    current_root: Option<&Dir>,
+    new_kargs: &mut Cmdline,
 ) -> Result<()> {
     let remote_kargs = get_kargs_in_root(new_fs, std::env::consts::ARCH)?;
-    cmdline.extend(&remote_kargs);
+
+    let existing_kargs = match current_root {
+        Some(root) => get_kargs_in_root(root, std::env::consts::ARCH)?,
+        None => Cmdline::new(),
+    };
+
+    compute_apply_kargs_diff(&existing_kargs, &remote_kargs, new_kargs);
+
     Ok(())
 }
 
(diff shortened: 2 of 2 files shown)


crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -408,16 +408,41 @@ pub(crate) fn setup_composefs_bls_boot(
...
+
+            let mut cmdline = match current_cfg.cfg_type {
+                BLSConfigType::NonEFI { options, .. } => {
+                    let options = options
+                        .ok_or_else(|| anyhow::anyhow!("No 'options' found in BLS Config"))?;
+
+                    Cmdline::from(options)
+                }
+
+                _ => anyhow::bail!("Found NonEFI config"),
```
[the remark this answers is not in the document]
Johan-Liebert1: We change the key `efi` to `uki` here https://github.com/bootc-dev/bootc/pull/1780
This would need to change anyway afterwards

Johan-Liebert1:
The big assumption here is that all composefs systems will boot via UEFI. That being said, we were already making this assumption in a few other places

Johan-Liebert1:
I think I overwrote some of my changes on force pushing. Marking this as draft for now

crates/lib/src/store/mod.rs
```diff
@@ -235,6 +252,15 @@ impl BootedStorage {
 pub(crate) struct Storage {
     /// Directory holding the physical root
     pub physical_root: Dir,
+
+    /// The 'boot' directory, useful and `Some` only for composefs systems
```
cgwalters, line 257: It's more that ostree::Sysroot handles this itself right now right?

And that kind of gets to the next thing. The way I'd been thinking about `struct Storage` is closer to an abstraction over `ostree::Repo` and `composefs::Repository`.

Maybe arguably it'd be slightly cleaner to actually have a `struct SystemRoot` that has these new members, and has an `impl Deref<Target=Storage>`?

But eh.

crates/lib/src/bootc_kargs.rs
```diff
@@ -45,6 +49,48 @@ impl Config {
...
+    let Some(kargs_d) = kargs_d else {
+        return Ok(());
+    };
+
+    for (fname, inode) in kargs_d.sorted_entries() {
+        let Inode::Leaf(..) = inode else { continue };
+
+        let fname_str = fname
+            .to_str()
+            .ok_or_else(|| anyhow::anyhow!("Failed to get filename as string"))?;
```
cgwalters: Perhaps arguably we should skip in this case?

I think we want a helper fn for this that includes the path in the error message if we *do* propagate an error.

(Non-UTF8 filenames is just a giant mess)

crates/lib/src/store/mod.rs
```diff
@@ -169,9 +171,22 @@ impl BootedStorage {
...
 
+                // NOTE: This is assuming that we'll only have composefs in a UEFI system
+                // We do have this assumptions in a lot of other places
+                let parent = get_sysroot_parent_dev(&physical_root)?;
+                let (esp_part, ..) = get_esp_partition(&parent)?;
+                let esp_mount = mount_esp(&esp_part)?;
+
+                let boot_dir = match get_bootloader()? {
+                    Bootloader::Grub => physical_root.open_dir("boot").context("Opening boot")?,
+                    Bootloader::Systemd => esp_mount.fd.try_clone().context("Cloning fd")?,
```
cgwalters, line 183: I don't think this will always be correct if XBOOTLDR is present; it's OK for now but can you add `// Note: Should also handle XBOOTLDR` or so?

crates/lib/src/bootc_composefs/gc.rs
```diff
@@ -44,28 +43,30 @@ fn list_erofs_images(sysroot: &Dir) -> Result<Vec<String>> {
 /// # Returns
 /// The fsverity of EROFS images corresponding to boot entries
 #[fn_error_context::context("Listing bootloader entries")]
-fn list_bootloader_entries(physical_root: &Dir) -> Result<Vec<String>> {
+fn list_bootloader_entries(storage: &Storage) -> Result<Vec<String>> {
     let bootloader = get_bootloader()?;
+    let boot_dir = storage
+        .boot_dir
+        .as_ref()
+        .ok_or_else(|| anyhow::anyhow!("Boot dir not found"))?;
```
cgwalters: This one is duplicated a lot, how about `storage.require_boot_dir()?`

crates/lib/src/store/mod.rs
Johan-Liebert1: > The way I'd been thinking about struct Storage is closer to an abstraction over ostree::Repo and composefs::Repository

We do have `physical_root` and `run` in `struct Storage` so not quite an abstraction over Repository... 
Maybe we could put these in `BootedComposefs`, but that doesn't sound right either

crates/lib/src/bootc_kargs.rs
Johan-Liebert1: > Perhaps arguably we should skip in this case?

I think we can skip it, and maybe have in the docs that we only support UTF8 filenames

crates/lib/src/bootc_composefs/gc.rs
Johan-Liebert1: Done

crates/lib/src/store/mod.rs
Johan-Liebert1: Ah yes. We don't use `XBOOTLDR` right now, but yes, we'll need to take it into account if and when we use it

crates/lib/src/bootc_kargs.rs
Johan-Liebert1: Skipping any non UTF-8 named files with a warning

crates/lib/src/store/mod.rs
```diff
@@ -169,9 +171,23 @@ impl BootedStorage {
                 }
                 let composefs = Arc::new(composefs);
 
+                // NOTE: This is assuming that we'll only have composefs in a UEFI system
```
cgwalters, line 174: I think you mean UKI here

crates/lib/src/bootc_kargs.rs
```diff
@@ -45,6 +50,96 @@ impl Config {
...
+    for arg in &removed_kargs {
+        new_kargs.remove_exact(arg);
+    }
+    for arg in &added_kargs {
+        new_kargs.add(arg);
+    }
+}
+
+/// Looks for files in usr/lib/bootc/kargs.d and parses cmdline agruments
+pub(crate) fn kargs_from_composefs_filesystem(
```
cgwalters, line 77: Don't we have the target composefs actually mounted at this point? I am pretty sure I just did something like that, and if so it's way simpler as we can just directly read the FS and not have a *third* path.

crates/lib/src/bootc_kargs.rs
```diff
@@ -45,6 +50,96 @@ impl Config {
...
+    fs: &ComposefsFilesystem,
+    repo: &ComposefsRepository,
+    new_kargs: &mut Cmdline,
+    fresh_install: bool,
```
cgwalters: How about `current_root: Option<&Dir>` instead

cgwalters:
BTW to pass CI this also needs a rebase :surfer: 
On that topic I did https://github.com/bootc-dev/infra/pull/46 so once that lands any maintainer should be able to add the `needs-rebase` label to do it.

crates/lib/src/bootc_kargs.rs
Johan-Liebert1: Updated

crates/lib/src/bootc_kargs.rs
Johan-Liebert1: Ah, I see. That definitely makes this a lot cleaner

crates/lib/src/store/mod.rs
Johan-Liebert1: I did mean UEFI. In a lot of places we assume that we're not installed on a BIOS system

crates/lib/src/bootc_kargs.rs
```diff
@@ -45,6 +45,52 @@ impl Config {
...
+    for arg in &removed_kargs {
+        new_kargs.remove_exact(arg);
+    }
+    for arg in &added_kargs {
+        new_kargs.add(arg);
+    }
+}
+
+/// Looks for files in usr/lib/bootc/kargs.d and parses cmdline agruments
+pub(crate) fn kargs_from_composefs_filesystem(
```
cgwalters, line 77: Nit, this is now unrelated to composefs and could probably just be called `compute_new_kargs` or something?

Johan-Liebert1: Updated in #1811

PR #1784, opened 2025-11-19
core: Fix usroverlay regression

cgwalters:
The `/sysroot` writability regressed this, and add a missing test case for this.

### DIFF
--- a/tmt/tests/booted/bootc_testlib.nu
+++ b/tmt/tests/booted/bootc_testlib.nu
@@ -1,0 +1,14 @@
+# A simple nushell "library" for the
+
+# This is a workaround for what must be a systemd bug
+# that seems to have appeared in C10S
+# TODO diagnose and fill in here
+export def reboot [] {
+    # Sometimes systemd daemons are still running old binaries and response "Access denied" when send reboot request
+    # Force a full sync before reboot
+    sync
+    # Allow more delay for bootc to settle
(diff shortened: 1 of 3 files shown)


jmarrero:
code lgtm, but it's weird that only stream10 is failing with:
```
  x Eval block failed with pipeline input
    ,-[/var/ARTIFACTS/work-readonly-testsjy4ax59d/tmt/plans/integration/readonly-tests/discover/default-0/tests/tmt/tests/booted/readonly/030-test-locking-read.nu:19:1]
 18 | # Await completion
 19 | 0..$n | each { |v|
    : ^^|^^
    :   `-- source value
 20 |     loop {
    `----

Error: nu::shell::non_zero_exit_code

  x External command had a non-zero exit code
    ,-[/var/ARTIFACTS/work-readonly-testsjy4ax59d/tmt/plans/integration/readonly-tests/discover/default-0/tests/tmt/tests/booted/readonly/030-test-locking-read.nu:26:9]
 25 |         # check status
 26 |         systemctl status $"bootc-status-($v)" out> /dev/null
    :         ^^^^|^^^^
    :             `-- exited with code 3
 27 |         # Clean it up
    `----
```
I am re-running these and see if it's some glitch as this nu test was not changed here.

cgwalters:
OK right we need the same workaround here that landed in the factory reset test.

PR #1787, opened 2025-11-19
Justfile: Bring back build-sealed target

jeckersb:
Except now this wraps the "variant" usage of `just build`.  For one,
this is a convenient little piece of shorthand.  But perhaps more
importantly, we have this referenced in the docs.  Although I could go
change the docs to reflect the new usage, I think it just reads nicer
in the docs for it to remain `just build-sealed`.


cgwalters (approved):
I think we should try hard to have all of our Justfile entrypoints have at least a single line documentation comment.

jeckersb:
> I think we should try hard to have all of our Justfile entrypoints have at least a single line documentation comment.

Added.

PR #1790, opened 2025-11-19
Various tmt + testing improvements

cgwalters:
See individual patches.

### DIFF
--- a/crates/xtask/src/tmt.rs
+++ b/crates/xtask/src/tmt.rs
@@ -18,8 +18,19 @@ const COMMON_INST_ARGS: &[&str] = &[
     // TODO: Pass down the Secure Boot keys for tests if present
     "--firmware=uefi-insecure",
     "--label=bootc.test=1",
-    "--bind-storage-ro",
 ];
+
+// Metadata field names
+const FIELD_TRY_BIND_STORAGE: &str = "try_bind_storage";
+const FIELD_SUMMARY: &str = "summary";
+const FIELD_ADJUST: &str = "adjust";
+
+// bcvk options
+const BCVK_OPT_BIND_STORAGE_RO: &str = "--bind-storage-ro";
+const ENV_BOOTC_UPGRADE_IMAGE: &str = "BOOTC_upgrade_image";
+
+// Distro identifiers
+const DISTRO_CENTOS_9: &str = "centos-9";
 
 // Import the argument types from xtask.rs
 use crate::{RunTmtArgs, TmtProvisionArgs};
@@ -56,13 +67,38 @@ fn sanitize_plan_name(plan: &str) -> String {
 /// Check that required dependencies are available
 #[context("Checking dependencies")]
 fn check_dependencies(sh: &Shell) -> Result<()> {
-    for tool in ["bcvk", "tmt", "rsync"] {
+    for tool in ["bcvk", "tmt", "rsync", "podman"] {
         cmd!(sh, "which {tool}")
             .ignore_stdout()
             .run()
             .with_context(|| format!("{} is not available in PATH", tool))?;
     }
     Ok(())
+}
+
+/// Detect distro from container image by reading os-release
+/// Returns distro string like "centos-9" or "fedora-42"
+#[context("Detecting distro from image")]
+fn detect_distro_from_image(sh: &Shell, image: &str) -> Result<String> {
+    let distro = cmd!(
+        sh,
+        "podman run --rm {image} bash -c '. /usr/lib/os-release && echo $ID-$VERSION_ID'"
+    )
+    .read()
+    .context("Failed to run image as container to detect distro")?;
+
+    let distro = distro.trim();
+    if distro.is_empty() {
+        anyhow::bail!("Failed to extract distro from os-release");
+    }
+
+    Ok(distro.to_string())
+}
+
+/// Check if a distro supports --bind-storage-ro
+/// CentOS 9 lacks systemd.extra-unit.* support required for bind-storage-ro
+fn distro_supports_bind_storage_ro(distro: &str) -> bool {
+    !distro.starts_with(DISTRO_CENTOS_9)
 }
 
 /// Wait for a bcvk VM to be ready and return SSH connection info
@@ -142,6 +178,43 @@ fn verify_ssh_connectivity(sh: &Shell, port: u16, key_path: &Utf8Path) -> Result
     )
 }
 
+/// Parse integration.fmf to extract extra-try_bind_storage for all plans
+#[context("Parsing integration.fmf")]
+fn parse_plan_metadata(plans_file: &Utf8Path) -> Result<std::collections::HashMap<String, bool>> {
+    let content = std::fs::read_to_string(plans_file)?;
+    let yaml = serde_yaml::from_str::<serde_yaml::Value>(&content)
+        .context("Failed to parse integration.fmf YAML")?;
+
+    let Some(mapping) = yaml.as_mapping() else {
+        anyhow::bail!("Expected YAML mapping in integration.fmf");
+    };
+
+    let mut plan_metadata = std::collections::HashMap::new();
+
+    for (key, value) in mapping {
+        let Some(plan_name) = key.as_str() else {
+            continue;
+        };
+        if !plan_name.starts_with("/plan-") {
+            continue;
+        }
+
+        let Some(plan_data) = value.as_mapping() else {
+            continue;
+        };
+        if let Some(try_bind) = plan_data.get(&serde_yaml::Value::String(format!(
+            "extra-{}",
+            FIELD_TRY_BIND_STORAGE
+        ))) {
+            if let Some(b) = try_bind.as_bool() {
+                plan_metadata.insert(plan_name.to_string(), b);
+            }
+        }
+    }
+
+    Ok(plan_metadata)
+}
+
 /// Run TMT tests using bcvk for VM management
 /// This spawns a separate VM per test plan to avoid state leakage between tests.
 #[context("Running TMT tests")]
@@ -151,16 +224,21 @@ pub(crate) fn run_tmt(sh: &Shell, args: &RunTmtArgs) -> Result<()> {
 
     let image = &args.image;
     let filter_args = &args.filters;
+
+    // Detect distro from the image
+    let distro = detect_distro_from_image(sh, image)?;
+
     let context = args
         .context
         .iter()
-        .map(|v| v.as_str())
-        .chain(std::iter::once("running_env=image_mode"))
-        .map(|v| format!("--context={v}"))
+        .map(|v| format!("--context={}", v))
+        .chain(std::iter::once(format!("--context=running_env=image_mode")))
+        .chain(std::iter::once(format!("--context=distro={}", distro)))
         .collect::<Vec<_>>();
     let preserve_vm = args.preserve_vm;
 
     println!("Using bcvk image: {}", image);
+    println!("Detected distro: {}", distro);
 
     // Create tmt-workdir and copy tmt bits to it
     // This works around https://github.com/teemtee/tmt/issues/4062
@@ -175,6 +253,10 @@ pub(crate) fn run_tmt(sh: &Shell, args: &RunTmtArgs) -> Result<()> {
 
     // Change to workdir for running tmt commands
     let _dir = sh.push_dir(workdir);
+
+    // Parse plan metadata from integration.fmf
+    let plans_file = Utf8Path::new("tmt/plans/integration.fmf");
+    let plan_metadata = parse_plan_metadata(plans_file)?;
 
     // Get the list of plans
     println!("Discovering test plans...");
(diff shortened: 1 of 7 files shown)


henrywang:
The failure comes from `test-26-examples-build` in aarch64. This test has not been included into tmt plan before, so it only runs on github runner, but not on packit. I'm working on it to fix the issue.

jeckersb:
> xtask: Run tmt report on failure
>
> Why tmt doesn't do this by default, I have no idea. It's also baffling that it takes three levels of verbose i.e. `-vvv` to get the actual test error.

This tmt behavior also drives me crazy.  Glad to see this change :+1:

tmt/tests/examples/bootc-uki/Containerfile.stage2
```diff
@@ -25,11 +25,13 @@ RUN --mount=type=secret,id=key \
         --measure \
         --json pretty \
         --output "/boot/$kver.efi"
+    # To support multi arch
+    systemd_boot_file=$(ls /usr/lib/systemd/boot/efi/*.efi | xargs -n 1 basename)
```
cgwalters: Right though we're going to need the same fix in Dockerfile.cfsuki. Which I think again is a driving use case for having a higher level wrapper for this.

cgwalters:
Strange, I don't understand why the soft reboot test is failing with testing-farm here but [works in previous runs](https://artifacts.dev.testing-farm.io/4a09be8b-673d-4838-8d57-58e48523ad57/#work-test-25-soft-rebootksz1audh_tmt-plans-integration-test-25-soft-reboot).

henrywang:
> Strange, I don't understand why the soft reboot test is failing with testing-farm here but [works in previous runs](https://artifacts.dev.testing-farm.io/4a09be8b-673d-4838-8d57-58e48523ad57/#work-test-25-soft-rebootksz1audh_tmt-plans-integration-test-25-soft-reboot).

Already worked on it. Not figure out the root cause in TF. Will debug it in pure TF to see what happens.

cgwalters:
Wait a minute...does testing-farm+tmt run all tests from a single plan on the same provisioned environment? It's sure looking like that's the case, see e.g. [this log](https://artifacts.dev.testing-farm.io/775b92a2-b5c0-4365-a3e3-c61378f1e0ff/work-all-testsz6hd6ff2/tmt-run.log) from the run here.

It's quite impressive actually that it's *just* that test that fails because some of these tests are going to leak state.

Yeah this seems to be the case. OK so this PR needs a rework.

cgwalters:
> because some of these tests are going to leak state.

(Of course, we could optimize this if we relied on a factory reset; I'm sure we could do that for most tests)

cgwalters:
I've come to the reluctant conclusion there's no sane option to fix the "triple test definition" problem we have other than going to *generating* the tmt stuff. Working on that.

cgwalters:
One thing I will say is I am so glad that AI exists to learn things like fmf for me

cgwalters:
>         content: /bin/sh: error while loading shared libraries: /lib64/libc.so.6: cannot apply additional memory protection after relocation: Permission denied

Hmm, I saw that once locally but it went away and I didn't debug. Dang it...I think this is a complicated side effect of us exporting the host container storage via virtiofs - basically the host selinux policy is tripping over the `virtiofs_t` type. Hum...I have an idea

cgwalters:
> Hum...I have an idea

:arrow_right: https://github.com/bootc-dev/bcvk/pull/159

Haven't reproduced the bug yet locally though but it at least doesn't hurt

tmt/tests/booted/readonly/017-test-bound-storage.nu
```diff
@@ -0,0 +1,27 @@
...
+
+if not (bootc_testlib have_hostexports) {
+    print "No host exports, skipping"
+    exit 0
+}
+
+bootc status
+let st = bootc status --json | from json
+let is_composefs = ($st.status.booted.composefs? != null)
+if is_composefs {
```
henrywang: One failure comes from `is_composefs`. It should be `$is_composefs`.

henrywang:
Another failure comes from podman build and podman run in `image-pushpull-upgrade`, `logically-bound-switch`, `soft-reboot` and `custom-selinux-policy`. I can reproduce this issue locally and the fix works in my local env.
```diff
diff --git i/tmt/tests/booted/test-image-pushpull-upgrade.nu w/tmt/tests/booted/test-image-pushpull-upgrade.nu
index 5367dcc8..eca7e9dd 100644
--- i/tmt/tests/booted/test-image-pushpull-upgrade.nu
+++ w/tmt/tests/booted/test-image-pushpull-upgrade.nu
@@ -49,9 +49,9 @@ COPY usr/ /usr/
 RUN echo test content > /usr/share/blah.txt
 " | save Dockerfile
     # Build it
-    podman build -t localhost/bootc-derived .
+    podman build --security-opt label=disable -t localhost/bootc-derived .
     # Just sanity check it
-    let v = podman run --rm localhost/bootc-derived cat /usr/share/blah.txt | str trim
+    let v = podman run --rm --security-opt label=disable localhost/bootc-derived cat /usr/share/blah.txt | str trim
     assert equal $v "test content"

     let orig_root_mtime = ls -Dl /ostree/bootc | get modified
@@ -158,7 +158,7 @@ COPY usr/ /usr/
 RUN echo test content2 > /usr/share/blah.txt
 " | save Dockerfile
     # Build it
-    podman build -t localhost/bootc-derived .
+    podman build --security-opt label=disable -t localhost/bootc-derived .
     let booted_digest = $booted.imageDigest
     print $"booted_digest = ($booted_digest)"
     # We should already be fetching updates from container storage
```

cgwalters:
Thanks for picking this up! I hope that https://github.com/bootc-dev/bcvk/pull/159 fixes the selinux issue. That said, I haven't yet reproduced the failure locally, and I'm not understanding why.

henrywang:
cs10 flaky issue appears again. So my workaround `sync` and `sleep` does not work. Investigate this issue now, have to fix it.

henrywang:
github test failures:

1. `test-24-image-upgrade-reboot` failed on CS9 only:
```
content: local image push + pull + upgrade
content: Applying localhost/bootc-integration-upgrade
content: error: Switching: Switching (ostree): Preparing import: Fetching manifest: failed to invoke method OpenImage: reference "[overlay@/var/lib/containers/storage+/run/containers/storage:overlay.mountopt=nodev,metacopy=on]localhost/bootc-integration-upgrade:latest" does not resolve to an image ID
```

2. `test-26-examples-build` failed on Fedora 42 only:
```
content: + ./bootc internals cfs --repo tmp/sysroot/composefs --insecure oci pull containers-storage:17c41e1c378b4616c1f17ccbffa0ec55ebed6a6e9ff384ad99f81c0b5dda1d0a
content: error: Unable to pull container image containers-storage:17c41e1c378b4616c1f17ccbffa0ec55ebed6a6e9ff384ad99f81c0b5dda1d0a: Failed to pull config Descriptor { media_type: ImageConfig, digest: Digest { algorithm: Sha256, value: "sha256:17c41e1c378b4616c1f17ccbffa0ec55ebed6a6e9ff384ad99f81c0b5dda1d0a", split: 6 }, size: 11833, urls: None, annotations: None, platform: None, artifact_type: None, data: None }: failed to invoke method GetBlob: open /run/host-container-storage/overlay/7f7c23aa7156afa1e272fdfa3506ba2f1947fc7224e6f3d377c99365d678c47e/diff/sysroot/ostree/repo/objects/68/7b65d1fc247722ad67c6fb16bbef684d45e1ec46c4d535ff0ba430a26bffda.file-xattrs-link: too many open files
```

Testing farm/Packit test failure:

1. `test-23-usroverlay` failed on CS10 only:

```
err: Call to Reboot failed: Access denied
```

cgwalters:
> localhost/bootc-integration-upgrade:latest" does not resolve to an image ID

Hmmm...it might be older podman doesn't handle additional image stores in the same way? I'll take a look.

Nah it's really simple, `--bind-storage-ro` doesn't work with C9S because it's missing https://www.freedesktop.org/software/systemd/man/latest/systemd-debug-generator.html#systemd.extra-unit.*

It's pretty tempting to try to "backfill" this...

>too many open files

I think this one is a flake...but if it keeps happening we'll have to debug.

cgwalters:
> Nah it's really simple, --bind-storage-ro doesn't work with C9S because it's missing https://www.freedesktop.org/software/systemd/man/latest/systemd-debug-generator.html#systemd.extra-unit.*

I mean, the *problem* was simple to understand...the fix is definitely uglier and more involved as it requires knowledge about the target environment capabilities "up front" in our tests. Anyways, done now and tested both cases locally.

crates/xtask/src/tmt.rs
```diff
@@ -0,0 +1,1118 @@
...
+    // Run each plan in its own VM
+    for plan in plans {
+        let plan_name = sanitize_plan_name(plan);
+        let vm_name = format!("bootc-tmt-{}-{}", random_suffix, plan_name);
+
+        println!("\n========================================");
+        println!("Running plan: {}", plan);
+        println!("VM name: {}", vm_name);
+        println!("========================================\n");
+
```
henrywang, line 313: Failure comes from the plan name.
```diff
diff --git i/crates/xtask/src/tmt.rs w/crates/xtask/src/tmt.rs
index 0ba7a1a2..4aeab60d 100644
--- i/crates/xtask/src/tmt.rs
+++ w/crates/xtask/src/tmt.rs
@@ -308,10 +308,35 @@ pub(crate) fn run_tmt(sh: &Shell, args: &RunTmtArgs) -> Result<()> {
         println!("VM name: {}", vm_name);
         println!("========================================\n");

+        // Extract the tail component of the plan path
+        // "/tmt/plans/integration/plan-01-readonly" → "plan-01-readonly"
+        let plan_basename = plan
+            .rsplit('/')
+            .next()
+            .unwrap_or(plan)
+            .trim_start_matches('/');
+
+        // Possible metadata keys it might match
+        let candidate_keys = [
+            plan,                        // full path
+            plan.trim_start_matches('/'),// full path w/o leading slash
+            plan_basename,               // "plan-01-readonly"
+            &format!("/{}", plan_basename), // "/plan-01-readonly"
+        ];
+
         // Get bcvk-opts based on plan metadata and distro support
         let plan_bcvk_opts = {
             let supports_bind_storage_ro = distro_supports_bind_storage_ro(&distro);
-            let try_bind_storage = plan_metadata.get(plan).copied().unwrap_or(false);
+            let try_bind_storage = candidate_keys
+                .iter()
+                .find_map(|k| plan_metadata.get(*k).copied())
+                .unwrap_or(false);
+
+            println!(
+                "Metadata lookup for plan='{}', basename='{}' → try_bind_storage={}",
+                plan, plan_basename, try_bind_storage
+            );
+            println!("support bind storage ro: {}", supports_bind_storage_ro);

             let mut opts = Vec::new();

@@ -332,6 +357,8 @@ pub(crate) fn run_tmt(sh: &Shell, args: &RunTmtArgs) -> Result<()> {

             opts
         };
+        println!("plan metadata: {:?}", plan_metadata);
+        println!("bcvk args: {:?}", plan_bcvk_opts);
```

crates/xtask/src/tmt.rs
```diff
@@ -0,0 +1,1118 @@
...
+            let mut opts = Vec::new();
+
+            // If test wants bind storage and distro supports it, add --bind-storage-ro
+            if try_bind_storage && supports_bind_storage_ro {
+                opts.push(BCVK_OPT_BIND_STORAGE_RO.to_string());
+
+                // If upgrade image is provided, set it as an environment variable
+                if let Some(ref upgrade_img) = args.upgrade_image {
+                    opts.push(format!("--env={}={}", ENV_BOOTC_UPGRADE_IMAGE, upgrade_img));
```
henrywang: In my environment, `bcvk` does not support `--env`.
```shell
$ bcvk libvirt run --name bootc-tmt-am5rk6c2-tmt-plans-integration-plan-01-readonly --detach --firmware=uefi-insecure --label=bootc.test=1 --bind-storage-ro --env=BOOTC_upgrade_image=localhost/bootc-integration-upgrade localhost/bootc-integration
error: unexpected argument '--env' found

  tip: to pass '--env' as a value, use '-- --env'

Usage: bcvk libvirt run --name <NAME> --detach --firmware <FIRMWARE> --label <LABEL> --bind-storage-ro <IMAGE>

For more information, try '--help'.
```

crates/xtask/src/tmt.rs
cgwalters: Thanks, yep I also forgot to regenerate the `integration.fmf`.

cgwalters:
OK this one should finally pass all tests!

cgwalters:
OK one more test fix for readonly+c9s

cgwalters:
> /tmt/plans/integration/plan-20-image-pushpull-upgrade
> content: error: System transaction in progress

Hum, odd first I've seen that. There shouldn't be anything else concurrent happening here...

henrywang (approved):

PR #1791, opened 2025-11-20
various: adjust to composefs-rs changes

allisonkarlitskaya:
This is mostly about the new splitstream API, but it also adopts the new library API exposed by the `cfsctl` crate, removing our private copy.

### DIFF
--- a/crates/lib/src/bootc_composefs/export.rs
+++ b/crates/lib/src/bootc_composefs/export.rs
@@ -3,17 +3,15 @@ use std::{fs::File, os::fd::AsRawFd};
 use anyhow::{Context, Result};
 use cap_std_ext::cap_std::{ambient_authority, fs::Dir};
 use composefs::splitstream::SplitStreamData;
+use composefs_oci::open_config;
 use ocidir::{OciDir, oci_spec::image::Platform};
+use ostree_ext::container::Transport;
 use ostree_ext::container::skopeo;
-use ostree_ext::{container::Transport, oci_spec::image::ImageConfiguration};
 use tar::EntryType;
 
 use crate::image::get_imgrefs_for_copy;
 use crate::{
-    bootc_composefs::{
-        status::{get_composefs_status, get_imginfo},
-        update::str_to_sha256digest,
-    },
+    bootc_composefs::status::{get_composefs_status, get_imginfo},
     store::{BootedComposefs, Storage},
 };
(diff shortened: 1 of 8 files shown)


cgwalters:
Needs DCO, also looks like some of the hashes changed

cgwalters:
I've rebased and tweaked this a bit, so it should probably be reviewed by someone other than me now.

crates/lib/src/cli.rs
```diff
@@ -1737,7 +1739,7 @@ async fn run_from_opt(opt: Opt) -> Result<()> {
                     Ok(())
                 }
             },
-            InternalsOpts::Cfs { args } => crate::cfsctl::run_from_iter(args.iter()).await,
+            InternalsOpts::Cfs { args } => Err(Command::new("cfsctl").args(args).exec().into()),
```
jeckersb: CI caught this, but this means we need to ensure `cfsctl` is installed alongside (and in CI right now it currently isn't)

jeckersb (changes requested):

crates/lib/src/cli.rs
cgwalters: Right, thanks for looking at that. I think what happened here is Allison wanted to make cfsctl into a library, but that didn't yet happen.

crates/lib/src/bootc_composefs/export.rs
```diff
@@ -120,7 +114,7 @@ pub async fn export_repo_to_image(
 
             let size = header.entry_size()?;
 
-            match layer_stream.read_exact(size as usize, ((size + 511) & !511) as usize)? {
+            match layer_stream.read_exact(size as usize, ((size as usize) + 511) & !511)? {
```
allisonkarlitskaya, line 117: . o O ( Reminds me that I wanted to get rid of the magic handling of padding in the writer side . . )

crates/lib/src/cli.rs
allisonkarlitskaya: I apparently lost the commit for that.  Are you interested in me doing it or do you prefer to keep things the way they are here?

cgwalters: Makes sense to me to make cfsctl a library crate, when we do it there's also at least one change here I'd like to keep - the write-dumpfile-to bit.

PR #1794, opened 2025-11-20
More cfs testing2

cgwalters:
This one rolls in https://github.com/bootc-dev/bootc/pull/1790/ which we should land first

The goal here is to have upgrade testing for composefs too. But also this paves the way for having more upgrade builds done from the host.


cgwalters:
I rolled this into https://github.com/bootc-dev/bootc/pull/1790

PR #1797, opened 2025-11-21
Make update and switch operations idempotent

Johan-Liebert1:
Pulls in https://github.com/bootc-dev/bootc/pull/1783
    
   Handle the following cases we can encounter on `bootc udpate` and `bootc switch`
    
    1. The verity is the same as that of the currently booted deployment
       - Nothing to do here in case of update as we're currently booted. But if we're switching
         then we update the target imageref in the .origin file for the deployment
    
    2. The verity is the same as that of the staged deployment
       - Nothing to do, as we only get a "staged" deployment if we have
       /run/composefs/staged-deployment which is the last thing we create
       while upgrading
    
    3. The verity is the same as that of the rollback deployment or any
       other deployment we have already deployed
       - Nothing to do since this is a rollback deployment which means
       this was unstaged at some point
    
    4. The verity is not found
       - The update/switch might've been canceled before
       /run/composefs/staged-deployment was created, or at any other point
       in time, or it's a new one.
    
       Any which way, we can overwrite everything. In this case we remove
       all the staged bootloader entries, if any, and remove the entire
       state directory, as it would most probably be in an inconsistent
       state.

There is a lot here, and I managed to test the main code paths and the above four cases

Related: https://github.com/bootc-dev/bootc/pull/1735

### DIFF
--- a/crates/lib/src/bootc_composefs/update.rs
+++ b/crates/lib/src/bootc_composefs/update.rs
@@ -15,7 +15,7 @@ use crate::{
         boot::{setup_composefs_bls_boot, setup_composefs_uki_boot, BootSetupType, BootType},
         repo::{get_imgref, pull_composefs_repo},
         service::start_finalize_stated_svc,
-        state::{update_target_imgref_in_origin, write_composefs_state},
+        state::write_composefs_state,
         status::{get_bootloader, get_composefs_status, get_container_manifest_and_config},
     },
     cli::UpgradeOpts,
@@ -49,7 +49,7 @@ pub fn str_to_sha256digest(id: &str) -> Result<Sha256Digest> {
 /// * The container image manifest
 /// * The container image configuration
 #[context("Checking if image {} is pulled", imgref.image)]
-async fn is_image_pulled(
+pub(crate) async fn is_image_pulled(
     repo: &ComposefsRepository,
     imgref: &ImageReference,
 ) -> Result<(Option<Sha512HashValue>, ImageManifest, ImageConfiguration)> {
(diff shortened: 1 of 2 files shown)


Johan-Liebert1:
This will need a rebase after #1783 lands

Johan-Liebert1:
Thought the TMT failure was a flake and reran the test, but it seems like reboot seems to be failing on centos-10?

```
    Cause number 1:

        Command 'reboot' returned 1.

        # stderr (1 lines)
        Call to Reboot failed: Access denied
```

crates/lib/src/bootc_composefs/update.rs
```diff
@@ -59,7 +62,201 @@ async fn is_image_pulled(
...
+            .mount(&id.to_hex())
+            .context("Failed to mount composefs image")?,
+    )?;
+
+    let boot_type = BootType::from(entry);
+    let mut boot_digest = None;
+
+    match boot_type {
+        BootType::Bls => {
+            boot_digest = Some(setup_composefs_bls_boot(
```
cgwalters, line 233: Only tangential to this but I think we want to more carefully tease apart `setup_composefs_bls_boot` so that we don't need the enum anymore.

crates/lib/src/bootc_composefs/update.rs
```diff
@@ -59,7 +62,201 @@ async fn is_image_pulled(
...
+    let image_id = fs.compute_image_id();
+
+    // Case1
+    //
+    // "update" image has the same verity as the one currently booted
+    // This could be someone trying to `bootc switch <remote_image>` where
+    // remote_image is the exact same image as the one currently booted, but
+    // they are wanting to change the target
+    //
+    // We could simply update the image origin file here
```
cgwalters, line 148: We *are* doing that right?

cgwalters (approved):

crates/lib/src/bootc_composefs/update.rs
Johan-Liebert1: Oh yeah, we are doing that. This comment was for implementation reference

Johan-Liebert1: Updated in #1811

PR #1802, opened 2025-11-24
Minor changes to UKI install flow

Johan-Liebert1:
Split from https://github.com/bootc-dev/bootc/pull/1780

- bls-config: Rename 'efi' key to 'uki'    
Key 'uki' is more appropriate for us as the PE Binary that the key
points to is a UKI

- composefs/uki: Install all UKIs in EFI/Linux/bootc
We were making a distinction based on the bootloader and installing UKIs
in EFI/Linux for Grub and EFI/Linux/bootc for sd-boot. IMO it's better
if we use the same directory for both bootloaders

### DIFF
--- a/crates/lib/src/bootc_composefs/boot.rs
+++ b/crates/lib/src/bootc_composefs/boot.rs
@@ -526,7 +526,7 @@ pub(crate) fn setup_composefs_bls_boot(
                 .with_title(title)
                 .with_sort_key(default_sort_key.into())
                 .with_version(version)
-                .with_cfg(BLSConfigType::NonEFI {
+                .with_cfg(BLSConfigType::NonUKI {
                     linux: entry_paths.abs_entries_path.join(&id_hex).join(VMLINUZ),
                     initrd: vec![entry_paths.abs_entries_path.join(&id_hex).join(INITRD)],
                     options: Some(cmdline_refs),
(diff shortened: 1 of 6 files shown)


cgwalters:
Looks like this breaks test-integration-cfs, also needs a rebase

Johan-Liebert1:
Yeah, there seems to be an issue with `systemd-boot` when using the `uki` key. It does not list the boot entry at all. Curiously, using the `efi` key works fine

cgwalters:
> Yeah, there seems to be an issue with systemd-boot when using the uki key. It does not list the boot entry at all. Curiously, using the efi key works fine

OK, marking as draft then.

Johan-Liebert1:
See: https://github.com/systemd/systemd/issues/39886

Johan-Liebert1:
This was happening due to "uki" keyword being introduced in v258 and but all the images had v257.x

cgwalters:
To use the uki keyword we'd need to do version detection then, which seems a bit complex.

But we can merge the second commit though right?

Johan-Liebert1:
Yes, we can push the second commit. It will break existing Grub UKI systems, though, which should be hopefully fine as we recommend UKI with systemd-boot anyway. We could make it backwards compatible, but I feel like that quite misses the point of having a single directory to put UKIs

Johan-Liebert1:
The second commit is in main.
The first one isn't too important and adds extra complexity of detecting systemd-boot version

PR #1809, opened 2025-11-26
Fix systemd-boot signing on sealed image test

gerblesh:
Fixes systemd-boot signing, before the systemd-boot binary was signed on the buildroot but not on the target image, resulting in an unbootable image with secure boot enabled and the proper keys enrolled. This PR fixes it by first signing the systemd-boot on image (assumes it is installed), copying it over to the final image, then computing the digest, and then finally signing and creating the UKI with a different multi stage build. Definitely a little jank and https://github.com/bootc-dev/bootc/issues/1498 does look like a better solution in the long term, however this at least gets the image in a bootable state on secure boot and allows for testing the secure boot in VMs. Would be happy to take a stab at proper image building UX but I'm not sure if that already has work done or if y'all have a particular vision in mind for the build system


Dockerfile.cfsuki
```diff
@@ -75,7 +66,7 @@ cp /run/kernel/boot/$kver.efi $target
 rm -v /usr/lib/modules/${kver}/{vmlinuz,initramfs.img}
 # Symlink into the /usr/lib/modules location
 ln -sr $target /usr/lib/modules/${kver}/$(basename $kver.efi)
-bootc container lint --fatal-warnings
+bootc container lint # --fatal-warnings no fatal warning
```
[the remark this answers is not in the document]
gerblesh: whoops! That was for testing, I'll go ahead and add that flag back

gerblesh:
added automatic key enrollment (with sdboot) in `bootc install` but one problem is that in `Dockerfile.sdboot` we need `efitools` for signing and preparing the certs for enrollment, which centos/RHEL doesn't seem to package. Let me know if it would also make sense to break out these changes (autoenrollment)

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -1,3 +1,4 @@
+<<<<<<< HEAD
```
cgwalters: Looks like you committed merge conflicts

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -1142,6 +1145,42 @@ pub(crate) fn setup_composefs_uki_boot(
     Ok(())
 }
 
+pub struct AuthFile {
+    pub filename: Utf8PathBuf,
+    pub file: RegularFile<Sha512HashValue>,
+}
+
+fn get_autoenroll_keys(root: &Directory<RegularFile<Sha512HashValue>>) -> Result<Vec<AuthFile>> {
```
cgwalters: Let's split this as a clearly distinct PR even. There's a lot to discuss for this.

(On an implementation level I think it's a bit cleaner if we operate on a mounted fd, then we can use regular FS APIs)

cgwalters (commented):
Thanks so much for starting this! Basically let's get the buildsystem rework in to properly sign systemd-boot first, and then look at the key autoenrollment as a distinct second step.

I can take a look at the first one, starting from the work here if you prefer!

crates/lib/src/bootc_composefs/boot.rs
gerblesh: Whoops! I'll fix that

crates/lib/src/bootc_composefs/boot.rs
gerblesh: Alright, I'll split it out. Yeah the implementation of the auto enrollment is pretty rough and was just to get a PoC working, still pretty new to Rust/bootc. Would you want a tracking issue for auto enrollment as well or would just a PR suffice for discussing how it should work?

gerblesh:
alright split out the code for autoenroll to just focus on signing systemd-boot in CI

gerblesh:
https://github.com/bootc-dev/bootc/pull/1818 new PR for auto enrollment here

tests/build-sealed
```diff
@@ -61,9 +56,30 @@ if test -z "${secureboot}"; then
   cd -
 fi
 
-runv podman build -t $output_image \
-  --build-arg=COMPOSEFS_FSVERITY=${cfs_digest} \
+# handle sdboot signing before building the sealed UKI
+# moves db, PK, and KEK test keys into /usr/lib/bootc/keys
```
cgwalters: This PR doesn't do that anymore

Dockerfile.sdboot
```diff
@@ -0,0 +1,50 @@
...
+    --mount=type=bind,from=base-unsigned,target=/target \
+    <<EORUN
+    set -xeuo pipefail
+
+    # pesign uses NSS database so create it from input cert/key
+    mkdir pesign
+    certutil -N -d pesign --empty-password
+    openssl pkcs12 -export -password 'pass:' -inkey /run/secrets/key -in /run/secrets/cert -out db.p12
+    pk12util -i db.p12 -W '' -d pesign
+    subject=$(openssl x509 -in /run/secrets/cert -subject | grep '^subject=CN=' | sed 's/^subject=CN=//')
```
cgwalters: Having a second copy of this code isn't fatal, but I think we could factor out a shared helper for this too.

tests/build-sealed
```diff
@@ -61,9 +56,30 @@ if test -z "${secureboot}"; then
...
+runv podman build -t $sdboot_signed \
   --build-arg=base=${input_image} \
   --build-arg=buildroot=${buildroot_base} \
   --secret=id=key,src=${secureboot}/db.key \
-  --secret=id=cert,src=${secureboot}/db.crt -f Dockerfile.cfsuki .
+  --secret=id=cert,src=${secureboot}/db.crt \
+  --secret=id=cert,src=${secureboot}/db.crt \
+  --secret=id=pk_key,src=${secureboot}/PK.key \
+  --secret=id=pk_cert,src=${secureboot}/PK.crt \
+  --secret=id=kek_key,src=${secureboot}/KEK.key \
```
cgwalters: We're only using `key` and `cert` in this one too right?

gerblesh: yup that's leftover, should be fixed now

gerblesh:
> Thanks again for starting this! Is it OK if I try some force pushing to this PR and we co-author?

Yeah go for it, just was cleaning up the comments and extra mounts

cgwalters:
OK what this is blocking on is I think we're not detecting the OVMF firmware correctly in bcvk on Debian derivatives. Looking at that

crates/xtask/src/tmt.rs
```diff
@@ -597,11 +628,36 @@ pub(crate) fn tmt_provision(sh: &Shell, args: &TmtProvisionArgs) -> Result<()> {
     println!("  Image: {}", image);
     println!("  VM name: {}\n", vm_name);
 
+    // Detect if this is a sealed image and build firmware args accordingly
```
cgwalters: Dedup this code

crates/xtask/src/tmt.rs
```diff
@@ -597,11 +628,36 @@ pub(crate) fn tmt_provision(sh: &Shell, args: &TmtProvisionArgs) -> Result<()> {
...
+        println!(
+            "Sealed image detected, using secure boot with keys from: {}",
+            sb_keys_dir
+        );
+        vec![
+            "--firmware=uefi-secure".to_string(),
+            format!("--secure-boot-keys={}", sb_keys_dir),
+        ]
+    } else {
+        if is_sealed {
```
cgwalters: This fallback doesn't make sense

cgwalters:
Thanks for the contribution @gerblesh! This looks like the right approach to fix the systemd-boot signing issue. I'm going through the code now and will leave more specific comments if I have any.

cgwalters:
Sorry, those comments were a misfire from my non-sandboxed agent; I'm going to ensure that doesn't happen again.

cgwalters:
This one is ready to merge I think but as always a lot of potential followup here.

Hum looks like the rawhide failures might be a nushell regression/semantic change, but not related.

cgwalters:
@jmarrero you're assigned reviewer looks like

cgwalters:
(Technically I can merge this since it's someone else's PR but I added quite a lot of code here so someone else should at least do a medium-level review)

tmt/plans/integration.fmf
```diff
@@ -102,17 +102,6 @@ execute:
     test:
       - /tmt/tests/tests/test-25-download-only-upgrade
 
-/plan-26-examples-build:
-  summary: Test bootc examples build scripts
```
cgwalters, line 106: cc @jeckersb sorry I should have called out explicitly here for you that I was dropping this.

The core problem is we had *two* sealing impls going on, one out of the main Dockerfile and one in here, and it was hard to juggle ensuring I'd applied the correct fixes to both of them.

There was also some incorrect cruft here (copying in initramfs services etc.)

I think per discussion let's aim to do https://github.com/bootc-dev/bootc/issues/1498#issuecomment-3612481162

Crucially if we do that after https://github.com/bootc-dev/bootc/pull/1858 merges it becomes a lot more clear that the sealing aspect of the build is separate from building bootc, so it should then be easier to ensure that the sealing logic can also be applied generically outside of building the code here, and then we migrate it to say https://gitlab.com/fedora/bootc/examples/ ?

This relates to https://gitlab.com/fedora/bootc/docs/-/merge_requests/143

PR #1810, opened 2025-11-26
test: Add more distros for composefs test

henrywang:
This PR adds CS9, Fedora 42/43/44 distros for composefs test.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -136,7 +136,7 @@ jobs:
 
       - name: Setup env
         run: |
-          BASE=$(just pullspec-for-os ${{ matrix.test_os }})
+          BASE=$(just pullspec-for-os base ${{ matrix.test_os }})
           echo "BOOTC_base=${BASE}" >> $GITHUB_ENV
 
       - name: Build container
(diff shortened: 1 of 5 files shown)


cgwalters:
> Let's just merge as is and debug that as a followup.

Ah well no can't do that because that's a required job.

henrywang:
Okey, AI tells me the issue on C9S is older kernel and older virtiofs/virtiofsd userspace. The composefs install does many small-file metadata ops (walk /etc/pki/..., copy many small cert files, systemd unit symlinks, etc.). That pattern reliably triggers older virtiofs/virtiofsd bugs that return EUCLEAN (“Structure needs cleaning”) during stat/cp operations.

cgwalters:
Maybe. In this case virtiofsd is on the host, but yeah it could be the guest virtiofs support.

Anyways...well, let's skip it for now with a TODO? Or we could do this https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax?versionId=free-pro-team%40latest&productId=actions&search-overlay-input=continue-on-error+docs&search-overlay-ask-ai=true#example-preventing-a-specific-failing-matrix-job-from-failing-a-workflow-run

Or if we want to narrow in on if it's actually a virtiofs problem, we could run the test in TF (though we'd need to control the filesystem, which really leads to us supporting a kickstart or Ignition flow)

henrywang:
> BTW though, one thing we should do soon is split up a "build rpm" phase for the GHA because now we're building from source fully twice per OS.

Do you mean in `test-integration-cfs`, the workflow build rpm twice? One by `- name: Build container`, another by `just test-composefs` in `- name: Run TMT tests`?

henrywang:
Yeah, I file an issue for tracking and use `continue-on-error` for centos-9 test.

henrywang:
Two failures can be tracked by issues https://github.com/bootc-dev/bootc/issues/1814 and https://github.com/bootc-dev/bootc/issues/1812. Both of them will be debugged and fixed later.

PR #1813, opened 2025-11-26
ci: Split RPM building into separate job

cgwalters:
This splits the RPM package building into a separate CI job that runs
before the integration tests. The built packages are then downloaded
and used by the integration test jobs, avoiding redundant builds.

### DIFF
--- a/tmt/tests/booted/test-image-pushpull-upgrade.nu
+++ b/tmt/tests/booted/test-image-pushpull-upgrade.nu
@@ -86,6 +86,11 @@ RUN echo test content > /usr/share/blah.txt
     # Test for https://github.com/ostreedev/ostree/issues/3544
     # Add a quoted karg using rpm-ostree if available
     if not $is_composefs {
+        # Check rpm-ostree and rpm-ostreed service status before run rpm-ostree
+        # And collect info for flaky error "error: System transaction in progress"
+        rpm-ostree status
+        journalctl -u rpm-ostreed
+
         print "Adding quoted karg via rpm-ostree to test ostree issue #3544"
         rpm-ostree kargs --append=($quoted_karg)
     }


cgwalters:
OK pushed this too early, this one has some broken bits, still iterating. But hopefully illustrates the idea

henrywang:
Hi @cgwalters, can we land this PR? Thanks.

cgwalters:
Hm, did the rebase bot not work?

henrywang:
> Hm, did the rebase bot not work?

I'm working on a PR to grant our bot `workflow` `write` permission. We use our bot token, not github token. It should be secure.

cgwalters:
> I'm working on a PR to grant our bot workflow write permission. We use our bot token, not github token. It should be secure.

Yep agree. That said, I think it can't be a PR, an organization admin needs to edit the bot permissions right now (unless we switch to something to git-ops it in the infra repo).

Also on this topic, while I agree workflow edit perms should be fine, we also do want to be sure we actually *are* careful that no code submitted as part of a pull request gets access to that bot token.

henrywang:
Granting workflow write permission to our bot is safe because rebase.yml only checks out the base repo. It's not checking out untrusted PR code — it's checking out the base repository. No attacker can inject code into that unless they already have write access.

I think we can grant our bot `workflow write` permission from organization admin UI.

.github/workflows/ci.yml
```diff
@@ -72,6 +72,8 @@ jobs:
           sudo just build-install-test-image
           sudo podman build -t localhost/bootc-fsverity -f ci/Containerfile.install-fsverity
 
+          # Grant permission
+          sudo chown -R "$(id -u):$(id -g)" /home/runner/work/bootc/bootc
```
cgwalters, line 75: Yeah, we need to get these tests out of live mutating the host. There's no need for it anymore; the nested virt stuff is working well.

henrywang:
> I'd like to get https://github.com/bootc-dev/bootc/pull/1809 in first though

OK, let's land https://github.com/bootc-dev/bootc/pull/1809 first and I'll make rebase.

henrywang:
@cgwalters, I changed this PR to `ready to review`. Two CS10 flaky issues should be fixed by https://github.com/bootc-dev/bootc/pull/1817. May I approve and land (auto-merge after approve) this PR? Thanks.

cgwalters:
Looks sane to me, but this is technically my PR, so I can't approve - please go ahead and do so!

cgwalters:
Turns out this didn't quite work because `just test-container` currently rebuilds. I think what I want to do is have some automation for syncing "justfile" => GHA effectively - i.e. enforce that each GHA step maps to something invokable via `just` but the individual stages are rendered as such.

henrywang:
> Turns out this didn't quite work because `just test-container` currently rebuilds. I think what I want to do is have some automation for syncing "justfile" => GHA effectively - i.e. enforce that each GHA step maps to something invokable via `just` but the individual stages are rendered as such.

Right, I'm working on this now.

PR #1815, opened 2025-11-28
Fix undefined BOOTC_buildroot_base in build-sealed script

shi2wei3:
The build-sealed script introduced in PR #1810 referenced BOOTC_buildroot_base which is only defined in GitHub Actions CI, causing failures when running 'just build-sealed' manually.

This allows manual execution while maintaining CI compatibility.

### DIFF
--- a/tests/build-sealed
+++ b/tests/build-sealed
@@ -11,7 +11,7 @@ shift
 output_image=$1
 shift
 # Buildroot base image for UKI build environment
-buildroot_base=${1:-quay.io/centos/centos:stream10}
+buildroot_base=$1
 shift
 # Optional directory with secure boot keys; if none are provided, then we'll
 # generate some under target/


Justfile
```diff
@@ -22,6 +22,7 @@ integration_upgrade_img := integration_img + "-upgrade"
 # composefs-sealeduki-sdboot: A system with a sealed composefs using systemd-boot
 variant := env("BOOTC_variant", "ostree")
 base := env("BOOTC_base", "quay.io/centos-bootc/centos-bootc:stream10")
+buildroot_base := env("BOOTC_buildroot_base", "quay.io/centos/centos:stream10")
```
cgwalters, line 25: I think this is a correct fix, but it is a bit of an ergonomic problem to have two bases and it'd probably simplify things to just reuse `base` for this by default.

cgwalters (approved):
See below

Issue #1559, opened 2025-08-28
install to-filesystem: look into the container for grub configs with `--src-imgref`


jbtrystram:
When using `--src-imgref`, bootupd fetches the static configs from the host rather than the destination container. 

An easy way around would be to `chroot` to the deployed path before the bootupd invocation (this would have the added bonus of using the bootupd binary that is shipped into the image rather than the one in the host).

e.g like in https://github.com/osbuild/osbuild/blob/main/stages/org.osbuild.bootupd

When building FCOS i got only the configs that exist in COSA rather than what is in the container (thus missing the ignition fragment)

Related: https://github.com/bootc-dev/bootc/issues/1455

jbtrystram:
Hmmm maybe it's more a bug in bootupd because we already specify `--src-root`.

cgwalters:
Yeah this is probably a bootupd bug.

That said it would also be really nice always run bootupd from the target root, we can totally do that, it just requires some plumbing work.

cgwalters:
This was fixed by https://github.com/coreos/bootupd/pull/990

Issue #1455, opened 2025-07-29
install : run bootupd with ` --src-root` pointing to the deployed container


jbtrystram:
Currently the bootc install makes assumptions about being run from the container. It mostly work outside of it, but also assume the buildroot is the same as the container content. (see https://github.com/bootc-dev/bootc/issues/1438 and https://github.com/bootc-dev/bootc/issues/1410).

the bootloader setup runs into the same pitfall : it expects `/usr/lib/bootupd/updates/<component>` to exists, but it may not if we are using a different buildroot. 
If i understand correctly, the call to bootupd is done after the container content is deployed, so bootupd could be pointed to the deployment.
The docs even outline the ostree use case : 

https://github.com/coreos/bootupd/blob/3acb6742f74be63c655c64c6adff57b2c2fb93cc/README-devel.md?plain=1#L44-L53

In CoreOS builds through osbuild we `chroot` before calling bootupd to work around this: 

https://github.com/osbuild/osbuild/blob/841e89fe08cb80d115871aa45696e37fae41cf23/stages/org.osbuild.bootupd#L67-L72

Related discussion in https://github.com/bootc-dev/bootc/issues/879

jbtrystram:
I tried something naive to see if that would work : 
```
diff --git a/crates/lib/src/bootloader.rs b/crates/lib/src/bootloader.rs
index fe7aa46d..fd67e4aa 100644
--- a/crates/lib/src/bootloader.rs
+++ b/crates/lib/src/bootloader.rs
@@ -20,10 +20,19 @@ pub(crate) fn install_via_bootupd(
     let bootupd_opts = (!configopts.generic_image).then_some(["--update-firmware", "--auto"]);
 
     let devpath = device.path();
+    let srcroot = format!(
+        "{}/deploy/{}/deploy/<ostree-ref>/",
+        rootfs.as_str(),
+        configopts
+            .stateroot
+            .clone()
+            .unwrap_or("default".to_string())
+    );
     let args = ["backend", "install", "--write-uuid"]
         .into_iter()
         .chain(verbose)
         .chain(bootupd_opts.iter().copied().flatten())
+        .chain(["--src-root", srcroot.as_str()])
         .chain(["--device", devpath.as_str(), rootfs.as_str()]);
     Task::new("Running bootupctl to install bootloader", "bootupctl")
         .args(args)
```

but i don't know how I can resolve the ostree deployment ID

cgwalters:
[...]

PR #1816, opened 2025-11-29
install/bootupd: chroot to deployment

jbtrystram:
When `--src-imgref` is passed, the deployed systemd does not match the running environnement. In this case, let's chroot into the deployment before calling bootupd. This makes sure we are using the binaries shipped in the image (and relevant config files such as grub fragements).

We could do that in all cases but i kept it behind the `--src-imgref` option since when using the target container as the buildroot it will have no impact, and we expect this scenario to be the most common.

In CoreOS we have a specific test that checks if the bootloader was installed with the `grub2-install` of the image.

Fixes https://github.com/bootc-dev/bootc/issues/1559 Also see https://github.com/bootc-dev/bootc/issues/1455

### DIFF
--- a/crates/lib/src/bootloader.rs
+++ b/crates/lib/src/bootloader.rs
@@ -2,7 +2,7 @@ use std::fs::create_dir_all;
 use std::process::Command;
 
 use anyhow::{Context, Result, anyhow, bail};
-use bootc_utils::CommandRunExt;
+use bootc_utils::{BwrapCmd, CommandRunExt};
 use camino::Utf8Path;
 use cap_std_ext::cap_std::fs::Dir;
 use cap_std_ext::dirext::CapStdExtDirExt;
@@ -91,22 +91,71 @@ pub(crate) fn install_via_bootupd(
     // bootc defaults to only targeting the platform boot method.
     let bootupd_opts = (!configopts.generic_image).then_some(["--update-firmware", "--auto"]);
 
-    let abs_deployment_path = deployment_path.map(|v| rootfs.join(v));
-    let src_root_arg = if let Some(p) = abs_deployment_path.as_deref() {
-        vec!["--src-root", p.as_str()]
+    // When not running inside the target container (through `--src-imgref`) we use
+    // will bwrap as a chroot to run bootupctl from the deployment.
+    // This makes sure we use binaries from the target image rather than the buildroot.
+    // In that case, the target rootfs is replaced with `/` because this is just used by
+    // bootupd to find the backing device.
+    let rootfs_mount = if deployment_path.is_none() {
+        rootfs.as_str()
     } else {
-        vec![]
+        "/"
     };
-    let devpath = device.path();
+
     println!("Installing bootloader via bootupd");
-    Command::new("bootupctl")
-        .args(["backend", "install", "--write-uuid"])
-        .args(verbose)
-        .args(bootupd_opts.iter().copied().flatten())
-        .args(src_root_arg)
-        .args(["--device", devpath.as_str(), rootfs.as_str()])
-        .log_debug()
-        .run_inherited_with_cmd_context()
+
+    // Build the bootupctl arguments
+    let mut bootupd_args: Vec<&str> = vec!["backend", "install", "--write-uuid"];
+    if let Some(v) = verbose {
+        bootupd_args.push(v);
+    }
+
+    if let Some(ref opts) = bootupd_opts {
+        bootupd_args.extend(opts.iter().copied());
+    }
+    bootupd_args.extend(["--device", device.path().as_str(), rootfs_mount]);
+
+    // Run inside a bwrap container. It takes care of mounting and creating
+    // the necessary API filesystems in the target deployment and acts as
+    // a nicer `chroot`.
+    if let Some(deploy) = deployment_path {
+        let target_root = rootfs.join(deploy);
+        let boot_path = rootfs.join("boot");
+
+        tracing::debug!("Running bootupctl via bwrap in {}", target_root);
+
+        // Prepend "bootupctl" to the args for bwrap
+        let mut bwrap_args = vec!["bootupctl"];
+        bwrap_args.extend(bootupd_args);
+
+        let mut cmd = BwrapCmd::new(&target_root)
+            // Bind mount /boot from the physical target root so bootupctl can find
+            // the boot partition and install the bootloader there
+            .bind(&boot_path, &"/boot")
+            // Bind the target block device inside the bwrap container so bootupctl can access it
+            .bind_device(device.path().as_str());
+
+        // Also bind all partitions of the tafet block device
+        for partition in &device.partitions {
+            cmd = cmd.bind_device(&partition.node);
+        }
+
+        // The $PATH in the bwrap env is not complete enough for some images
+        // so we inject a reasonnable default.
+        // This is causing bootupctl and/or sfdisk binaries
+        // to be not found with fedora 43.
+        cmd.setenv(
+            "PATH",
+            "/bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin",
+        )
+        .run(bwrap_args)
+    } else {
+        // Running directly without chroot
+        Command::new("bootupctl")
+            .args(&bootupd_args)
+            .log_debug()
+            .run_inherited_with_cmd_context()
+    }
 }
 
 #[context("Installing bootloader")]
--- a/crates/utils/src/bwrap.rs
+++ b/crates/utils/src/bwrap.rs
@@ -1,0 +1,106 @@
+/// Builder for running commands inside a target os tree using bubblewrap (bwrap).
+use std::borrow::Cow;
+use std::ffi::OsStr;
+use std::os::fd::AsRawFd;
+use std::process::Command;
+
+use anyhow::Result;
+use cap_std_ext::camino::{Utf8Path, Utf8PathBuf};
+use cap_std_ext::cap_std::fs::Dir;
+
+use crate::CommandRunExt;
+
+/// Builder for running commands inside a target directory using bwrap.
+#[derive(Debug)]
+pub struct BwrapCmd<'a> {
+    /// The target directory to use as root for the container
+    chroot_path: Cow<'a, Utf8Path>,
+    /// Bind mounts in format (source, target)
+    bind_mounts: Vec<(&'a str, &'a str)>,
+    /// Device nodes to bind into the container
+    devices: Vec<&'a str>,
+    /// Environment variables to set
+    env_vars: Vec<(&'a str, &'a str)>,
+}
+
+impl<'a> BwrapCmd<'a> {
+    /// Create a new BwrapCmd builder with a root directory as a File Descriptor.
+    #[allow(dead_code)]
+    pub fn new_with_dir(path: &'a Dir) -> Self {
+        let fd_path: String = format!("/proc/self/fd/{}", path.as_raw_fd());
+        Self {
+            chroot_path: Cow::Owned(Utf8PathBuf::from(&fd_path)),
+            bind_mounts: Vec::new(),
+            devices: Vec::new(),
+            env_vars: Vec::new(),
+        }
+    }
+
+    /// Create a new BwrapCmd builder with a root directory
+    pub fn new(path: &'a Utf8Path) -> Self {
+        Self {
+            chroot_path: Cow::Borrowed(path),
+            bind_mounts: Vec::new(),
+            devices: Vec::new(),
+            env_vars: Vec::new(),
+        }
+    }
+
+    /// Add a bind mount from source to target inside the container.
+    pub fn bind(
+        mut self,
+        source: &'a impl AsRef<Utf8Path>,
+        target: &'a impl AsRef<Utf8Path>,
+    ) -> Self {
+        self.bind_mounts
+            .push((source.as_ref().as_str(), target.as_ref().as_str()));
+        self
+    }
+
+    /// Bind a device node into the container.
+    pub fn bind_device(mut self, device: &'a str) -> Self {
+        self.devices.push(device);
+        self
+    }
+
+    /// Set an environment variable for the command.
+    pub fn setenv(mut self, key: &'a str, value: &'a str) -> Self {
+        self.env_vars.push((key, value));
+        self
+    }
+
+    /// Run the specified command inside the container.
+    pub fn run<S: AsRef<OsStr>>(self, args: impl IntoIterator<Item = S>) -> Result<()> {
+        let mut cmd = Command::new("bwrap");
+
+        // Bind the root filesystem
+        cmd.args(["--bind", self.chroot_path.as_str(), "/"]);
+
+        // Setup API filesystems
+        // See https://systemd.io/API_FILE_SYSTEMS/
+        cmd.args(["--proc", "/proc"]);
+        cmd.args(["--dev", "/dev"]);
+        cmd.args(["--ro-bind", "/sys", "/sys"]);
+
+        // Add bind mounts
+        for (source, target) in &self.bind_mounts {
+            cmd.args(["--bind", source, target]);
+        }
+
+        // Add device bind mounts
+        for device in self.devices {
+            cmd.args(["--dev-bind", device, device]);
+        }
+
+        // Add environment variables
+        for (key, value) in &self.env_vars {
+            cmd.args(["--setenv", key, value]);
+        }
+
+        // Command to run
+        cmd.arg("--");
+        cmd.args(args);
+
+        cmd.log_debug().run_inherited_with_cmd_context()
+    }
+}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,6 +89,10 @@ jobs:
           # Install tests
           sudo bootc-integration-tests install-alongside localhost/bootc-install
 
+          # inspect system state after the install tests.
+          sudo lsblk
+          sudo mount
+          
           # system-reinstall-bootc tests
           cargo build --release -p system-reinstall-bootc
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -194,6 +194,7 @@ version = "0.1.0"
 dependencies = [
  "anstream",
  "anyhow",
+ "cap-std-ext",
  "chrono",
  "owo-colors",
  "rustix",
--- a/crates/utils/Cargo.toml
+++ b/crates/utils/Cargo.toml
@@ -10,6 +10,7 @@ repository = "https://github.com/bootc-dev/bootc"
 # Workspace dependencies
 anstream = { workspace = true }
 anyhow = { workspace = true }
+cap-std-ext = {workspace = true, features = ["fs_utf8"] }
 chrono = { workspace = true, features = ["std"] }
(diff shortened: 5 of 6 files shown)


crates/lib/src/bootloader.rs
```diff
@@ -50,22 +51,60 @@ pub(crate) fn install_via_bootupd(
     // bootc defaults to only targeting the platform boot method.
     let bootupd_opts = (!configopts.generic_image).then_some(["--update-firmware", "--auto"]);
 
-    let abs_deployment_path = deployment_path.map(|v| rootfs.join(v));
-    let src_root_arg = if let Some(p) = abs_deployment_path.as_deref() {
-        vec!["--src-root", p.as_str()]
+    let abs_deployment_path = deployment_path.map(|deploy| rootfs.join(deploy));
+    // When not running inside the target container (through `--src-imgref`) we chroot
```
cgwalters: There's other threads were we talked about offering a `bootc install mount` as a general ability to mount a deployment outside of booting it; were we to do that it would make a lot of sense for this code to use it.

In ostree we resisted doing that for a long time but eventually did just internally for selinux, see https://github.com/ostreedev/ostree/blob/c6f0b5b2bc26b22fbceee0dc28a0f31349c28d41/src/libostree/ostree-sysroot-deploy.c#L3308

On that topic, it'd be a lot cleaner even here to use a more proper containerization than just setting up the mounts. It's a bit tricky though because we actually do need to e.g. pass through all of `/dev` and `/sys` here (i.e. `--privileged` in docker/podman terms) in order to update the ESP if desired.

I haven't looked at which of bwrap/{runc,crun}/nspawn/podman would make the most sense for this use case.

crates/lib/src/bootloader.rs
```diff
@@ -50,22 +51,60 @@ pub(crate) fn install_via_bootupd(
...
         .args(["backend", "install", "--write-uuid"])
         .args(verbose)
         .args(bootupd_opts.iter().copied().flatten())
-        .args(src_root_arg)
         .args(["--device", devpath.as_str(), rootfs.as_str()])
         .log_debug()
-        .run_inherited_with_cmd_context()
+        .run_inherited_with_cmd_context();
+
+    // Clean up the mounts after ourselves
```
cgwalters: We could entirely avoid the need to clean up by using the new mount API to get file descriptors instead, and then use https://docs.rs/cap-std-ext/latest/cap_std_ext/cmdext/trait.CapStdExtCommandExt.html#tymethod.cwd_dir with `chroot .` or so

crates/lib/src/bootloader.rs
jbtrystram: > On that topic, it'd be a lot cleaner even here to use a more proper containerization than just setting up the mounts. It's a bit tricky though because we actually do need to e.g. pass through all of `/dev` and `/sys` here (i.e. `--privileged` in docker/podman terms) in order to update the ESP if desired.
> 
> I haven't looked at which of bwrap/{runc,crun}/nspawn/podman would make the most sense for this use case.

I am not sure of what you mean with this comment. Do you want to block this change until there are more proper containerization helpers in bootc, or are you just making a note that this should be revisited later on ?

cgwalters: We had a live chat about this and agreed to merge as is and file a tracker followup issue for improving the mount setup.

cgwalters:
OK there's some legit failures here like `  content: error: boot data installation failed: installing component EFI: Listing partitions of /dev/loop0: No such file or directory (os error 2)`.

cgwalters (changes requested):
Marking as requested changes due to failing CI

crates/lib/src/bootloader.rs
```diff
@@ -50,22 +52,70 @@ pub(crate) fn install_via_bootupd(
...
+    // from the target image rather than the buildroot
+    // In some cases (e.g. --write-uuid), bootupd needs to find the underlying device
+    // for /boot. But since we don't control where the destination rootfs is mounted
+    // let's bind mount it to a temp mountpoint under /run
+    // so it gets carried over in the chroot.
+
+    let rootfs_mount = if rootfs.starts_with("/run") {
+        rootfs.to_path_buf().into_std_path_buf()
+    } else {
+        let rootfs_mount = tempfile::tempdir_in("/run")?.keep();
```
cgwalters: Let's hoist the tempdir up and let its Drop handler deallocate

crates/lib/src/bootloader.rs
```diff
@@ -50,22 +52,70 @@ pub(crate) fn install_via_bootupd(
...
+            let dest = target_root
+                // joining an absolute path
+                // makes it replace self, so we strip the prefix
+                .join_os(src.strip_prefix("/").unwrap());
+            tracing::debug!("bind mounting {}", dest.display());
+            rustix::mount::mount_bind_recursive(src, dest)?;
+        }
+        // Append the `bootupctl` command, it will be passed as
+        // an argument to chroot
+        vec![target_root.as_str(), "bootupctl"]
```
cgwalters: Can't we just pass `/` here instead of target_root?

crates/lib/src/bootloader.rs
```diff
@@ -50,22 +52,70 @@ pub(crate) fn install_via_bootupd(
...
+    // so it gets carried over in the chroot.
+
+    let rootfs_mount = if rootfs.starts_with("/run") {
+        rootfs.to_path_buf().into_std_path_buf()
+    } else {
+        let rootfs_mount = tempfile::tempdir_in("/run")?.keep();
+        rustix::mount::mount_bind_recursive(rootfs.as_std_path(), &rootfs_mount)?;
+        rootfs_mount
+    };
+    let bind_mount_dirs = ["/dev", "/run", "/proc", "/sys"];
```
cgwalters: Worth xref to https://systemd.io/API_FILE_SYSTEMS/

crates/lib/src/bootloader.rs
```diff
@@ -50,22 +52,70 @@ pub(crate) fn install_via_bootupd(
...
+    let abs_deployment_path = deployment_path.map(|deploy| rootfs.join(deploy));
+    // When not running inside the target container (through `--src-imgref`) we chroot
+    // into the deployment before running bootupd. This makes sure we use binaries
+    // from the target image rather than the buildroot
+    // In some cases (e.g. --write-uuid), bootupd needs to find the underlying device
+    // for /boot. But since we don't control where the destination rootfs is mounted
+    // let's bind mount it to a temp mountpoint under /run
+    // so it gets carried over in the chroot.
+
+    let rootfs_mount = if rootfs.starts_with("/run") {
```
cgwalters: Thinking about this it may be a lot simpler to just setup the API filesystems by default in the install flow.

(and per discussion mount e.g. `/boot` in the target rootfs)

cgwalters:
Note that it's only Fedora variants that are failing; should reproduce locally via e.g. `just base=quay.io/fedora/fedora-bootc:42 test-tmt install-unified`

jbtrystram:
> Note that it's only Fedora variants that are failing; should reproduce locally via e.g. `just base=quay.io/fedora/fedora-bootc:42 test-tmt install-unified`

Yes, i can reproduce that locally ! Ok so I dug deeper and I think I figured out something :
These tmt test run the install process through a [systemd transient unit with `MountFlags=Slave`](https://github.com/bootc-dev/bootc/blob/64ad5c539f1fd779fe7b4887cbf681108fc352d8/tmt/tests/booted/test-install-unified-flag.nu#L30), which cause the `/dev/loop` device to not get mounted inside the chroot target. 

I am not sure what happens because removing the `MountFlags` does not fix it, but removing the systemd-run wrapper does help : 

```
truncate -s 10G disk.img
systemd-run  -qdPG -- /bin/sh -c $"./bootc install to-disk --disable-selinux --via-loopback --filesystem xfs  --source-imgref docker://quay.io/centos-bootc/centos-bootc:stream10 ./disk.img"
....
error: boot data installation failed: installing component EFI: Listing partitions of /dev/loop0: No such file or directory (os error 2)
```

Without the systemd wrapper : 
```
truncate -s 10G disk.img
./bootc install to-disk --disable-selinux --via-loopback --filesystem xfs  --source-imgref docker://quay.io/centos-bootc/centos-bootc:stream10 ./disk.img
.....
Bootloader: grub
Added 01_users.cfg
Added 10_blscfg.cfg
Added 14_menu_show_once.cfg
Added 30_uefi-firmware.cfg
Added 41_custom.cfg
Installed: grub.cfg
Installed: bootuuid.cfg
Installed: "centos/grub.cfg"
Installed: "centos/bootuuid.cfg"
```

In the second case the chroot works.

Another thing to note, and i cannot figure this out yet : the error finding the `/dev/loop0` device is yielded from [bootc code](https://github.com/bootc-dev/bootc/blob/64ad5c539f1fd779fe7b4887cbf681108fc352d8/crates/blockdev/src/blockdev.rs#L221-L227) and not bootupd. 
Ah, nevermind : it's pulled in bootupd : https://github.com/coreos/bootupd/blob/78dc9baea7e8418e040ba0217caba63b25dcfb75/src/blockdev.rs#L39-L53

jbtrystram:
Thinking more about this : in both cases the mounting and formating of the block device (`disk.img`) through the loopback device works. So it's systemd not letting us bind mount inside the chroot ?

jbtrystram:
Ok so after doing a bunch of testing around `systemd-run` it's not the cause. 
I ended up changing the `bootloader.rs` code to simply do : `chroot <deployment_path> /usr/sbin/sfdisk -J /dev/loop0` and it works !? But it does not work without specifying `sfdisk` full path. So maybe it's down to a `$PATH` issue ?

I will try to add the full path but that requires a rebuild of bootupd as well, so that's for tomorrow. 

edit: ok I did ask bootc to run `which sfdisk` in the chroot and sure enough: `no sfdisk in (/usr/local/bin:/usr/bin)`

crates/lib/src/bootloader.rs
jbtrystram: we would then need to have another bind mount `rootfs/boot` into `deployment_path/boot` I think.

jbtrystram: Ok so i ended up doing just that and now more the tests are passing.

jbtrystram:
Alright so I finally figured out a working solution ! For posterity i'll leave a small writeup :

I figured out why the CI was failing in my previous manual `chroot` + `bind-mount` attempts : the recursive bind-mount of `/dev/` inside the chroot was messing up with the host's `/dev/pts`.
This lead the `install to-exising-root` tests to be unable to allocate a PTY to answer the prompts.
 
The chroot was technically working, but leaving the host in a broken state, requiring a reboot. 
I tried to fix `/dev/pts` manually but then hit another issue with `/dev/shm` and realized I was probably re-inventing the wheel. 

I tried systemd-nspawn to handle all the API filesystems better than me manually, which seemed to work just fine in my initial testing, until i figured out the base bootc images don't ship the `systemd-container` rpm. 

I then tried `podman run --rootfs` but containers inside containers are weird and require a bunch more mounts in the initial podman invocation. I thought that would lead to a terrible UX. 

`bwrap` was the easy answer in the end.

The `bwrap` code was added in the util crate in the hope that it get reused with the composefs backend to achieve the same effect for composefs + systemd boot. This however requires mounting the EROFS first in a temporary directory.

crates/utils/src/bwrap.rs
```diff
@@ -0,0 +1,83 @@
...
+
+use anyhow::Result;
+
+use crate::CommandRunExt;
+
+/// Builder for running commands inside a target directory using bwrap.
+#[derive(Debug, Default)]
+pub struct BwrapCmd<'a> {
+    /// The target directory to use as root for the container
+    chroot_path: &'a str,
```
cgwalters: Fine as is but I would like to add an option that takes a `Dir` instead.

crates/lib/src/bootloader.rs
```diff
@@ -91,22 +91,67 @@ pub(crate) fn install_via_bootupd(
...
+            // the boot partition and install the bootloader there
+            .bind(boot_path.as_str(), "/boot")
+            // Bind the target block device inside the bwrap container so bootupctl can access it
+            .bind_device(device.path().as_str());
+
+        // Also bind all partitions of the tafet block device
+        for partition in &device.partitions {
+            cmd = cmd.bind_device(&partition.node);
+        }
+        // // TODO : is it needed ?
```
cgwalters: In the general case - maybe, and especially when talking about cross-distro.

Especially when run outside of a container in the general case we could get all sorts of stuff leaking into our environment - like `LD_LIBRARY_PATH` or `LD_PRELOAD` and that could easily break because now the environment is different from the target root.

But in practice, we can just let it be inherited for now.

crates/utils/src/bwrap.rs
```diff
@@ -0,0 +1,83 @@
...
+    /// Create a new BwrapCmd builder with the given root directory.
+    pub fn new(path: &'a str) -> Self {
+        Self {
+            chroot_path: path,
+            ..Default::default()
+        }
+    }
+
+    /// Add a bind mount from source to target inside the container.
+    pub fn bind(mut self, source: &'a str, target: &'a str) -> Self {
```
cgwalters: I would make these `impl AsRef<Utf8Path>` or so (or if we wanted to be looser, `impl AsRef<Path>`)

crates/utils/src/bwrap.rs
```diff
@@ -0,0 +1,83 @@
...
+        cmd.args(["--dev", "/dev"]);
+        cmd.args(["--ro-bind", "/sys", "/sys"]);
+
+        // Add bind mounts
+        for (source, target) in &self.bind_mounts {
+            cmd.args(["--bind", source, target]);
+        }
+
+        // Add device bind mounts
+        for device in self.devices {
```
cgwalters, line 91: What would actually probably work most reliably in general is for us to pass the block device as a file descriptor.

cgwalters:
There was one failure in fedora-43/ostree which *might* have been related to this; I restarted it. Clearly one thing that would help again is for us to lock the blockdev and pass it down as a fd and not a path through the stack.

crates/lib/src/bootloader.rs
jbtrystram: So this is a leftover from the manual `chroot`, but I noticed that bwrap does set this up for us so I commented it out.

You're right about other things that could leak in though

jbtrystram:
> There was one failure in fedora-43/ostree which _might_ have been related to this; I restarted it. Clearly one thing that would help again is for us to lock the blockdev and pass it down as a fd and not a path through the stack.

I think that's because of the commented out $PATH. I'm testing locally to verify.

crates/utils/src/bwrap.rs
jbtrystram: IIUC that would require bootupd to accept the device as a FD rather than a path. so we can't change that here for now, correct ?

jbtrystram:
> > There was one failure in fedora-43/ostree which _might_ have been related to this; I restarted it. Clearly one thing that would help again is for us to lock the blockdev and pass it down as a fd and not a path through the stack.
> 
> I think that's because of the commented out $PATH. I'm testing locally to verify.

So the PATH fix I tried did not help. 
Right now `30-install-unified-flag` and `23-install-outside-container` are failing .

crates/utils/src/bwrap.rs
jbtrystram: Oh, or we can pass `/proc/ns/fd/{rawFd}` I guess ?

crates/utils/src/bwrap.rs
jbtrystram: Fixed in the last push

crates/utils/src/bwrap.rs
jbtrystram: done in the last push

crates/utils/src/bwrap.rs
HuijingHei: If passing `/proc/ns/fd/{rawFd}`, when mount ESP, bootupd will be updated to get uuid instead of node, as the node will not exist.
E.g.
```
[root@cosa-devsh core]# exec 3</dev/vda
[root@cosa-devsh core]# sfdisk -J /proc/self/fd/3
{
   "partitiontable": {
      "label": "gpt",
      "id": "77B17F5E-C5DA-419A-B200-BCD3BAE39D35",
      "device": "/proc/self/fd/3",
      "unit": "sectors",
      "firstlba": 2048,
      "lastlba": 33554398,
      "sectorsize": 512,
      "partitions": [
         {
            "node": "/proc/self/fd/3p1",
            "start": 2048,
            "size": 2048,
            "type": "21686148-6449-6E6F-744E-656564454649",
            "uuid": "431E16C0-3582-4C42-B8A5-38FF61E588BA",
            "name": "BIOS-BOOT"
         },{
            "node": "/proc/self/fd/3p2",
            "start": 4096,
            "size": 260096,
            "type": "C12A7328-F81F-11D2-BA4B-00A0C93EC93B",
            "uuid": "7295F72A-E852-CC4F-8606-444EDA295B72",
            "name": "EFI-SYSTEM"
         },{
            "node": "/proc/self/fd/3p3",
            "start": 264192,
            "size": 786432,
            "type": "BC13C2FF-59E6-4262-A352-B275FD6F7172",
            "uuid": "B2EFF31A-187F-B14F-BCDF-A8C07BDEFD04",
            "name": "boot"
         },{
            "node": "/proc/self/fd/3p4",
            "start": 1050624,
            "size": 32503775,
            "type": "4F68BCE3-E8CD-4DB1-96E7-FBCAF984B709",
            "uuid": "57D12571-0713-4A48-92BA-6010C1A59C10",
            "name": "root"
         }
      ]
   }
}
[root@cosa-devsh core]# ls -l /proc/self/fd/3p2
ls: cannot access '/proc/self/fd/3p2': No such file or directory
```

jbtrystram:
Ok, so the fix is to always inject a reasonnable default path. I really don't understand why though.
I started multiples images and VMs to try to figure it out.
In the fedora case, `sfdisk` path cannot be resolved, although it's installed in `/usr/bin` and all symlinks point to there anyway. I suppose that's because of the [sbin and bin](https://fedoraproject.org/wiki/Changes/Unify_bin_and_sbin) merge but since this is all symlinks pointing to the same place, I don't see how the value of $PATH have an impact.

crates/lib/src/bootloader.rs
```diff
@@ -91,22 +91,71 @@ pub(crate) fn install_via_bootupd(
...
+            cmd = cmd.bind_device(&partition.node);
+        }
+
+        // The $PATH in the bwrap env is not complete enough for some images
+        // so we inject a reasonnable default.
+        // This is causing bootupctl and/or sfdisk binaries
+        // to be not found with fedora 43.
+        cmd.setenv(
+            "PATH",
+            "/bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin",
```
cgwalters, line 149: We should link to https://github.com/systemd/systemd/blame/055ceec66672289eed76a2b8d9ad4e80c053010b/src/basic/path-util.h#L12 or so as a reference.

cgwalters:
So one thing I belatedly discovered here is that bootc isn't defined to depend on bwrap and some other images (bootcrew/debian) don't have it. (also bcvk hard depends on it too)

This is...messy. I guess perhaps simplest is to just require it. Alternatively...we could detect this case and try crun/runc perhaps.

Issue #1814, opened 2025-11-27
`usroverlay` test failed to reboot on C10S


henrywang:
`usroverlay` test failed to reboot on C10S with error `err: Call to Reboot failed: Access denied`. This error only exists on Packit(Testing Farm). The same test passed in Github runner environment.

Failure log: https://artifacts.dev.testing-farm.io/f5bb7937-c31c-4f18-bffe-8c2ac6788f9e/work-plan-23-usroverlayqb1fsmzi/log.txt

PR #1817, opened 2025-12-01
test: Fix flaky reboot issue in C10S

henrywang:
This PR is to fix `tmt-reboot` failed in `useroverlay` test in C10S. I tried 3 times, all passed. `sleep` is not the best solution, but it works in this case. I checked avc log, journal log and login permission. All looks good.

I also tried 60 seconds, filed one times in 2 runs. So 120 seconds is more stable from my try runs.

### DIFF
--- a/tmt/tests/booted/bootc_testlib.nu
+++ b/tmt/tests/booted/bootc_testlib.nu
@@ -4,11 +4,8 @@
 # that seems to have appeared in C10S
 # TODO diagnose and fill in here
 export def reboot [] {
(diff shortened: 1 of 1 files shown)


henrywang:
`error: System transaction in progress` flaky issue has been fixed by https://github.com/bootc-dev/bootc/pull/1813/commits/4e1d6ad1a8fe69b7c042c23439e4afb5884f76d1 in PR https://github.com/bootc-dev/bootc/pull/1813

cgwalters:
OK right the tracker issue is https://github.com/bootc-dev/bootc/issues/1814 so let's link that here

PR #1818, opened 2025-12-02
sdboot: add support for autoenroll in `bootc install`

gerblesh:
add autoenroll to sdboot

installs `.auth` files from `/usr/lib/bootc/keys` into `ESP/loader/keys/auto`. Currently uses cfs file APIs and is a little rough. Not sure if this should be in `composefs-boot` rather than `bootc`. The directory `/usr/lib/bootc/keys` probably isn't the best choice although at least it means that the auth files are "measured" when the cfs digest is taken.

Note that this is missing some of the bits from CI for this to be properly tested. I will try to get that done when I can though

### DIFF
--- a/crates/lib/src/bootc_composefs/boot.rs
+++ b/crates/lib/src/bootc_composefs/boot.rs
@@ -66,7 +66,7 @@ use std::fs::create_dir_all;
 use std::io::Write;
 use std::path::Path;
 
-use anyhow::{anyhow, Context, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use bootc_blockdev::find_parent_devices;
 use bootc_kernel_cmdline::utf8::{Cmdline, Parameter};
 use bootc_mount::inspect_filesystem_of_dir;
@@ -136,6 +136,10 @@ const SYSTEMD_LOADER_CONF_PATH: &str = "loader/loader.conf";
 
 const INITRD: &str = "initrd";
 const VMLINUZ: &str = "vmlinuz";
+
+const BOOTC_AUTOENROLL_PATH: &str = "usr/lib/bootc/install/secureboot-keys";
+
+const AUTH_EXT: &str = "auth";
 
 /// We want to be able to control the ordering of UKIs so we put them in a directory that's not the
 /// directory specified by the BLS spec. We do this because we want systemd-boot to only look at
@@ -1146,6 +1150,52 @@ pub(crate) fn setup_composefs_uki_boot(
     Ok(())
 }
 
+pub struct SecurebootKeys {
+    pub dir: Dir,
+    pub keys: Vec<Utf8PathBuf>,
+}
+
+fn get_secureboot_keys(fs: &Dir, p: &str) -> Result<Option<SecurebootKeys>> {
+    let mut entries = vec![];
+
+    // if the dir doesn't exist, return None
+    let keys_dir = match fs.open_dir_optional(p)? {
+        Some(d) => d,
+        _ => return Ok(None),
+    };
+
+    // https://github.com/systemd/systemd/blob/26b2085d54ebbfca8637362eafcb4a8e3faf832f/man/systemd-boot.xml#L392
+
+    for entry in keys_dir.entries()? {
+        let dir_e = entry?;
+        let dirname = dir_e.file_name();
+        if !dir_e.file_type()?.is_dir() {
+            bail!("/{p}/{dirname:?} is not a directory");
+        }
+
+        let dir_path: Utf8PathBuf = dirname.try_into()?;
+        let dir = dir_e.open_dir()?;
+        for entry in dir.entries()? {
+            let e = entry?;
+            let local: Utf8PathBuf = e.file_name().try_into()?;
+            let path = dir_path.join(local);
+
+            if path.extension() != Some(AUTH_EXT) {
+                continue;
+            }
+
+            if !e.file_type()?.is_file() {
+                bail!("/{p}/{path:?} is not a file");
+            }
+            entries.push(path);
+        }
+    }
+    return Ok(Some(SecurebootKeys {
+        dir: keys_dir,
+        keys: entries,
+    }));
+}
+
 #[context("Setting up composefs boot")]
 pub(crate) async fn setup_composefs_boot(
     root_setup: &RootSetup,
@@ -1185,6 +1235,7 @@ pub(crate) async fn setup_composefs_boot(
             &root_setup.physical_root_path,
             &state.config_opts,
             None,
+            get_secureboot_keys(&mounted_fs, BOOTC_AUTOENROLL_PATH)?,
         )?;
     }
 
--- a/crates/lib/src/bootloader.rs
+++ b/crates/lib/src/bootloader.rs
@@ -1,3 +1,4 @@
+use std::fs::create_dir_all;
 use std::process::Command;
 
 use anyhow::{anyhow, bail, Context, Result};
@@ -9,7 +10,7 @@ use fn_error_context::context;
 use bootc_blockdev::{Partition, PartitionTable};
 use bootc_mount as mount;
 
-use crate::bootc_composefs::boot::mount_esp;
+use crate::bootc_composefs::boot::{mount_esp, SecurebootKeys};
 use crate::{discoverable_partition_specification, utils};
 
 /// The name of the mountpoint for efi (as a subdirectory of /boot, or at the toplevel)
@@ -18,6 +19,9 @@ pub(crate) const EFI_DIR: &str = "efi";
 /// Path to the bootupd update payload
 #[allow(dead_code)]
 const BOOTUPD_UPDATES: &str = "usr/lib/bootupd/updates";
+
+// from: https://github.com/systemd/systemd/blob/26b2085d54ebbfca8637362eafcb4a8e3faf832f/man/systemd-boot.xml#L392
+const SYSTEMD_KEY_DIR: &str = "loader/keys";
 
 #[allow(dead_code)]
 pub(crate) fn esp_in(device: &PartitionTable) -> Result<&Partition> {
@@ -74,6 +78,7 @@ pub(crate) fn install_systemd_boot(
     _rootfs: &Utf8Path,
     _configopts: &crate::install::InstallConfigOpts,
     _deployment_path: Option<&str>,
+    autoenroll: Option<SecurebootKeys>,
 ) -> Result<()> {
     let esp_part = device
         .find_partition_of_type(discoverable_partition_specification::ESP)
@@ -87,7 +92,35 @@ pub(crate) fn install_systemd_boot(
     Command::new("bootctl")
         .args(["install", "--esp-path", esp_path.as_str()])
         .log_debug()
-        .run_inherited_with_cmd_context()
+        .run_inherited_with_cmd_context()?;
+
+    if let Some(SecurebootKeys { dir, keys }) = autoenroll {
+        let path = esp_path.join(SYSTEMD_KEY_DIR);
+        create_dir_all(&path)?;
+
+        let keys_dir = esp_mount
+            .fd
+            .open_dir(SYSTEMD_KEY_DIR)
+            .with_context(|| format!("Opening {path}"))?;
+
+        for filename in keys.iter() {
+            let p = path.join(&filename);
+
+            // create directory if it doesn't already exist
+            if let Some(parent) = p.parent() {
+                create_dir_all(parent)?;
+            }
+
+            dir.copy(&filename, &keys_dir, &filename)
+                .with_context(|| format!("Copying secure boot key: {p}"))?;
+            println!("Wrote Secure Boot key: {p}");
+        }
+        if keys.is_empty() {
+            tracing::debug!("No Secure Boot keys provided for systemd-boot enrollment");
+        }
+    }
+
+    Ok(())
 }
 
 #[context("Installing bootloader using zipl")]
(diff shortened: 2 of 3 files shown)


crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -1141,6 +1143,42 @@ pub(crate) fn setup_composefs_uki_boot(
...
 }
 
+pub struct AuthFile {
+    pub filename: Utf8PathBuf,
+    pub file: RegularFile<Sha512HashValue>,
+}
+
+fn get_autoenroll_keys(root: &Directory<RegularFile<Sha512HashValue>>) -> Result<Vec<AuthFile>> {
+    let mut entries = vec![];
+    match root.get_directory("/usr/lib/bootc/keys".as_ref()) {
```
Johan-Liebert1: We mount the EROFS image before adding boot entries. Working with fs APIs might be a bit cleaner here.

https://github.com/bootc-dev/bootc/blob/2c34df68607ebcaef4768191af4fb3a852f5b695/crates/lib/src/bootc_composefs/boot.rs#L1154

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -1142,6 +1143,42 @@ pub(crate) fn setup_composefs_uki_boot(
...
+                entries.push(AuthFile {
+                    filename: path,
+                    file: file.clone(),
+                });
+            }
+        }
+        Err(other) => Err(other)?,
+    };
+    Ok(entries)
+}
```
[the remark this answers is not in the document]
Johan-Liebert1: This sounds fair

crates/lib/src/bootc_composefs/boot.rs
gerblesh: Now using the regular FS APIs, thanks!

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -1141,6 +1144,46 @@ pub(crate) fn setup_composefs_uki_boot(
...
+    pub dir: Dir,
+    pub keys: Vec<Utf8PathBuf>,
+}
+
+fn get_systemd_boot_autoenroll(fs: &Dir, p: &str) -> Result<Option<AutoEnroll>> {
+    let mut entries = vec![];
+
+    let keys_dir = match fs.open_dir(p) {
+        Ok(d) => d,
+        Err(_) => return Ok(None), // if the directory doesn't exist it just means we don't have any keys to enroll
```
Johan-Liebert1: I think we should throw an error if this error is anything other than `ErrorKind::NotFound`. We could use `fs.open_dir_optional`

Johan-Liebert1:
We also use "topic prefixes" for commits, so if you could prefix your commit messages with `sd-boot:` that would be great. Also, I think most of these commits could be squashed  under `add support for key autoenroll to sdboot`

crates/lib/src/bootc_composefs/boot.rs
gerblesh: Using `fs.open_dir_optional` here now, thanks!

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -132,6 +132,8 @@ const SYSTEMD_LOADER_CONF_PATH: &str = "loader/loader.conf";
 const INITRD: &str = "initrd";
 const VMLINUZ: &str = "vmlinuz";
 
+const BOOTC_AUTOENROLL_PATH: &str = "usr/lib/bootc/keys";
```
cgwalters: If these are only copied from the container to the bootloader state at install time, then I think this should be `/usr/lib/bootc/install/secureboot-keys` or so.

(Note `install/` subdir)

Also can you add a section to the install manpage about this?

crates/lib/src/bootloader.rs
```diff
@@ -87,7 +92,29 @@ pub(crate) fn install_systemd_boot(
...
+
+            let keys_dir = esp_mount
+                .fd
+                .open_dir(SYSTEMD_AUTOENROLL)
+                .with_context(|| format!("Opening {path}"))?;
+            for filename in keys.iter() {
+                let p = path.join(filename.clone());
+                keys_dir
+                    .atomic_write(&filename, dir.read(&filename)?)
+                    .with_context(|| format!("Writing secure boot key: {p}"))?;
```
cgwalters: I think this can just use https://docs.rs/cap-std/3.4.4/cap_std/fs/struct.Dir.html#method.copy

crates/lib/src/bootloader.rs
```diff
@@ -19,6 +21,8 @@ pub(crate) const EFI_DIR: &str = "efi";
 #[allow(dead_code)]
 const BOOTUPD_UPDATES: &str = "usr/lib/bootupd/updates";
 
+const SYSTEMD_AUTOENROLL: &str = "loader/keys/auto";
```
cgwalters: Let's link to https://github.com/systemd/systemd/blob/26b2085d54ebbfca8637362eafcb4a8e3faf832f/man/systemd-boot.xml#L392 in a comment here.

Also let's not require `auto` here, I think that should be the choice of the container image.

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -1141,6 +1144,47 @@ pub(crate) fn setup_composefs_uki_boot(
...
+        }
+
+        if !name.as_bytes().ends_with(b".auth") {
+            continue;
+        }
+
+        let path = match Utf8PathBuf::from_os_string(name.clone()) {
+            Ok(p) => p,
+            Err(_) => bail!("couldn't get pathbuf: /{p}/{name:?}"),
+        };
```
cgwalters: There's more idiomatic ways to do this. One is to use `DirUtf8` which hard enforces this, or use `let path: Utf8PathBuf = name.try_into()?;`

Or on this case we could probably just use `PathBuf` (or `OSString`).

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -1141,6 +1144,47 @@ pub(crate) fn setup_composefs_uki_boot(
...
+    let keys_dir = match fs.open_dir_optional(p)? {
+        Some(d) => d,
+        None => return Ok(None),
+    };
+
+    for entry in keys_dir.entries()? {
+        let file = entry?;
+
+        let name = file.file_name();
+        if !file.file_type()?.is_file() {
```
cgwalters: Let's move this check to *after* us checking the file name vs `.auth`

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -1141,6 +1144,47 @@ pub(crate) fn setup_composefs_uki_boot(
...
+            bail!("/{p}/{name:?} is a not a regular file");
+        }
+
+        if !name.as_bytes().ends_with(b".auth") {
```
cgwalters: I also like `const` for this type of thing (.auth). And note if we move this after the utf8 check, then we don't need `as_bytes()` and `b`

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -1141,6 +1144,47 @@ pub(crate) fn setup_composefs_uki_boot(
...
+            continue;
+        }
+
+        let path = match Utf8PathBuf::from_os_string(name.clone()) {
+            Ok(p) => p,
+            Err(_) => bail!("couldn't get pathbuf: /{p}/{name:?}"),
+        };
+        entries.push(path);
+    }
+    if entries.len() > 0 {
```
cgwalters: Converting an empty result to None isn't wrong, but it's not clear we actually gain much versus just handling the empty vector in the callers instead.

To *really* get the benefit of combining Option + Vec, we'd use https://docs.rs/vec1/latest/vec1/ instead I think (but let's not do that yet, not clear it's worth the nontrivial external dep just for this)

cgwalters (changes requested):
As far as tests; we could unit test the key parsing easily, but not a blocker.

For the actual functionality - it looks like (combined with the systemd-boot signing fixes in https://github.com/bootc-dev/bootc/pull/1809 ) we could actually have a test flow that does this in addition to/instead of setting up the libvirt firmware too, which would be cool. (Actually we do need to test both paths)

crates/lib/src/bootc_composefs/boot.rs
gerblesh: moved to the install function, not really sure what the idiomatic Rust way of doing this is but things don't break with no entries just nothing is really copied over. Added some logging for when the directory exists but no keys are to be enrolled

crates/lib/src/bootc_composefs/boot.rs
gerblesh: Used `try_into()?` here. `esp_path` is already a `Utf8Path` so I figure it's better to keep the types consistent but I dunno if there's a more preferred type here

crates/lib/src/bootloader.rs
gerblesh: fixed, thanks

crates/lib/src/bootc_composefs/boot.rs
gerblesh: sounds good, the path I chose was pretty arbitrary. Added a section in the manpage about this as well

cgwalters:
To xref I also did https://github.com/systemd/systemd/pull/39993 while investigating this

crates/lib/src/bootloader.rs
```diff
@@ -87,7 +92,39 @@ pub(crate) fn install_systemd_boot(
     Command::new("bootctl")
         .args(["install", "--esp-path", esp_path.as_str()])
         .log_debug()
-        .run_inherited_with_cmd_context()
+        .run_inherited_with_cmd_context()?;
+
+    match autoenroll {
+        None => return Ok(()),
+        Some(SecurebootKeys { dir, keys }) => {
```
cgwalters: Slightly more idiomatic as 
```
if let Some(SecurebootKeys { dir, keys}) = autoenroll {
  ...
}
```

Then in the None case we just naturally fall through.

One reason this is more idiomatic is that if someone wanted to add more code after this, that'd work just fine.

crates/lib/src/bootloader.rs
```diff
@@ -87,7 +92,39 @@ pub(crate) fn install_systemd_boot(
...
+                    create_dir_all(parent)?;
+                }
+
+                dir.copy(&filename, &keys_dir, &filename)
+                    .with_context(|| format!("Copying secure boot key: {p}"))?;
+                println!("Wrote secure boot key: {p}");
+            }
+            if keys.len() < 1 {
+                println!("No secure boot keys to copy over");
+            }
```
cgwalters: Well it's weird to print "Autoenrolling keys". let's consolidate with the above println.

crates/lib/src/bootloader.rs
```diff
@@ -87,7 +92,39 @@ pub(crate) fn install_systemd_boot(
     Command::new("bootctl")
         .args(["install", "--esp-path", esp_path.as_str()])
         .log_debug()
-        .run_inherited_with_cmd_context()
+        .run_inherited_with_cmd_context()?;
+
+    match autoenroll {
+        None => return Ok(()),
+        Some(SecurebootKeys { dir, keys }) => {
+            println!("Autoenrolling keys");
```
cgwalters: How about:
```
if keys.is_empty() {
    tracing::debug!("No Secure Boot keys provided for systemd-boot enrollment"); 
}
```

crates/lib/src/bootloader.rs
```diff
@@ -87,7 +92,39 @@ pub(crate) fn install_systemd_boot(
...
+                let p = path.join(&filename);
+
+                // create directory if it doesn't already exist
+                if let Some(parent) = p.parent() {
+                    create_dir_all(parent)?;
+                }
+
+                dir.copy(&filename, &keys_dir, &filename)
+                    .with_context(|| format!("Copying secure boot key: {p}"))?;
+                println!("Wrote secure boot key: {p}");
```
cgwalters: Let's capitalize Secure Boot as it's a formally known thing

docs/src/man/bootc-install.8.md
```diff
@@ -28,6 +28,10 @@ updates.
 An installation is not simply a copy of the container filesystem, but
 includes other setup and metadata.
 
+## Secure Boot Keys
+
+When installing with `systemd-boot`, bootc can let `systemd-boot` can handle enrollment of secureboot keys by putting signed efi signature lists in `/usr/lib/bootc/install/secureboot-keys` which will copy over into `ESP/loader/keys` after bootloader installation. The keys will be copied to `ESP/loader/keys` after installing `systemd-boot` to the system. More information on how key enrollment works with `systemd-boot` is available  [systemd-boot](https://github.com/systemd/systemd/blob/26b2085d54ebbfca8637362eafcb4a8e3faf832f/man/systemd-boot.xml#L392) man page
```
cgwalters: - capitalize EFI
- The ESP can be mounted in a few places, but none of them include "ESP". We could say `/efi/loader/keys` or so, though the [BLS](https://uapi-group.org/specifications/specs/boot_loader_specification/) uses `$BOOT`. Maybe simplest to say "The `loader/keys` subdirectory of the ESP"
- available *in the*"
- Add trailing `.`

gerblesh:
Alright should have addressed all of the nits, lmk if there's more I should fix/tweak before merge

cgwalters (approved):
Thanks so much! I'm ok getting this in without integration tests for now as the whole composefs is still experimental; I hope we'll continue momentum on fleshing out a lot of those though.

cgwalters:
@Johan-Liebert1 you need to clear your requested changes state here

cgwalters:
- Rebased
- Added a commit message

PR #1820, opened 2025-12-02
cli: add `install print-configuration --all`

mvo5:
When `install print-configuration` is run some options (notably the kargs) are currently filtered out. This makes sense because in general `bootc install to-filesystem` takes care of them.

However with the recent work in image-builder/osbuild to use bootc containers directly as inputs to build ISOs [0],[1] we would like to get access to the kernel args too because when constructing a bootable ISO we also want to add the bootc container kargs.

[0] https://github.com/orgs/osbuild/discussions/45
[1] https://github.com/osbuild/images/pull/1906

### DIFF
--- a/crates/tests-integration/src/container.rs
+++ b/crates/tests-integration/src/container.rs
@@ -1,3 +1,7 @@
+use indoc::indoc;
+use scopeguard::defer;
+use serde::Deserialize;
+use std::fs;
 use std::process::Command;
 
 use anyhow::{Context, Result};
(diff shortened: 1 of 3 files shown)


crates/lib/src/cli.rs
```diff
@@ -289,7 +289,7 @@ pub(crate) enum InstallOpts {
     ///
     /// At the current time, the only output key is `root-fs-type` which is a string-valued
     /// filesystem name suitable for passing to `mkfs.$type`.
-    PrintConfiguration,
+    PrintConfiguration(crate::install::InstallPrintConfigurationOpts),
```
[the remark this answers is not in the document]
cgwalters: Tests are good, but we can't really do this in unit tests because it operates on ambient data.

We can do it in `tests-integration/src/container.rs` though I think.

cgwalters:
cc https://github.com/bootc-dev/bootc/issues/1827 - I think image builder actually wants that because we do want to capture kargs from kargs.d, not just the install config (which is scoped to day 1 only).

crates/tests-integration/src/container.rs
```diff
@@ -36,8 +40,34 @@ pub(crate) fn test_bootc_install_config() -> Result<()> {
...
+    Ok(())
+}
+
+pub(crate) fn test_bootc_install_config_all() -> Result<()> {
+    #[derive(Deserialize)]
+    struct TestInstallConfig {
+        kargs: Vec<String>,
+    }
+
+    let config_d = std::path::Path::new("/run/bootc/install/");
```
cgwalters, line 54: This is fine, though I guess probably what we really need here is a workflow that combines container build + test.

mvo5: Agreed, it would be nice to have a combined build+test here. I can remove this part of the code again, its a bit unnecessary because the default container already has some kargs I can test for. I wanted to make this is a test-as-example thing so wanted to include the code that writes the toml but maybe I was overthinking it :) Either way is fine with me!

PR #1828, opened 2025-12-04
Composefs soft reboot

Johan-Liebert1:
- [x] Requires: https://github.com/containers/composefs-rs/pull/198 and a bump to `composefs-rs`

Add an internal command for soft rebooting the system. Similar to how
it's done for ostree, we only allow soft reboot if the other deployment
has the same kernel state, i.e. the SHASum of kernel + initrd is the
same as that of the current deployment.

Similar to soft reboots for Type1 entries, we compute the SHA256Sum of
.linux + .initrd sections in the UKI, and compare them to check for
kernel skew

Next, compare the .cmdline section skipping the `composefs=` parameter
as that will always be different

After a soft reboot the kernel cmdline doesn't change so we can't rely
on the `composefs=` parameter in the cmdline. Instead, we check the
source of the root mount point

### DIFF
--- a/crates/lib/src/bootc_composefs/state.rs
+++ b/crates/lib/src/bootc_composefs/state.rs
@@ -162,7 +162,10 @@ pub(crate) fn update_target_imgref_in_origin(
         "origin",
         &[(
             ORIGIN_CONTAINER,
-            &format!("ostree-unverified-image:{imgref}"),
+            &format!(
+                "ostree-unverified-image:{}",
+                get_imgref(&imgref.transport, &imgref.image)
+            ),
         )],
     )
 }


crates/lib/src/bootc_composefs/status.rs
```diff
@@ -76,7 +77,29 @@ pub(crate) fn composefs_booted() -> Result<Option<&'static ComposefsCmdline>> {
...
+        .source
+        .strip_prefix("composefs:")
+        .ok_or_else(|| anyhow::anyhow!("Root not mounted using composefs"))?;
+
+    let r = if *verity_from_mount_src != *v.digest {
+        // soft rebooted into another deployment
+        CACHED_DIGEST_VALUE.get_or_init(|| Some(ComposefsCmdline::new(verity_from_mount_src)))
+    } else {
+        CACHED_DIGEST_VALUE.get_or_init(|| Some(v))
+    };
```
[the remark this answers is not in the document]
Johan-Liebert1: It's always going to be either/or. The kernel cmdline won't change unless we reboot. On soft reboot we restart userspace anyway. Am I missing something here?

crates/lib/src/bootc_composefs/soft_reboot.rs
```diff
@@ -0,0 +1,79 @@
...
+        target: Some(NEXTROOT.into()),
+    };
+
+    setup_root(args)?;
+
+    if reboot {
+        // Replacing the current process should be fine as we restart userspace anyway
+        let res = Command::new("systemctl").arg("soft-reboot").exec();
+        println!("Soft reboot error: {res:#?}");
+        return Ok(());
```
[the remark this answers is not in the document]
Johan-Liebert1: This is definitely better. Updated

Johan-Liebert1:
`plan-20-image-pushpull-upgrade` seems to have failed with

```
content: Failed to kill unit test-cat-progress.service: Unit test-cat-progress.service not loaded.
content: Running as unit: test-cat-progress.service; invocation ID: ad2c2ea1c85a4253a2fc81619b9ade72
content: layers already present: 65; layers needed: 11 (267.8 MB)
content: Deploying...done (5 seconds)
content: Bound images stored: 3
content: Queued for next boot: ostree-unverified-image:containers-storage:localhost/bootc-derived
content:   Version: 10
content:   Digest: sha256:38e79b454a170138c58e2add8b6106ff374d63c03c33f8ef693590f0e28adf17
content: -- No entries --
content: Adding quoted karg via rpm-ostree to test ostree issue #3544
content: error: System transaction in progress
content: Connection to localhost closed.
```

Is this related to `rpm-ostree`? https://github.com/coreos/rpm-ostree/issues/210 seems related, but not sure

crates/lib/src/bootc_composefs/status.rs
```diff
@@ -76,7 +77,29 @@ pub(crate) fn composefs_booted() -> Result<Option<&'static ComposefsCmdline>> {
...
-    let r = CACHED_DIGEST_VALUE.get_or_init(|| Some(v));
+
+    // Find the source of / mountpoint as the cmdline doesn't change on soft-reboot
+    let findmnt = run_findmnt(&[], None, None)?;
+
+    let root_mnt = findmnt
+        .filesystems
+        .iter()
+        .find(|fs| fs.target == "/")
+        .ok_or_else(|| anyhow::anyhow!("Failed to get root mountpoint"))?;
```
cgwalters: `inspect_filesystem("/")` is a lot more efficient than querying everything and filtering in memory

crates/lib/src/bootc_composefs/status.rs
```diff
@@ -76,7 +77,29 @@ pub(crate) fn composefs_booted() -> Result<Option<&'static ComposefsCmdline>> {
...
+
+    // This is of the format composefs:<composefs_hash>
+    let verity_from_mount_src = root_mnt
+        .source
+        .strip_prefix("composefs:")
+        .ok_or_else(|| anyhow::anyhow!("Root not mounted using composefs"))?;
```
cgwalters, line 102: Hum...I'm OK with this but ultimately I think it'd be even nicer to have truly structured metadata associated with the mount point somehow.

crates/lib/src/bootc_composefs/status.rs
```diff
@@ -76,7 +77,29 @@ pub(crate) fn composefs_booted() -> Result<Option<&'static ComposefsCmdline>> {
...
+    let root_mnt = findmnt
+        .filesystems
+        .iter()
+        .find(|fs| fs.target == "/")
+        .ok_or_else(|| anyhow::anyhow!("Failed to get root mountpoint"))?;
+
+    // This is of the format composefs:<composefs_hash>
+    let verity_from_mount_src = root_mnt
+        .source
+        .strip_prefix("composefs:")
```
cgwalters, line 101: Can we strongly associate this string with the thing that creates it via a `const`?

crates/lib/src/bootc_composefs/status.rs
```diff
@@ -413,6 +442,29 @@ pub(crate) async fn composefs_deployment_status_from(
         host.spec.boot_order = BootOrder::Rollback
     };
 
+    // Can only soft reboot non UKI boot entries
```
cgwalters: Why?

cgwalters:
> content: error: System transaction in progress

It's a flake (well, a bug somewhere), I'll ensure we have an issue for it

crates/lib/src/bootc_composefs/status.rs
Johan-Liebert1: It is a little bit more complex, as we'd have to extract `.linux` and `.initrd` sections to compute the hash. Definitely possible though. That being said, in the confidential clusters use case, would soft rebooting be a thing anyway?

crates/lib/src/bootc_composefs/status.rs
Johan-Liebert1: I wanted to, but it's done in `composefs-rs`. Maybe we can export it from there

crates/lib/src/bootc_composefs/status.rs
Johan-Liebert1: Wouldn't the `fsconfig` string be that metadata?

crates/lib/src/bootc_composefs/status.rs
Johan-Liebert1: Updated

crates/lib/src/bootc_composefs/status.rs
cgwalters: We want to support sealed systems outside of just confidential workloads. 

> It is a little bit more complex, as we'd have to extract .linux and .initrd sections to compute the hash.

There are other UKI components too - such as the critical `.cmdline`. I think what we really need to do here is a complete diff the two, just dropping out `composefs=` from the `.cmdline` or so.

cgwalters:
Hmm...right I think we need to use pull_request_target for the rebase helper

Johan-Liebert1:
I have rebased it locally. Working on soft-reboot for UKIs

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -338,6 +338,30 @@ fn compute_boot_digest(
...
+/// * repo - The composefs repository
+#[context("Computing boot digest")]
+pub(crate) fn compute_boot_digest_uki(uki: &[u8]) -> Result<String> {
+    let vmlinuz = composefs_boot::uki::get_section(uki, ".linux")
+        .ok_or_else(|| anyhow::anyhow!(".linux not present"))??;
+
+    let initramfs = composefs_boot::uki::get_section(uki, ".initrd")
+        .ok_or_else(|| anyhow::anyhow!(".initrd not present"))??;
+
+    let mut hasher = openssl::hash::Hasher::new(openssl::hash::MessageDigest::sha256())
```
cgwalters, line 354: We should perhaps use sha512 just on general principle for stuff like this given the post-quantum crypto stuff. It's not like this logic is seriously performance sensitive anyways.

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -338,6 +338,30 @@ fn compute_boot_digest(
...
+///
+/// # Arguments
+/// * entry - BootEntry containing VMlinuz and Initrd
+/// * repo - The composefs repository
+#[context("Computing boot digest")]
+pub(crate) fn compute_boot_digest_uki(uki: &[u8]) -> Result<String> {
+    let vmlinuz = composefs_boot::uki::get_section(uki, ".linux")
+        .ok_or_else(|| anyhow::anyhow!(".linux not present"))??;
+
+    let initramfs = composefs_boot::uki::get_section(uki, ".initrd")
```
cgwalters, line 351: Not a blocker but I wonder if we should hard require an initrd? In theory some use cases could probably go without.

OK for *us* we obviously hard require it to mount composefs, but since this is a generic interface; per the spec https://uapi-group.org/specifications/specs/unified_kernel_image/#uki-file-format

> Optionally, the initrd that the kernel shall unpack and invoke, in the .initrd section.

crates/lib/src/bootc_composefs/state.rs
```diff
@@ -151,6 +150,36 @@ pub(crate) fn update_target_imgref_in_origin(
...
+        booted_cfs.cmdline.digest.as_ref(),
+        "origin",
+        &[(
+            ORIGIN_CONTAINER,
+            &format!("ostree-unverified-image:{imgref}"),
+        )],
+    )
+}
+
+pub(crate) fn update_boot_digest_in_origin(
```
cgwalters, line 170: Do we really need to cache the digest in the origin? It doesn't seem like it should be that expensive to compute dynamically from the original source of truth.

BTW also related to this, if we are worried about speed we can just check the *sizes* of the kernel/initrd sections first, and if those differ then we know the kernel state differs...

crates/lib/src/bootc_composefs/utils.rs
```diff
@@ -0,0 +1,58 @@
...
+
+fn get_uki(storage: &Storage, deployment_verity: &str) -> Result<Vec<u8>> {
+    let uki_dir = storage
+        .esp
+        .as_ref()
+        .ok_or_else(|| anyhow::anyhow!("ESP not mounted"))?
+        .fd
+        .open_dir(SYSTEMD_UKI_DIR)?;
+
+    let req_fname = format!("{deployment_verity}.efi");
```
cgwalters, line 20: Hmm, should that be a required format?

crates/lib/src/bootc_composefs/utils.rs
```diff
@@ -0,0 +1,58 @@
...
+};
+use anyhow::Result;
+use bootc_kernel_cmdline::utf8::Cmdline;
+use fn_error_context::context;
+
+fn get_uki(storage: &Storage, deployment_verity: &str) -> Result<Vec<u8>> {
+    let uki_dir = storage
+        .esp
+        .as_ref()
+        .ok_or_else(|| anyhow::anyhow!("ESP not mounted"))?
```
cgwalters, line 16: How about a `.require_esp()?` method on storage?

cgwalters (approved):
Overall looks reasonable but biggest question: can we now add `test-soft-reboot` underneath `just test-composefs`?

crates/lib/src/bootc_composefs/boot.rs
Johan-Liebert1: True, outside of confidential use case, initrd would definitely be optional

crates/lib/src/bootc_composefs/state.rs
Johan-Liebert1: > Do we really need to cache the digest in the origin

Not really, and I haven't really tested the speed without the cache, but it would be fairly quick, I'd assume. I had done this way earlier for the Type1 (kernel + initrd), and I just ended up following that for the UKI approach as well.

crates/lib/src/bootc_composefs/utils.rs
Johan-Liebert1: We control the names of the UKI, and while writing to the ESP, we do use this format. Maybe it makes sense to put this in a helper function

Johan-Liebert1:
Going ahead and merging this. I'll create a follow up PR addressing all the comments

Issue #1320, opened 2025-05-15
Use finalization locking


jlebon:
This was mentioned in https://github.com/bootc-dev/bootc/issues/165 as a desirable feature, but AFAICT there isn't a card tracking it specifically.

Obviously, this isn't hard to do I think with the current OSTree backend. Though the expectation is that we'll want to retain this semantic when we eventually move to a pure composefs-based backend.

cgwalters:
This issue also relates to https://github.com/bootc-dev/bootc/issues/1375

urld:
@jlebon is there any chance to also allow this flag for `bootc-switch`?
Working on a project where we are currently using `bootc-switch` rather than `bootc-upgrade` as we need to be able switch to exact  versions, rather than relying on the latest available.
At the moment we download the images outside of bootc and then use  `bootc switch --transport oci`. Having the --download-only flag would make this more streamlined.

jlebon:
@urld I didn't implement the logic for this but agree this should exist. (I have also the same use case, i.e. exact digests, driven by a higher-level agent.)

Filed your request as https://github.com/bootc-dev/bootc/issues/2137

PR #1831, opened 2025-12-05
lib: Add --download-only flag for upgrade (ostree backend)

shi2wei3:
Resolve https://github.com/bootc-dev/bootc/issues/1320 for the ostree backend

Add support for downloading and staging updates without automatic application on reboot. This allows users to prepare updates and apply them at a controlled time.

User-facing changes:
- Add --download-only flag to bootc upgrade command
- bootc upgrade --download-only: locks staged deployment
- bootc upgrade (no flags): unlocks locked staged deployment
- bootc upgrade --apply: unlocks and immediately reboots
- bootc upgrade --check: read-only, doesn't change lock state
- bootc status --verbose: shows "Locked: yes/no" for staged deployments

Implementation details:
- Internally uses OSTree finalization locking APIs
- Sets opts.locked in SysrootDeployTreeOpts when staging deployments
- Added change_finalization() method to SysrootLock wrapper
- Handles lock state transitions for existing staged deployments
- Field name in BootEntry is finalization_locked for clarity
- Display name is shortened to "Locked" for user convenience
- Uses deployment.is_finalization_locked() API (OSTree v2023.8+)

Testing and documentation:
- Added TMT integration test (test-upgrade-download-only.nu)
- Test verifies 3-boot workflow: lock, reboot (stays old), unlock, reboot (applies)
- Updated docs/src/upgrades.md with comprehensive workflow examples
- Generated man pages and JSON schemas updated

The download-only flag is only available for upgrade, not switch. The implementation is designed to support future composefs backend.

### DIFF
--- a/crates/lib/src/cli.rs
+++ b/crates/lib/src/cli.rs
@@ -100,6 +100,14 @@ pub(crate) struct UpgradeOpts {
     #[clap(long = "soft-reboot", conflicts_with = "check")]
     pub(crate) soft_reboot: Option<SoftRebootMode>,
 
+    /// Download and stage the update without applying it.
+    ///
+    /// Download the update and ensure it's retained on disk for the lifetime of this system boot,
+    /// but it will not be applied on reboot. If the system is rebooted without applying the update,
+    /// the image will be eligible for garbage collection again.
+    #[clap(long, conflicts_with_all = ["check", "apply"])]
+    pub(crate) download_only: bool,
+
     #[clap(flatten)]
     pub(crate) progress: ProgressOptions,
 }
@@ -956,7 +964,35 @@ async fn upgrade(
             .map(|img| &img.manifest_digest == fetched_digest)
             .unwrap_or_default();
         if staged_unchanged {
-            println!("Staged update present, not changed.");
+            let staged_deployment = storage.get_ostree()?.staged_deployment();
+            let mut download_only_changed = false;
+
+            if let Some(staged) = staged_deployment {
+                // Handle download-only mode based on flags
+                if opts.download_only {
+                    // --download-only: set download-only mode
+                    if !staged.is_finalization_locked() {
+                        storage.get_ostree()?.change_finalization(&staged)?;
+                        println!("Image downloaded, but will not be applied on reboot");
+                        download_only_changed = true;
+                    }
+                } else if !opts.check {
+                    // --apply or no flags: clear download-only mode
+                    // (skip if --check, which is read-only)
+                    if staged.is_finalization_locked() {
+                        storage.get_ostree()?.change_finalization(&staged)?;
+                        println!("Staged deployment will now be applied on reboot");
+                        download_only_changed = true;
+                    }
+                }
+            } else if opts.download_only || opts.apply {
+                anyhow::bail!("No staged deployment found");
+            }
+
+            if !download_only_changed {
+                println!("Staged update present, not changed");
+            }
+
             handle_staged_soft_reboot(booted_ostree, opts.soft_reboot, &host)?;
             if opts.apply {
                 crate::reboot::reboot()?;
@@ -966,7 +1002,15 @@ async fn upgrade(
         } else {
             let stateroot = booted_ostree.stateroot();
             let from = MergeState::from_stateroot(storage, &stateroot)?;
-            crate::deploy::stage(storage, from, &fetched, &spec, prog.clone()).await?;
+            crate::deploy::stage(
+                storage,
+                from,
+                &fetched,
+                &spec,
+                prog.clone(),
+                opts.download_only,
+            )
+            .await?;
             changed = true;
             if let Some(prev) = booted_image.as_ref() {
                 if let Some(fetched_manifest) = fetched.get_manifest(repo)? {
@@ -1071,7 +1115,7 @@ async fn switch_ostree(
 
     let stateroot = booted_ostree.stateroot();
     let from = MergeState::from_stateroot(storage, &stateroot)?;
-    crate::deploy::stage(storage, from, &fetched, &new_spec, prog.clone()).await?;
+    crate::deploy::stage(storage, from, &fetched, &new_spec, prog.clone(), false).await?;
 
     storage.update_mtime()?;
 
@@ -1200,7 +1244,7 @@ async fn edit_ostree(
 
     let stateroot = booted_ostree.stateroot();
     let from = MergeState::from_stateroot(storage, &stateroot)?;
-    crate::deploy::stage(storage, from, &fetched, &new_spec, prog.clone()).await?;
+    crate::deploy::stage(storage, from, &fetched, &new_spec, prog.clone(), false).await?;
 
     storage.update_mtime()?;
 
--- a/crates/lib/src/spec.rs
+++ b/crates/lib/src/spec.rs
@@ -227,6 +227,10 @@ pub struct BootEntry {
     /// This is true if (relative to the booted system) this is a possible target for a soft reboot
     #[serde(default)]
     pub soft_reboot_capable: bool,
+    /// Whether this deployment is in download-only mode (prevented from automatic finalization on shutdown).
+    /// This is set via --download-only on the CLI.
+    #[serde(default)]
+    pub download_only: bool,
     /// The container storage backend
     #[serde(default)]
     pub store: Option<Store>,
@@ -628,6 +632,7 @@ mod tests {
                 incompatible: false,
                 soft_reboot_capable: false,
                 pinned: false,
+                download_only: false,
                 store: None,
                 ostree: None,
                 composefs: None,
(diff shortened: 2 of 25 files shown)


crates/lib/src/cli.rs
```diff
@@ -957,6 +964,26 @@ async fn upgrade(
             .unwrap_or_default();
         if staged_unchanged {
             println!("Staged update present, not changed.");
```
cgwalters: But it now could change if we change download-only state

crates/lib/src/spec.rs
```diff
@@ -227,6 +227,10 @@ pub struct BootEntry {
     /// This is true if (relative to the booted system) this is a possible target for a soft reboot
     #[serde(default)]
     pub soft_reboot_capable: bool,
+    /// Whether this deployment is locked from automatic finalization on shutdown
```
cgwalters: Let's note `. This is set via --download-only on the CLI for example.`

crates/lib/src/spec.rs
```diff
@@ -227,6 +227,10 @@ pub struct BootEntry {
...
+    #[serde(default)]
+    #[serde(skip_serializing_if = "is_false")]
+    pub finalization_locked: bool,
```
cgwalters: Well, we could also rename this for consistency. Maybe `downloaded_only`?

crates/lib/src/spec.rs
```diff
@@ -227,6 +227,10 @@ pub struct BootEntry {
     /// This is true if (relative to the booted system) this is a possible target for a soft reboot
     #[serde(default)]
     pub soft_reboot_capable: bool,
+    /// Whether this deployment is locked from automatic finalization on shutdown
+    #[serde(default)]
+    #[serde(skip_serializing_if = "is_false")]
```
cgwalters: I would actually lean towards emitting this always out of consistency, even though yes...it would require changing all the test fixtures (again).

crates/lib/src/status.rs
```diff
@@ -563,6 +565,33 @@ fn write_soft_reboot(
...
+            out,
+            "{}",
+            if entry.finalization_locked {
+                "yes"
+            } else {
+                "no"
+            }
+        )?;
+    } else if entry.finalization_locked {
+        // For non-staged deployments, only show if locked
```
cgwalters: I don't believe this case can ever be true

crates/lib/src/status.rs
```diff
@@ -563,6 +565,33 @@ fn write_soft_reboot(
...
+/// Helper function to render finalization lock status
+fn write_finalization_locked(
+    mut out: impl Write,
+    slot: Option<Slot>,
+    entry: &crate::spec::BootEntry,
+    prefix_len: usize,
+) -> Result<()> {
+    // Always show lock status for staged deployments
+    if matches!(slot, Some(Slot::Staged)) {
+        write_row_name(&mut out, "Locked", prefix_len)?;
```
cgwalters: "DownloadOnly"?

crates/lib/src/cli.rs
```diff
@@ -100,6 +100,13 @@ pub(crate) struct UpgradeOpts {
     #[clap(long = "soft-reboot", conflicts_with = "check")]
     pub(crate) soft_reboot: Option<SoftRebootMode>,
 
+    /// Download and stage the update without applying it.
+    ///
+    /// The update will be downloaded and staged, but will not be applied automatically on reboot.
+    /// Use `bootc upgrade --apply` to apply the staged update.
```
cgwalters: How about

> Download the update and ensure it's retained on disk for the lifetime of this system boot, but it will not be applied on reboot. If the system is rebooted without applying the update, the image will be eligible for garbage collection again.

docs/src/upgrades.md
```diff
@@ -14,6 +14,88 @@ changed by default.
...
+
+The `--download-only` flag allows you to prepare updates without automatically applying
+them on the next reboot:
+
+```shell
+bootc upgrade --download-only
+```
+
+This will pull the new container image from the registry and create a staged deployment
+that is **locked** from automatic application. The deployment will not be applied on
```
cgwalters: This relates to conversation above; I think we should try to avoid the term "locked".

docs/src/upgrades.md
```diff
@@ -14,6 +14,88 @@ changed by default.
...
+
+# 4. Apply the update (choose one):
+# Option A: Unlock and let it apply on next shutdown
+bootc upgrade
+
+# Option B: Apply immediately with reboot
+bootc upgrade --apply
+```
+
+**Important**: If you reboot before applying a locked update, the system will boot into the
```
cgwalters: Also note here that if one switches to a different image, then that will replace the previous cached download-only image.

tmt/tests/booted/test-upgrade-download-only.nu
```diff
@@ -0,0 +1,117 @@
...
+bootc status
+journalctl --list-boots
+
+let st = bootc status --json | from json
+let booted = $st.status.booted.image
+
+# Parse the kernel commandline into a list.
+# This is not a proper parser, but good enough
+# for what we need here.
+def parse_cmdline []  {
```
cgwalters: Followup: Move this into bootc_testlib.nu

tmt/tests/booted/test-upgrade-download-only.nu
```diff
@@ -0,0 +1,117 @@
...
+# Run on the first boot
+def initial_build [] {
+    tap begin "upgrade --download-only test"
+
+    let imgsrc = imgsrc
+    # For the packit case, we build locally right now
+    if ($imgsrc | str ends-with "-local") {
+        bootc image copy-to-storage
+
+        # A simple derived container that adds a file
```
cgwalters: I'm not sure we need another copy of this. How about integrating `--download-only` alongside say `test-image-upgrade-reboot.nu`? We could do a download-only first, then reboot without applying, verify it isn't applied etc

shi2wei3: Right, reuse `test-image-upgrade-reboot.nu` could be a good idea, but after check the test logic of `test-image-upgrade-reboot.nu` I found it could use both pre-built upgrade image from Containerfile.upgrade and local build upgrade image, I guess this is for the composefs sdboot scenarios, but this make it hard to consolidate with download-only test which need to switch and then upgrade, so it need to build two derived images at least.
Let me try to keep this in a separate test case to prove the test case coverage is okay, and then think about how to merge them into a single test.

cgwalters: Yep a separate test case is cool by me for now, but if we grow 3-4 more like these we probably would need to try to consolidate.

crates/lib/src/cli.rs
```diff
@@ -956,7 +964,35 @@ async fn upgrade(
...
+            let staged_deployment = storage.get_ostree()?.staged_deployment();
+            let mut download_only_changed = false;
+
+            if let Some(staged) = staged_deployment {
+                // Handle download-only mode based on flags
+                if opts.download_only {
+                    // --download-only: set download-only mode
+                    if !staged.is_finalization_locked() {
+                        storage.get_ostree()?.change_finalization(&staged)?;
+                        println!("Staged deployment in download-only mode.");
```
cgwalters: How about "Image downloaded, but will not be applied on reboot"

Also stylistically we tend not to use `.` on printed messages like this

crates/lib/src/cli.rs
```diff
@@ -956,7 +964,35 @@ async fn upgrade(
...
+                        download_only_changed = true;
+                    }
+                } else if !opts.check {
+                    // --apply or no flags: clear download-only mode
+                    // (skip if --check, which is read-only)
+                    if staged.is_finalization_locked() {
+                        storage.get_ostree()?.change_finalization(&staged)?;
+                        println!("Cleared download-only mode.");
```
cgwalters: How about "Staged deployment will now be applied on reboot" ?

shi2wei3:
@cgwalters I still cannot find a good solution to integrate this upgrade-download-only test into the existing image-upgrade-reboot test, that's because image-upgrade-reboot design to work both ways (the upgrade image is pre-built and passthough via `bcvk --bind-storage-ro`, or the upgrade image is built locally from copy-to-storage), the existing image-upgrade-reboot test use `bootc switch` to simulate the `bootc upgrade`  (I didn't find a way to switch the imgref w/o pull and stage, --mutate-in-place won't work for this case)  to avoid do another newer pre-built upgrade image to facilitate the real `bootc upgrade` operation (I cannot find a way to build this newer upgrade image from the host after the testcase run bootc switch)
So I tend to treat the existing image-upgrade-reboot test as what it is, mainly for this pre-built mode, and the upgrade-download-only test just for local upgrade image build mode.

Let me know if anything else need to be done to get this PR merged.

Issue #1827, opened 2025-12-03
Add ability to render our own kargs in container


jeckersb:
Right now in the composefs build test/example flow, we are [hardcoding](https://github.com/bootc-dev/bootc/blob/a4bda72d48b98b484cb67bb48da8e5532dba42fa/Dockerfile.cfsuki#L31) the kargs that are injected into the UKI.  However we need to properly render the kargs given the context of the contents of the container image, and also let users supply additional kargs and merge them the same way one would do with the `bootc install` flow.  This is a prerequisite to generating the full UKI since we need to know this ahead of time, the same that we need to know the composefs fsverity hash.

This should be something like (naming bikesheddable):

```
bootc container render-kargs --additional-kargs "foo bar=baz"
```

This will print the kargs to stdout the same way it would be added to the cmdline in the normal `bootc install`.  So this would also take into account things like installconfig, kargs.d, etc.

One outstanding question is what to do about the "root filesystem" kargs from the install flow?  I guess we just ignore that in this case, since we're running in an arbitrary container and don't have the final root filesystem available for introspection.  Will that cause any problems down the line?

cgwalters:
What would the output format be? Plain text whitespace separated? Should probably have `--json` as a general rule.

This also relates a bit to https://github.com/bootc-dev/bootc/pull/1820 (that use case actually wants this instead really).

Maybe we add `bootc container inspect` which is generalized to outputting more information, like say the detected kernel version and whether it's UKI or not? I've wanted that elsewhere.

>  --additional-kargs "foo bar=baz"

What would be the use case for that versus writing them as kargs.d files?

jeckersb:
> What would the output format be? Plain text whitespace separated? Should probably have `--json` as a general rule.

Both, probably plain text by default (basically `Cmdline::to_string()` on the final computed value).  Primarily we want to just feed this into `ukify`.

> This also relates a bit to [#1820](https://github.com/bootc-dev/bootc/pull/1820) (that use case actually wants this instead really).
> 
> Maybe we add `bootc container inspect` which is generalized to outputting more information, like say the detected kernel version and whether it's UKI or not? I've wanted that elsewhere.
>

[...]

PR #1832, opened 2025-12-05
cli: Add `container inspect`

jeckersb:
Prints JSON of container metadata/attributes of interest.

For now this just renders out the kargs embedded in the container
under the kargs.d drop-in.  Future ideas for enhancements would be to
include kernel version and whether or not the image uses a UKI.

Closes: #1827

### DIFF
--- a/docs/src/man/bootc-container-inspect.8.md
+++ b/docs/src/man/bootc-container-inspect.8.md
@@ -23,7 +23,9 @@ Output JSON to stdout containing the container image metadata
 
 # EXAMPLES
 
-TODO: Add practical examples showing how to use this command.
+Inspect container image metadata:
+
+    bootc container inspect
 
 # SEE ALSO
 


jeckersb:
One outstanding question is that as-is this only looks at kargs.d, but for the container inspection case do we also want to consider install-config embedded in the container as well?  I lean slightly towards no (which is why I omitted it initially) but could pretty easily be persuaded otherwise.

tmt/tests/booted/readonly/003-test-container-inspect.nu
```diff
@@ -0,0 +1,15 @@
...
+tap begin "verify bootc container inspect"
+
+# Run container inspect and parse the JSON output
+let inspect = bootc container inspect | from json
+
+# Verify we get the expected kargs from the test image
+# These are defined in hack/test-kargs/*.toml
+assert ($inspect.kargs | any { |k| $k == "kargsd-test=1" }) "expected kargsd-test=1"
+assert ($inspect.kargs | any { |k| $k == "kargsd-othertest=2" }) "expected kargsd-othertest=2"
+assert ($inspect.kargs | any { |k| $k == "testing-kargsd=3" }) "expected testing-kargsd=3"
```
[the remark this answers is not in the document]
jeckersb: It very specifically doesn't check it that way because there's other kargs that get pulled in as part of other pieces of the test suite.  Under `hack/test-kargs` we have:

```
jeckersb@toolbx bootc> ls hack/test-kargs/                                                                                                                                                                                                                  12/05/2025 04:37:51 PM
╭───┬───────────────────────────────┬──────┬──────┬────────────╮
│ # │             name              │ type │ size │  modified  │
├───┼───────────────────────────────┼──────┼──────┼────────────┤
│ 0 │ hack/test-kargs/10-test.toml  │ file │ 48 B │ a year ago │
│ 1 │ hack/test-kargs/20-test2.toml │ file │ 29 B │ a year ago │
╰───┴───────────────────────────────┴──────┴──────┴────────────╯
jeckersb@toolbx bootc> cat hack/test-kargs/*                                                                                                                                                                                                                12/05/2025 04:38:00 PM
kargs = ["kargsd-test=1", "kargsd-othertest=2"]
kargs = ["testing-kargsd=3"]
```

But in the final integration image we get other things:

```
jeckersb@toolbx bootc> podman run --rm -it localhost/bootc-integration bootc container inspect                                                                                                                                                              12/05/2025 04:39:02 PM
{
  "kargs": [
    "kargsd-test=1",
    "kargsd-othertest=2",
    "console=ttyS0,115200n8",
    "testing-kargsd=3",
    "console=hvc0"
  ]
}
```

crates/lib/src/cli.rs
```diff
@@ -295,6 +303,12 @@ pub(crate) enum InstallOpts {
 /// Subcommands which can be executed as part of a container build.
 #[derive(Debug, clap::Subcommand, PartialEq, Eq)]
 pub(crate) enum ContainerOpts {
+    /// Output JSON to stdout containing the container image metadata.
```
cgwalters, line 298: Since this is a proposed stable API can you generate the initial manpages with `cargo xtask update-generated` and then populate it

cgwalters:
> do we also want to consider install-config embedded in the container as well? I lean slightly towards no (which is why I omitted it initially) but could pretty easily be persuaded otherwise.

Yes it's a complicated question. I think there's very few use cases for the install config kargs nowadays and we should generally treat them as legacy. So we can omit them here.

crates/lib/src/cli.rs
```diff
@@ -52,6 +52,14 @@ use crate::store::{BootedOstree, ComposefsRepository, Storage};
 use crate::store::{BootedStorage, BootedStorageKind};
 use crate::utils::sigpolicy_from_opt;
 
+/// The result of a `bootc container inspect` command.
```
cgwalters: I think let's put this in say spec.rs to make clearer it's public API, as distinct from the rust structs for CLI parsing?

tmt/tests/booted/readonly/003-test-container-inspect.nu
```diff
@@ -0,0 +1,15 @@
+use std assert
+use tap.nu
+
+tap begin "verify bootc container inspect"
```
cgwalters: This is totally OK, however I think what we really want to validate is that this command works in a container. That's a distinct test suite, see `just test-container` and https://github.com/bootc-dev/bootc/blob/main/crates/tests-integration/src/container.rs

cgwalters (changes requested):

tmt/tests/booted/readonly/003-test-container-inspect.nu
jeckersb: Yeah that makes a lot more sense, I'll move it there.

crates/lib/src/cli.rs
jeckersb: Done

crates/lib/src/cli.rs
jeckersb: Moved!

crates/lib/src/cli.rs
```diff
@@ -295,6 +295,12 @@ pub(crate) enum InstallOpts {
 /// Subcommands which can be executed as part of a container build.
 #[derive(Debug, clap::Subcommand, PartialEq, Eq)]
 pub(crate) enum ContainerOpts {
+    /// Output JSON to stdout containing the container image metadata.
```
jeckersb, line 298: Oops I left the comment in :(

Edit: oh wait nevermind this isn't where I moved the struct 🤦‍♂️ 

Ignore me, too late on Friday!

jeckersb:
Filed https://github.com/bootc-dev/bootc/issues/1839 to track CI failure

cgwalters:
I was working on https://github.com/bootc-dev/bootc/pull/1873 and then realized it'd be useful to have a human-readable format too, and that wants to share code with `bootc status` and that made me realize...why not just have instead:

`bootc status --container`
?

Or maybe it's fine as is...and we should just share code.

cgwalters:
Updated in https://github.com/bootc-dev/bootc/pull/1873

Issue #1833, opened 2025-12-07
bootc usr-overlay mounts with improper permissions on composefs backend


hecknt:
I'm not sure if this applies to the ostree backend, but on the composefs backend, when running `bootc usr-overlay`, /usr is mounted in a way that only the root user is allowed to read the files.

<img width="468" height="600" alt="Image" src="https://github.com/user-attachments/assets/98585491-1333-468c-a5cd-1d3b5c97e3ce" />

I am running bootc 1.11.0

Johan-Liebert1:
Thanks for reporting. The linked PR should fix this

PR #1836, opened 2025-12-08
composefs/usr: Fix /usr permissions on overlay mount

Johan-Liebert1:
The upper,work directories being created for `/usr` transient mount always had the mode `0o700` hence only being accessible to root

Update `bootc_initramfs_setup::ensure_dir` to accept an optional `mode` argument

Fixes: https://github.com/bootc-dev/bootc/issues/1833

### DIFF
--- a/crates/initramfs/src/lib.rs
+++ b/crates/initramfs/src/lib.rs
@@ -169,8 +169,8 @@ pub fn open_dir(dirfd: impl AsFd, name: impl AsRef<Path> + Debug) -> Result<Owne
 }
 
 #[context("Ensure dir")]
-fn ensure_dir(dirfd: impl AsFd, name: &str) -> Result<OwnedFd> {
-    match mkdirat(dirfd.as_fd(), name, 0o700.into()) {
+fn ensure_dir(dirfd: impl AsFd, name: &str, mode: Option<rustix::fs::Mode>) -> Result<OwnedFd> {
+    match mkdirat(dirfd.as_fd(), name, mode.unwrap_or(0o700.into())) {
         Ok(()) | Err(Errno::EXIST) => {}
         Err(err) => Err(err).with_context(|| format!("Creating dir {name}"))?,
     }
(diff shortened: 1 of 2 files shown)


crates/lib/src/bootc_composefs/state.rs
```diff
@@ -246,7 +246,7 @@ pub(crate) fn composefs_usr_overlay() -> Result<()> {
         return Ok(());
     }
 
-    overlay_transient(usr)?;
+    overlay_transient(usr, Some(0o755))?;
```
cgwalters: Let's not hardcode the mode here - while it's unlikely to change, if someone wants to make it (or more likely all of /usr including subdirs) group writable (775), or alternatively not user writable (555) that'd be up to the distro/OS.

crates/initramfs/src/lib.rs
```diff
@@ -169,8 +169,8 @@ pub fn open_dir(dirfd: impl AsFd, name: impl AsRef<Path> + Debug) -> Result<Owne
 }
 
 #[context("Ensure dir")]
-fn ensure_dir(dirfd: impl AsFd, name: &str) -> Result<OwnedFd> {
-    match mkdirat(dirfd.as_fd(), name, 0o700.into()) {
+fn ensure_dir(dirfd: impl AsFd, name: &str, mode: Option<u32>) -> Result<OwnedFd> {
```
cgwalters: minor: more type safety by using `rustix::fs::Mode`

Johan-Liebert1: Updated

crates/lib/src/bootc_composefs/state.rs
Johan-Liebert1: > if someone wants to make it (or more likely all of /usr including subdirs) group writable (775), or alternatively not user writable (555)

How would we want to do this? Would we take mode as an optional cli opt?

cgwalters: Ah yeah I was unclear. I'm just saying we should inherit the mode of the underlying directory - seems that simple to me

PR #1846, opened 2025-12-11
lib: Add --from-downloaded flag for bootc upgrade

shi2wei3:
Add a new --from-downloaded flag to bootc upgrade that allows users to unlock a staged deployment created with --download-only without checking for newer updates from the registry.

This provides a way to apply already-downloaded updates without triggering a registry check, which is useful for scheduled maintenance workflows where the update was downloaded earlier and should now be applied at a scheduled time.

Usage:
  # Download update without applying
  bootc upgrade --download-only

  # Later: Apply the staged update (no registry check)
  bootc upgrade --from-downloaded

  # Or: Apply staged update and reboot immediately
  bootc upgrade --from-downloaded --apply

The flag conflicts with --check and --download-only as those operations have different purposes. It can be combined with --apply to immediately reboot after unlocking the staged deployment.

This commit also updates the documentation (upgrades.md) to describe all three ways to apply a download-only update, and updates the download-only test case (test-25) to use --apply-staged instead of plain `bootc upgrade` when clearing the download-only flag.

### DIFF
--- a/crates/lib/src/cli.rs
+++ b/crates/lib/src/cli.rs
@@ -108,6 +108,14 @@ pub(crate) struct UpgradeOpts {
     #[clap(long, conflicts_with_all = ["check", "apply"])]
     pub(crate) download_only: bool,
 
+    /// Apply a staged deployment that was previously downloaded with --download-only.
+    ///
+    /// This unlocks the staged deployment without fetching updates from the container image source.
+    /// The deployment will be applied on the next shutdown or reboot. Use with --apply to
+    /// reboot immediately.
+    #[clap(long, conflicts_with_all = ["check", "download_only"])]
+    pub(crate) from_downloaded: bool,
+
     #[clap(flatten)]
     pub(crate) progress: ProgressOptions,
 }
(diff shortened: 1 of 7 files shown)


crates/lib/src/cli.rs
```diff
@@ -108,6 +108,14 @@ pub(crate) struct UpgradeOpts {
     #[clap(long, conflicts_with_all = ["check", "apply"])]
     pub(crate) download_only: bool,
 
+    /// Apply a staged deployment that was previously downloaded with --download-only.
+    ///
+    /// This unlocks the staged deployment without checking for newer updates from the registry.
+    /// The deployment will be applied on the next shutdown or reboot. Use with --apply to
+    /// reboot immediately.
+    #[clap(long, conflicts_with_all = ["check", "download_only"])]
+    pub(crate) apply_staged: bool,
```
cgwalters: `--apply-staged --apply` "stutters" i.e. at first it looks redundant but it's not.

How about calling this `--enqueue`?

But maybe we can also add a shorthand `--enqueue --apply` ➡️  `--apply-from-download` ?

Or maybe the option is called `--from-downloaded` ? I kind of like that more than `--enqueue`.

shi2wei3: Thanks for the suggestion, `--from-downloaded` sounds more natural, updated.

docs/src/upgrades.md
```diff
@@ -84,15 +96,23 @@ bootc status --verbose
...
 
-# Option B: Apply immediately with reboot
-bootc upgrade --apply
+# Option C: Check for newer updates first, then apply
+bootc upgrade
 ```
 
 **Important notes**:
 
+- **Image source check difference**: `bootc upgrade --from-downloaded` does NOT fetch from the
```
cgwalters, line 111: This paragraph feels redundant? We explained it just above right?

cgwalters (approved):

PR #1847, opened 2025-12-11
WIP: Use a registry with TMT tests

ckyrouac:
Most of this was written by Claude and needs to be cleaned up. It's very much a WIP, but I want to get feedback on the high level design before going too far into the cleanup. Please focus on the README changes (contained in a separate commit) that explain the design. I have tested the code at least works and you can try it out if you want.

The key design points (explained in detail in the READMEs)
- A separate registry VM is started before the test VMs
- The registry connection details are passed as environment variables to each test
- The base bootc image under test is pushed to the registry
- Individual tests can be defined as a directory or a single file
- Additional images required for a test must be defined in an `images` sub directory of the test
- The pre-built images are passed to the tests as environment variables

The primary reason for this change is to support building sealed images on the test runner without needing to copy keys into the test VM and copy-to-storage. An added benefit is our tests will run in a more realistic environment (i.e. using a registry instead of building on the system under test).


crates/xtask/src/tmt.rs
```diff
@@ -35,6 +35,47 @@ const DISTRO_CENTOS_9: &str = "centos-9";
 // Import the argument types from xtask.rs
 use crate::{RunTmtArgs, TmtProvisionArgs};
 
+/// RAII guard to ensure VM cleanup on drop unless explicitly preserved
+struct VmCleanupGuard<'a> {
```
cgwalters, line 39: Fine as is but I filed https://github.com/bootc-dev/bcvk/issues/173 related to this

crates/xtask/src/tmt.rs
```diff
@@ -118,7 +159,17 @@ fn wait_for_vm_ready(sh: &Shell, vm_name: &str) -> Result<(u16, String)> {
                     json.get("ssh_private_key").and_then(|v| v.as_str()),
                 ) {
                     let ssh_port = ssh_port as u16;
-                    return Ok((ssh_port, ssh_key.to_string()));
+                    // Try to get IP address from network interfaces
+                    let ip = json
```
cgwalters, line 163: I think this can use https://docs.rs/serde_json/1.0.145/serde_json/enum.Value.html#method.pointer

crates/xtask/src/tmt.rs
```diff
@@ -178,6 +229,27 @@ fn verify_ssh_connectivity(sh: &Shell, port: u16, key_path: &Utf8Path) -> Result
...
+/// Save SSH private key to a temporary file with proper permissions (0600)
+/// Returns the temporary file (which must be kept alive) and its path
+#[context("Saving SSH key")]
+fn save_ssh_key(key: &str) -> Result<(tempfile::NamedTempFile, Utf8PathBuf)> {
+    let key_file = tempfile::NamedTempFile::new().context("Failed to create SSH key file")?;
+
+    let key_path = Utf8PathBuf::try_from(key_file.path().to_path_buf())
+        .context("Failed to convert key path to UTF-8")?;
+
+    std::fs::write(&key_path, key).context("Failed to write SSH key")?;
```
cgwalters, line 241: key_file.write_all(key.as_bytes())

crates/xtask/src/tmt.rs
```diff
@@ -215,6 +287,274 @@ fn parse_plan_metadata(plans_file: &Utf8Path) -> Result<std::collections::HashMa
...
+
+        println!(
+            "  Building {} from {}/images/{}...",
+            image_ref, test_name_prefix, image.image_dir
+        );
+
+        // Build context is images/<tag>/ subdirectory
+        cmd!(
+            sh,
+            "podman build -t {image_ref} -f {containerfile_path} {image_dir}"
```
cgwalters, line 339: Hmm. I'd prefer that we do things like build container images from `Justfile`. Arguably `just build` should build *all* images by default?

crates/xtask/src/tmt.rs
```diff
@@ -215,6 +287,274 @@ fn parse_plan_metadata(plans_file: &Utf8Path) -> Result<std::collections::HashMa
...
+        return Ok(());
+    }
+
+    println!("\nCleaning up {} custom image(s)...", images.len());
+
+    for image in images {
+        // Remove from local storage
+        // We don't remove from registry since it's a test registry that gets torn down anyway
+        let full_tag = &image.full_tag;
+        let _ = cmd!(sh, "podman rmi -f {full_tag}")
```
cgwalters, line 375: How about `just image-cleanup` or something? (which could call into this) also on this topic...I tried to ensure that every image we build is tagged with a label, so we should make it easy to remove all local images we built.

hack/Containerfile.registry
```diff
@@ -0,0 +1,105 @@
...
+Restart=always
+
+[Install]
+WantedBy=multi-user.target
+EOF
+
+ln -s /usr/share/containers/systemd/registry.container /usr/lib/bootc/bound-images.d/registry.container
+
+cat > /etc/selinux/config <<'EOF'
+SELINUX=permissive
```
cgwalters, line 101: I think this needs a comment for why we're doing this

crates/xtask/src/tmt.rs
```diff
@@ -215,6 +287,274 @@ fn parse_plan_metadata(plans_file: &Utf8Path) -> Result<std::collections::HashMa
...
+    /// SSH private key
+    ssh_key: String,
+    /// IP address where test VMs can reach the registry (host gateway IP)
+    registry_ip: String,
+}
+
+/// Setup and configure a persistent registry VM for test images
+/// Returns registry VM info and cleanup guard (which will cleanup VM on drop unless disarmed)
+#[context("Setting up registry VM")]
+fn setup_registry_vm<'a>(
```
cgwalters, line 401: Hmm just so I understand here - is this only used for non-testing-farm cases i.e. in GHA? A toplevel goal of this effort is that with testing-farm we'd be spinning up two VMs in the IaaS it uses (e.g. AWS or whatever) right?

crates/xtask/src/tmt.rs
```diff
@@ -215,6 +287,274 @@ fn parse_plan_metadata(plans_file: &Utf8Path) -> Result<std::collections::HashMa
...
+    // Save SSH key to temporary file
+    let (_registry_key_file, registry_key_path) = save_ssh_key(&key)?;
+
+    // Verify SSH connectivity
+    println!("Verifying SSH connectivity to registry VM...");
+    verify_ssh_connectivity(sh, port, &registry_key_path)
+        .context("SSH verification failed for registry VM")?;
+    println!("Registry VM SSH connectivity verified");
+
+    // The registry is exposed on the host via port forwarding (--port 5000:5000)
```
cgwalters, line 440: The hardcoded port will preclude concurrent test runs in the same environment. It'd be a bit cleaner to allocate a dynamic port probably, which argues for bcvk supporting something much like `podman run -P` here (which in turn kind of argues for libvirt supporting it)

crates/xtask/src/tmt.rs
```diff
@@ -215,6 +287,274 @@ fn parse_plan_metadata(plans_file: &Utf8Path) -> Result<std::collections::HashMa
...
+    // Test VMs need to reach the host's IP, which is the default gateway
+    // Get the gateway IP dynamically from the registry VM's default route
+    println!("Getting host gateway IP from registry VM...");
+    let port_str = port.to_string();
+    let gateway_output = cmd!(
+        sh,
+        "ssh -i {registry_key_path} -p {port_str} -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o ConnectTimeout=5 -o IdentitiesOnly=yes root@localhost ip route show default"
```
cgwalters, line 447: Er...but isn't that just the host IP?

Hmm does testing-farm multihost networking inject hostnames or anything? Is all we have https://tmt.readthedocs.io/en/stable/spec/plans.html#spec-plans-guest-topology ?

crates/xtask/src/tmt.rs
```diff
@@ -215,6 +287,274 @@ fn parse_plan_metadata(plans_file: &Utf8Path) -> Result<std::collections::HashMa
...
+    let registry_ip = gateway_output
+        .split_whitespace()
+        .skip_while(|&word| word != "default")
+        .skip(1) // skip "default"
+        .find(|&word| word != "via") // find first word after "via"
+        .unwrap_or("")
+        .trim();
+
+    // Validate that it's a valid IP address
+    let _ip_addr: std::net::IpAddr = registry_ip.parse().with_context(|| {
```
cgwalters, line 462: Maybe a bit cleaner to drop the `_` and have `let registry_ip = ip_addr.to_string()` or so? Then we're actually using the value we parsed

crates/xtask/src/tmt.rs
```diff
@@ -215,6 +287,274 @@ fn parse_plan_metadata(plans_file: &Utf8Path) -> Result<std::collections::HashMa
...
+    println!(
+        "Registry will be accessible to test VMs at {} (host gateway)",
+        registry_ip
+    );
+
+    // Push base image to registry
+    // Host pushes to localhost:5000
+    // Test VMs use hostname (configured via TMT prepare)
+    println!("Pushing base image to registry...");
+    let registry_url = "localhost:5000".to_string();
```
cgwalters, line 480: Let's use a `const` instead of 5000 in multiple places

.gitignore
```diff
@@ -5,3 +5,6 @@ bootc.tar.zst
 
 # Added by cargo
 /target
+
+# Registry TLS certificates (generated at build time)
+/hack/.registry-certs
```
cgwalters, line 10: I'm ok with this but note the secure-boot keys which are similar are in `target` which I feel is cleaner; basically it's the default directory for build-time artifacts.

Justfile
```diff
@@ -140,18 +145,28 @@ validate:
...
     ./tests/build-sealed {{variant}} {{integration_upgrade_img}}-bin {{integration_upgrade_img}} {{buildroot_base}}
 
+# Build a registry VM image for multi-VM testing
+# Uses Podman Quadlet for idiomatic container-as-service setup
+_build-registry-image:
+    # Generate TLS certificates for the registry (idempotent - skips if exists)
+    ./hack/setup-registry-certs.sh
+    # Build registry image with Quadlet configuration
+    # Pre pull registry container to be used as a LBI
+    podman pull quay.io/libpod/registry:2.8.2
```
cgwalters, line 163: Not a blocker but we'll need to ensure stuff like this is tagged so that renovate can handle bumping it. Probably best as a Justfile variable at the top too.

I would also say we should not use that specific container image which I think is only for the podman team's CI. https://github.com/distribution/distribution is part of CNCF too, but obviously there's a lot of choices for registries.

Dockerfile
```diff
@@ -74,5 +74,22 @@ RUN --mount=type=bind,from=packaging,target=/run/packaging \
...
+if [ -f /run/src/src/hack/.registry-certs/ca.pem ]; then
+    echo "Installing registry CA certificate to trust store..."
+    cp /run/src/src/hack/.registry-certs/ca.pem /usr/share/pki/ca-trust-source/anchors/bootc-registry-ca.crt
+    update-ca-trust
+    echo "✓ Registry CA certificate installed"
+else
+    echo "Note: Registry CA certificate not found - registry will need --tls-verify=false"
+    echo "To enable secure registry access, run: hack/setup-registry-certs.sh"
+    exit 1
+fi
```
cgwalters, line 92: This is all 100% fine as is but I think it'd be cleaner and clearer written like this:

```
if test '!' -f /run/src/src/hack/.registry-certs/ca.pem; then
  echo "hack/setup-registry-certs.sh must be invoked" 1>&2; exit 1
fi
cp /run/src/src/hack/.registry-certs/ca.pem /usr/share/pki/ca-trust-source/anchors/bootc-registry-ca.crt
update-ca-trust
```

The idea is we check for exceptions/errors first, and the "happy path" is the default one.

tmt/tests/booted/test-20-image-pushpull-upgrade/images/01-bootc-derived/usr/lib/bootc/kargs.d/05-testkargs.toml
```diff
@@ -0,0 +1 @@
+kargs = ["testarg=foo", "othertestkarg", "thirdkarg=bar"]
```
cgwalters, line 1: Not related to this specific line but just a note: since this PR is changing *all* of the tests it's going to be an ongoing merge conflict fest.

One thing that would likely here is to split up any "prep" changes that can land independently first

cgwalters (commented):
Thanks for working on this!

cgwalters:
xref https://github.com/bootc-dev/bootc/pull/1852

PR #1848, opened 2025-12-11
tests: Remove composefs-bcvk test infrastructure

cgwalters:
The composefs-bcvk tests were a separate test path that ran integration tests directly via bcvk. This functionality is now covered by the tmt-based testing framework, so remove the redundant infrastructure.


henrywang:
The fedora rawhide failure should be related with rawhide change https://fedoraproject.org/wiki/Changes/BootLoaderUpdatesPhase1. We have discussion in slack.

PR #1852, opened 2025-12-14
lib: Pass absolute authfile path when pulling LBIs

csssuf:
ostree-ext explicitly handles authfile paths as relative; this works fine for most callers of get_global_authfile, as they only read the returned open file descriptor, and ignore the path. However, pulling logically bound images requires passing the actual authfile path to Podman, so we must resolve the absolute path in this case - otherwise, we see errors like the following:

```
[root@fedora ~]# bootc upgrade
layers already present: 69; layers needed: 1 (242.2 MB)
Fetched layers: 230.95 MiB in 3 seconds (90.88 MiB/s)
  Deploying: done (3 seconds)
  Fetching bound image: quay.io/prometheus/node-exporter:v1.10.2: done (0 seconds)
error: Upgrading: Staging: Pulling bound images: Pulling bound images: Failed to pull image: Subprocess failed: ExitStatus(unix_wait_status(32000))
Error: credential file is not accessible: faccessat etc/ostree/auth.json: no such file or directory
```

Since cap_std::fs::Dir intentionally does not expose its filesystem path, we must resort to reconstructing it from a file descriptor. We could do this by inspectingthe file descriptor for `sysroot` and combining that with the relative path returned by get_global_authfile, but since get_global_authfile returns the descriptor of the actual authfile, we can simply read that directly.

### DIFF
--- a/crates/lib/src/podstorage.rs
+++ b/crates/lib/src/podstorage.rs
@@ -9,7 +9,7 @@
 //! At the current time, this is only used for Logically Bound Images.
 
 use std::collections::HashSet;
-use std::io::Seek;
+use std::io::{Seek, Write};
 use std::os::unix::process::CommandExt;
 use std::process::{Command, Stdio};
 use std::sync::Arc;
@@ -17,14 +17,14 @@ use std::sync::Arc;
 use anyhow::{Context, Result};
 use bootc_utils::{AsyncCommandRunExt, CommandRunExt, ExitStatusExt};
 use camino::{Utf8Path, Utf8PathBuf};
-use cap_std_ext::cap_std;
 use cap_std_ext::cap_std::fs::Dir;
 use cap_std_ext::cap_tempfile::TempDir;
 use cap_std_ext::cmdext::CapStdExtCommandExt;
 use cap_std_ext::dirext::CapStdExtDirExt;
+use cap_std_ext::{cap_std, cap_tempfile};
 use fn_error_context::context;
 use ostree_ext::ostree::{self};
-use std::os::fd::{AsRawFd, OwnedFd};
+use std::os::fd::{AsFd, AsRawFd, OwnedFd};
 use tokio::process::Command as AsyncCommand;
 
 // Pass only 100 args at a time just to avoid potentially overflowing argument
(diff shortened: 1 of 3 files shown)


cgwalters:
Hm, this is a CI gap for sure in terms of LBIs with auth. https://github.com/bootc-dev/bootc/pull/1847 would help here.

crates/lib/src/podstorage.rs
```diff
@@ -353,10 +353,15 @@ impl CStorage {
...
         cmd.stdout(Stdio::null());
         cmd.args(["pull", image]);
-        let authfile = ostree_ext::globals::get_global_authfile(&self.sysroot)?
-            .map(|(authfile, _fd)| authfile);
-        if let Some(authfile) = authfile {
-            cmd.args(["--authfile", authfile.as_str()]);
+        let authfile_fd =
+            ostree_ext::globals::get_global_authfile(&self.sysroot)?.map(|(_authfile, fd)| fd);
+        if let Some(fd) = authfile_fd {
+            let authfile_path = std::fs::read_link(format!("/proc/self/fd/{}", fd.as_raw_fd()))
```
cgwalters: I don't think this will work because the fd should have `O_CLOEXEC` and won't be passed to the child by default.

The general pattern we use here instead is using https://docs.rs/cap-std-ext/latest/cap_std_ext/cmdext/trait.CapStdExtCommandExt.html#tymethod.take_fd_n 

In fact we should just make this code identical to the one in containers-image-proxy-rs: https://github.com/bootc-dev/containers-image-proxy-rs/blob/b5e0861ad5065f47eaf9cda0d48da3529cc1bc43/src/imageproxy.rs#L310

cgwalters (changes requested):
Thanks for finding this bug and trying a patch!

Can you explain if you built and tested this, or was this more of a drive-by?

csssuf:
I have built and tested it! From an earlier testing iteration with a debug print still in there:

```
[root@fedora ~]# bootc-local upgrade
Fetched layers: 0 B in 0 seconds (0 B/s)
  Deploying: done (3 seconds)
  Fetching bound image: quay.io/prometheus/node-exporter:v1.10.2
authfile_path: "/etc/ostree/auth.json"
  Fetching bound image: quay.io/prometheus/node-exporter:v1.10.2: done (1 second)
Bound images stored: 1
Pruned images: 0 (layers: 0, objsize: 726.8 MB)
Queued for next boot: 10.0.2.2:8000/localhost/test-bound-image
  Version: 43.20251213.0
  Digest: sha256:5cf3430f72d9dae0fc223f41ee13b5eec04e5d270369688a5b1415ce5047448b
Total new layers: 70    Size: 2.1 GB
Removed layers:   0     Size: 0 bytes
Added layers:     0     Size: 0 bytes
[root@fedora ~]#
```

crates/lib/src/podstorage.rs
csssuf: I don't believe we rely on passing it to the child here - we only use the fd in the current process to look up the absolute path through `/proc`, and just pass the path to the child since podman's `--authfile` option expects a path.

Using `/proc` here _does_ feel a little hacky, but as far as I can tell `cap_std::fs::Dir` is very deliberate in not exposing the underlying path, but open to other suggestions here - would copying the contents as in the imageproxy implementation be preferable to resolving the absolute path this way?

cgwalters: I pushed a followup commit here, can you review it?

cgwalters:
Since we have two people who wrote code here I think SOP should be for a 3rd to review so passing this over to @ckyrouac for final review/approval

Issue #1849, opened 2025-12-12
rawhide tests "failing with error: Status: loading sysroot: Unexpected state: /run/ostree-booted found, but no /boot/loader directory"


jmarrero:
```
error: Status: loading sysroot: Unexpected state: /run/ostree-booted found, but no /boot/loader directory
Error: nu::shell::incompatible_path_access

  × Data cannot be accessed with a cell path
    ╭─[/var/ARTIFACTS/work-plan-01-readonlynp2u0i0j/tmt/plans/integration/plan-01-readonly/discover/default-0/tests/tmt/tests/booted/readonly/001-test-status.nu:17:25]
 16 │ # Detect composefs by checking if composefs field is present
 17 │ let is_composefs = ($st.status.booted.composefs? != null)
```

An full example: https://artifacts.dev.testing-farm.io/5011fe3d-91ea-4bb2-8f81-f4e670f7b41e/

openning this as we have seen these on: https://github.com/bootc-dev/bootc/pull/1809 and now on https://github.com/bootc-dev/bootc/pull/1601 which unified storage does not touch the bootloader.  

This is only rawhide so something might have changed recently affecting how the boot/loader directory is created?

cgwalters:
@henrywang said this might be related to https://fedoraproject.org/wiki/Changes/BootLoaderUpdatesPhase1

HuijingHei:
Can reproduce locally on rawhide, but looks wired about the mount, maybe we do not handle correctly the mount on btrfs

- Start Fedora-rawhide and check `/etc/fstab` before running bootc install

```
[fedora@citest-1 ~]$ cat /etc/fstab 
UUID=999de14f-18cf-40ba-99cc-c229a1225da0 / btrfs compress=zstd:1,defaults,subvol=root 0 1
UUID=999de14f-18cf-40ba-99cc-c229a1225da0 /boot btrfs compress=zstd:1,subvol=boot 0 0
UUID=999de14f-18cf-40ba-99cc-c229a1225da0 /home btrfs compress=zstd:1,subvol=home 0 0
UUID=999de14f-18cf-40ba-99cc-c229a1225da0 /var btrfs compress=zstd:1,subvol=var 0 0
UUID=4824-0C80 /boot/efi vfat defaults,umask=0077,shortname=winnt 0 0
```

- Run bootc install
```
podman run --rm --privileged -v /dev:/dev -v /var/lib/containers:/var/lib/containers -v /:/target \
[...]

PR #1856, opened 2025-12-16
install: read boot entry from `/etc/fstab` and write to new created bootc deployment

HuijingHei:
On rawhide, the /boot is subvol instead of separate partition, we need to use the same option in the created deployment.

Fixes https://github.com/bootc-dev/bootc/issues/1849

### DIFF
--- a/crates/lib/src/install.rs
+++ b/crates/lib/src/install.rs
@@ -2235,9 +2235,15 @@ pub(crate) async fn install_to_filesystem(
             Some(MountSpec::new(&spec, "/boot"))
         }
     } else {
-        boot_uuid
-            .as_deref()
-            .map(|boot_uuid| MountSpec::new_uuid_src(boot_uuid, "/boot"))
+        // Read /etc/fstab to get boot entry, but only use it if it's UUID-based
+        // Otherwise fall back to boot_uuid
+        read_boot_fstab_entry(&rootfs_fd)?
+            .filter(|spec| spec.get_source_uuid().is_some())
+            .or_else(|| {
+                boot_uuid
+                    .as_deref()
+                    .map(|boot_uuid| MountSpec::new_uuid_src(boot_uuid, "/boot"))
+            })
     };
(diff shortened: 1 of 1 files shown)


crates/lib/src/install.rs
```diff
@@ -2235,9 +2235,15 @@ pub(crate) async fn install_to_filesystem(
             Some(MountSpec::new(&spec, "/boot"))
         }
     } else {
-        boot_uuid
-            .as_deref()
-            .map(|boot_uuid| MountSpec::new_uuid_src(boot_uuid, "/boot"))
+        // Read /etc/fstab to get boot entry, but only use it if it's UUID-based
```
cgwalters, line 2238: Hum...I think we can unconditionally propagate this; I don't think we need to filter to UUID only.

cgwalters (approved):
Thanks so much for digging into this! It looks like there's a backup in the testing farm queue but we can merge after seeing the rawhide branch be green.

crates/lib/src/install.rs
HuijingHei: Yes, I tried without condition firstly, but the install test failed with
```
$ sudo podman run --rm --privileged --pid=host -v /:/target -v /tmp/xshell-tmp-dir-1/test_authorized_keys:/test_authorized_keys localhost/bootc-integration-install bootc install to-filesystem --acknowledge-destructive --karg=foo=bar --replace=alongside --root-ssh-authorized-keys=/test_authorized_keys /target
test replace=alongside with ssh keys and a karg, and SELinux disabled ... Installing image: docker://localhost/bootc-integration-install:latest
Digest: sha256:f718c825e84eb8f5f3854b815a282ef80773a5f2320e5effe0c885a3146a5e83
error: Installing to filesystem: /boot is not specified via UUID= (this is currently required)
```
See https://github.com/bootc-dev/bootc/actions/runs/20265544961/job/58187786591

PR #1858, opened 2025-12-16
packaging: Enable reproducible RPM builds and add verification

cgwalters:
Pass SOURCE_DATE_EPOCH from git commit timestamp through to rpmbuild, enabling bit-for-bit reproducible RPM builds. This is useful for verification and caching.

Key changes:
- Justfile: Set SOURCE_DATE_EPOCH from git log and pass to podman build
- Dockerfile: Export SOURCE_DATE_EPOCH as ENV for rpmbuild visibility
- build-rpm: Add rpmbuild defines for reproducibility:
  - use_source_date_epoch_as_buildtime
  - clamp_mtime_to_source_date_epoch
  - _buildhost=reproducible
- bootc.spec: Sort find output for deterministic file ordering

Add `just check-buildsys` command that builds packages twice and verifies checksums match, confirming reproducibility. The CI package job now uses this to catch regressions.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -8,7 +8,11 @@
 # Rules written here are *often* used by the Github Action flows,
 # and should support being configurable where that makes sense (e.g.
 # the `build` rule supports being provided a base image).
-
+#
+# By default the layering should be thus:
+# Github Actions -> Justfile -> podman -> make -> rustc
+#                            -> podman -> dnf|apt ...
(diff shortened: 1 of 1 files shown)


cgwalters:
OK, yeah this is WAY better for local iteration, and is correctly building the binary just once per distro/OS in CI, saving ~10m per job right now.

PR #1859, opened 2025-12-16
ostree-ext: Fix warning on rust 1.9.2

jeckersb:
This started warning with:

warning: value assigned to `no_signature_verification` is never read
    --> crates/ostree-ext/src/cli.rs:1265:21
     |
1265 |                     mut no_signature_verification,
     |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = help: maybe it is overwritten before being read?
     = note: `#[warn(unused_assignments)]` (part of `#[warn(unused)]`) on by default

Which is true in this case, we never actually read the pattern-matched
value.  Change it to be explicitly unused with '_' and then shadow it
with a `let` binding.

### DIFF
--- a/crates/ostree-ext/src/cli.rs
+++ b/crates/ostree-ext/src/cli.rs
@@ -1262,7 +1262,7 @@ async fn run_from_opt(opt: Opt) -> Result<()> {
                     imgref,
                     image,
                     transport,
-                    mut no_signature_verification,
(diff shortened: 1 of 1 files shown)


jeckersb:
Well, rust 1.92, but I'm not going to retrigger CI on an obvious typo :smile:

PR #1860, opened 2025-12-17
ci: add crates-release.yml workflow to publish to crates.io

HuijingHei:
Publish two crates:
`bootc-internal-utils` and `bootc-internal-blockdev`

Part of https://github.com/bootc-dev/infra/issues/20

### DIFF
--- a/.github/workflows/crates-release.yml
+++ b/.github/workflows/crates-release.yml
@@ -1,0 +1,22 @@
+# See https://crates.io/docs/trusted-publishing
+name: Publish to crates.io
+on:
+  push:
+    tags: ['v*']  # Triggers when pushing tags starting with 'v'
+jobs:
+  publish:
+    runs-on: ubuntu-24.04
+    permissions:
+      id-token: write     # Required for OIDC token exchange
+    steps:
+    - uses: actions/checkout@v6
+    - uses: rust-lang/crates-io-auth-action@v1
+      id: auth
+    - run: |
+        for crate in bootc-internal-utils bootc-internal-blockdev; do
+          echo "Publishing $crate..."
+          cargo publish -p "$crate"
+          echo "Successfully published $crate"
+        done
+      env:
+        CARGO_REGISTRY_TOKEN: ${{ steps.auth.outputs.token }}


.github/workflows/crates-release.yml
```diff
@@ -0,0 +1,30 @@
+# See https://crates.io/docs/trusted-publishing
+name: Publish to crates.io
+on:
+  workflow_dispatch:
```
ckyrouac: I think we'll want to automate this, similar to https://github.com/bootc-dev/containers-image-proxy-rs/blob/b5e0861ad5065f47eaf9cda0d48da3529cc1bc43/.github/workflows/cratesio-release.yml

.github/workflows/crates-release.yml
```diff
@@ -0,0 +1,30 @@
...
+jobs:
+  publish:
+    runs-on: ubuntu-24.04
+    permissions:
+      id-token: write     # Required for OIDC token exchange
+    steps:
+    - uses: actions/checkout@v6
+    - uses: rust-lang/crates-io-auth-action@v1
+      id: auth
+    - run: cargo publish -p ${{ inputs.crate }} --dry-run
```
ckyrouac: I don't think running `--dry-run` first is necessary

.github/workflows/crates-release.yml
cgwalters: Yeah agree, I think we can auto-publish the crates on release tags.

.github/workflows/crates-release.yml
```diff
@@ -0,0 +1,30 @@
...
+name: Publish to crates.io
+on:
+  workflow_dispatch:
+    inputs:
+      crate:
+        description: 'Crate to publish'
+        required: true
+        type: choice
+        options:
+          - bootc-internal-utils
```
cgwalters: Combining with the previous I think we can just publish all the relevant crates on git tags.

.github/workflows/crates-release.yml
HuijingHei: Removed that, thank you!

.github/workflows/crates-release.yml
HuijingHei: Updated, and now only publish two crates: `bootc-internal-utils` and `bootc-internal-blockdev`

henrywang:
Failure has been fixed by PR https://github.com/bootc-dev/bootc/pull/1867. This PR needs a rebase. Thanks.

HuijingHei:
Updated, thank you!

.github/workflows/crates-release.yml
```diff
@@ -0,0 +1,22 @@
...
+  publish:
+    runs-on: ubuntu-24.04
+    permissions:
+      id-token: write     # Required for OIDC token exchange
+    steps:
+    - uses: actions/checkout@v6
+    - uses: rust-lang/crates-io-auth-action@v1
+      id: auth
+    - run: |
+        for crate in bootc-internal-utils bootc-internal-blockdev; do
```
cgwalters, line 16: I think we also need to add a step to our release workflow which bumps the versions of these crates right?

cgwalters:
Overall though I will say...we have huge amounts of mechanical overhead from keeping bootc and bootupd separate and I think https://github.com/coreos/bootupd/issues/432 is still very relevant.

Issue #1862, opened 2025-12-17
`bootc internals cfs compute-id` requires a booted system


jeckersb:
This should be a thing that works:

```
jeckersb@toolbx bootc> podman run --rm -it --mount=type=image,src=localhost/bootc,dst=/target localhost/bootc bootc internals cfs compute-id /target                                                                                      12/17/2025 11:19:54 AM
error: Initializing storage: Preparing for write: Detected container; this command requires a booted host system.
```

cgwalters:
Yes, though we might as well scope in to this having `bootc container compute-composefs-digest` default to operating on a `/target` (and have the current thing become `bootc container compute-composefs-digest-from-storage` which we `hide=true`?)

PR #1863, opened 2025-12-17
container: Add path-based compute-composefs-digest command

jeckersb:
Add a new `bootc container compute-composefs-digest` command that computes
the bootable composefs digest directly from a filesystem directory path,
defaulting to `/target`. This enables computing digests in container
environments without requiring access to container storage or a booted
host system.

The existing container-storage-based behavior is preserved and renamed
to `compute-composefs-digest-from-storage` (hidden). The `hack/compute-composefs-digest`
script is updated to use the renamed command.

Computing the digest on `/` is explicitly rejected with a clear error
message, as this is not supported.

Closes: https://github.com/bootc-dev/bootc/issues/1862

### DIFF
--- a/crates/lib/src/cli.rs
+++ b/crates/lib/src/cli.rs
@@ -7,7 +7,6 @@ use std::fs::File;
 use std::io::{BufWriter, Seek};
 use std::os::unix::process::CommandExt;
 use std::process::Command;
-use std::sync::Arc;
 
 use anyhow::{anyhow, ensure, Context, Result};
 use camino::{Utf8Path, Utf8PathBuf};
@@ -32,10 +31,10 @@ use ostree_ext::ostree;
 use ostree_ext::sysroot::SysrootLock;
 use schemars::schema_for;
 use serde::{Deserialize, Serialize};
-use tempfile::tempdir_in;
 
 use crate::bootc_composefs::delete::delete_composefs_deployment;
 use crate::bootc_composefs::{
+    digest::{compute_composefs_digest, new_temp_composefs_repo},
     finalize::{composefs_backend_finalize, get_etc_diff},
     rollback::composefs_rollback,
     state::composefs_usr_overlay,
@@ -48,7 +47,7 @@ use crate::podstorage::set_additional_image_store;
 use crate::progress_jsonl::{ProgressWriter, RawProgressFd};
 use crate::spec::Host;
 use crate::spec::ImageReference;
-use crate::store::{BootedOstree, ComposefsRepository, Storage};
+use crate::store::{BootedOstree, Storage};
 use crate::store::{BootedStorage, BootedStorageKind};
 use crate::utils::sigpolicy_from_opt;
 
@@ -358,9 +357,20 @@ pub(crate) enum ContainerOpts {
         #[clap(long)]
         no_truncate: bool,
     },
-    /// Output the bootable composefs digest.
+    /// Output the bootable composefs digest for a directory.
     #[clap(hide = true)]
     ComputeComposefsDigest {
+        /// Path to the filesystem root
+        #[clap(default_value = "/target")]
+        path: Utf8PathBuf,
+
+        /// Additionally generate a dumpfile written to the target path
+        #[clap(long)]
+        write_dumpfile_to: Option<Utf8PathBuf>,
+    },
+    /// Output the bootable composefs digest from container storage.
+    #[clap(hide = true)]
+    ComputeComposefsDigestFromStorage {
         /// Additionally generate a dumpfile written to the target path
         #[clap(long)]
         write_dumpfile_to: Option<Utf8PathBuf>,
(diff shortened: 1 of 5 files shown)


jeckersb:
Worth noting explicitly that these two approaches do *not* agree, which is mostly (or entirely?) due to https://github.com/containers/composefs-rs/issues/132 and all that mess.

```
jeckersb@toolbx bootc> podman run --rm -it --mount=type=image,src=localhost/bootc,dst=/target localhost/bootc bootc container compute-composefs-digest                                                                                    12/17/2025 12:08:17 PM
66a8f3ba996bdfa70dd908ffe78412bae1102189bc4207cdb5990904004551f83d897a421fd6f1f9b81bae256c45068e28ad02a7bb10c57a0d5a0da77f59525a
jeckersb@toolbx bootc> hack/compute-composefs-digest localhost/bootc                                                                                                                                                                      12/17/2025 12:08:30 PM
71203ee0983088b9d50b1f909cfff16b569826b2fcc7a764f9b1c06e2215a9f50f24c94e4569187937fa9a0f28cf690f5f8912d49c024db672c5653ae655229f
```

But... progress.

crates/lib/src/cli.rs
```diff
@@ -1499,6 +1510,49 @@ async fn run_from_opt(opt: Opt) -> Result<()> {
...
+                let td = td.path();
+                let td = &Dir::open_ambient_dir(td, cap_std::ambient_authority())?;
+
+                td.create_dir("repo")?;
+                let repo = td.open_dir("repo")?;
+                let mut repo =
+                    ComposefsRepository::open_path(&repo, ".").context("Init cfs repo")?;
+                // We don't need to hard require verity on the *host* system, we're just computing a checksum here
+                repo.set_insecure(true);
+                let repo = Arc::new(repo);
```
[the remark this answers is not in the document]
jeckersb: I'm less worried about DRYing this up since (famous last words) we want to remove `ComputeComposefsDigestFromStorage` once it's no longer needed.

crates/lib/src/cli.rs
```diff
@@ -1499,6 +1510,49 @@ async fn run_from_opt(opt: Opt) -> Result<()> {
                 Ok(())
             }
             ContainerOpts::ComputeComposefsDigest {
+                path,
+                write_dumpfile_to,
+            } => {
+                if path.as_str() == "/" {
+                    anyhow::bail!(
+                        "Computing composefs-digest on '/' is not supported at this time"
```
cgwalters: How about `Cannot operate on active root filesystem; mount separate target instead`?

crates/lib/src/cli.rs
```diff
@@ -358,9 +358,20 @@ pub(crate) enum ContainerOpts {
...
+        /// Path to the filesystem root
+        #[clap(default_value = "/target")]
+        path: Utf8PathBuf,
+
+        /// Additionally generate a dumpfile written to the target path
+        #[clap(long)]
+        write_dumpfile_to: Option<Utf8PathBuf>,
+    },
+    /// Output the bootable composefs digest from container storage.
+    #[clap(hide = true, name = "compute-composefs-digest-from-storage")]
```
cgwalters: `name` is unnecessary here right as the default kebab-case should apply I think?

crates/lib/src/cli.rs
```diff
@@ -1499,6 +1510,49 @@ async fn run_from_opt(opt: Opt) -> Result<()> {
                 Ok(())
             }
             ContainerOpts::ComputeComposefsDigest {
+                path,
+                write_dumpfile_to,
+            } => {
```
cgwalters, line 1514: Fine as is but I'd say let's move this into the bootc_composefs module?

crates/lib/src/cli.rs
```diff
@@ -358,9 +358,20 @@ pub(crate) enum ContainerOpts {
         #[clap(long)]
         no_truncate: bool,
     },
-    /// Output the bootable composefs digest.
+    /// Output the bootable composefs digest for a directory.
     #[clap(hide = true)]
     ComputeComposefsDigest {
```
cgwalters, line 362: I believe we can trivially unit test i.e. `#[test]` this because we don't depend on anything related to containers-storage here unlike the other path (we can't assert on the exact value of the digest as we don't control the invoking uid).

So unit test could cover the internals.

But then to gain even more coverage we could have a variant in tests-integration `Opt::Container` say that makes a temp rootfs there and because we know our uid we can have an active assertion on the digest even.

crates/lib/src/cli.rs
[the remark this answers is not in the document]
cgwalters: Yes...though that cleanup looks pretty trivial to do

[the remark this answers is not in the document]
jeckersb: Latest version splits this out.

crates/tests-integration/src/container.rs
```diff
@@ -125,6 +125,61 @@ fn test_variant_base_crosscheck() -> Result<()> {
...
+    let hello_path = usr_bin.join("hello");
+    fs::write(&hello_path, "test\n")?;
+    fs::set_permissions(&hello_path, fs::Permissions::from_mode(0o755))?;
+
+    // Create etc/config (regular file)
+    let etc = root.join("etc");
+    fs::create_dir_all(&etc)?;
+    let config_path = etc.join("config");
+    fs::write(&config_path, "test\n")?;
+    fs::set_permissions(&config_path, fs::Permissions::from_mode(0o644))?;
```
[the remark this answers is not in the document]
cgwalters: Per discussion I'd actually like to adapt the fixture.rs code from the ostree side - another option though is composefs dumpfile format.

crates/tests-integration/src/container.rs
```diff
@@ -125,6 +125,61 @@ fn test_variant_base_crosscheck() -> Result<()> {
     Ok(())
 }
 
+/// Test that compute-composefs-digest works on a directory
+pub(crate) fn test_compute_composefs_digest() -> Result<()> {
+    use std::os::unix::fs::PermissionsExt;
+
+    // Create temp directory with test filesystem structure
+    let td = tempfile::tempdir()?;
```
cgwalters, line 133: This is fine as is but I think cap-std is a great choice for the test use cases. It's *really* easy to accidentally forget to prepend the rootdir and then suddenly your unit tests are writing to the current working directory etc.

cgwalters:
We have this as `hide = true` for now, but it's probably worth adding to the composefs.md experimental docs at least.

(edit I mean as a followup)

Issue #1225, opened 2025-03-24
copy-to-storage: Fails for non-ostree base image


cgwalters:
The `bootc image copy-to-storage` code currently requires the `final-diffid` from an ostree-container base.

Typed this up, but didn't test yet

```
diff --git a/ostree-ext/src/container/store.rs b/ostree-ext/src/container/store.rs
index 74500f36..bb6fee11 100644
--- a/ostree-ext/src/container/store.rs
+++ b/ostree-ext/src/container/store.rs
@@ -1401,13 +1401,7 @@ pub(crate) fn export_to_oci(
     let srcinfo = query_image(repo, imgref)?.ok_or_else(|| anyhow!("No such image"))?;
     let (commit_layer, component_layers, remaining_layers) =
         parse_manifest_layout(&srcinfo.manifest, &srcinfo.configuration)?;
-    let commit_layer = commit_layer.ok_or_else(|| anyhow!("Missing {DIFFID_LABEL}"))?;
-    let commit_chunk_ref = ref_for_layer(commit_layer)?;
-    let commit_chunk_rev = repo.require_rev(&commit_chunk_ref)?;
-    let mut chunking = chunking::Chunking::new(repo, &commit_chunk_rev)?;
-    for layer in component_layers {
-        chunking_from_layer_committed(repo, layer, &mut chunking)?;
-    }
+
     // Unfortunately today we can't guarantee we reserialize the same tar stream
     // or compression, so we'll need to generate a new copy of the manifest and config
     // with the layers reset.
@@ -1417,8 +1411,6 @@ pub(crate) fn export_to_oci(
     new_config.history_mut().clear();
     new_config.rootfs_mut().diff_ids_mut().clear();
 
-    let mut dest_oci = ocidir::OciDir::ensure(dest_oci)?;
-
     let opts = ExportOpts {
         skip_compression: opts.skip_compression,
         authfile: opts.authfile,
@@ -1427,19 +1419,36 @@ pub(crate) fn export_to_oci(
 
     let mut labels = HashMap::new();
 
-    // Given the object chunking information we recomputed from what
-    // we found on disk, re-serialize to layers (tarballs).
-    export_chunked(
-        repo,
-        &srcinfo.base_commit,
-        &mut dest_oci,
-        &mut new_manifest,
-        &mut new_config,
-        &mut labels,
-        chunking,
-        &opts,
-        "",
-    )?;
[...]

PR #1864, opened 2025-12-17
build-sys: Many changes

cgwalters:
See individual commits.

This is mainly with an eye to reworking our sealing, for which we needed to do a "scratch" build and as soon as we do that we might as well just take the step of consolidating our integration test and base image etc.

This was developed iteratively, and I did at least do a build test of each individual stage.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -206,17 +206,6 @@ jobs:
           fi
           just clean-local-images
 
-      - name: Run TMT test about bootc install on coreos
-        run: |
-          # Only test fedora-43 on fedora-coreos:testing-devel
-          if [ "${{ matrix.test_os }}" = "fedora-43" ] && [ "${{ matrix.variant }}" = "ostree" ]; then
-            just build-testimage-coreos target/packages
-            just test-tmt-on-coreos plan-bootc-install-on-coreos
-            just clean-local-images
-          else
-            echo "skipped"
-          fi
-
       - name: Archive TMT logs
         if: always()
         uses: actions/upload-artifact@v6
@@ -224,10 +213,52 @@ jobs:
           name: tmt-log-PR-${{ github.event.number }}-${{ matrix.test_os }}-${{ matrix.variant }}-${{ env.ARCH }}
           path: /var/tmp/tmt
 
+  # Test bootc install on Fedora CoreOS (separate job to avoid disk space issues
+  # when run in the same job as test-integration).
+  # Uses fedora-43 as it's the current stable Fedora release matching CoreOS.
+  test-coreos:
+    needs: package
+    runs-on: ubuntu-24.04
+
+    steps:
+      - uses: actions/checkout@v6
+      - name: Bootc Ubuntu Setup
+        uses: ./.github/actions/bootc-ubuntu-setup
+        with:
+          libvirt: true
+      - name: Install tmt
+        run: pip install --user "tmt[provision-virtual]"
+
+      - name: Setup env
+        run: |
+          BASE=$(just pullspec-for-os base fedora-43)
+          echo "BOOTC_base=${BASE}" >> $GITHUB_ENV
+          echo "BOOTC_variant=ostree" >> $GITHUB_ENV
+
+      - name: Download package artifacts
+        uses: actions/download-artifact@v7
+        with:
+          name: packages-fedora-43
+          path: target/packages/
+
+      - name: Build container and test on CoreOS
+        run: |
+          BOOTC_SKIP_PACKAGE=1 just build
+          just build-testimage-coreos target/packages
+          just test-tmt-on-coreos plan-bootc-install-on-coreos
+          just clean-local-images
+
+      - name: Archive TMT logs
+        if: always()
+        uses: actions/upload-artifact@v6
+        with:
+          name: tmt-log-PR-${{ github.event.number }}-fedora-43-coreos-${{ env.ARCH }}
+          path: /var/tmp/tmt
+
   # Sentinel job for required checks - configure this job name in repository settings
   required-checks:
     if: always()
-    needs: [cargo-deny, validate, package, test-integration]
+    needs: [cargo-deny, validate, package, test-integration, test-coreos]
     runs-on: ubuntu-latest
     steps:
       - run: exit 1
@@ -235,4 +266,5 @@ jobs:
           needs.cargo-deny.result != 'success' ||
           needs.validate.result != 'success' ||
           needs.package.result != 'success' ||
-          needs.test-integration.result != 'success'
+          needs.test-integration.result != 'success' ||
+          needs.test-coreos.result != 'success'


jeckersb:
I won't have time to test this but the eyeball test looks sane, especially grateful for breaking it up into individual commits so I can more easily follow your train of thought :clap:

.github/actions/bootc-ubuntu-setup/action.yml
```diff
@@ -65,24 +65,19 @@ runs:
     - name: Install libvirt and virtualization stack
       if: ${{ inputs.libvirt == 'true' }}
       shell: bash
-      env:
-        GH_TOKEN: ${{ github.token }}
       run: |
         set -xeuo pipefail
-        export BCVK_VERSION=0.8.0
-        /bin/time -f '%E %C' sudo apt install -y libkrb5-dev pkg-config libvirt-dev genisoimage qemu-utils qemu-kvm virtiofsd libvirt-daemon-system python3-virt-firmware
```
Johan-Liebert1, line 72: We're hitting

```
Error: 
   0: Failed to create libvirt domain
   1: Failed to setup secure boot
   2: virt-fw-vars tool not found
```
We need `python3-virt-firmware` for this right?

henrywang:
Our bot updated bcvk version in PR https://github.com/bootc-dev/bootc/pull/1867.

cgwalters:
OK, fixed some more things; ended up adding an env var to shortcut things and skip package builds, it was just too fragile without that. The main thing I overlooked is that the unit/integration test depchains ended up pulling in package builds too.

cgwalters:
Hum, not sure yet why most but not *all* of the cfs tests failed.

And now I see this blocks on https://github.com/bootc-dev/bootc/issues/1225

<details>
# PR #1864 CI Failure Investigation - Composefs Boot Failures

## Summary

PR #1864 ("build-sys: Many changes") introduces a "from scratch" image build approach
using `bootc-base-imagectl build-rootfs --manifest=standard`. This causes composefs
sealed UKI boot failures on certain distros.

## Test Results Pattern

| Distro | ostree | composefs-sealeduki-sdboot |
|--------|--------|---------------------------|
| Fedora 42 | PARTIAL (upgrade tests fail with `ostree.final-diffid`) | **PASS** |
| Fedora 43 | PARTIAL | **FAIL** - SSH timeout, VM doesn't boot |
| Fedora 44 | PARTIAL | **FAIL** - SSH timeout, VM doesn't boot |
| CentOS 9 | PARTIAL | (excluded from matrix) |
| CentOS 10 | PARTIAL | **FAIL** - SSH timeout, VM doesn't boot |

**Key observation**: Fedora 42 composefs works, but Fedora 43/44 and CentOS 10 don't.

## Main Branch Status

On main branch (run 20343724695), **ALL composefs and ostree tests PASS**.
This confirms the issue is introduced by PR #1864.

## Key Changes in PR #1864

1. **"From scratch" image build** - Uses `bootc-base-imagectl build-rootfs --manifest=standard`
   instead of directly layering on top of the base image. This is to fix
   https://github.com/containers/composefs-rs/issues/132

2. **Cloud-init always enabled** - Removed the conditional `cloudinit` argument from
   `provision-derived.sh`. Cloud-init is now always installed AND enabled via:
   ```bash
   ln -s ../cloud-init.target /usr/lib/systemd/system/default.target.wants
   ```

3. **Dockerfile consolidation** - Removed separate `hack/Containerfile` for integration
   tests, consolidated into main Dockerfile.

## Composefs Sealed UKI Build Process

The sealed composefs image is built via `Dockerfile.cfsuki`:

1. Compute composefs digest: `hack/compute-composefs-digest $input_image`
2. Build UKI with hardcoded cmdline:
   ```
   cmdline="composefs=${COMPOSEFS_FSVERITY} console=ttyS0,115200n8 console=hvc0 enforcing=0 rw"
   ```
3. Sign UKI with test secureboot keys
4. Place UKI in `/boot/EFI/Linux/`

**Note**: The cmdline is hardcoded and does NOT read from `/usr/lib/bootc/kargs.d/`.

## Failure Details

For failing distros (Fedora 43/44, CentOS 10):
- VM is created successfully by bcvk
- Installation completes ("Installation completed successfully!")
- VM starts but never becomes SSH-accessible
- SSH timeout after 60 attempts (~3 minutes of waiting)
- No kernel panic or boot error visible in logs

## Potential Causes to Investigate

1. **Initramfs differences** - The "from scratch" build may produce different initramfs
   content that doesn't work correctly with composefs sealed boot on newer distros.

2. **Cloud-init interaction** - Cloud-init being enabled might interfere with bcvk's
   SSH key injection, though Fedora 42 works with the same cloud-init setup.

3. **Kernel/dracut version differences** - Fedora 43/44 and CentOS 10 may have different
   kernel or dracut versions that behave differently with the sealed composefs boot.

4. **Missing kargs** - The hardcoded cmdline in `Dockerfile.cfsuki` may be missing
   required kernel arguments for newer distros.

5. **Network configuration** - The "from scratch" build may be missing network
   configuration that the base images normally provide.

## Cloud-init Version Differences

- Fedora 42: `cloud-init-24.2-5.fc42` - services: `cloud-init.service`
- Fedora 43: `cloud-init-25.2-7.fc43` - services: `cloud-init-main.service` (renamed!)
- CentOS 10: `cloud-init-24.4-6.el10` - services: `cloud-init.service`

The service rename in cloud-init 25 (Fedora 43) may cause issues if something
expects the old service name, but CentOS 10 uses cloud-init 24.x and still fails.

## Files to Examine

- `Dockerfile` - New from-scratch build logic
- `Dockerfile.cfsuki` - Sealed composefs UKI build
- `hack/build-sealed` - Script that builds sealed image
- `hack/provision-derived.sh` - Provisioning script (cloud-init changes)
- `hack/compute-composefs-digest` - Computes composefs fsverity digest

## Next Steps

1. Check if main branch composefs tests use the same bcvk/test infrastructure
2. Compare initramfs contents between working (F42) and failing (F43/C10) images
3. Try to get console output from failing VMs to see where boot hangs
4. Consider reverting the cloud-init enabling change to test if that's the cause
5. Test if the "from scratch" build works correctly on failing distros for non-sealed boot
</details>

cgwalters:
OK and then the next problem is the "from scratch" is really painful for composefs installs due to https://github.com/containers/composefs-rs/issues/62

cgwalters:
OK, I think now we're just down to the failures in those selinux-policy tests. Not sure what's up with that.

cgwalters:
OK so this almost worked *finally* - the last thing was I messed up the merge with https://github.com/bootc-dev/bootc/pull/1752 - hopefully fixed now

cgwalters:
Hum...not sure what was going on with that previous failure. OpenCode+Opus 4.5 says

---

     Summary: The job test-integration (fedora-43, ostree) is marked as failure but shows inconsistent state - the step "Run TMT test about bootc install on coreos" is still in_progress with no completion time or conclusion. The job ran for ~1h42m before failing.                                                                         
                                                                                                                                                                                                                                                                                                                                                
     This looks like either:                                                                                                                                                                                                                                                                                                                    
     1. Infrastructure failure - the runner died or lost connection                                                                                                                                                                                                                                                                             
     2. Timeout - the step may have hit an internal timeout                                                                                                                                                                                                                                                                                     
     3. Cancellation - the job was cancelled externally                                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                                                
     The GitHub API is in an inconsistent state (job marked failed but step still in_progress), which often happens when a runner dies unexpectedly. You might want to re-run the job - this appears to be a transient infrastructure issue rather than a test failure.                                                                         
                                                                                                                                                                                                                                                                            
---

I restarted the job...

jeckersb:
Hm reproduced the CI failure, doesn't make any sense to me either :shrug:

cgwalters:
Ah hah! With AI help I noticed there's an "annotations" section at the top of the GHA run (I had never seen this before) and

image

So...it feels cleanest to split this coreos thing off to its own distinct CI context, will look at that

jeckersb:
> Ah hah! With AI help I noticed there's an "annotations" section at the top of the GHA run (I had never seen this before)

:exploding_head: 

Never noticed that either.  Seems like the sort of thing that should be big and red and scary by default and not tucked behind a collapsed dropdown.

PR #1870, opened 2025-12-18
renovate: Ignore bootc-dev/infra synced GitHub workflows

henrywang:
To avoid duplicate github action and workflow update.

The renovate can update action and workflows in `bootc-dev/infra`. When the update PR in `bootc-dev/infra` merged, the push triggers the `syn common` to send PR to update action and workflows in this repo.


henrywang:
Right. Good suggestion. Close this PR and use https://github.com/bootc-dev/infra/pull/62 instead.

PR #1876, opened 2025-12-22
Use setup-rust action in GHA workflows

sideeffffect:
Seems like wasn't used anywhere yet.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,6 +117,8 @@ jobs:
       - uses: actions/checkout@v6
       - name: Bootc Ubuntu Setup
         uses: ./.github/actions/bootc-ubuntu-setup
+      - name: Setup Rust
+        uses: ./.github/actions/setup-rust
       - name: Build mdbook
         run: just build-mdbook
   # Build packages for each test OS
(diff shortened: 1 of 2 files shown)


henrywang:
Hi, thanks for working on this.
We build bootc inside Container with https://github.com/bootc-dev/bootc/blob/main/Dockerfile. So we don't need setup rust in github runner.

sideeffffect:
Oh, I see. So does that mean `./.github/actions/setup-rust` doesn't need to be used anywhere? Or just somewhere?

How does it handle caching or the Rust build?

If nowhere, should it get deleted?

henrywang:
This action is synced from https://github.com/bootc-dev/infra by workflow https://github.com/bootc-dev/infra/blob/main/.github/workflows/sync-common.yml.

This action has been used by the other repos, like https://github.com/bootc-dev/bcvk.

henrywang:
And need a PR to update https://github.com/bootc-dev/infra/blob/d5a5a62c9810a416e4cc98f377c05343393f7c14/renovate-shared-config.json#L33. Not sync `setup-rust` for this repo.

Thanks for catching this.

cgwalters:
Hi, thanks for submitting this!

Yeah, I agree with henrywang - the general direction I would like to go is building code inside a container image - and in this flow it's the container build that should own the Rust toolchain (and other dependency) installation.

This is particularly true for bootc.

Now in this organization we do have some Rust crates, for which it's a bit more "natural" to just directly build them on the GHA runner. But even there I'd personally prefer to cut over those to use the `container:` verb in GHA and pull one of our devenv containers https://github.com/bootc-dev/infra/blob/main/devenv/README.md

---

Either way since we now have multiple crates in the org, I think it argues to have shared workflows sync'd from the bootc-dev/infra repo.

Thanks again though!

PR #1880, opened 2025-12-23
ostree-ext: Make new packing if using prior build fails

HastD:
Under some circumstances—for example, if a new exclusive component has been added since the prior build—packing with a prior build structure can fail. When this happens, we can simply discard the prior build data and make a new packing structure, rather than having chunking fail entirely.

This should fix coreos/rpm-ostree#5545, and I think it'll also fix the test failure that was observed in CI for coreos/rpm-ostree#5535.

### DIFF
--- a/crates/ostree-ext/src/chunking.rs
+++ b/crates/ostree-ext/src/chunking.rs
@@ -605,14 +605,17 @@ fn get_partitions_with_threshold<'a>(
 ///  required packages
 /// else if pkg structure to be changed || prior build not specified
 ///  Recompute optimal packaging structure (Compute partitions, place packages and optimize build)
+///
+/// A return value of `Ok(None)` indicates there was not an error, but the prior packing structure
+/// is incompatible in some way with the current build.
 fn basic_packing_with_prior_build<'a>(
     components: &'a [ObjectSourceMetaSized],
     bin_size: NonZeroU32,
     prior_build: &oci_spec::image::ImageManifest,
-) -> Result<Vec<Vec<&'a ObjectSourceMetaSized>>> {
+) -> Result<Option<Vec<Vec<&'a ObjectSourceMetaSized>>>> {
     let before_processing_pkgs_len = components.len();
 
-    tracing::debug!("Keeping old package structure");
+    tracing::debug!("Attempting to use old package structure");
 
     // The first layer is the ostree commit, which will always be different for different builds,
     // so we ignore it.  For the remaining layers, extract the components/packages in each one.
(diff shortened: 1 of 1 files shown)


henrywang:
Failure has been fixed by https://github.com/bootc-dev/bootc/pull/1882. Thanks!

HastD:
@henrywang Sorry, I'm not sure I understand—what does that PR have to do with this one?

henrywang:
> @henrywang Sorry, I'm not sure I understand—what does that PR have to do with this one?

There's a failure on `fedora 42`. It's fixed in https://github.com/bootc-dev/bootc/pull/1882. Could you please make a rebase to make `fedora 42` test green? Thanks.

henrywang:
Could you please sign off commit which is required by `DCO` test. Thank you and Happy New Year!

crates/ostree-ext/src/chunking.rs
```diff
@@ -634,6 +634,12 @@ fn basic_packing_with_prior_build<'a>(
         .collect();
     let mut curr_build = curr_build?;
 
+    if let Ok(bin_size) = usize::try_from(bin_size.get()) {
```
cgwalters: How about just:

```
if bin_size.get() as usize < curr_build.len() {
    bail!("{bin_size} is too small to be compatible with the prior build");
}
```

Note that since we don't care about 16 bit microcontrollers, usize is at least u32, so we can safely just `as usize`.

crates/ostree-ext/src/chunking.rs
```diff
@@ -714,7 +720,18 @@ fn basic_packing<'a>(
 
     // If we have a prior build, then use that
     if let Some(prior_build) = prior_build_metadata {
-        return basic_packing_with_prior_build(components, bin_size, prior_build);
+        match basic_packing_with_prior_build(components, bin_size, prior_build) {
```
cgwalters: A problem with this is it "swallows" *all* errors (including e.g. I/O errors, failure to parse JSON etc).

What I think would be a cleaner fix here is to change `basic_packing_with_prior_build` to return `Result<Option<T>>` instead, and then instead of `bail!` for incompatible setups we `return Ok(None);`

cgwalters (changes requested):
Thanks for working on this!

crates/ostree-ext/src/chunking.rs
HastD: But what if I want to created a chunked OCI image with tens of thousands of layers on a 16-bit microcontroller... okay, point taken. :sweat_smile:

crates/ostree-ext/src/chunking.rs
HastD: Good idea, done.

cgwalters (approved):
Thanks!

HastD:
Sorry, realized right after I pushed that those comments should've used `tracing::debug` instead--I'm done making changes now.

PR #1881, opened 2025-12-24
cli: Add shell completion generation command

tnk4on:
## Summary

Add a hidden `bootc completion <shell>` subcommand that generates shell completion scripts for bash, zsh, and fish.

## Motivation

This enables distributions (e.g., Fedora, CentOS Stream, RHEL) to generate and package shell completions during RPM/deb builds by calling:

```bash
bootc completion bash > /usr/share/bash-completion/completions/bootc
bootc completion zsh > /usr/share/zsh/site-functions/_bootc
bootc completion fish > /usr/share/fish/vendor_completions.d/bootc.fish
```

## Implementation Details

- Added `CompletionShell` enum with Bash, Zsh, and Fish variants
- The completion scripts only include visible (non-hidden) subcommands intended for end users
- Completed subcommands: edit, image, install, status, switch, update, upgrade, usroverlay
- Uses `clap::CommandFactory` to introspect the CLI structure
- Added `clap_complete` as a dev-dependency

## Testing

- Tested on Fedora container with cargo build
- Tested RPM build on RHEL10-compatible environment
- Verified completion works correctly in bash, zsh, and fish shells

### DIFF
--- a/crates/lib/Cargo.toml
+++ b/crates/lib/Cargo.toml
@@ -33,6 +33,7 @@ cap-std-ext = { workspace = true, features = ["fs_utf8"] }
 cfg-if = { workspace = true }
 chrono = { workspace = true, features = ["serde"] }
 clap = { workspace = true, features = ["derive","cargo"] }
+clap_complete = "4"
 clap_mangen = { workspace = true, optional = true }
 composefs = { workspace = true }
 composefs-boot = { workspace = true }
--- a/crates/lib/src/cli.rs
+++ b/crates/lib/src/cli.rs
@@ -12,6 +12,7 @@ use anyhow::{anyhow, ensure, Context, Result};
 use camino::{Utf8Path, Utf8PathBuf};
 use cap_std_ext::cap_std;
 use cap_std_ext::cap_std::fs::Dir;
+use clap::CommandFactory;
 use clap::Parser;
 use clap::ValueEnum;
 use composefs::dumpfile;
@@ -733,6 +734,15 @@ pub(crate) enum Opt {
     /// Diff current /etc configuration versus default
     #[clap(hide = true)]
     ConfigDiff,
+    /// Generate shell completion script for supported shells.
+    ///
+    /// Example: `bootc completion bash` prints a bash completion script to stdout.
+    #[clap(hide = true)]
+    Completion {
+        /// Shell type to generate (bash, zsh, fish)
+        #[clap(value_enum)]
+        shell: clap_complete::aot::Shell,
+    },
     #[clap(hide = true)]
     DeleteDeployment {
         depl_id: String,
@@ -1573,6 +1583,15 @@ async fn run_from_opt(opt: Opt) -> Result<()> {
                 Ok(())
             }
         },
+        Opt::Completion { shell } => {
+            use clap_complete::aot::generate;
+
+            let mut cmd = Opt::command();
+            let mut stdout = std::io::stdout();
+            let bin_name = "bootc";
+            generate(shell, &mut cmd, bin_name, &mut stdout);
+            Ok(())
+        }
         Opt::Image(opts) => match opts {
             ImageOpts::List {
                 list_type,
@@ -1977,5 +1996,30 @@ mod tests {
             "pull",
         ]));
         assert_eq!(args.as_slice(), ["container", "image", "pull"]);
+    }
+
+    #[test]
+    fn test_generate_completion_scripts_contain_commands() {
+        use clap_complete::aot::{generate, Shell};
+
+        // For each supported shell, generate the completion script and
+        // ensure obvious subcommands appear in the output. This mirrors
+        // the style of completion checks used in other projects (e.g.
+        // podman) where the generated script is examined for expected
+        // tokens.
+
+        // `completion` is intentionally hidden from --help / suggestions;
+        // ensure other visible subcommands are present instead.
+        let want = ["install", "upgrade"];
+
+        for shell in [Shell::Bash, Shell::Zsh, Shell::Fish] {
+            let mut cmd = Opt::command();
+            let mut buf = Vec::new();
+            generate(shell, &mut cmd, "bootc", &mut buf);
+            let s = String::from_utf8(buf).expect("completion should be utf8");
+            for w in &want {
+                assert!(s.contains(w), "{shell:?} completion missing {w}");
+            }
+        }
     }
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -239,6 +239,7 @@ dependencies = [
  "cfg-if",
  "chrono",
  "clap",
+ "clap_complete",
  "clap_mangen",
  "comfy-table",
  "composefs",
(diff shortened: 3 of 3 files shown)


henrywang:
Could you please sign off commit which is required by `DCO` test and make a rebase to fix `fedora-42` failure? Thank you and Happy New Year.

crates/lib/src/cli.rs
```diff
@@ -1986,4 +2018,55 @@ mod tests {
...
+        // ensure obvious subcommands appear in the output. This mirrors
+        // the style of completion checks used in other projects (e.g.
+        // podman) where the generated script is examined for expected
+        // tokens.
+
+        // `completion` is intentionally hidden from --help / suggestions;
+        // ensure other visible subcommands are present instead.
+        let want = ["install", "upgrade"];
+
+        // Bash
```
cgwalters: This would be 1/3 the size as a table-driven test that loops over the 3 different shells.

crates/lib/src/cli.rs
```diff
@@ -1581,6 +1600,19 @@ async fn run_from_opt(opt: Opt) -> Result<()> {
...
             }
         },
+        Opt::Completion { shell } => {
+            use clap_complete::{generate, shells};
+
+            let mut cmd = Opt::command();
+            let mut stdout = std::io::stdout();
+            let bin_name = "bootc";
+            match shell {
+                CompletionShell::Bash => generate(shells::Bash, &mut cmd, bin_name, &mut stdout),
```
cgwalters: According to the docs `shells` is deprecated in favor of https://docs.rs/clap_complete/latest/clap_complete/aot/index.html it looks like. Any reason you didn't use that?

Did you write this from scratch or did you base it on an existing example in another Rust project?

crates/lib/src/cli.rs
```diff
@@ -1581,6 +1600,19 @@ async fn run_from_opt(opt: Opt) -> Result<()> {
                 Ok(())
             }
         },
+        Opt::Completion { shell } => {
```
cgwalters, line 1586: It looks like a totally different option is to generate this data at compile time, per https://docs.rs/clap_complete/latest/clap_complete/aot/fn.generate_to.html

I don't have a strong sense of the tradeoffs - if this doesn't notably increase the binary size, then doing it dynamically seems completely fine.

cgwalters:
Also, thanks so much for working on this!!

crates/lib/src/cli.rs
tnk4on: Refactored to be table-driven, reducing code size by ~1/3.

crates/lib/src/cli.rs
tnk4on: Thanks for catching that! I wasn't aware that `clap_complete::shells` was deprecated. I've updated the code to use `clap_complete::aot` instead. The code was written with Cursor in a pair programming session, and I didn't reference any specific project for this implementation.

crates/lib/src/cli.rs
tnk4on: I've kept the runtime generation approach primarily for two reasons:

1. Flexibility: Users can generate completions on-demand via `bootc completion <shell>`, which is convenient for both end users and developers who want to quickly set up completions without needing to locate pre-generated files.

2. Distribution-friendly: This approach allows distributions (Fedora, RHEL, etc.) to generate completion scripts during RPM builds by simply running `bootc completion <shell>` commands. This gives distributions full control over when and how completions are generated, without requiring changes to the build script or build dependencies.

If you think compile-time generation would be better for this use case, I can refactor it. What are your thoughts?

tnk4on:
I've made a few updates before pushing:

1. Local validation: Ran validation tests locally and fixed any issues before pushing to ensure the code is ready for review.

2. Focused changes: Removed unrelated file changes that were accidentally included. The PR now focuses only on the shell completion feature.

crates/lib/src/cli.rs
cgwalters: > he code was written with Cursor in a pair programming session, and I didn't reference any specific project for this implementation.

Using AI is fine, but did Cursor not read our [AGENTS.md](https://github.com/bootc-dev/bootc/blob/main/AGENTS.md)? You should do so as well as there are AI usage disclosure requirements there. Please update your commit to match that.

crates/lib/src/cli.rs
```diff
@@ -744,6 +754,15 @@ pub(crate) enum Opt {
...
     #[clap(hide = true)]
     ConfigDiff,
+    /// Generate shell completion script for supported shells.
+    ///
+    /// Example: `bootc completion bash` prints a bash completion script to stdout.
+    #[clap(hide = true)]
+    Completion {
+        /// Shell type to generate (bash, zsh, fish)
+        #[clap(value_enum)]
+        shell: CompletionShell,
```
cgwalters: Can't we just directly use https://docs.rs/clap_complete/latest/clap_complete/aot/enum.Shell.html and avoid a custom enum?

crates/lib/src/cli.rs
tnk4on: Updated the commit message to include `Assisted-by: Cursor (Auto)` per AGENTS.md requirements.Updated the commit message to include `Assisted-by: Cursor (Auto)` per AGENTS.md requirements.

crates/lib/src/cli.rs
tnk4on: I've updated the code to use `clap_complete::aot::Shell` directly instead of a custom enum.

crates/tests-integration/src/container.rs
```diff
@@ -201,10 +201,10 @@ pub(crate) fn test_compute_composefs_digest() -> Result<()> {
 
     // Verify it's a valid hex string of expected length (SHA-512 = 128 hex chars)
     assert_eq!(
-        digest.len(),
+        digest.as_bytes().len(),
```
cgwalters: Looks unrelated - why?

Ohh wait is this related to our custom clippy lint?

crates/lib/Cargo.toml
cgwalters: This change looks unnecessary.

crates/lib/src/status.rs
```diff
@@ -832,7 +832,11 @@ fn container_inspect_print_human(
     rows.push(("Kargs", kargs));
 
     // Find the max label width for right-alignment
-    let max_label_len = rows.iter().map(|(label, _)| label.len()).max().unwrap_or(0);
+    let max_label_len = rows
+        .iter()
+        .map(|(label, _)| label.as_bytes().len())
+        .max()
+        .unwrap_or(0);
```
cgwalters: See below

cgwalters (changes requested):

cgwalters:
> This enables distributions (e.g., Fedora, CentOS Stream, RHEL) to generate and package shell completions during RPM/deb builds by calling:

Also we do have a spec file here...so it'd likely be useful to actually do that change in it.

It looks like at least podman ships bash and fish completions by default. Also they ship a man page `man podman-completion`, which might make sense here.

---

Also I tested this out, it seems to work well but one thing I noticed I don't like is that it shows our `hide=true` subcommands. Digging in there's a ton of related issues/PRs in clap, of which https://github.com/clap-rs/clap/pull/5283 is most relevant.

crates/tests-integration/src/container.rs
tnk4on: I opened another PR https://github.com/bootc-dev/bootc/pull/1900

PR #1882, opened 2025-12-24
test/tmt: Use `bootc usroverlay` for /usr overlay mount

Johan-Liebert1:
Previously we were mounting a rw overlay on top of /usr using `mount -t overlay -olowerdir=/usr,workdir=...,upperdir=... overlay /usr` which caused the kernel to throw
`overlayfs: maximum fs stacking depth exceeded`
possibly because the mountpoint was the same as the lowerdir

Also, move the overlay mount BEFORE we mask off `/sysroot/ostree` else bootc throws `error: Read only filesystem`


cgwalters:
> overlayfs: maximum fs stacking depth exceeded

This is https://github.com/coreos/fedora-coreos-tracker/issues/2087

PR #1888, opened 2026-01-05
composefs/status: resolve rollback entry correctly

### LINKED ISSUE #1887: composefs backend: `bootc status` fails with `Setting soft reboot capability for Type1 entries: Entry not found` when order of deployement list is messed up
I am getting this from my installation:
```
[hkj@kagumine Linux]$ sudo bootc status
error: Status: Getting composefs deployment status: Getting composefs deployment status: Checking soft reboot capability: Setting soft reboot capability for Type1 entries: Entry not found
```

### To reproduce:

1. `upgrade` or `switch` to a new image
2. after reboot, enter rollback boot option
3. in rollback boot, `upgrade` or `switch` to a third image
4. reboot and the system will be in this state

### Context:

So I use a customized image in my local self-hosted registry with baked in configurations with this installation. What happened is that I misconfigured one revision and completely broke my network software. I made a revision that fixes the error, but without an internet connection I cannot pull the upgrade to the host. So I rebooted into the rollback deployment, and applied upgrade from there. However after rebooting into the now working new revision, bootc is completely broken and just gives me the above error message for virtually any sub-command I want to run.

### Some analysis:

I think the problem is [here](https://github.com/bootc-dev/bootc/blob/5ec2c5ce690a01c35cd6cfb43cc03e176ab8e96a/crates/lib/src/bootc_composefs/status.rs#L615), when status is trying to determine which deployment is rollback, it just tries to go through all deployments and match against booted and staged - then the leftover is rollback. But if you have multiple stale old deployments present, this just gives the last enumerated deployment that is not booted nor staged. On my system this apparently gets resolved to the broken image deployment whose BLS entry already got removed by the upgrade run in  the rollback boot.

Funny enough, this still works under most installations if you just upgrade normally without rollbacks, since `opendir` (underlying syscall for `Dir:read_dir`) just reads the underlying directory's on fs index to get the entries and return entries in order as stored in the index. And most fs *happens* to implement their index as a simple linear data structure that puts new entries at the end of the enumeration. So everything just works if the order is followed, until one rollback, the last order got reversed and everything goes haywire.

We probably need to fix that rollback determination code by ensuring only deployment that has a corresponding boot entry will be recognized as rollback

chaserhkj:
Previous implementation had undefined behavior and was coincidentally correct under conditions where no rollback was performed, see #1887

Matches deployment entries in composefs deploy folder that are neither staged nor booted against entires defined in /boot to find out rollback entry.

Fixes #1887

### DIFF
--- a/crates/lib/src/bootc_composefs/status.rs
+++ b/crates/lib/src/bootc_composefs/status.rs
@@ -712,21 +712,22 @@ pub(crate) async fn composefs_deployment_status_from(
                 .map(|menu| menu.get_verity()),
         )
         .collect::<Result<HashSet<_>>>()?;
-    for entry in extra_deployment_boot_entries {
-        // SAFETY: boot_entry.composefs will always be present
-        let verity = &entry.composefs.as_ref().unwrap().verity;
-        if bootloader_configured_verity.contains(verity) {
-            match host.status.rollback {
-                Some(ref _entry) => {
-                    anyhow::bail!(
-                        "Multiple extra entires in /boot, could not determine rollback entry"
-                    );
-                }
-                None => {
-                    host.status.rollback = Some(entry);
-                }
-            }
-        }
+    let rollback_candidates: Vec<_> = extra_deployment_boot_entries
+        .into_iter()
+        .filter(|entry| {
+            let verity = &entry
+                .composefs
+                .as_ref()
+                .expect("composefs is always Some for composefs deployments")
+                .verity;
+            bootloader_configured_verity.contains(verity)
+        })
+        .collect();
+
+    if rollback_candidates.len() > 1 {
+        anyhow::bail!("Multiple extra entries in /boot, could not determine rollback entry");
+    } else if let Some(rollback_entry) = rollback_candidates.into_iter().next() {
+        host.status.rollback = Some(rollback_entry);
     }
 
     host.status.rollback_queued = is_rollback_queued;


chaserhkj:
@cgwalters @Johan-Liebert1 Can anyone take a look at this change?  I feel this change makes sense even without reproduction of #1887 since current implemented behavior relies on undefined behaviors from the underlying library/syscall ( syscall `opendir` from `Dir::read_dir` ) and that could be counted as a bug on itself.

Johan-Liebert1:
> since current implemented behavior relies on undefined behaviors from the underlying library/syscall

Could you please highlight the UB that might occur?

I might be getting this wrong since I'm unable to reproduce #1887, but wouldn't `bootc status` now fail at 
https://github.com/bootc-dev/bootc/pull/1888/changes#diff-0c2404bde8d45d9665ce40004291992c9503391924082c74ef6ef515e1eedf50R728?

chaserhkj:
UB comes from these code that reads the entries from directory `/sysroot/state/deploy`:
https://github.com/bootc-dev/bootc/blob/ad60763ff290b0fbf9d605b9d3d5f022a7b239ff/crates/lib/src/bootc_composefs/status.rs#L545-L549
The `read_dir` function [doc](https://doc.rust-lang.org/std/fs/fn.read_dir.html) describes the order of the returned entry list to be UB.
Later in the same function, this list with undefined order is enumerated through:
https://github.com/bootc-dev/bootc/blob/ad60763ff290b0fbf9d605b9d3d5f022a7b239ff/crates/lib/src/bootc_composefs/status.rs#L573-L621
Within this loop, `host.status.booted` and `host.status.staged` are both set and early continues the loop, and the last statement catches all fall-through cases and sets `host.status.rollback`.

But if there are more than two entires in `/sysroot/state/deploy` that are neither booted nor staged, the loop still goes through all of them, and `rollback` will eventually settle on the later entry in the list, so what gets resolved to be the rollback image is UB here.

If the UB here picked up an entry that does have a corresponding `/boot` BLS entry, this won't create a problem as that entry could then be presumed to be the rollback entry and pass all subsequent checks.
If the UB here picked up a stale entry in `/sysroot/state/deploy` that does not have a corresponding `/boot` BLS entry, later here in function `set_reboot_capable_type1_deployments`:
https://github.com/bootc-dev/bootc/blob/ad60763ff290b0fbf9d605b9d3d5f022a7b239ff/crates/lib/src/bootc_composefs/status.rs#L444-L452
`find_bls_entry` would fail on `host.status.rollback` due to the missing BLS entry, causing #1887

As for [here](https://github.com/bootc-dev/bootc/pull/1888/changes#diff-0c2404bde8d45d9665ce40004291992c9503391924082c74ef6ef515e1eedf50R728) in my change, that section is trying to find boot entries that are neither the booted entry nor the staged entry, and are **in both `/sysroot/state/deploy` and `/boot`**. The state that there are two entries satisfying this criteria is currently impossible to be in, since bootc under normal state only keeps one booted entry and one rollback entry in /boot, and under staged state only keeps one booted entry, one old rollback entry and one staged entry in /boot

PR #1890, opened 2026-01-06
Implement `bootc image copy-to-storage` for composefs backend

Johan-Liebert1:
Export a composefs repository as an OCI image. This would be helpful for running tests that we have for ostree backend. Tests for sealed UKI would be separate, I think, but tests for Type1 entries can be shared between the two backends

### DIFF
--- a/crates/lib/src/bootc_composefs/export.rs
+++ b/crates/lib/src/bootc_composefs/export.rs
@@ -26,7 +26,7 @@ pub async fn export_repo_to_image(
 ) -> Result<()> {
     let host = get_composefs_status(storage, booted_cfs).await?;
 
-    let (source, dest_imgref) = get_imgrefs_for_copy(source, target).await?;
+    let (source, dest_imgref) = get_imgrefs_for_copy(&host, source, target).await?;
 
     let mut depl_verity = None;
 
--- a/crates/lib/src/cli.rs
+++ b/crates/lib/src/cli.rs
@@ -47,6 +47,7 @@ use crate::podstorage::set_additional_image_store;
 use crate::progress_jsonl::{ProgressWriter, RawProgressFd};
 use crate::spec::Host;
 use crate::spec::ImageReference;
+use crate::status::get_host;
 use crate::store::{BootedOstree, Storage};
 use crate::store::{BootedStorage, BootedStorageKind};
 use crate::utils::sigpolicy_from_opt;
@@ -1588,12 +1589,16 @@ async fn run_from_opt(opt: Opt) -> Result<()> {
             } => crate::image::list_entrypoint(list_type, list_format).await,
 
             ImageOpts::CopyToStorage { source, target } => {
+                // We get "host" here to avoid deadlock in the ostree path
+                let host = get_host().await?;
+
                 let storage = get_storage().await?;
 
                 match storage.kind()? {
                     BootedStorageKind::Ostree(..) => {
                         crate::image::push_entrypoint(
                             &storage,
+                            &host,
                             source.as_deref(),
                             target.as_deref(),
                         )
--- a/crates/lib/src/image.rs
+++ b/crates/lib/src/image.rs
@@ -15,7 +15,7 @@ use crate::{
     boundimage::query_bound_images,
     cli::{ImageListFormat, ImageListType},
     podstorage::CStorage,
-    status::get_host,
+    spec::Host,
     store::Storage,
     utils::async_task_with_spinner,
 };
@@ -145,6 +145,7 @@ pub(crate) async fn list_entrypoint(
 /// If the source isn't specified, we use booted image
 /// If the target isn't specified, we push to containers-storage with our default image
 pub(crate) async fn get_imgrefs_for_copy(
+    host: &Host,
     source: Option<&str>,
     target: Option<&str>,
 ) -> Result<(ImageReference, ImageReference)> {
@@ -169,18 +170,17 @@ pub(crate) async fn get_imgrefs_for_copy(
             .context("Parsing source image")?,
 
         None => {
-            let host = get_host().await?;
-
             let booted = host
                 .status
                 .booted
+                .as_ref()
                 .ok_or_else(|| anyhow::anyhow!("Booted deployment not found"))?;
 
-            let booted_image = booted.image.unwrap().image;
+            let booted_image = &booted.image.as_ref().unwrap().image;
 
             ImageReference {
                 transport: Transport::try_from(booted_image.transport.as_str()).unwrap(),
-                name: booted_image.image,
+                name: booted_image.image.clone(),
             }
         }
     };
@@ -192,10 +192,11 @@ pub(crate) async fn get_imgrefs_for_copy(
 #[context("Pushing image")]
 pub(crate) async fn push_entrypoint(
     storage: &Storage,
+    host: &Host,
     source: Option<&str>,
     target: Option<&str>,
 ) -> Result<()> {
-    let (source, target) = get_imgrefs_for_copy(source, target).await?;
+    let (source, target) = get_imgrefs_for_copy(host, source, target).await?;
 
     let ostree = storage.get_ostree()?;
     let repo = &ostree.repo();


crates/lib/src/bootc_composefs/export.rs
```diff
@@ -0,0 +1,256 @@
...
+                TarItem::Leaf(leaf_content) => {
+                    match &leaf_content {
+                        LeafContent::Regular(reg) => match reg {
+                            RegularFile::Inline(items) => {
+                                v[..hsize].copy_from_slice(items);
+                            }
+
+                            RegularFile::External(obj_id, ..) => {
+                                let mut file = File::from(booted_cfs.repo.open_object(obj_id)?);
+                                file.read_exact(&mut v)?;
```
cgwalters: I don't think we need to buffer the whole file into memory, we can instead just pass the `File` instance here to the `append` operation below.

Although to handle the inline case which is a different type, we'd need to raise the `append` into each branch instead (or have a `Box<dyn Read>` but eh, easier to just move append here)

crates/lib/src/bootc_composefs/export.rs
```diff
@@ -0,0 +1,256 @@
...
+            transport: Transport::ContainerStorage,
+            name: target.to_owned(),
+        },
+        None => ostree_ext::container::ImageReference {
+            transport: Transport::ContainerStorage,
+            name: IMAGE_DEFAULT.into(),
+        },
+    };
+
+    // If the source isn't specified, we use the booted image
```
cgwalters: Not a blocker but I think we could pretty easily share this code with the ostree case. It's basically just the inner mechanics of writing to the OCI dir that's different, so the shared helper could allocate that too.

crates/lib/src/bootc_composefs/export.rs
```diff
@@ -0,0 +1,256 @@
...
+                            }
+                        },
+
+                        // we don't need to write the data for symlinks.
+                        // Same goes for devices, fifos and sockets
+                        _ => {}
+                    }
+                }
+
+                // we don't need to write the data for hardlinks/dirs
```
cgwalters: I found this comment confusing, thinking we were skipping these items. Related to the above (and matching other uses of the tar crate in this project but also more generally) i think we should instead call `layer_writer.append(&header, std::io::empty())` here instead.

crates/lib/src/bootc_composefs/export.rs
```diff
@@ -0,0 +1,256 @@
...
+    },
+    image::IMAGE_DEFAULT,
+    store::{BootedComposefs, Storage},
+};
+
+fn get_entry_with_header<R: Read, ObjectID: FsVerityHashValue>(
+    reader: &mut SplitStreamReader<R, ObjectID>,
+) -> anyhow::Result<Option<(Header, TarItem<ObjectID>)>> {
+    let mut buf = [0u8; 512];
+    if !reader.read_inline_exact(&mut buf)? || buf == [0u8; 512] {
```
cgwalters: It should only be corrupted tar streams for which `read_inline_exact` would return `false` right? I wonder if that should instead be an error?

I would also find it helpful to separate these conditionals like

```
anyhow::ensure!(reader.read_inline_exact(&mut buf)?);
// All-zero header signals end of archive
if buf.iter().all(|b| b == 0) {
  ....
```

Except it's actually *two* all-zero headers that signal EOF canonically
https://www.gnu.org/software/tar/manual/html_node/Standard.html

> Physically, an archive consists of a series of file entries terminated by an end-of-archive entry, which consists of two 512 blocks of zero bytes.

Of course I think it's very likely that container tar parsers are liberal in what they accept, and maybe we're even relying on that...but it seems like perhaps worth a warning?

crates/lib/src/bootc_composefs/export.rs
```diff
@@ -0,0 +1,256 @@
...
+        return Ok(None);
+    }
+
+    let header = tar::Header::from_byte_slice(&buf);
+
+    let size = header.entry_size()?;
+
+    let item = match reader.read_exact(size as usize, ((size + 511) & !511) as usize)? {
```
cgwalters: This seems to be relying on `size = 0` for anything except regular files. I think we should probably have at least an `anyhow::ensure!` for that.

crates/lib/src/bootc_composefs/export.rs
cgwalters: Also isn't there a bigger problem in that iterating by raw headers in this way means we aren't handling e.g. GNU long links and PAX extended headers?

crates/lib/src/bootc_composefs/export.rs
```diff
@@ -0,0 +1,256 @@
...
+            EntryType::Regular | EntryType::Continuous => {
+                TarItem::Leaf(LeafContent::Regular(RegularFile::External(id, size)))
+            }
+            _ => anyhow::bail!("Unsupported external-chunked entry {header:?} {id:?}"),
+        },
+
+        SplitStreamData::Inline(content) => match header.entry_type() {
+            EntryType::Directory => TarItem::Directory,
+            // We do not care what the content is as we're re-archiving it anyway
+            _ => TarItem::Leaf(LeafContent::Regular(RegularFile::Inline(content))),
```
cgwalters: Humm...so I guess this is actually the thing that means we just pass through those GNU long links and PAX headers? And for that matter, symlinks.

It seems cleaner to change the logic here so we check *just* if the tar entry is a regular file, and handle that specially - otherwise we just return the raw bytes to the caller.

Or honestly are we gaining much by having this separate helper function? It'd simplify things even more I think to just inline this and directly have it write to the output tar writer.

Note that we have *two* on-stack 512b buffers when we only need one really.

crates/lib/src/bootc_composefs/export.rs
```diff
@@ -0,0 +1,256 @@
...
+        return Ok(None);
+    }
+
+    let header = tar::Header::from_byte_slice(&buf);
+
+    let size = header.entry_size()?;
+
+    let item = match reader.read_exact(size as usize, ((size + 511) & !511) as usize)? {
+        SplitStreamData::External(id) => match header.entry_type() {
+            EntryType::Regular | EntryType::Continuous => {
```
cgwalters: I had never before seen the `EntryType::Continuous`, I wonder if it actually is used in the wild

crates/lib/src/bootc_composefs/export.rs
Johan-Liebert1: yeah, this is a better way to do things

crates/lib/src/bootc_composefs/export.rs
Johan-Liebert1: updated. Since we're moving the writer inside the match itself, I moved writes for everything else inside the match

crates/lib/src/bootc_composefs/export.rs
Johan-Liebert1: That's definitely an improvement. Updated

crates/lib/src/bootc_composefs/export.rs
Johan-Liebert1: I did run into the truncation problem when using the parsing code in `composefs-rs` (hence the copying of part of the code here). The PAX and GNU extended headers are just passed through

Johan-Liebert1: > It should only be corrupted tar streams for which read_inline_exact would return false right?

It'll throw an error if the tar is malformed, afaiu

And yeah, we should be checking two consecutive zero blocks. Updated

crates/lib/src/bootc_composefs/export.rs
Johan-Liebert1: Updated

crates/lib/src/bootc_composefs/export.rs
Johan-Liebert1: Apparently it's something to do with continuous disk allocation for a file. Not sure if it's used very often

crates/lib/src/bootc_composefs/export.rs
cgwalters: Which parsing code? You mean https://github.com/containers/composefs-rs/blob/1498349e11d0e4bd545feab533f0b7ab5294ae9a/crates/composefs-oci/src/tar.rs#L202 ?

crates/lib/src/bootc_composefs/repo.rs
```diff
@@ -61,7 +61,7 @@ pub(crate) fn get_imgref(transport: &str, image: &str) -> String {
     let img = image.strip_prefix(":").unwrap_or(&image);
     let transport = transport.strip_suffix(":").unwrap_or(&transport);
 
-    if transport == "registry" {
+    if transport == "registry" || transport == "docker://" {
```
cgwalters, line 64: Unrelated to this change but it'd be cleaner to parse this back into the `Transport` enum which already handles this canonicalization

Johan-Liebert1:
The tests seem to be timing out, but I can't find the actual error. Weird. I'll try it locally once

crates/lib/src/bootc_composefs/export.rs
Johan-Liebert1: Yes, that's the one

cgwalters: Ah, I think the logic we have here should replace that in `get_entry` ideally...there's no reason AFAIK for us to synthesize a *new* tarstream.

Actually though in this case there's the `cat` function which I think already does what we want in just giving us back the raw byte stream.

So: I think all of this just be replaced by `reader.cat(layer_writer, || repository.load_object())`

Though note the `cat` implementation has the same suboptimal "read entire file into memory" thing but that's fixable.

Johan-Liebert1: I think the issue lies with GNU long names and PAX headers which need to be parsed and then reconstructed to the full path. That could be inlined, but it's a bit cleaner as a separate struct

cgwalters:
BTW not a big problem but something I noticed post-merge: my personal preference is to squash critical fixes to prior commits. "critical" has a sliding scale from "compiles" to "unit tests pass" to "integration tests pass" and beyond that of course to "docs are coherent" etc.

In this case the earlier commit didn't pass our integration tests, and I think that argues for squashing.

Johan-Liebert1:
Ah, my bad, it would've been better to squash. I'll keep this in mind

PR #1892, opened 2026-01-07
docs: Publish rustdoc for internal crates alongside mdbook

cgwalters:
The docs workflow now also generates rustdoc for all workspace crates and publishes them as a subdirectory of the main documentation site. This makes internal API documentation available at bootc-dev.github.io/bootc/internals.html with links to each crate.

Changes:
- Switch docs container to CentOS Stream 10 for newer Rust (1.91)
- Add native library dependencies needed for cargo doc
- Build rustdoc with --document-private-items for full coverage
- Add internals.md page listing all crates with rustdoc links
- Update Justfile to build from repo root for cargo access

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -233,7 +233,7 @@ pullspec-for-os TYPE NAME:
     @jq -r --arg v "{{NAME}}" '."{{TYPE}}"[$v]' < hack/os-image-map.json
 
 build-mdbook:
-    cd docs && podman build {{base_buildargs}} -t localhost/bootc-mdbook -f Dockerfile.mdbook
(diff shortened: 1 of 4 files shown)


cgwalters:
>  tested with a local build 

Thanks for doing that! I had done so myself as well, but with things like this having human eyes on it is useful as it's harder to "statically" check.

BTW I had AI do the initial pass at this and it ended up doing lots of stuff in the GHA workflow and I told it "no all work needs to be in the Justfile and container build" for *precisely* this reason of making it easy to run locally too.

I'm pretty happy with how the Justfile stuff is working out.

cgwalters:
Cool, it's live https://bootc-dev.github.io/bootc/internals.html

PR #1895, opened 2026-01-08
docs: Add composefs-rs crates to internals rustdoc

cgwalters:
These external git dependencies don't have docs on docs.rs, so include them in the internal documentation alongside our workspace crates.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -23,6 +23,8 @@ jobs:
         uses: ./.github/actions/bootc-ubuntu-setup
       - name: Build mdbook
         run: mkdir target && just build-mdbook-to target/docs
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - name: Setup Pages
         id: pages
         uses: actions/configure-pages@v5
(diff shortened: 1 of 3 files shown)


cgwalters:
It's because

> INFO has_release_artifact{release=GhRelease { repo: GhRepo { owner: "cargo-bins", repo: "cargo-quickinstall" }, tag: "mdbook-linkcheck-0.7.7" } artifact_name="mdbook-linkcheck-0.7.7-x86_64-unknown-linux-gnu.tar.gz"}:do_send_request{request=Request { method: GET, url: Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("api.github.com")), port: None, path: "/repos/cargo-bins/cargo-quickinstall/releases/tags/mdbook-linkcheck-0.7.7", query: None, fragment: None }, headers: {"accept": "application/vnd.github+json", "x-github-api-version": "2022-11-28"} } url=[https://api.github.com/repos/cargo-bins/cargo-quickinstall/releases/tags/mdbook-linkcheck-0.7.7}:](https://api.github.com/repos/cargo-bins/cargo-quickinstall/releases/tags/mdbook-linkcheck-0.7.7%7D:) Received status code 403 Forbidden, will wait for 120s and retry

So cargo-binstall falls back to building from source which for bad reasons also includes building openssl from source which for historical reasons requires Perl stuff we don't ship in the devenv.

cgwalters:
I tried adding a commit here to propagate the github token into the podman build, which I *think* will mean that cargo-binstall provides it when doing fetches which should avoid having us be ratelimited as if we were doing anonymous requests.

PR #1904, opened 2026-01-13
install: Fix bug in mount point check

ckyrouac:
This fixes a regression from https://github.com/bootc-dev/bootc/pull/1727 by removing the unnecessary mount point check prior to the recursive function call. Also adds some tracing statements and updates the integration test to validate the mount check works for this scenario:

/boot/efi mounted with contents in /boot/efi/EFI/firmware/foo

### DIFF
--- a/crates/tests-integration/src/install.rs
+++ b/crates/tests-integration/src/install.rs
@@ -109,7 +109,7 @@ pub(crate) fn run_alongside(image: &str, mut testargs: libtest_mimic::Arguments)
 
                 // Create a disk image with partitions for root and var
                 let disk_img = work_dir.join("disk.img");
-                let size = 12 * 1024 * 1024 * 1024;
+                let size = 1024 * 1024 * 1024; // 1 GiB
                 let disk_file = std::fs::File::create(&disk_img)?;
                 disk_file.set_len(size)?;
                 drop(disk_file);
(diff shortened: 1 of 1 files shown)


dustymabe:
This seemed to work for me with a mounted /boot/efi:

```
core@skalnik:~$ find /mnt/sysimage/
/mnt/sysimage/
/mnt/sysimage/boot
/mnt/sysimage/boot/efi
core@skalnik:~$ df -kh /mnt/sysimage/boot/efi
Filesystem      Size  Used Avail Use% Mounted on
/dev/vda1      1022M  4.0K 1022M   1% /var/mnt/sysimage/boot/efi
```

crates/tests-integration/src/install.rs
```diff
@@ -99,6 +99,10 @@ pub(crate) fn run_alongside(image: &str, mut testargs: libtest_mimic::Arguments)
                 let sh = &xshell::Shell::new()?;
                 reset_root(sh, image)?;
 
+                // Clean up any leftover LVM state from previous failed runs
```
cgwalters, line 102: For the future: This seems like it could have been a separate prep commit - then if there's a bug in this change and we need to revert it we wouldn't revert this cleanup.

cgwalters (approved):

henrywang:
The failure comes from grub2 regression. https://github.com/bootc-dev/bootc/pull/1905#issuecomment-3747458878

cgwalters:
Oh wow it was disk space again? We should probably add something to our GHA runs by default that shows the free space at end or so.

cgwalters (approved):

PR #1908, opened 2026-01-14
Drop fedora-44 from gating

cgwalters:
Add a `gating` matrix property to test-integration jobs. Jobs with `gating: false` use `continue-on-error: true`, allowing them to fail without blocking PR merges.

Mark fedora-44 as non-gating due to a grub2 regression in the base image (https://bugzilla.redhat.com/show_bug.cgi?id=2429501).

Assisted-by: OpenCode (Claude Sonnet 4)


cgwalters:
What we really need to do is https://gitlab.com/fedora/bootc/tracker/-/issues/18

henrywang:
> What we really need to do is https://gitlab.com/fedora/bootc/tracker/-/issues/18

Right. Will do later this week.

Issue #481, opened 2024-04-22
bootc install to-existing-root fails with multiple backing devices


henrywang:
Run `podman run --rm --privileged --pid=host -v /:/target -v /dev:/dev -v /var/lib/containers:/var/lib/containers --security-opt label=type:unconfined_t quay.io/redhat_emp1/bootc-workflow-test:jjj0 bootc install to-existing-root` on nvidia-jetson-agx-orin board with `LVM` configured failed with error `Installing to filesystem: Found multiple parent devices /dev/mmcblk0 and /dev/nvme0n1; not currently supported`.

This nvidia board has one SD/mmc and one nvme as disks. Bootc install does not support two disks?

bootc version: `bootc-0.1.9-4.el9.aarch64`

fdisk log:
```
Disk /dev/nvme0n1: 465.76 GiB, 500107862016 bytes, 976773168 sectors
    Disk model: Force MP600
    Units: sectors of 1 * 512 = 512 bytes
    Sector size (logical/physical): 512 bytes / 512 bytes
    I/O size (minimum/optimal): 512 bytes / 512 bytes
    Disklabel type: gpt
    Disk identifier: 957723F9-C6E8-4189-8CF9-D456F380FFFC
  
    Device         Start       End   Sectors   Size Type
    /dev/nvme0n1p1  2048 976773119 976771072 465.8G Linux LVM
  
  
    Disk /dev/mmcblk0: 59.28 GiB, 63652757504 bytes, 124321792 sectors
    Units: sectors of 1 * 512 = 512 bytes
    Sector size (logical/physical): 512 bytes / 512 bytes
    I/O size (minimum/optimal): 512 bytes / 512 bytes
    Disklabel type: gpt
    Disk identifier: E629D5BD-48A9-4041-9A5E-E04E9615BD5A
  
    Device           Start       End   Sectors  Size Type
    /dev/mmcblk0p1    2048   1230847   1228800  600M EFI System
    /dev/mmcblk0p2 1230848   3327999   2097152    1G Linux filesystem
    /dev/mmcblk0p3 3328000 124319743 120991744 57.7G Linux LVM
  
  
    Disk /dev/mmcblk0boot0: 31.5 MiB, 33030144 bytes, 64512 sectors
    Units: sectors of 1 * 512 = 512 bytes
    Sector size (logical/physical): 512 bytes / 512 bytes
    I/O size (minimum/optimal): 512 bytes / 512 bytes
  
  
[...]

PR #1911, opened 2026-01-14
install: Enable installing to multi device parents

ckyrouac:
When the root filesystem spans multiple backing devices (e.g., LVM across multiple disks), discover all parent devices and find ESP partitions on each. For bootupd/GRUB, install the bootloader to all devices with an ESP partition, enabling boot from any disk in a multi-disk setup. systemd-boot and zipl only support single-device configurations.

This adds a new integration test validating both single-ESP and dual-ESP multi-device scenarios.

Fixes: #481

Assisted-by: Claude Code (Opus 4.5)

### DIFF
--- a/tmt/tests/booted/test-multi-device-esp.nu
+++ b/tmt/tests/booted/test-multi-device-esp.nu
@@ -404,6 +404,9 @@ def test_no_esp_failure [] {
         } | complete)
 
         assert ($result.exit_code != 0) "Expected install to fail with no ESP partitions"
+        # Verify the failure is ESP-related, not an unrelated podman/runtime error
+        let combined = $"($result.stdout)\n($result.stderr)"
+        assert ($combined | str contains "ESP") $"Expected ESP-related error message, got: ($combined | str substring 0..200)"
         print $"Install failed as expected with exit code ($result.exit_code)"
     } catch {|e|
         cleanup $vg_name [$loop1, $loop2] $mountpoint


crates/lib/src/bootloader.rs
```diff
@@ -97,26 +97,61 @@ pub(crate) fn install_via_bootupd(
...
+            .log_debug()
+            .run_inherited_with_cmd_context();
+    }
+
+    // Install bootloader to each device
+    for dev in devices {
+        let devpath = dev.path();
+        println!("Installing bootloader via bootupd to {devpath}");
+        Command::new("bootupctl")
+            .args(["backend", "install", "--write-uuid"])
```
cgwalters: I think this will lead to a kind of last-one wins behavior for `bootupd.json` - but in the end they should be identical I guess?

cc @HuijingHei 

We probably want to document the right way to do multi-device installs there. (and have man pages in general)

Alternatively it might be nicer to explicitly support this in bootupd by just passing each device?

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -520,8 +520,13 @@ pub(crate) fn setup_composefs_bls_boot(
 
             cmdline_options.extend(&Cmdline::from(&composefs_cmdline));
 
-            // Locate ESP partition device
-            let esp_part = esp_in(&root_setup.device_info)?;
+            // Locate ESP partition device (use first device)
+            // TODO: Handle multiple devices (RAID, LVM, etc)
```
cgwalters: AFAIK *non-redundant* multi-device composefs setups (with systemd-boot e.g.) should work where there's just one ESP.

So I think it should work here to walk the blockdevs until we find an ESP, but we would need to error out if there are multiple.

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -1242,8 +1253,9 @@ pub(crate) async fn setup_composefs_boot(
             None,
         )?;
     } else {
+        // systemd-boot only supports single device
```
cgwalters: only a single *ESP*

cgwalters (dismissed):

ckyrouac:
waiting to merge until the patch release goes out

crates/lib/src/bootloader.rs
HuijingHei: > I think this will lead to a kind of last-one wins behavior for `bootupd.json` - but in the end they should be identical I guess?

Agree, but we need this like RAID.

> Alternatively it might be nicer to explicitly support this in bootupd by just passing each device?

That will be cleaner, and we could do this only if we make bootupd not fail if the passed device does not have the esp device.

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -520,8 +520,13 @@ pub(crate) fn setup_composefs_bls_boot(
 
             cmdline_options.extend(&Cmdline::from(&composefs_cmdline));
 
-            // Locate ESP partition device
-            let esp_part = esp_in(&root_setup.device_info)?;
+            // Locate ESP partition device (use first device)
+            // TODO: Handle multiple devices (RAID, LVM, etc)
+            let device_info = root_setup
+                .device_info
+                .first()
```
cgwalters: Can we change this to find all ESPs on all backing devs and error if there's more than one?

HuijingHei:
Maybe this PR could help to consider https://github.com/bootc-dev/bootc/issues/1916 ? @ckyrouac

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -520,8 +520,23 @@ pub(crate) fn setup_composefs_bls_boot(
...
+            }
+            let esp_part = match esp_parts.len() {
+                0 => {
+                    anyhow::bail!("Cannot locate ESP: no ESP partition found on any backing device")
+                }
+                1 => esp_parts.into_iter().next().unwrap(),
+                n => anyhow::bail!(
+                    "Found {n} ESP partitions across backing devices; only a single ESP is supported for composefs boot"
+                ),
+            };
```
cgwalters: We can do this without an `unwrap()` more like this


```
let Some(esp_part) = esp_parts.pop() else { bail!("empty") };
if esp_parts.len() > 0 { bail!("multiple"); }
```
or so

crates/lib/src/bootloader.rs
```diff
@@ -97,26 +97,61 @@ pub(crate) fn install_via_bootupd(
...
+
+    // No backing devices with ESP found. Run bootupd without --device and let it
+    // try to auto-detect. This works for:
+    // - BIOS boot (uses MBR, not ESP)
+    // - Systems where bootupd can find ESP via mounted /boot/efi
+    // UEFI boot will fail if bootupd cannot locate the ESP.
+    if devices.is_empty() {
+        tracing::warn!(
+            "No backing device with ESP found; UEFI boot may fail if ESP cannot be auto-detected"
+        );
```
cgwalters: The more I look at this the more it feels like the right fix here is for us to push multi-device installation down into bootupd.

It's really bootupd that understands e.g. "update payload only has BIOS so that's OK and that's all we should install".

It will need to be bootupd that handles non-EFI platforms too (like android boot).

crates/lib/src/install.rs
```diff
@@ -1141,7 +1141,10 @@ pub(crate) fn exec_in_host_mountns(args: &[std::ffi::OsString]) -> Result<()> {
 pub(crate) struct RootSetup {
     #[cfg(feature = "install-to-disk")]
     luks_device: Option<String>,
-    pub(crate) device_info: bootc_blockdev::PartitionTable,
+    /// Information about the backing block device partition tables.
+    /// Contains all devices that have an ESP partition when the root filesystem
+    /// spans multiple backing devices (e.g., LVM across multiple disks).
+    pub(crate) device_info: Vec<bootc_blockdev::PartitionTable>,
```
cgwalters: Related to above: We can't define this as being filtered to the ESP for e.g. s390x and other non-EFI platforms.

I think this just needs to be all backing partitioned blockdevs, and then we filter it later.

tmt/tests/booted/test-multi-device-esp.nu
```diff
@@ -0,0 +1,431 @@
...
+
+    print "No ESP failure test completed successfully"
+    tap ok
+}
+
+def main [] {
+    # See https://tmt.readthedocs.io/en/stable/stories/features.html#reboot-during-test
+    match $env.TMT_REBOOT_COUNT? {
+        null | "0" => test_single_esp,
+        "1" => { test_dual_esp; test_three_devices_partial_esp; tmt-reboot },
```
HuijingHei, line 450: How about add another test scenario like RAID1 that using the whole disks (see https://github.com/coreos/bootupd/issues/1059), for example:
```
root@localhost-live:/home/fedora# lsblk
NAME        MAJ:MIN RM  SIZE RO TYPE  MOUNTPOINTS
vda         253:0    0   30G  0 disk  
└─md126       9:126  0   30G  0 raid1 
  ├─md126p1 259:0    0  477M  0 part  
  ├─md126p2 259:1    0  954M  0 part  /boot
  └─md126p3 259:2    0 28.6G  0 part  /
vdb         253:16   0   30G  0 disk  
└─md126       9:126  0   30G  0 raid1 
  ├─md126p1 259:0    0  477M  0 part  
  ├─md126p2 259:1    0  954M  0 part  /boot
  └─md126p3 259:2    0 28.6G  0 part  /
```
To create RAID1 using command:
`sudo mdadm -CR /dev/md126 -e 1 -l1 -n 2 /dev/loop0 /dev/loop1 --assume-clean`

crates/lib/src/bootloader.rs
```diff
@@ -67,6 +76,48 @@ pub(crate) fn supports_bootupd(root: &Dir) -> Result<bool> {
...
+/// output. When `deployment_path` is set the command runs inside a bwrap
+/// container so we probe the binary from the target image.
+fn bootupd_supports_filesystem(rootfs: &Utf8Path, deployment_path: Option<&str>) -> Result<bool> {
+    let help_args = ["bootupctl", "backend", "install", "--help"];
+    let output = if let Some(deploy) = deployment_path {
+        let target_root = rootfs.join(deploy);
+        BwrapCmd::new(&target_root)
+            .setenv(
+                "PATH",
+                "/bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin",
```
cgwalters: Let's use a shared `const` for this or perhaps better have `set_default_path()` on `BwrapCmd`

hack/provision-derived.sh
```diff
@@ -69,6 +69,48 @@ resize_rootfs: false
 CLOUDEOF
 fi
 
+# Temporary: update bootupd from @CoreOS/continuous copr until
```
cgwalters, line 75: This is a bit tricky as we probably also want to CI test the case without the new bootupd too...

I know, another matrix entry in theory...

I guess it'll be tested in post submits by the workflow test.

tmt/tests/booted/test-multi-device-esp.nu
```diff
@@ -0,0 +1,430 @@
+# number: 32
+# tmt:
+#   summary: Test multi-device ESP detection for to-existing-root
+#   duration: 60m
+#
+# Test that bootc install to-existing-root can find and use ESP partitions
+# when the root filesystem spans multiple backing devices (e.g., LVM across disks).
+#
+# Five scenarios are tested across three reboot cycles:
```
cgwalters, line 9: We're just testing installation to loopback, AFAICS there's no reason to reboot the host.

That said what would be a stronger test here is for us to test these setups probably via Anaconda where we boot into a guest VM with that. I tried to streamline this in https://github.com/bootc-dev/bcvk/pull/202 but yeah it needs a bit of work.

I guess one thing we could do is extend the anaconda stuff we already have here in our integration tests.

Yet a different alternative is we ask tmt to directly support kickstart (and copy over some of the bcvk smarts there re mounting host container storage via virtiofs etc).

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -529,8 +529,18 @@ pub(crate) fn setup_composefs_bls_boot(
...
+            let esp_part = root_setup
+                .device_info
+                .find_colocated_esps()?
+                .and_then(|mut v| {
+                    if v.is_empty() {
+                        None
+                    } else {
+                        Some(v.remove(0))
+                    }
+                })
```
cgwalters: This construction looks strange, isn't it the same as just `find_colocated_esps()?.first()`?

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -567,10 +577,18 @@ pub(crate) fn setup_composefs_bls_boot(
...
             cmdline.add_or_modify(&param);
 
-            // Locate ESP partition device
-            let root_dev =
-                bootc_blockdev::list_dev_by_dir(&storage.physical_root)?.require_single_root()?;
-            let esp_dev = root_dev.find_partition_of_esp()?;
+            // Locate ESP partition device by walking up to the root disk(s)
+            let root_dev = bootc_blockdev::list_dev_by_dir(&storage.physical_root)?;
+            let esp_dev = root_dev
+                .find_colocated_esps()?
```
cgwalters: May as well have a `find_first_esp()?`

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -687,7 +705,14 @@ pub(crate) fn setup_composefs_bls_boot(
                     options: Some(cmdline_refs),
                 });
 
-            match find_vmlinuz_initrd_duplicates(&boot_digest)? {
+            // Only check for shared boot binaries during upgrades. During a
+            // fresh install the target has no existing entries, and the host's
+            // /sysroot/state/deploy would incorrectly match.
```
cgwalters: Wait a minute this looks like a bad unrelated bug - we shouldn't be opening up the *absolute* path `/boot` during install times, need to operate on the target root.

crates/lib/src/bootloader.rs
```diff
@@ -23,7 +23,13 @@ const BOOTUPD_UPDATES: &str = "usr/lib/bootupd/updates";
...
 const SYSTEMD_KEY_DIR: &str = "loader/keys";
 
-/// Mount ESP part at /boot/efi
+/// Mount the first ESP found among backing devices at /boot/efi.
+///
+/// This is used by the install-alongside path to clean stale bootloader
+/// files before reinstallation.  On multi-device setups only the first
+/// ESP is mounted and cleaned; stale files on additional ESPs are left
+/// in place (bootupd will overwrite them during installation).
+// TODO: clean all ESPs on multi-device setups
```
cgwalters, line 32: Hmm yeah maybe should move this logic into bootupd

hack/provision-derived.sh
```diff
@@ -69,6 +69,44 @@ resize_rootfs: false
...
+# base images include a version supporting --filesystem
+. /usr/lib/os-release
+case $ID in
+    fedora) copr_distro="fedora" ;;
+    *) copr_distro="centos-stream" ;;
+esac
+# Update bootc from rhcontainerbot copr; the new bootupd
+# requires a newer bootc than what ships in some base images.
```
cgwalters, line 83: But wait we're building this as part of our CI here right?

hack/provision-derived.sh
```diff
@@ -69,6 +69,44 @@ resize_rootfs: false
...
+baseurl=https://download.copr.fedorainfracloud.org/results/@CoreOS/continuous/${copr_distro}-\$releasever-\$basearch/
+type=rpm-md
+skip_if_unavailable=True
+gpgcheck=1
+gpgkey=https://download.copr.fedorainfracloud.org/results/@CoreOS/continuous/pubkey.gpg
+repo_gpgcheck=0
+enabled=1
+enabled_metadata=1
+REPOEOF
+dnf -y install bootupd-0.2.32.41.gb788553
```
cgwalters: Isn't this the main thing we need?

crates/lib/src/bootc_composefs/boot.rs
cgwalters: ➡️  https://github.com/bootc-dev/bootc/pull/2115

cgwalters:
OK I pushed a few cleanups here, when you get back tomorrow can you look @ckyrouac ?

hack/provision-derived.sh
ckyrouac: yea, this is definitely sub-optimal. This is in place to unblock the bootupd install which will fail due to the bootc version being <= 1.14.1. I might be able to move the bootupd copr install to later in the Dockerfile, after the new bootc build is installed...

ckyrouac:
> OK I pushed a few cleanups here, when you get back tomorrow can you look @ckyrouac ?

lgtm, fixed a doc string bug and removed the composefs commit.

cgwalters:
Needs a rebase :surfer: otherwise looks sane to me!

jmarrero:
s390x builds have been broken for a while too, and we know about the ZIPL/systemd boot limitations no. Maybe we should add those to the docs in a follow up.

PR #1912, opened 2026-01-14
lints: Add nonempty-run-tmp warning for runtime-only directories

cgwalters:
Add a lint that warns when /run or /tmp contain any content. These
directories are tmpfs at runtime and should be empty in container images.

Common causes of content in these directories include:
- podman/buildah's RUN --mount leaving directory stubs
- Build tools leaving temporary files

This is particularly important for bootc with composefs because content
in these directories can cause digest mismatches between build-time
(mounted filesystem) and install-time (OCI tar layers) views, leading
to sealed boot failures.

The lint uses the walk API with noxdev() to automatically skip mount
points, and filters out content injected by container runtimes
(.containerenv, secrets, packages).

### DIFF
--- a/crates/lib/src/lints.rs
+++ b/crates/lib/src/lints.rs
@@ -29,6 +29,16 @@ use serde::Serialize;
 
 use crate::bootc_composefs::boot::EFI_LINUX;
 
+/// Create a default WalkConfiguration with noxdev enabled.
+///
+/// This ensures we skip directory mount points when walking,
+/// which is important to avoid descending into bind mounts, tmpfs, etc.
+/// Note that non-directory mount points (e.g. bind-mounted regular files)
+/// will still be visited.
+fn walk_configuration() -> WalkConfiguration<'static> {
+    WalkConfiguration::default().noxdev()
+}
+
 /// Reference to embedded default baseimage content that should exist.
 const BASEIMAGE_REF: &str = "usr/share/doc/bootc/baseimage/base";
 // https://systemd.io/API_FILE_SYSTEMS/ with /var added for us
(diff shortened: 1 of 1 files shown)


jeckersb:
I guess moving the linting to the end to allow the bind mounts to release in the sealed case isn't working:

```
[3/4] STEP 3/3: RUN --network=none bootc container lint --fatal-warnings
error: Linting: Checks failed: 1
Lint warning: nonempty-run-tmp: Found content in runtime-only directories (/run, /tmp):
  /run/packages
```

jeckersb:
Hm if I run `just build-sealed` locally (which is roughly what the CI job is doing) that works ok for me.  Maybe something that's fixed in newer podman?  Looks like the ubuntu runner ends up with `5.4.1`:

```
Unpacking podman (5.4.1+ds1-1) over (4.9.3+ds1-1ubuntu0.2) ...
```

Whereas locally I have `5.7.1`.

cgwalters:
> Lint warning: nonempty-run-tmp: Found content in runtime-only directories (/run, /tmp):
  /run/packages

This should be fixed now, had to avoid the global `-v /run/packages` injected in the podman build.

jeckersb (approved):

jeckersb:
Now:

```
Lint warning: nonempty-run-tmp: Found content in runtime-only directories (/run, /tmp):
  /run/systemd
  /run/systemd/resolve
  /run/systemd/resolve/stub-resolv.conf
```

:sob:

cgwalters:
Hmm yes I wasn't reproducing that locally; I think it may indeed be a difference in the podman networking setup. I tried enforcing a tmpfs for `/run` across *all* `RUN` invocations.

cgwalters:
No you were right it is a podman version difference, my agent research turned up https://github.com/containers/buildah/pull/6233 which looks like the likely fix.

jeckersb:
> No you were right it is a podman version difference, my agent research turned up [containers/buildah#6233](https://github.com/containers/buildah/pull/6233) which looks like the likely fix.

Good job agent! :robot:

cgwalters:
Disabling automerge for this until we do a point release

PR #1913, opened 2026-01-15
Composefs test parity with ostree

Johan-Liebert1:


Johan-Liebert1:
Requires https://github.com/bootc-dev/infra/pull/95

Johan-Liebert1:
Requires https://github.com/bootc-dev/actions/pull/11

Johan-Liebert1:
Hmmm, it's still picking up bcvk v0.9.0

.github/workflows/ci.yml
```diff
@@ -157,7 +157,7 @@ jobs:
       matrix:
         # No fedora-44 due to https://bugzilla.redhat.com/show_bug.cgi?id=2429501
         test_os: [fedora-43, centos-9, centos-10]
-        variant: [ostree, composefs-sealeduki-sdboot]
+        variant: [ostree, composefs-sealeduki-sdboot, composefs-sdboot, composefs-grub]
```
cgwalters, line 164: Exciting!

.github/workflows/ci.yml
```diff
@@ -178,7 +178,18 @@ jobs:
         run: |
           BASE=$(just pullspec-for-os base ${{ matrix.test_os }})
           echo "BOOTC_base=${BASE}" >> $GITHUB_ENV
-          echo "BOOTC_variant=${{ matrix.variant }}" >> $GITHUB_ENV
+
+          case "${{ matrix.variant }}" in
+            composefs-grub|composefs-sdboot)
```
cgwalters: I'd prefer we do this kind of thing in Justfile so it's easier to do locally.

.github/workflows/ci.yml
```diff
@@ -207,11 +218,12 @@ jobs:
 
       - name: Run TMT integration tests
         run: |
-          if [ "${{ matrix.variant }}" = "composefs-sealeduki-sdboot" ]; then
-            just test-composefs
```
cgwalters, line 215: I think at this point it'd be cleaner to have a mechainism to skip some tests (tmt already has metadata) so it's always `just test-tmt integration` and the variant detection does the right thing internally.

crates/lib/src/bootloader.rs
```diff
@@ -131,6 +132,73 @@ pub(crate) fn install_systemd_boot(
         .log_debug()
         .run_inherited_with_cmd_context()?;
 
+    // Check for systemd-boot binaries in the EFI/systemd directory
+    // systemd v258 won't copy the binary if an EFI booted system is not detected
```
cgwalters: I think we really want `bootctl --force-if-non-efi-booted` or so...did you file a bug report about this? We should chase down the change and find the rationale.

The logic here feels too manual. (I mean if we had to I'd argue we just fake it out by creating `/sys/firmware/efi` as mount point or so - or wahtever is needed to make bootctl work)

crates/xtask/src/tmt.rs
```diff
@@ -430,6 +432,17 @@ pub(crate) fn run_tmt(sh: &Shell, args: &RunTmtArgs) -> Result<()> {
                     opts.push("--filesystem=xfs".to_string());
                 }
             }
+
+            if args.composefs_backend {
+                opts.push("--filesystem=ext4".into());
```
cgwalters, line 490: I think this shouldn't be hardcoded here...actually "filesystem choice" is a whole new dimension to the matrix in general.

We absolutely *should* be testing (and support!) e.g. "insecure" (fsverity disabled) composefs-rs storage too.

I think we don't even have a mechanism to specify that, but we should. I guess it could be part of the install config?

Justfile
```diff
@@ -105,9 +105,32 @@ test-container: build build-units
 
 # Build and test sealed composefs images
 [group('core')]
-test-composefs:
+test-composefs-sealeduki-sdboot:
     just variant=composefs-sealeduki-sdboot test-tmt readonly local-upgrade-reboot
 
+[group('core')]
+test-composefs bootloader:
+    just variant=composefs test-tmt --composefs-backend --bootloader {{bootloader}} \
```
cgwalters: Per above I think we can attach "supports composefs" as metadata to tests instead

Johan-Liebert1:
Okay, at least the grub tests partially pass

```
========================================
Test Summary
========================================
/tmt/plans/integration/plan-01-readonly: PASSED
/tmt/plans/integration/plan-20-image-pushpull-upgrade: PASSED
/tmt/plans/integration/plan-23-install-outside-container: PASSED
/tmt/plans/integration/plan-23-usroverlay: PASSED
/tmt/plans/integration/plan-24-image-upgrade-reboot: PASSED
/tmt/plans/integration/plan-25-soft-reboot: PASSED
/tmt/plans/integration/plan-26-download-only-upgrade: PASSED
/tmt/plans/integration/plan-29-soft-reboot-selinux-policy: FAILED
/tmt/plans/integration/plan-32-install-to-filesystem-var-mount: FAILED
/tmt/plans/integration/plan-33-bib-build: PASSED
========================================
```

Johan-Liebert1:
All tests pass except for

```
/tmt/plans/integration/plan-29-soft-reboot-selinux-policy: FAILED
```

which should be fixed by https://github.com/bootc-dev/bootc/pull/1983

.github/workflows/ci.yml
Johan-Liebert1: Updated. Added a new extra field `works_for_composefs`. Maybe the name could be a bit better...

crates/lib/src/bootloader.rs
Johan-Liebert1: We don't have systemd v258 in any of our images afaik, but I faced this when testing locally. I haven't filed a bug report as I'm not a 100% sure that it's bootctl's issue. I'll debug this

Johan-Liebert1: Have created a tracker for now https://github.com/bootc-dev/bootc/issues/1993

Justfile
Johan-Liebert1: Done

crates/xtask/src/tmt.rs
Johan-Liebert1: While trying to add this option, I realized we don't properly support installing without fsverity support :|

Johan-Liebert1: I guess I'll open a separate PR with the fixes + test

crates/xtask/src/tmt.rs
```diff
@@ -22,15 +22,19 @@ const FIELD_TRY_BIND_STORAGE: &str = "try_bind_storage";
...
+const FIELD_WORKS_FOR_COMPOSEFS: &str = "works_for_composefs";
+
 // bcvk options
 const BCVK_OPT_BIND_STORAGE_RO: &str = "--bind-storage-ro";
 const ENV_BOOTC_UPGRADE_IMAGE: &str = "BOOTC_upgrade_image";
 
 // Distro identifiers
 const DISTRO_CENTOS_9: &str = "centos-9";
 
+const COMPOSEFS_KERNEL_ARGS: [&str; 1] = ["--karg=enforcing=0"];
```
cgwalters, line 34: We landed this here https://github.com/bootc-dev/bootc/pull/1960/changes#diff-3e600a58ee853843259808340a37be363ee3545fdae1eb5f8263ea27ac176527R30 are you sure this is necessary?

crates/xtask/src/tmt.rs
```diff
@@ -22,15 +22,19 @@ const FIELD_TRY_BIND_STORAGE: &str = "try_bind_storage";
 const FIELD_SUMMARY: &str = "summary";
 const FIELD_ADJUST: &str = "adjust";
 
+const FIELD_WORKS_FOR_COMPOSEFS: &str = "works_for_composefs";
```
cgwalters: How about the inverse `fixme_skip_if_composefs` - then the goal is clear: fix the `fixme_`s.

tmt/tests/booted/test-image-pushpull-upgrade.nu
```diff
@@ -68,24 +74,25 @@ RUN echo test content > /usr/share/blah.txt
...
-    # Now, let's do some checking of the progress json
-    let progress = open --raw $progress_json | from json -o
-    sanity_check_switch_progress_json $progress
 
-    # Check that /run/reboot-required exists and is not empty
-    let rr_meta = (ls /run/reboot-required | first)
-    assert ($rr_meta.size > 0b)
+    if not $is_composefs {
+        # Now, let's do some checking of the progress json
+        let progress = open --raw $progress_json | from json -o
```
cgwalters, line 80: *Ideally* whenever we do `if composefs` we have a comment for why that is. In this case it's https://github.com/containers/composefs-rs/issues/140

However we would *also* fix this with https://github.com/bootc-dev/bootc/issues/20 doing "podman pull => composefs reflink".

tmt/tests/booted/test-install-outside-container.nu
```diff
@@ -28,6 +30,15 @@ umount /var/mnt
 # so we mask off /sysroot/ostree
 # And using systemd-run here breaks our install_t so we disable SELinux enforcement
 setenforce 0
+
+let install_cmd = if (tap is_composefs) {
+    let st = bootc status --json | from json
+    let bootloader = ($st.status.booted.composefs.bootloader | str downcase)
+    $"bootc install to-disk --disable-selinux --via-loopback --composefs-backend --bootloader=($bootloader) --filesystem ext4 --source-imgref ($target_image) ./disk.img"
```
cgwalters: Let's ideally factor out the common args

tmt/tests/booted/test-install-to-filesystem-var-mount.sh
```diff
@@ -56,13 +59,13 @@ parted -s "$LOOP_DEV" mklabel gpt
 # BIOS boot partition (for GRUB on GPT)
 parted -s "$LOOP_DEV" mkpart primary 1MiB 2MiB
 parted -s "$LOOP_DEV" set 1 bios_grub on
-# EFI partition (200 MiB)
-parted -s "$LOOP_DEV" mkpart primary fat32 2MiB 202MiB
+# EFI partition (1 GiB)
```
cgwalters, line 60: Ideally btw in the future our base images come with a repart.d that defines a default for this

Dockerfile
```diff
@@ -34,7 +34,8 @@ WORKDIR /src
 # We aren't using the full recommendations there, just the simple bits.
 # First we download all of our Rust dependencies
 # Note: Local path dependencies (from [patch] sections) are auto-detected and bind-mounted by the Justfile
-RUN --mount=type=tmpfs,target=/run --mount=type=tmpfs,target=/tmp --mount=type=cache,target=/src/target --mount=type=cache,target=/var/roothome cargo fetch
+RUN --mount=type=tmpfs,target=/run --mount=type=tmpfs,target=/tmp --mount=type=cache,target=/src/target --mount=type=cache,target=/var/roothome \
+    rm -rf /var/roothome/.cargo/registry; cargo fetch
```
cgwalters, line 38: Why?

crates/xtask/src/tmt.rs
Johan-Liebert1: That's only for UKI right? And we wouldn't want to add it to `contrib/packaging/**/kargs.d` as that will also apply for ostree

crates/xtask/src/tmt.rs
Johan-Liebert1: Yeah, that sounds good. Also, we will have some composfs specific test, like the insecure verity one, which wouldn't run for ostree. I have a separate branch for that, as that'll require this PR first. For that case, would `only_composefs` or `skip_if_ostree` make sense?

Dockerfile
Johan-Liebert1: When trying locally, I was running into

```
[3/7] STEP 6/6: RUN --mount=type=tmpfs,target=/run --mount=type=tmpfs,target=/tmp --mount=type=cache,target=/src/target --mount=type=cache,target=/var/roothome cargo fetch 
Downloading crates ... error: failed to download iana-time-zone-haiku v0.1.2 Caused by: unable to get packages from source 
Caused by: failed to parse manifest at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/iana-time-zone-haiku-0.1.2/Cargo.toml 
Caused by: no targets specified in the manifest either src/lib.rs, src/main.rs, a [lib] section, or [[bin]] section must be present 
Error: building at STEP "RUN --mount=type=tmpfs,target=/run --mount=type=tmpfs,target=/tmp --mount=type=cache,target=/src/target --mount=type=cache,target=/var/roothome cargo fetch": while running runtime: exit status 101 error: Recipe package failed with exit code 101 error: Recipe test-composefs failed on line 120 with exit code 101
```

tmt/tests/booted/test-install-outside-container.nu
Johan-Liebert1: Done

crates/xtask/src/tmt.rs
Johan-Liebert1: Updated

Dockerfile
cgwalters: Yeah I've hit that in the past, it's the cargo cache getting corrupted.

crates/xtask/src/tmt.rs
cgwalters: Let's also link this one to https://github.com/containers/composefs-rs/issues/183

PR #1915, opened 2026-01-15
Composefs changes

Johan-Liebert1:
Most of this is in prep for #1913 

1. Add option to reset soft reboot state

2. Don't soft-reboot automatically
Aligning with ostree API, now we only initiate soft-reboot if `--apply`
is passed to `bootc update`, `bootc switch`, else we only prepare the
soft reboot

3. Update image digest query format
After bootc/commit/49d753f996747a9b1f531abf35ba4e207cf4f020,
composefs-rs saves config in the format `oci-config-sha256:`.
Update to match the same

4. composefs/update: Handle --download-only flag
When `--download-only` is passed, only download the image into the
composefs repository but don't finalize it.
Conver the /run/composefs/staged-deployment to a JSON file and Add a
finalization_locked field depending upon which the finalize service will
either finalize the staged deployment or leave it as is for garbage
collection (even though GC is not fully implemented right now).

5. composefs/soft-reboot: Check for SELinux policy divergence
Until now while checking if a deployment is capable of being soft
rebooted, we were not taking into account any differences in SELinux
policies between the two deployments. This commit adds such a check
We only check for policy diff if SELinux is enabled

Some commits are split from https://github.com/bootc-dev/bootc/pull/1913

### DIFF
--- a/crates/lib/src/bootc_composefs/selinux.rs
+++ b/crates/lib/src/bootc_composefs/selinux.rs
@@ -13,6 +13,57 @@ const SELINUX_CONFIG_PATH: &str = "etc/selinux/config";
 const SELINUX_TYPE: &str = "SELINUXTYPE=";
 const POLICY_FILE_PREFIX: &str = "policy.";
 
+/// Find the highest versioned policy file in the given directory
+fn find_latest_policy_file(policy_dir: &Dir) -> Result<String> {
+    let mut highest_policy_version = -1;
+    let mut latest_policy_name = None;
+
+    for entry in policy_dir
+        .entries_utf8()
+        .context("Getting policy dir entries")?
+    {
+        let entry = entry?;
+
+        if !entry.file_type()?.is_file() {
+            // We don't want symlinks, another directory etc
+            continue;
+        }
+
+        let filename = entry.file_name()?;
+
+        match filename.strip_prefix(POLICY_FILE_PREFIX) {
+            Some(version) => {
+                let v_int = version
+                    .parse::<i32>()
+                    .with_context(|| anyhow::anyhow!("Parsing {version} as int"))?;
+
+                if v_int < highest_policy_version {
+                    continue;
+                }
+
+                highest_policy_version = v_int;
+                latest_policy_name = Some(filename.to_string());
+            }
+
+            None => continue,
+        };
+    }
+
+    latest_policy_name.ok_or_else(|| anyhow::anyhow!("Failed to get latest SELinux policy"))
+}
+
+/// Compute SHA256 hash of a policy file
+fn compute_policy_file_hash(deployment_root: &Dir, full_path: &str) -> Result<String> {
+    let mut file = deployment_root
+        .open(full_path)
+        .context("Opening policy file")?;
+    let mut hasher = openssl::hash::Hasher::new(openssl::hash::MessageDigest::sha256())?;
+    std::io::copy(&mut file, &mut hasher)?;
+
+    let hash = hex::encode(hasher.finish().context("Computing hash")?);
+    Ok(hash)
+}
+
 #[context("Getting SELinux policy for deployment {depl_id}")]
 fn get_selinux_policy_for_deployment(
     storage: &Storage,
@@ -50,56 +101,15 @@ fn get_selinux_policy_for_deployment(
 
     let policy_dir_path = format!("etc/selinux/{type_}/policy");
 
-    let mut highest_policy_version = -1;
-    let mut latest_policy_name = None;
-
     let policy_dir = deployment_root
         .open_dir(&policy_dir_path)
         .context("Opening selinux policy dir")?;
 
-    for entry in policy_dir
-        .entries_utf8()
-        .context("Getting policy dir entries")?
-    {
-        let entry = entry?;
-
-        if !entry.file_type()?.is_file() {
-            // We don't want symlinks, another directory etc
-            continue;
-        }
-
-        let filename = entry.file_name()?;
-
-        match filename.strip_prefix(POLICY_FILE_PREFIX) {
-            Some(version) => {
-                let v_int = version
-                    .parse::<i32>()
-                    .with_context(|| anyhow::anyhow!("Parsing {version} as int"))?;
-
-                if v_int < highest_policy_version {
-                    continue;
-                }
-
-                highest_policy_version = v_int;
-                latest_policy_name = Some(filename.to_string());
-            }
-
-            None => continue,
-        };
-    }
-
-    let policy_name =
-        latest_policy_name.ok_or_else(|| anyhow::anyhow!("Failed to get latest SELinux policy"))?;
+    let policy_name = find_latest_policy_file(&policy_dir)?;
 
     let full_path = format!("{policy_dir_path}/{policy_name}");
 
-    let mut file = deployment_root
-        .open(full_path)
-        .context("Opening policy file")?;
-    let mut hasher = openssl::hash::Hasher::new(openssl::hash::MessageDigest::sha256())?;
-    std::io::copy(&mut file, &mut hasher)?;
-
-    let hash = hex::encode(hasher.finish().context("Computing hash")?);
+    let hash = compute_policy_file_hash(&deployment_root, &full_path)?;
 
     Ok(Some(hash))
 }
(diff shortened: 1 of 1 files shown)


Johan-Liebert1:
Setting up boot components seem to be failing in centos9, f42 with 

```
Bootloader: grub
error: boot data installation failed: installing component EFI: No such file or directory (os error 2)
error: Installing to disk: Installing bootloader: Failed to run command: Command {
    program: "bootupctl",
    args: [
        "bootupctl",
        "backend",
        "install",
        "--write-uuid",
        "--src-root",
        "/run/bootc/mounts/rootfs/ostree/deploy/default/deploy/3c393b017a7d9c64b3cab4c74372e0226024e26c15983479c3e70455ba0ac62a.0",
        "--device",
        "/dev/loop1",
        "/run/bootc/mounts/rootfs",
    ],
    create_pidfd: false,
}
```

Johan-Liebert1:
I guess https://github.com/bootc-dev/bootc/pull/1926 is related

crates/lib/src/bootc_composefs/soft_reboot.rs
```diff
@@ -10,25 +10,70 @@ use bootc_initramfs_setup::setup_root;
...
+use cap_std_ext::dirext::CapStdExtDirExt;
 use fn_error_context::context;
 use ostree_ext::systemd_has_soft_reboot;
+use rustix::mount::{unmount, UnmountFlags};
 use std::{fs::create_dir_all, os::unix::process::CommandExt, path::PathBuf, process::Command};
 
 const NEXTROOT: &str = "/run/nextroot";
 
+#[context("Resetting soft reboot state")]
+fn reset_soft_reboot() -> Result<()> {
```
cgwalters: I think functions like this should accept their global state, like Storage.

crates/lib/src/bootc_composefs/soft_reboot.rs
```diff
@@ -10,25 +10,70 @@ use bootc_initramfs_setup::setup_root;
...
+
+    let run_dir = Dir::open_ambient_dir("/run", ambient_authority()).context("Opening run")?;
+
+    let nextroot = run_dir
+        .open_dir_optional("nextroot")
+        .context("Opening nextroot")?;
+
+    let Some(nextroot) = nextroot else {
+        tracing::debug!("Nextroot is not a directory");
+        println!("No deployment staged for soft rebooting");
```
cgwalters, line 42: Ideally we share these output messages between the two backends

crates/lib/src/bootc_composefs/update.rs
```diff
@@ -267,7 +267,7 @@ pub(crate) async fn do_upgrade(
     }
 
     if opts.soft_reboot.is_some() {
-        prepare_soft_reboot_composefs(storage, booted_cfs, &id.to_hex(), true).await?;
+        prepare_soft_reboot_composefs(storage, booted_cfs, Some(&id.to_hex()), true, false).await?;
```
cgwalters: Having more than one boolean value to functions I think leads to confusion as to what they mean. I think this can become an `enum` since they're mutually exclusive no?

Johan-Liebert1: I think this is from an older commit? I had changed this to an enum

crates/lib/src/bootc_composefs/soft_reboot.rs
Johan-Liebert1: Feels a bit weird to have an argument when they don't use it. We could use `run` from the Storage, but we need to bind mount `/run` from host any to unmount nextroot, unless I'm missing something here

crates/lib/src/bootc_composefs/soft_reboot.rs
Johan-Liebert1: That would be great, but afaik the implementation for soft reboot for ostree backend is in the ostree repo and not bootc. In the ostree path, we don't print anything if there is no deployment staged for soft rebooting. I have updated the final messages, i.e. soft reboot prep complete and soft reboot state reset messages, to match the ostree ones

crates/lib/src/bootc_composefs/export.rs
```diff
@@ -50,7 +50,8 @@ pub async fn export_repo_to_image(
 
     let imginfo = get_imginfo(storage, &depl_verity, None).await?;
 
-    let config_digest = imginfo.manifest.config().digest().digest();
+    // We want the digest in the form of "sha256:abc123"
+    let config_digest = format!("{}", imginfo.manifest.config().digest());
```
cgwalters, line 54: Hmm isn't this us losing type safety in that basically composefs-rs is using a different type alias for digests than exists in https://docs.rs/oci-spec/latest/oci_spec/image/struct.Digest.html ?

I think we could probably `impl From<oci_spec::Digest> for composefs_rs::Sha256Digest` or so?

Having to go through strings is definitely a big trap.

crates/lib/src/bootc_composefs/soft_reboot.rs
```diff
@@ -10,25 +11,66 @@ use bootc_initramfs_setup::setup_root;
...
 use ostree_ext::systemd_has_soft_reboot;
+use rustix::mount::{UnmountFlags, unmount};
 use std::{fs::create_dir_all, os::unix::process::CommandExt, path::PathBuf, process::Command};
 
 const NEXTROOT: &str = "/run/nextroot";
 
+#[context("Resetting soft reboot state")]
+pub(crate) fn reset_soft_reboot() -> Result<()> {
+    let run = Utf8Path::new("/run");
+    bind_mount_from_pidns(PID1, &run, &run, true).context("Bind mounting /run")?;
```
cgwalters, line 32: I don't think this is necessary, we only need to do that when run under podman.

crates/lib/src/bootc_composefs/soft_reboot.rs
```diff
@@ -10,25 +11,66 @@ use bootc_initramfs_setup::setup_root;
...
+
+    let run_dir = Dir::open_ambient_dir("/run", ambient_authority()).context("Opening run")?;
+
+    let nextroot = run_dir
+        .open_dir_optional("nextroot")
+        .context("Opening nextroot")?;
+
+    let Some(nextroot) = nextroot else {
+        tracing::debug!("Nextroot is not a directory");
```
cgwalters: It doesn't exist, not just not a directory

crates/lib/src/bootc_composefs/soft_reboot.rs
```diff
@@ -10,25 +11,66 @@ use bootc_initramfs_setup::setup_root;
...
+        .is_mountpoint(".")?
+        .ok_or_else(|| anyhow::anyhow!("Failed to get mount info"))?;
+
+    if !nextroot_mounted {
+        tracing::debug!("Nextroot is not a mountpoint");
+        println!("No deployment staged for soft rebooting");
+        return Ok(());
+    }
+
+    unmount(NEXTROOT, UnmountFlags::DETACH).context("Unmounting nextroot")?;
```
cgwalters, line 56: Does this work though? We always enter a mount namespace, so this is where...ohhhh, I see that's why you did the bind mount of `/run` from the host mountns above.

And indeed I think we have similar code in ostree.

But...hmmm I think it'd be cleaner to avoid mutating global state this way - we could fork a child process that does the `umount` (look at similar code in `bind_mount_from_pidns`) *or* we could always pull in the real `/run` right after we do the re-exec to enter a new mountns.

crates/lib/src/bootc_composefs/soft_reboot.rs
```diff
@@ -10,25 +11,66 @@ use bootc_initramfs_setup::setup_root;
...
+}
+
 /// Checks if the provided deployment is soft reboot capable, and soft reboots the system if
 /// argument `reboot` is true
 #[context("Soft rebooting")]
 pub(crate) async fn prepare_soft_reboot_composefs(
     storage: &Storage,
     booted_cfs: &BootedComposefs,
-    deployment_id: &String,
+    deployment_id: Option<&String>,
```
cgwalters: Prefer `Option<&str>`

crates/lib/src/bootc_composefs/status.rs
```diff
@@ -79,6 +79,12 @@ impl std::fmt::Display for ComposefsCmdline {
     }
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub(crate) struct StagedDeployment {
```
cgwalters, line 93: This one seems like it could use a comment

crates/lib/src/bootc_composefs/update.rs
```diff
@@ -279,18 +306,60 @@ pub(crate) async fn upgrade_composefs(
...
+
+        let staged_depl_dir =
+            Dir::open_ambient_dir(COMPOSEFS_TRANSIENT_STATE_DIR, ambient_authority())
+                .context("Opening transient state directory")?;
+
+        staged_depl_dir
+            .atomic_replace_with(
+                COMPOSEFS_STAGED_DEPLOYMENT_FNAME,
+                |f| -> std::io::Result<()> {
+                    f.write_all(new_staged.to_canon_json_string()?.as_bytes())
```
cgwalters: Doesn't matter here but we can optimize this by not serializing to an intermediate string by using the canon-json-rs formatter with `serde_json::to_writer`

crates/lib/src/bootc_composefs/soft_reboot.rs
Johan-Liebert1: > we could always pull in the real /run right after we do the re-exec to enter a new mountns.

I think we do have an `fd` for `/run` in the `pid/1` mount ns inside of global storage Struct, iirc. I might be imagining things though...

> we could fork a child process that does the umount

wdyt about a function similar to `bind_mount_from_pidns` but it takes a lambda to run and only runs that instead of doing `move_mount` afterwards?

Johan-Liebert1: > takes a lambda to run and only runs that

this would be in the forked child process

crates/lib/src/bootc_composefs/export.rs
Johan-Liebert1: > Hmm isn't this us losing type safety in that basically composefs-rs is using a different type alias for digests

I don't think that's the case, as composefs streams are now named as `oci-layer-sha256:<digest>`. The `<digest>` here is the same as the config digest. It's just that previously the streams were stored only as `<digest>` without the `sha256:` or any other prefix.

Maybe it'd be cleaner to export https://github.com/containers/composefs-rs/blob/6cbdb428b06d6121993c469337a57390c5895544/crates/composefs-oci/src/lib.rs#L31 from `composefs-rs`

crates/lib/src/bootc_composefs/status.rs
Johan-Liebert1: Added doc comments

crates/lib/src/bootc_composefs/update.rs
Johan-Liebert1: Updated

crates/lib/src/bootc_composefs/soft_reboot.rs
Johan-Liebert1: Updated

crates/lib/src/bootc_composefs/status.rs
```diff
@@ -422,8 +443,145 @@ fn compare_cmdline_skip_cfs(first: &Cmdline<'_>, second: &Cmdline<'_>) -> bool {
     return true;
 }
 
+const SELINUX_CONFIG_PATH: &str = "etc/selinux/config";
```
cgwalters: Let's split this out into a `mod selinux`. And while we're here...deduplicate with the similar code in composefs-rs

crates/lib/src/bootc_composefs/status.rs
```diff
@@ -422,8 +443,145 @@ fn compare_cmdline_skip_cfs(first: &Cmdline<'_>, second: &Cmdline<'_>) -> bool {
...
+
+        if !entry.file_type()?.is_file() {
+            // We don't want symlinks, another directory etc
+            continue;
+        }
+
+        let filename = entry.file_name();
+        let filename = filename
+            .to_str()
+            .ok_or_else(|| anyhow::anyhow!("Filename is not proper UTF-8"))?;
```
cgwalters: We can use `Dir::entries_utf8` (probably have other places like this)

crates/lib/src/bootc_composefs/status.rs
```diff
@@ -422,8 +443,145 @@ fn compare_cmdline_skip_cfs(first: &Cmdline<'_>, second: &Cmdline<'_>) -> bool {
...
+
+        if !filename.starts_with(POLICY_FILE_PREFIX) {
+            continue;
+        }
+
+        match filename.split_once(".") {
```
cgwalters: I think moving the `.` into the prefix then use `filename.strip_prefix(POLICY_FILE_PREFIX)` is more elegant

crates/lib/src/bootc_composefs/status.rs
```diff
@@ -448,46 +612,62 @@ fn set_reboot_capable_type1_deployments(
...
-    if depl_cmdline.as_bytes().len() != booted_cmdline.as_bytes().len() {
+    if target.full_cmdline.as_bytes().len() != booted.full_cmdline.as_bytes().len() {
         tracing::debug!("Soft reboot not allowed due to differing cmdline");
-        return false;
+        return Ok(false);
     }
 
-    return compare_cmdline_skip_cfs(depl_cmdline, booted_cmdline)
-        && compare_cmdline_skip_cfs(booted_cmdline, depl_cmdline);
+    return Ok(
```
cgwalters: Minor style personally I prefer
```
let r = <complex computation>;
Ok(r)
```
in cases like this

crates/lib/src/bootc_composefs/status.rs
```diff
@@ -422,8 +443,145 @@ fn compare_cmdline_skip_cfs(first: &Cmdline<'_>, second: &Cmdline<'_>) -> bool {
...
+        let erofs_tmp_mnt = TempMount::mount_fd(&composefs_fd)?;
+
+        (erofs_tmp_mnt.fd.try_clone()?, Some(erofs_tmp_mnt))
+    };
+
+    if !deployment_root.exists(SELINUX_CONFIG_PATH) {
+        return Ok(None);
+    }
+
+    let selinux_config = deployment_root
```
cgwalters: Stuff below here is easily unit testable (again in the selinux mod) operating on a `Dir` without any booted bits etc

crates/lib/src/bootc_composefs/soft_reboot.rs
cgwalters, line 32: Can you add a comment like `/// Note: By default bootc runs in an unshared mount namespace; this sets up our `/run` to actually be the same as the host `/run` so the `umount` below actually affects the host`

cgwalters (changes requested):

crates/lib/src/bootc_composefs/status.rs
Johan-Liebert1: Split into a separate module. About the deduplication, the composefs-rs version works with `Filesystem` which is quite different

crates/lib/src/bootc_composefs/status.rs
Johan-Liebert1: Oh, I wasn't aware that selinux policy files always have a suffix. Seems like they do. Updated

crates/lib/src/bootc_composefs/status.rs
Johan-Liebert1: Updated

crates/lib/src/bootc_composefs/soft_reboot.rs
Johan-Liebert1: Done

crates/lib/src/bootc_composefs/status.rs
Johan-Liebert1: Added tests using Claude Code

PR #1917, opened 2026-01-15
ostree-ext/tar: Unset G_MESSAGES_DEBUG before spawning ostree commit

jmarrero:
When G_MESSAGES_DEBUG is set (e.g., 'all'), GLib and OSTree emit debug messages to stdout instead of stderr. This corrupts the commit hash that we parse from the ostree commit subprocess output, causing derived layer content to be silently lost during container imports.

The issue manifests as packages installed via 'dnf install' in a Containerfile not appearing in the deployed system's rpmdb after a rebase/upgrade operation.

Fixes: https://issues.redhat.com/browse/OCPBUGS-64692

### DIFF
--- a/crates/ostree-ext/src/tar/write.rs
+++ b/crates/ostree-ext/src/tar/write.rs
@@ -400,6 +400,12 @@ pub async fn write_tar(
         None
     };
     let mut c = std::process::Command::new("ostree");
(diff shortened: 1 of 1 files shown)


jmarrero:
> (don't need to amend the commit and rerun CI though)

Thanks, created https://github.com/bootc-dev/bootc/issues/1918

Issue #1907, opened 2026-01-14
error: Installing to filesystem: Verifying empty rootfs: Requiring directory contains only mount points: Found entry in boot: efi


supakeen:
After the latest bootc update (1.12.0) `image-builder` and `bootc-image-builder` are failing to produce any disk images out of containers that include this new version with:

```
mount/- (org.osbuild.xfs): mounting /dev/loop0p4 -> /store/tmp/buildroot-tmp-p1p9zjbo/mounts/
mount/boot-efi (org.osbuild.fat): mounting /dev/loop0p2 -> /store/tmp/buildroot-tmp-p1p9zjbo/mounts/boot/efi
Installing image: docker://quay.io/centos-bootc/centos-bootc:stream9
error: Installing to filesystem: Verifying empty rootfs: Requiring directory contains only mount points: Found entry in boot: efi
```

This was already mentioned in matrix by @dustymaybe; but I thought an upstream issue to track this would be benificial so I can direct our users to this.

It seems there's already an open PR to address this: #1904.

---

As a workaround users can build containers that pin `bootc` to 1.10.0 for now.

alexlarsson:
This is affecting automotive too

cgwalters:
OK, another one with us missing gating on bib in this repo. Will work on a point release.

supakeen:
> OK, another one with us missing gating on bib in this repo. Will work on a point release.

@travier  and me started discussing how to have integration tests in `bootc` (and `bootupd`) here: https://github.com/coreos/bootupd/issues/1050

cgwalters:
I'm not immediately reproducing this problem. What precise container images are in play here?

For me a simple

```
$ sudo podman run \                                                                                                                                                                                                                                                                                     
         --rm \                                                                                                                                                                                                                                                                                            
         --privileged \                                                                                                                                                                                                                                                                                    
         --pull=newer \                                                                                                                                                                                                                                                                                    
         --security-opt label=type:unconfined_t \                                                                                                                                                                                                                                                          
         -v /tmp/output:/output \                                                                                                                                                                                                                                                                          
         -v /var/lib/containers/storage:/var/lib/containers/storage \                                                                                                                                                                                                                                      
         quay.io/centos-bootc/bootc-image-builder:latest \                                                                                                                                                                                                                                                 
         --type qcow2 \                                                                                                                                                                                                                                                                                    
         quay.io/centos-bootc/centos-bootc:stream10
```

works. And to be clear this is with

```
[...]

PR #1922, opened 2026-01-16
tmt: Add bootc-image-builder integration test for issue #1907

cgwalters:
Add a TMT test that exercises the mount point check fix from PR #1904. The test builds a container image with an embedded disk.yaml that creates a partition layout WITHOUT a separate /boot partition - just root (/) with /boot/efi as a separate mount point.

This partition layout triggers the bug from issue #1907 where bootc's empty rootfs verification would fail with:
  "Found entry in boot: efi"

The issue was that when /boot is a directory on the root filesystem (not a separate partition), but /boot/efi IS a mount point on a different device, the old code incorrectly saw "efi" as a regular directory entry rather than recognizing it was a mount point boundary.

Verified that temporarily reverting the fix from PR #1904 causes this test to fail with the expected error message.

This was already fixed by https://github.com/bootc-dev/bootc/pull/1904/changes/ab65078675f4a3875d3063f5ea43f1a89eeb52d4 but we didn't realize at the time the scope.

Closes: https://github.com/bootc-dev/bootc/issues/1907

### DIFF
--- a/tmt/plans/integration.fmf
+++ b/tmt/plans/integration.fmf
@@ -159,4 +159,11 @@ execute:
     how: fmf
     test:
       - /tmt/tests/tests/test-32-install-to-filesystem-var-mount
+
+/plan-33-bib-build:
+  summary: Test building a qcow2 disk image with bootc-image-builder
+  discover:
+    how: fmf
+    test:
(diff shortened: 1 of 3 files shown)


jeckersb (dismissed):
Yay tests 😻

PR #1898, opened 2026-01-08
build-sys: Rework sealing to be one build step

cgwalters:
Now that we're doing a "from scratch" build we don't
have the mtime issue, and so we can change our build system
to do everything in a single step.

Assisted-by: OpenCode (Opus 4.5)

PR #1926, opened 2026-01-16
build-sys: Enable CentOS Stream compose repos to avoid version skew

cgwalters:
This relates to https://github.com/coreos/bootupd/pull/995

The base image may be built from a compose that has newer packages than what's available on the public mirrors. This causes version skew where packages like bootupd have different versions between the base image and our built image.

For example, bootupd 0.2.32 changed the EFI file layout from /usr/lib/bootupd/updates/EFI/ to /usr/lib/efi/, and if we build with an older bootupd from mirrors while the target image has the newer layout, bootloader installation fails.

Enable the CentOS Stream compose repos with higher priority to ensure we get matching versions.

xref https://gitlab.com/redhat/centos-stream/containers/bootc/-/issues/1174

### DIFF
--- a/tmt/tests/booted/test-install-outside-container.nu
+++ b/tmt/tests/booted/test-install-outside-container.nu
@@ -10,7 +10,7 @@ use tap.nu
 # this test in theory independent of starting from a bootc host,
 # but also because it's useful to test "skew" between the bootc binary
 # doing the install and the target image.
-let target_image = "docker://quay.io/centos-bootc/centos-bootc:stream10"
+let target_image = "docker://quay.io/centos-bootc/centos-bootc:stream9"
 
 # setup filesystem
 mkdir /var/mnt
--- a/tmt/tests/booted/test-install-unified-flag.nu
+++ b/tmt/tests/booted/test-install-unified-flag.nu
@@ -13,7 +13,7 @@ use tap.nu
 
 # Use a generic target image to test skew between the bootc binary doing
 # the install and the target image
-let target_image = "docker://quay.io/centos-bootc/centos-bootc:stream10"
+let target_image = "docker://quay.io/centos-bootc/centos-bootc:stream9"
 
 def main [] {
     tap begin "install with experimental unified storage flag"


cgwalters:
Heh, the centos gpg filenames changed because of the PQC work...it looked obviously right to me in a quick diff but...yeah. I did test this one now.

jeckersb (dismissed):

jeckersb:
> Heh, the centos gpg filenames changed because of the PQC work...it looked obviously right to me in a quick diff but...yeah. I did test this one now.

It's funny because I tested it with gpg keys after gemini mentioned it, but i had copied it out of the other repo file and didn't notice that it didn't match with the version here.  Then when you switched gpgcheck on I said "great i just tested that!"

cgwalters:
OK so c9s is some kind of tmpfile leaking but I don't know from where, investigating.
fedora-42 I think is another variant of the bootupd skew...also looking

cgwalters:
> fedora-42 I think is another variant of the bootupd skew...also looking

Ah no we actually need to roll https://github.com/bootc-dev/bootc/pull/1927 into this

cgwalters:
I see, can't do the unified storage test with containers-storage...

But I think the bootupd desynchronization is fixed now, let's see

cgwalters:
Hmm I think there's race condition in something invoked during our build in c9s that is leaking files in `/tmp`. But looking at this I messed this up during a refactoring because we ended up using `RUN --mount=type=tmpfs,target=/run bootc container lint` which completely undermines the point of the lint...

cgwalters:
Something else I notice is the "don't rebuild the packages in tests" broke. Something to fix after this

jeckersb:
On a related note, we might have to expedite https://github.com/bootc-dev/bootc/pull/1898/changes/308acb9580c4824bc65697393d6840f59c219e80 out of that big sealing rework PR because I think it's breaking our gating tests on [F42](https://bodhi.fedoraproject.org/updates/FEDORA-2026-400d56950a#comment-4521268)

cgwalters:
Yeah looks like the updated bootupd wasn't pushed to Fedora 42 per https://bodhi.fedoraproject.org/updates/?search=rust-bootupd (only F43+).

So bootc-f42 will fail to install any newer target containers. 

And yeah I didn't test that before pushing this but it's quite possible just that one fails.

If it does my inclination is to drop f42 from gating here for now.

But the fix would be to push an updated bootupd to f42.

The only other thing we can easily do is change the tests to use an older image, but I'm not a big fan of that.

The most comprehensive fix is https://github.com/bootc-dev/bootc/pull/1816 but I am a bit leery of rolling *that* into this...

cgwalters:
Ah yes and the new bootupd was never built for c9s either, so all the "install outside container" tests that are targeting a newer image will also fail there.

But hmm...we can change such tests to also target the c9s image instead.

jeckersb (approved):
👀

PR #1927, opened 2026-01-16
tmt: Fix var-mount test to use booted container image

ckyrouac:
Use copy-to-storage to add the booted container to podman storage instead of pulling a remote image. This matches the pattern used by other TMT tests and ensures we test the actual bootc under test.

Changes:
- Use localhost/bootc from copy-to-storage instead of remote image
- Disable LBIs via bind mount of /usr/share/empty
- Remove unnecessary host modifications (usr-overlay, dnf install, etc.)
- Use 100%FREE for root LV to ensure sufficient space for deployment

### DIFF
--- a/tmt/tests/booted/test-install-to-filesystem-var-mount.sh
+++ b/tmt/tests/booted/test-install-to-filesystem-var-mount.sh
@@ -16,28 +16,20 @@
 
 set -xeuo pipefail
 
-# Use a generic target image to test skew between the bootc binary doing
-# the install and the target image
-TARGET_IMAGE="docker://quay.io/centos-bootc/centos-bootc:stream10"
+# Build a derived image with LBIs removed for installation
+TARGET_IMAGE="localhost/bootc-install"
 
 echo "Testing bootc install to-filesystem with separate /var mount"
 
-# Disable SELinux enforcement for the install
-setenforce 0
+# Copy the currently booted image to container storage for podman to use
+bootc image copy-to-storage
 
-# Enable usr-overlay to allow modifications
-bootc usr-overlay
-
-# Install required packages (bootc images are immutable, so we need to install
-# after usr-overlay is enabled)
-dnf install -y parted lvm2 dosfstools e2fsprogs
-
-# Mask off conflicting ostree state
-if test -d /sysroot/ostree; then
-    mount --bind /usr/share/empty /sysroot/ostree
-fi
-rm -vrf /usr/lib/bootupd/updates
-rm -vrf /usr/lib/bootc/bound-images.d
+# Build a derived image that removes LBIs
+cat > /tmp/Containerfile.drop-lbis <<'EOF'
+FROM localhost/bootc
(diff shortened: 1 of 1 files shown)


tmt/tests/booted/test-install-to-filesystem-var-mount.sh
```diff
@@ -122,12 +107,13 @@ echo "Filesystem layout:"
 mount | grep /var/mnt/target || true
 df -h /var/mnt/target /var/mnt/target/boot /var/mnt/target/boot/efi /var/mnt/target/var
 
-# Run bootc install to-filesystem
-# This should succeed and handle the separate /var mount correctly
+# Run bootc install to-filesystem from within the container image under test
+# The bind mount of /usr/share/empty to /usr/lib/bootc/bound-images.d disables LBIs
```
cgwalters: Yeah...but that would be really likely to break if we changed how we handle LBIs.

And in the future I think `bootc fsck` or equiv *should* error out on this.

So if LBIs are a problem here we need to make a derived image which drops them.

tmt/tests/booted/test-install-to-filesystem-var-mount.sh
```diff
@@ -16,28 +16,13 @@
 
 set -xeuo pipefail
 
-# Use a generic target image to test skew between the bootc binary doing
-# the install and the target image
-TARGET_IMAGE="docker://quay.io/centos-bootc/centos-bootc:stream10"
```
cgwalters, line 21: Ah yes the "not actually testing the code one built" is a super easy to hit problem in what we're doing...

cgwalters (dismissed):

tmt/tests/booted/test-install-to-filesystem-var-mount.sh
```diff
@@ -16,28 +16,20 @@
...
-# after usr-overlay is enabled)
-dnf install -y parted lvm2 dosfstools e2fsprogs
-
-# Mask off conflicting ostree state
-if test -d /sysroot/ostree; then
-    mount --bind /usr/share/empty /sysroot/ostree
-fi
-rm -vrf /usr/lib/bootupd/updates
-rm -vrf /usr/lib/bootc/bound-images.d
+# Build a derived image that removes LBIs
```
cgwalters, line 27: Sorry I meant to ask before also - *why* are we doing this? How do LBIs relate to this issue?

Is it about saving disk space?

ckyrouac: just a small optimization to avoid pulling the LBIs into the TMT test VM's container storage for the bootc install.

cgwalters:
I cherry picked this into https://github.com/bootc-dev/bootc/pull/1926 because we need both to fix CI right now.

tmt/tests/booted/test-install-to-filesystem-var-mount.sh
cgwalters: OK. But let's try to ensure that we include rationale for things like this in comments and commit messages.

PR #1934, opened 2026-01-19
WIP: Fallback to running system's auth.json

ckyrouac:
This fixes a bug where running bootc upgrade/switch to an image that does not have an auth.json results in failing to pull new LBIs that are stored in an authenticated registry.

This also adds a test for this to the existing LBI switch TMT test. It does this by using cstor-dist as an authenticated registry, running on the TMT test VM. Marking as a WIP for now until I push the cstor-dist image somewhere permanent.

### DIFF
--- a/tmt/tests/booted/test-logically-bound-switch.nu
+++ b/tmt/tests/booted/test-logically-bound-switch.nu
@@ -12,10 +12,17 @@
 # <verify new bound images are pulled>
 # <reboot>
 # <verify booted state>
+#
+# This test also verifies that authenticated LBIs work in two scenarios:
+# 1. Auth credentials stored in the image itself
+# 2. Auth credentials stored on the running system
+#
+# The test uses cstor-dist to serve images from containers-storage via an
+# authenticated OCI registry endpoint.
 
 use std assert
 use tap.nu
-use bootc_testlib.nu [get_cstor_auth]
+use bootc_testlib.nu [CSTOR_DIST_REGISTRY, start_cstor_dist, get_cstor_auth, setup_insecure_registry, setup_system_auth]
 
 # This code runs on *each* boot.
 bootc status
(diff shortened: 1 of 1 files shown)


crates/lib/src/podstorage.rs
```diff
@@ -133,10 +142,22 @@ fn new_podman_cmd_in(sysroot: &Dir, storage_root: &Dir, run_root: &Dir) -> Resul
 
     // Keep this in sync with https://github.com/bootc-dev/containers-image-proxy-rs/blob/b5e0861ad5065f47eaf9cda0d48da3529cc1bc43/src/imageproxy.rs#L310
     // We always override the auth to match the bootc setup.
-    let authfile_fd = ostree_ext::globals::get_global_authfile(sysroot)?.map(|v| v.1);
-    if let Some(mut fd) = authfile_fd {
+    // First try to get the authfile from the sysroot (e.g., upgrade image), and if not found,
```
cgwalters: I had a long thing where I wanted to use the term "sysroot" for "physical root" but we're not doing that consistently.

Anyways I think this comment is wrong because it shouldn't be the new image - we basically (AFAIK) never look at the new container's root during an upgrade or other operation.

IOW I think this is probably a bug in `get_global_authfile`

tmt/tests/booted/bootc_testlib.nu
```diff
@@ -23,3 +23,88 @@ export def have_hostexports [] {
 export def parse_cmdline []  {
     open /proc/cmdline | str trim | split row " "
 }
+
+# cstor-dist configuration for authenticated registry testing
+# cstor-dist serves images from containers-storage via an authenticated OCI registry endpoint
+# https://github.com/ckyrouac/cstor-dist
+const CSTOR_DIST_IMAGE = "ghcr.io/ckyrouac/cstor-dist:latest"
```
cgwalters, line 30: I'd probably be good moving cstor-dist into the bootc-dev org

tmt/tests/booted/bootc_testlib.nu
```diff
@@ -23,3 +23,88 @@ export def have_hostexports [] {
...
+    print "Starting cstor-dist with basic auth..."
+
+    # Pull test images that cstor-dist will serve
+    print "Pulling test images for cstor-dist to serve..."
+    podman pull docker.io/library/alpine:latest
+    podman pull docker.io/library/busybox:latest
+
+    # Run cstor-dist container with auth enabled
+    # Mount the local containers storage so cstor-dist can serve images from it
+    let storage_path = if ("/var/lib/containers/storage" | path exists) {
```
cgwalters, line 50: This can be done by querying the `.graphRoot` parameter from `podman system info` we have some code for this

Issue #1935, opened 2026-01-21
bootc-destructive-cleanup.service fails when no old packages exist


henrywang:
Found by John Wyatt.

The fedora-bootc-destructive-cleanup script fails when rpm -qa returns no packages:

```
Failed Units: 1
  bootc-destructive-cleanup.service
[root@realtime-03 ~]# journalctl -u bootc-destructive-cleanup.service -b --no-pager
Jan 20 03:26:54 localhost systemd[1]: Starting bootc-destructive-cleanup.service - Cleanup previous the installation after an alongside installation...
Jan 20 03:26:54 localhost fedora-bootc-destructive-cleanup[2480]: + mount -o remount,rw /sysroot
Jan 20 03:26:54 localhost fedora-bootc-destructive-cleanup[2486]: + rpm -qa --root=/sysroot --dbpath=/usr/lib/sysimage/rpm
Jan 20 03:26:54 localhost fedora-bootc-destructive-cleanup[2487]: + xargs rpm -e --root=/sysroot --dbpath=/usr/lib/sysimage/rpm
Jan 20 03:26:55 localhost.localdomain fedora-bootc-destructive-cleanup[2516]: rpm: no packages given for erase
Jan 20 03:26:55 localhost.localdomain systemd[1]: bootc-destructive-cleanup.service: Main process exited, code=exited, status=123/n/a
Jan 20 03:26:55 localhost.localdomain systemd[1]: bootc-destructive-cleanup.service: Failed with result 'exit-code'.
Jan 20 03:26:55 localhost.localdomain systemd[1]: Failed to start bootc-destructive-cleanup.service - Cleanup previous the installation after an alongside installation.
```

### Root cause:
In contrib/scripts/fedora-bootc-destructive-cleanup line 8: `rpm -qa --root=/sysroot --dbpath=/usr/lib/sysimage/rpm | xargs rpm -e --root=/sysroot --dbpath=/usr/lib/sysimage/rpm`. When there are no packages to remove (valid scenario for bootc-to-bootc installations), xargs still invokes `rpm -e` with no arguments, causing it to fail with `exit code 123`.

PR #1936, opened 2026-01-21
Fix destructive-cleanup failure when no old packages exist

henrywang:
When rpm -qa returns no packages, xargs would still invoke rpm -e with no arguments, causing failure with "no packages given for erase" (exit code 123).

Add -r flag to xargs to skip execution when input is empty.

Resolves https://github.com/bootc-dev/bootc/issues/1935


henrywang:
Failure has been fixed by PR #1937

PR #1937, opened 2026-01-21
test: Add test required packages in Packit provision

henrywang:
`install-to-filesystem-var-mount` test requires packages which should be built into image for Packit provision (system-reinstall-bootc on package mode testing farm runner)


henrywang:
Yes, all green now.

PR #1938, opened 2026-01-21
Makefile/rpm: Add shell completions

renner0e:
works 🙂
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ manpages:
 	cargo run --release --package xtask -- manpages
 
 .PHONY: completion
-completion:
+completion: bin
 	mkdir -p target/completion
 	for shell in bash elvish fish powershell zsh; do \
 		target/release/bootc completion $$shell > target/completion/bootc.$$shell; \
--- a/contrib/packaging/bootc.spec
+++ b/contrib/packaging/bootc.spec
@@ -164,8 +164,8 @@ chmod +x %{?buildroot}/%{system_reinstall_bootc_install_podman_path}
 touch %{?buildroot}/%{_docdir}/bootc/baseimage/base/sysroot/.keepdir
 find %{?buildroot}/%{_docdir} ! -type d -printf '%{_docdir}/%%P\n' | sort > bootcdoclist.txt
 
-rm -f %{buildroot}/%{_prefix}/share/elvish/lib/bootc.elv
-rm -f %{buildroot}/%{_prefix}/share/powershell/Modules/Bootc/Bootc.psm1
+rm -f %{buildroot}/%{_datadir}/elvish/lib/bootc.elv
+rm -f %{buildroot}/%{_datadir}/powershell/Modules/Bootc/Bootc.psm1
 
 %if %{with check}
 %check
@@ -193,7 +193,7 @@ fi
 %endif
 %{_unitdir}/*
 %{_mandir}/man*/*bootc*
-%if 0%{?rhel} >= 9
+%if 0%{?rhel} && 0%{?rhel} <= 9
 %{_datadir}/bash-completion/completions/bootc
 %{_datadir}/zsh/site-functions/_bootc
 %{_datadir}/fish/vendor_completions.d/bootc.fish


contrib/packaging/bootc.spec
```diff
@@ -164,6 +168,10 @@ chmod +x %{?buildroot}/%{system_reinstall_bootc_install_podman_path}
 touch %{?buildroot}/%{_docdir}/bootc/baseimage/base/sysroot/.keepdir
 find %{?buildroot}/%{_docdir} ! -type d -printf '%{_docdir}/%%P\n' | sort > bootcdoclist.txt
 
+install -Dpm 0644 target/release/bootc.bash %{buildroot}%{bash_completions_dir}/bootc.bash
```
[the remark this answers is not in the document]
renner0e: seems like most packages on fedora don't use the extension and the docs are just outdated
https://docs.fedoraproject.org/en-US/packaging-guidelines/ShellCompletions/

contrib/packaging/bootc.spec
```diff
@@ -137,6 +137,10 @@ cargo build -j%{_smp_build_ncpus} --release %{?with_rhsm:--features rhsm} --bins
 %endif
 %endif
 
+for shell in bash zsh fish; do
```
cgwalters: A little bit unfortunate to hardcode the list of shells again here; the Rust code supports others too. But OTOH there's a need to deal with the hardcoded RPM macro dirs for these too.

contrib/packaging/bootc.spec
```diff
@@ -137,6 +137,10 @@ cargo build -j%{_smp_build_ncpus} --release %{?with_rhsm:--features rhsm} --bins
...
+    target/release/bootc completion $shell > target/release/bootc.$shell
```
cgwalters: Also: it is a goal that the build process is encapsulated in `Makefile`. Let's handle this the same way we do man pages, by having it be a `make completion` say which is part of `make install` by default.

cgwalters: i.e. when someone makes a Debian/arch/etc package they don't need to replicate this logic.

renner0e: should be addressed with: https://github.com/bootc-dev/bootc/pull/1938/changes/97f26f94a9c7ccde19d0926da7621f1b06a1b4c6

I am new to Makefiles, thanks for the patience :slightly_smiling_face:

Makefile
```diff
@@ -43,8 +43,15 @@ bin: manpages
 manpages:
 	cargo run --release --package xtask -- manpages
 
+.PHONY: completion
+completion:
+	mkdir -p target/completion
+	for shell in bash elvish fish powershell zsh; do \
```
cgwalters, line 49: This list...

Makefile
```diff
@@ -54,6 +61,9 @@ install:
 	install -d $(DESTDIR)$(prefix)/lib/bootc/install
 	install -D -m 0644 -t $(DESTDIR)$(prefix)/share/man/man5 target/man/*.5; \
 	install -D -m 0644 -t $(DESTDIR)$(prefix)/share/man/man8 target/man/*.8; \
+	install -D -m 0644 target/completion/bootc.bash $(DESTDIR)$(prefix)/share/bash-completion/completions/bootc
```
cgwalters, line 64: Is not consistent with this list.

contrib/packaging/bootc.spec
```diff
@@ -164,6 +164,10 @@ chmod +x %{?buildroot}/%{system_reinstall_bootc_install_podman_path}
 touch %{?buildroot}/%{_docdir}/bootc/baseimage/base/sysroot/.keepdir
 find %{?buildroot}/%{_docdir} ! -type d -printf '%{_docdir}/%%P\n' | sort > bootcdoclist.txt
 
+install -Dpm 0644 target/completion/bootc.bash %{buildroot}%{bash_completions_dir}/bootc
```
cgwalters: This shouldn't be necessary now that we're installing `make install`

Makefile
renner0e: fixed with: https://github.com/bootc-dev/bootc/pull/1938/commits/eaf920e5ea95bf633e84c8c110cbf26c1e7805f2. I would rather not package elvish and powershell completions in the rpm due to lack of docs and ecosystem around it. I also never used them. I hope that's okay.

contrib/packaging/bootc.spec
```diff
@@ -190,6 +193,15 @@ fi
 %endif
 %{_unitdir}/*
 %{_mandir}/man*/*bootc*
+%if 0%{?rhel} >= 9
```
cgwalters: AI: *Important*: This conditional appears inverted. The commit message says "old versions don't have these macros" but the condition uses explicit paths on RHEL 9+, where the macros (`%{bash_completions_dir}` etc.) *do* exist.

If the intent is to use explicit paths on old RHEL that lacks the macros, the condition should be `%if 0%{?rhel} && 0%{?rhel} < 9` (old RHEL without macros) with explicit paths in that branch.

Alternatively, since the macros expand to identical paths anyway, consider just always using explicit paths and removing the conditional entirely:
```rpm
%{_datadir}/bash-completion/completions/bootc
%{_datadir}/zsh/site-functions/_bootc
%{_datadir}/fish/vendor_completions.d/bootc.fish
```

contrib/packaging/bootc.spec
```diff
@@ -164,6 +164,9 @@ chmod +x %{?buildroot}/%{system_reinstall_bootc_install_podman_path}
 touch %{?buildroot}/%{_docdir}/bootc/baseimage/base/sysroot/.keepdir
 find %{?buildroot}/%{_docdir} ! -type d -printf '%{_docdir}/%%P\n' | sort > bootcdoclist.txt
 
+rm -f %{buildroot}/%{_prefix}/share/elvish/lib/bootc.elv
```
cgwalters: AI: (low) Minor inconsistency: using `%{_prefix}/share` here vs `%{_datadir}` elsewhere. Could use `%{_datadir}` for consistency, though functionally equivalent.

Makefile
```diff
@@ -43,8 +43,15 @@ bin: manpages
 manpages:
 	cargo run --release --package xtask -- manpages
 
+.PHONY: completion
+completion:
```
cgwalters: AI: The `completion` target calls `target/release/bootc` but doesn't depend on `bin`. If someone runs `make completion` or `make install` without first running `make bin`, this will fail. Consider adding the dependency:
```makefile
completion: bin
```

PR #1942, opened 2026-01-22
xtask: Add local-rust-deps command for auto-detecting path dependencies

cgwalters:
Add `cargo xtask local-rust-deps` which uses `cargo metadata` to find local path dependencies outside the workspace (e.g., from [patch] sections) and outputs podman bind mount arguments.

This enables a cleaner workflow for local development against modified dependencies like composefs-rs:

1. Add a [patch] section to Cargo.toml with real local paths
2. Run `just build` - the Justfile auto-detects and bind-mounts them

Benefits over the previous BOOTC_extra_src approach:
- No manual env var needed
- Paths work for both local `cargo build` and container builds
- No /run/extra-src indirection or Cargo.toml path munging required
- Auto-detection means it Just Works™

The Justfile's build target now calls `cargo xtask local-rust-deps` to get bind mount args, falling back gracefully if there are no external deps. The old BOOTC_extra_src mechanism is still supported for backwards compat.

Assisted-by: Claude Sonnet 4 (via OpenCode)

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -25,7 +25,10 @@ base := env("BOOTC_base", "quay.io/centos-bootc/centos-bootc:stream10")
 # Buildroot base image
 buildroot_base := env("BOOTC_buildroot_base", "quay.io/centos/centos:stream10")
 # Optional: path to extra source (e.g. composefs-rs) for local development
+# DEPRECATED: Use [patch] sections in Cargo.toml instead, which are auto-detected
 extra_src := env("BOOTC_extra_src", "")
+# Set to "1" to disable auto-detection of local Rust dependencies
+no_auto_local_deps := env("BOOTC_no_auto_local_deps", "")
 
 # Internal variables
 nocache := env("BOOTC_nocache", "")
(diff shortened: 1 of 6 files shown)


Justfile
```diff
@@ -231,7 +234,12 @@ package:
     fi
     eval $(just _git-build-vars)
     echo "Building RPM with version: ${VERSION}"
-    podman build {{base_buildargs}} --build-arg=SOURCE_DATE_EPOCH=${SOURCE_DATE_EPOCH} --build-arg=pkgversion=${VERSION} -t localhost/bootc-pkg --target=build .
+    # Auto-detect local Rust path dependencies (e.g., from [patch] sections)
+    local_deps_args=""
+    if [[ -z "{{no_auto_local_deps}}" ]]; then
+        local_deps_args=$(cargo xtask local-rust-deps 2>/dev/null || true)
+    fi
+    podman build {{base_buildargs}} --build-arg=SOURCE_DATE_EPOCH=${SOURCE_DATE_EPOCH} --build-arg=pkgversion=${VERSION} -t localhost/bootc-pkg --target=build $local_deps_args .
```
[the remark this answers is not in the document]
cgwalters: Man, the Gemini review rule for shell injection is one of those things where 20% of the time it's probably absolutely critical, and 50% it's pointless noise.

Rust crates I don't think can even have shell metacharacters.

tmt/plans/integration.fmf
```diff
@@ -71,19 +71,19 @@ execute:
     test:
       - /tmt/tests/tests/test-22-logically-bound-install
 
-/plan-23-usroverlay:
```
jeckersb, line 74: Unrelated to this but I'll do a follow-up to dedupe these `23` tests, or otherwise figure out why the sort order isn't stable and we keep getting these diffs showing up

jeckersb:
Not sure what's up with the CI failure

```
+ cargo build --release -p tests-integration
...
   Compiling fnv v1.0.7
error: couldn't read `/home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/fnv-1.0.7/lib.rs`: Permission denied (os error 13)
```

Doesn't seem like that should be related to anything in this, but with this touching cargo... maybe?

tmt/plans/integration.fmf
cgwalters: The fix was in this PR https://github.com/bootc-dev/bootc/pull/1942/changes/cf4efa350a4ab5f20a6c469645c308fd809cd963

Sorry I forgot to edit the PR description after adding that commit.

What I really want is for github to default to showing a list of commit messages and the PR description is like an optional extra...

cgwalters:
> Doesn't seem like that should be related to anything in this, but with this touching cargo... maybe?

No this PR does break it, I was going to dig in more but test-install isn't gating (yet)...though it probably should be.

Anyways I'll work on it.

tmt/plans/integration.fmf
jeckersb: Ah perfect, I (obviously) didn't look through commit-by-commit and just looked at the big diff so I missed it.  I even saw the `PartialOrd` stuff but just at a glance assumed it was boilerplate for sorting something related to the local rust deps.

Issue #1686, opened 2025-10-15
set user agent header


cgwalters:
The topic of usage metrics came up, I think we should override https://github.com/containers/container-libs/blob/a3b0f19c3ea12cb02f0b11788489c52d4fb02e96/image/internal/useragent/useragent.go#L6 (vs the skopeo proxy).

That'd allow registries to distinguish "image pulls for build" vs "image pulls for bootc client runs".

cgwalters:
Requires https://github.com/containers/skopeo/pull/2724

cgwalters:
Step 2 https://github.com/bootc-dev/containers-image-proxy-rs/pull/100

PR #1943, opened 2026-01-22
lib: Set user agent header for container image pulls

cgwalters:
This allows registries to distinguish "image pulls for bootc client runs" from other skopeo/containers-image users. The user agent will be in the format "bootc/<version> skopeo/<version>".

All places in bootc that create ImageProxyConfig now use a new helper function that sets the user_agent_prefix field.

Closes: https://github.com/bootc-dev/bootc/issues/1686
Assisted-by: OpenCode (Sonnet 4)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -689,9 +689,9 @@ dependencies = [
 
 [[package]]
 name = "containers-image-proxy"
-version = "0.9.0"
+version = "0.9.1"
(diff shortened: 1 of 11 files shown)


cgwalters:
One thing I realized here is when we get to unified storage we're probably going to need to patch podman with a similar user agent override.

Issue #1939, opened 2026-01-21
Support `{r,b}oot-mount-spec` and `stateroot` in install config


jbtrystram:
Let's make the `--boot-mount-spec` and `--root-mount-spec` and `--stateroot` options of bootc install supported by the install config file

See https://github.com/osbuild/bootc-image-builder/issues/1180

cgwalters:
@jbtrystram were you planning to pick this up? When filing issues like this can you also just for procedural clarity cross-link it to the larger tracker work that (I believe) you're aiming for of using `bootc install` for coreos?

jbtrystram:
> @jbtrystram were you planning to pick this up? 

Yes, already have some code, I'll just do a bit more testing before opening a PR.

PR #1945, opened 2026-01-23
Support stateroot and mount specs in install config file

jbtrystram:
install: Allow setting ostree stateroot in install config
    
    Support for configuring the stateroot name through the install
    configuration file under `[install.ostree]`.
    The CLI flag will override config file values, as for other options.
    
    Partial fix for https://github.com/bootc-dev/bootc/issues/1939
    
    Assisted-by: Opencode (Claude Opus 4.5)
    Signed-off-by: jbtrystram <jbtrystram@redhat.com>

   
   
   ----
   
    install: Allow root and boot mount-specs in config
    
    Allow configuring the root and boot filesystem mount
    specs via the install configuration file under [install].
    
    As for other options, CLI arguments take precedence.
    
    For the to-existing-root flow, mount specs from config are ignored.
    Example configuration:
    ```
    [install]
    root-mount-spec = "LABEL=rootfs"
    boot-mount-spec = "UUID=abcd-1234"
    ```
    
    Fixes https://github.com/bootc-dev/bootc/issues/1939
    
    Assisted-by: Opencode (Claude Opus 4.5)
    Signed-off-by: jbtrystram <jbtrystram@redhat.com>

### DIFF
--- a/docs/src/man/bootc-install-config.5.md
+++ b/docs/src/man/bootc-install-config.5.md
@@ -26,6 +26,13 @@ The `install` section supports these subfields:
 - `match_architectures`: An array of strings; this filters the install config.
 - `ostree`: See below.
 - `stateroot`: The stateroot name to use. Defaults to `default`.
+- `root-mount-spec`: A string specifying the root filesystem mount specification.
+   For example, `UUID=2e9f4241-229b-4202-8429-62d2302382e1` or `LABEL=rootfs`.
+   If not provided, the UUID of the target filesystem will be used.
+   An empty string signals to omit boot mount kargs entirely.
+- `boot-mount-spec`: A string specifying the /boot filesystem mount specification.
+   If not provided and /boot is a separate mount, its UUID will be used.
+   An empty string signals to omit boot mount kargs entirely.
 
 # filesystem
 
@@ -52,9 +59,12 @@ Configuration options for the ostree repository. There is one valid field:
 ```toml
 [install.filesystem.root]
 type = "xfs"
+
 [install]
 kargs = ["nosmt", "console=tty0"]
 stateroot = "myos"
+root-mount-spec = "LABEL=rootfs"
+boot-mount-spec = "UUID=abcd-1234"
 
 [install.ostree]
 bls-append-except-default = 'grub_users=""'
--- a/crates/lib/src/install.rs
+++ b/crates/lib/src/install.rs
@@ -2239,7 +2239,12 @@ pub(crate) async fn install_to_filesystem(
     // We support overriding the mount specification for root (i.e. LABEL vs UUID versus
     // raw paths).
     // We also support an empty specification as a signal to omit any mountspec kargs.
-    let root_info = if let Some(s) = fsopts.root_mount_spec {
+    // CLI takes precedence over config file.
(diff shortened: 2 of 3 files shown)


docs/src/man/bootc-install-config.5.md
```diff
@@ -25,6 +25,14 @@ The `install` section supports these subfields:
 - `kargs`: An array of strings; this will be appended to the set of kernel arguments.
 - `match_architectures`: An array of strings; this filters the install config.
 - `ostree`: See below.
+- `stateroot`: The stateroot name to use. Defaults to `default`.
+- `root-mount-spec`: A string specifying the root filesystem mount specification.
+   For example, `UUID=2e9f4241-229b-4202-8429-62d2302382e1` or `LABEL=rootfs`.
+   If not provided, the UUID of the target filesystem will be used.
+   An empty string signals to omit boot mount kargs entirely.
+- `boot-mount-spec`: A string specifying the /boot filesystem mount specification.
```
cgwalters, line 33: This one heavily relates to https://github.com/bootc-dev/bootc/issues/1388 among others - we need to change/strengthen how we do `/boot` in general.

Are you also just omitting this value?

docs/src/man/bootc-install-config.5.md
```diff
@@ -25,6 +25,14 @@ The `install` section supports these subfields:
 - `kargs`: An array of strings; this will be appended to the set of kernel arguments.
 - `match_architectures`: An array of strings; this filters the install config.
 - `ostree`: See below.
+- `stateroot`: The stateroot name to use. Defaults to `default`.
+- `root-mount-spec`: A string specifying the root filesystem mount specification.
```
cgwalters, line 29: Let's debate this one a bit. I am OK to keep this as is under the concept that we're basically just supporting what `bootc install` already does.

But *most* cases now ideally use the [DPS](https://uapi-group.org/specifications/specs/discoverable_partitions_specification/). When would one want to hardcode a different UUID?

For your use case you just need to support omitting it right? We could start with just `root-karg-omit: true` or so?

But I dunno.

docs/src/man/bootc-install-config.5.md
jbtrystram: > Are you also just omitting this value?

in the FCOS case yes, as we expect to find the partition with it's label during first boot.

docs/src/man/bootc-install-config.5.md
jbtrystram: > But most cases now ideally use the DPS. When would one want to hardcode a different UUID?

I guess some people may need to, in some very specific cases?  but yeah, I agree that simply having a `omit` flag probably works for most of the cases. It's also required for DPS, IIUC.

Also we [recently moved to DPS](https://github.com/coreos/coreos-assembler/commit/ca0cb56dc375378440fcc9e79803b0c1abbc2bd7) in coreos-assembler as well, so it's worth investigating if that would work as-is.

cgwalters: I merged this as is with the rationale that it's not really adding new API or semantics, just making available at container build time what is already accessible at installation time.

But if you (or someone else) wants to make changes based on the comments above let's do it!

PR #1755, opened 2025-11-12
Fix a typo in bootc-install doc

rvykydal:

PR #1948, opened 2026-01-23
docs: Add architecture documentation to internals page

cgwalters:
Extend our internals docs to have more information.

### DIFF
--- a/crates/ostree-ext/src/container/mod.rs
+++ b/crates/ostree-ext/src/container/mod.rs
@@ -1,29 +1,164 @@
 //! # APIs bridging OSTree and container images
 //!
-//! This module contains APIs to bidirectionally map between a single OSTree commit and a container image wrapping it.
-//! Because container images are just layers of tarballs, this builds on the [`crate::tar`] module.
+//! This module provides the core infrastructure for bidirectionally mapping between
+//! OCI/Docker container images and OSTree repositories. It enables bootable container
+//! images to be fetched from registries, stored efficiently, and deployed as ostree
+//! commits.
 //!
-//! To emphasize this, the current high level model is that this is a one-to-one mapping - an ostree commit
-//! can be exported (wrapped) into a container image, which will have exactly one layer.  Upon import
-//! back into an ostree repository, all container metadata except for its digested checksum will be discarded.
+//! ## Overview
+//!
+//! Container images are fundamentally layers of tarballs. This module leverages the
+//! [`crate::tar`] module to import container layers as ostree content, and exports
+//! ostree commits back to container images. The key insight is that ostree's
+//! content-addressed object storage maps naturally to OCI layer deduplication.
+//!
+//! When a container image is imported ("pulled"), each layer becomes an ostree commit.
+//! These layer commits are then merged into a single "merge commit" that represents
+//! the complete filesystem state. This merge commit is what gets deployed as a
+//! bootable system.
+//!
+//! ## On-Disk Storage Structure
+//!
+//! Container images are stored in the ostree repository (typically `/sysroot/ostree/repo/`)
+//! using a structured reference (ref) namespace:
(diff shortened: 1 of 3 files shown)


docs/src/bootc-install.md
```diff
@@ -26,7 +26,7 @@ This requires running the container via `--privileged`; it uses the running Linu
 on the host to write the file content from the running container image; not the kernel
 inside the container.
 
-There are two sub-commands: `bootc install to-disk` and `boot install to-filesystem`.
+There are two sub-commands: `bootc install to-disk` and `bootc install to-filesystem`.
```
SNThrailkill, line 29: good catch. With this change [this PR](https://github.com/bootc-dev/bootc/pull/1755/changes) can be closed then.

docs/src/bootc-install.md
```diff
@@ -346,6 +346,84 @@ This argument is mainly useful for 3rd-party tooling for building disk images fr
...
+### When is DPS used vs explicit kernel arguments?
+
+| Installation Mode | Root Discovery Method |
+|-------------------|----------------------|
+| `to-disk` | Explicit `root=UUID=...` karg |
+| `to-filesystem --root-mount-spec=""` | DPS auto-discovery (no `root=` karg) |
+| `to-filesystem` (default) | Uses filesystem UUID as `root=UUID=...` karg |
+| `to-existing-root` | Inherits from existing system |
+
+For `to-disk`, bootc always injects a `root=UUID=<uuid>` kernel argument for
```
SNThrailkill, line 391: Does this mean that since were injecting a karg that when using `to-disk` DPS will always be ignored in favor of the explicit `root=` karg? Or will DPS systems ignore this and find the root partition through auto-discovery?

docs/src/bootc-install.md
```diff
@@ -393,3 +471,74 @@ Installation software such as [Anaconda](https://github.com/rhinstaller/anaconda
...
+bootc install to-disk --root-ssh-authorized-keys /path/to/authorized_keys /dev/sda
+```
+
+This writes a systemd-tmpfiles configuration that ensures the SSH keys are
+present on every boot, even if `/root` is a tmpfs.
+
+### The `.bootc-aleph.json` file
+
+After installation, bootc writes a JSON file at the root of the physical
+filesystem (`.bootc-aleph.json`) containing installation provenance information:
```
SNThrailkill, line 535: This is a very cool feature. I'm sorry if this is a dumb question, but was does `aleph` mean? I don't think I've ever come across this

docs/src/internals.md
```diff
@@ -1,29 +1,117 @@
...
+Handles deployment lifecycle:
+
+- Staging new deployments from container images
+- Kernel argument management (`bootc_kargs`)
+- Three-way merge of `/etc` configuration
+- Rollback between deployments
+
+### Spec Module
+
+Defines core types: `Host`, `HostSpec`, `HostStatus`, `BootEntry`, `ImageReference`.
```
SNThrailkill: Granted I'm not the best Rust developer, but its not readily apparent to me what these Core types in the Spec module are. Can we get a link to where these are found or a little more context?

docs/src/bootc-install.md
cgwalters: Thanks for noticing that!

docs/src/bootc-install.md
cgwalters: Yes, the `root=` karg always wins.

docs/src/bootc-install.md
cgwalters: https://en.wikipedia.org/wiki/Aleph

docs/src/internals.md
cgwalters: I've clarified this, thanks for the review!

Issue #1929, opened 2026-01-18
bootc install to-filesystem uses the wrong efi partition


hecknt:
I'm running `bootc install to-filesystem /mnt`. I am doing this with the composefs backend, but the same issue occurs on both composefs and ostree. The mount structure I have is as follows:

- /dev/sda4 (btrfs): /mnt
- /dev/sda3 (vfat): /mnt/boot/efi

My partitions overall look like this:

- /dev/sda1: vfat, type EFI System
- /dev/sda2: btrfs, type Linux filesystem
- /dev/sda3: vfat, type EFI System
- /dev/sda4: btrfs, type Linux filesystem

When I run the `bootc install to-filesystem` command, it completes successfully, but it doesn't populate `/mnt/boot/efi` at all (/dev/sda3). That's because it's actually populating /dev/sda1, which is a completely different EFI partition. This means that it is creating boot entries and installing systemd-boot to a partition that is **not mounted under the directory that i invoked `bootc install to-filesystem` on.** This is very bad.

When I set the type of `/dev/sda1` to anything that isn't EFI System and then run `bootc install to-filesystem`, it uses the correct filesystem. My assumption was that whatever EFI partition was mounted in the directory that I invoked `bootc install to-filesystem` on would be used, since it requires that you mount the EFI partition `boot/efi` in advance. Instead what happens is it uses the first available partition on the same disk that has its type set to `EFI System`.

PR #1953, opened 2026-01-25
lib/install: Use mounted ESP for install to-filesystem

Guara92:
Previously, `bootc install to-filesystem` determined the ESP by scanning the partition table of the disk hosting the target root. This logic failed to respect an explicit `/boot/efi` mount if it was different from the first ESP on the disk.

This change updates the installation logic (both for systemd-boot and bootupd paths) to strictly require that `/boot/efi` is mounted under the target root. It inspects this mountpoint to determine the correct backing device, ensuring the installation targets the intended ESP.

Assisted-by: Zed Agent (GPT-5.2-Codex)

Fixes: https://github.com/bootc-dev/bootc/issues/1929

### DIFF
--- a/crates/lib/src/bootc_composefs/boot.rs
+++ b/crates/lib/src/bootc_composefs/boot.rs
@@ -520,11 +520,11 @@ pub(crate) fn setup_composefs_bls_boot(
 
             // Locate ESP partition device
             let esp_root = root_setup.open_target_root()?;
-            let esp_part = if root_setup.require_esp_mount {
-                crate::bootloader::require_boot_efi_mount(&esp_root)?
+            let esp_device = if root_setup.require_esp_mount {
+                crate::bootloader::find_esp_mount(&esp_root)?.device
             } else {
-                match crate::bootloader::require_boot_efi_mount(&esp_root) {
-                    Ok(p) => p,
+                match crate::bootloader::find_esp_mount(&esp_root) {
+                    Ok(esp) => esp.device,
                     Err(e) => {
                         tracing::debug!(
                             "ESP mount check failed in permissive mode: {e}; falling back to partition table scan"
@@ -537,7 +537,7 @@ pub(crate) fn setup_composefs_bls_boot(
 
             (
                 root_setup.physical_root_path.clone(),
-                esp_part,
+                esp_device,
                 cmdline_options,
                 fs,
                 postfetch.detected_bootloader.clone(),
@@ -1077,11 +1077,11 @@ pub(crate) fn setup_composefs_uki_boot(
 
             let esp_root = root_setup.open_target_root()?;
 
-            let esp_part = if root_setup.require_esp_mount {
-                crate::bootloader::require_boot_efi_mount(&esp_root)?
+            let esp_device = if root_setup.require_esp_mount {
+                crate::bootloader::find_esp_mount(&esp_root)?.device
             } else {
-                match crate::bootloader::require_boot_efi_mount(&esp_root) {
-                    Ok(p) => p,
+                match crate::bootloader::find_esp_mount(&esp_root) {
+                    Ok(esp) => esp.device,
                     Err(e) => {
                         tracing::debug!(
                             "ESP mount check failed in permissive mode: {e}; falling back to partition table scan"
@@ -1094,7 +1094,7 @@ pub(crate) fn setup_composefs_uki_boot(
 
             (
                 root_setup.physical_root_path.clone(),
-                esp_part,
+                esp_device,
                 postfetch.detected_bootloader.clone(),
                 state.composefs_options.insecure,
                 state.composefs_options.uki_addon.as_ref(),
--- a/crates/lib/src/bootloader.rs
+++ b/crates/lib/src/bootloader.rs
@@ -4,7 +4,7 @@ use std::process::Command;
 
 use anyhow::{Context, Result, anyhow, bail};
 use bootc_utils::{BwrapCmd, CommandRunExt};
-use camino::Utf8Path;
+use camino::{Utf8Path, Utf8PathBuf};
 use cap_std_ext::cap_std::fs::Dir;
 use cap_std_ext::dirext::CapStdExtDirExt;
 use fn_error_context::context;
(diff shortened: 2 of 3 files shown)


cgwalters:
Thanks for this! I haven't tried reviewing the code yet, but there's also https://github.com/bootc-dev/bootc/pull/1911 which touches things in this space. I also saw https://github.com/bootc-dev/bootc/issues/1929 which needs some analysis and may be related too.

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -214,6 +212,17 @@ fi
     )
 }
 
+fn open_target_root(root_setup: &RootSetup) -> Result<Dir> {
```
cgwalters: Slightly cleaner as a method on `impl RootSetup` I'd say

crates/lib/src/install/baseline.rs
```diff
@@ -496,5 +496,6 @@ pub(crate) fn install_create_rootfs(
         boot,
         kargs,
         skip_finalize: false,
+        require_esp_mount: true,
```
cgwalters: Hmm...I would say arguably `to-disk` should *not* mount the ESP and we should ensure that it's autodiscovered properly.

crates/lib/src/bootloader.rs
```diff
@@ -31,40 +31,45 @@ pub(crate) fn esp_in(device: &PartitionTable) -> Result<&Partition> {
         .ok_or(anyhow::anyhow!("ESP not found in partition table"))
 }
 
-/// Get esp partition node based on the root dir
-pub(crate) fn get_esp_partition_node(root: &Dir) -> Result<Option<String>> {
-    let device = get_sysroot_parent_dev(&root)?;
-    let base_partitions = bootc_blockdev::partitions_of(Utf8Path::new(&device))?;
-    let esp = base_partitions.find_partition_of_esp()?;
-    Ok(esp.map(|v| v.node.clone()))
+fn normalize_esp_source(source: String) -> String {
```
cgwalters: "normalize" here could use a definition.

Also if we're trying to micro-optimize this it can return a `Cow<str>`

crates/lib/src/bootloader.rs
```diff
@@ -31,40 +31,45 @@ pub(crate) fn esp_in(device: &PartitionTable) -> Result<&Partition> {
...
-    else {
-        return Ok(());
-    };
+        .ok_or_else(|| anyhow::anyhow!("/boot/efi does not exist"))?;
 
-    let Some(false) = esp_fd.is_mountpoint(".")? else {
-        return Ok(());
-    };
+    if !esp_fd.is_mountpoint(".")?.unwrap_or(false) {
+        anyhow::bail!("/boot/efi is not a mountpoint. This is required for installation.");
```
cgwalters: I don't want to make /boot/efi strictly hard required - among other things the [DPS](https://uapi-group.org/specifications/specs/discoverable_partitions_specification/) suggests mounting at `/efi` which makes sense to me and we should handle.

Guara92:
> Thank you so much for your contribution!
> 
> > Previously, bootc install to-filesystem determined the ESP by scanning the partition table of the disk hosting the target root. This logic failed to respect an explicit /boot/efi mount if it was different from the first ESP on the disk.
> 
> Can you explain in what circumstances this happens?

Basically whenever the active mount (/boot/efi) disagreed with the first physical candidate on the disk.

Some of the circumstances where the old "scan-the-disk" logic would fail:

- Multi-Boot: If a disk has multiple partitions flagged as an EFI System Partition (ESP), the old logic would always pick the first one it found in the partition table.
- Cross-Disk Configurations: The old logic assumed the ESP must reside on the same physical disk as the root filesystem (/), i.e. if you want to fully use your fastest drive for your root filesystem (/) saving space moving your bootloader and ESP (/boot/efi) to an older drive.
- LVM or Software RAID configuration: Your root is on a logical volume, the previous logic might fail  to find the "parent disk" might return the wrong device or fail.

crates/lib/src/install/baseline.rs
Guara92: Very reasonable, thanks for the feedback, implemented https://github.com/bootc-dev/bootc/pull/1953/changes/8d870fc814dcf22d526c101223961268ddf9114d

crates/lib/src/bootc_composefs/boot.rs
Guara92: moved https://github.com/bootc-dev/bootc/pull/1953/changes/79b8874bf14d6f23880175205a5f90905696224a thanks

crates/lib/src/bootloader.rs
Guara92: Thanks for pointing me to the spec, those low-level stuff are kinda new to me, I implemented a lookup by priority based on DPS https://github.com/bootc-dev/bootc/pull/1953/changes/f3def476eab45fae1d3035a201629745d43d3470. 
Am I still missing something?

crates/lib/src/bootloader.rs
Guara92: `Cow` used and added some rustdoc on the function https://github.com/bootc-dev/bootc/pull/1953/changes/f3def476eab45fae1d3035a201629745d43d3470, is it clear enough now?

cgwalters:
> LVM or Software RAID configuration: Your root is on a logical volume, the previous logic might fail to find the "parent disk" might return the wrong device or fail.
> Cross-Disk Configurations: The old logic assumed the ESP must reside on the same physical disk as the root filesystem (/), i.e. if you want to fully use your fastest drive for your root filesystem (/) saving space moving your bootloader and ESP (/boot/efi) to an older drive.

Yes, we need to support these and it's what https://github.com/bootc-dev/bootc/pull/1911 is trying to also fix. I think we need to rationalize these two PRs cc @ckyrouac 

> Multi-Boot: If a disk has multiple partitions flagged as an EFI System Partition (ESP), the old logic would always pick the first one it found in the partition table.

I think if one does this it starts to rely on NVRAM ordering, which can get messy. We do update NVMRAM on bootc install, so it will usually work. But still...there's just no good reason to be in this situation in general I think, although maybe some dual boot setups rely on it?

ckyrouac:
@Guara92 thanks for working on this! As part of #1911 I am working on refactoring blockdev.rs to better support multiple parent devices. I should have a PR for that up shortly. That will likely conflict with this PR, so I'd recommend holding off a bit then we can work on merging this with #1911.

PR #1958, opened 2026-01-28
ci: Build test binaries before sudo commands

henrywang:
CI failed on `Test Install` test: https://github.com/bootc-dev/bootc/actions/runs/21350622481/job/61598071031?pr=1954

Solution: 
- Move `cargo build --release -p tests-integration` to run before `sudo podman build` to avoid cargo registry permission issues.
- The tests-integration binary doesn't depend on the fsverity image, so this reordering has no functional impact on the test flow.


henrywang:
The CS9 failure is caused by issue `Pull image registry.access.redhat.com/ubi9/podman:latest failed with signature error`. Already have a blocker Jira ticket to track this issue.

Issue #1955, opened 2026-01-26
Add `bootc container ukify`


jeckersb:
This has come up in a few discussions but I don't see anywhere that we created an issue specifically for this, so here it is!

In a nutshell, right now in the sealed image build workflow we are computing some things in bash and then passing those to the `ukify` command to generate the UKI.  Instead, bootc already knows internally how to compute most of that (or can be tweaked to do so), so the idea is to replace the direct `ukify` invocation with `bootc container ukify` which will simplify the build flow.

PR #1960, opened 2026-01-28
feat: Add `bootc container ukify` command

jeckersb:
Add a new subcommand that builds a Unified Kernel Image (UKI) by
computing the necessary arguments from a container image and invoking
ukify. This simplifies the sealed image build workflow by having bootc
internally compute:

- The composefs digest (via existing compute-composefs-digest logic)
- Kernel arguments from /usr/lib/bootc/kargs.d/*.toml files
- Paths to kernel, initrd, and os-release

Any additional arguments are passed through to ukify unchanged, allowing
full control over signing, output paths, and other ukify options.

The seal-uki script is updated to use this new command instead of
manually computing these values and invoking ukify directly.

Also adds kargs.d configuration files for the sealed UKI workflow:
- 10-rootfs-rw.toml: Mount root filesystem read-write
- 21-console-hvc0.toml: Console configuration for QEMU/virtio

Closes: #1955

### DIFF
--- a/contrib/packaging/seal-uki
+++ b/contrib/packaging/seal-uki
@@ -12,15 +12,7 @@ shift
 secrets=$1
 shift
 
-# Compute the composefs digest from the target rootfs
-composefs_digest=$(bootc container compute-composefs-digest "${target}")
-
-# Build the kernel command line
-# enforcing=0: https://github.com/bootc-dev/bootc/issues/1826
-# TODO: pick up kargs from /usr/lib/bootc/kargs.d
-cmdline="composefs=${composefs_digest} console=ttyS0,115200n8 console=hvc0 enforcing=0 rw"
-
-# Find the kernel version
+# Find the kernel version (needed for output filename)
 kver=$(bootc container inspect --rootfs "${target}" --json | jq -r '.kernel.version')
 if [ -z "$kver" ] || [ "$kver" = "null" ]; then
   echo "Error: No kernel found" >&2
@@ -29,12 +21,14 @@ fi
 
 mkdir -p "${output}"
 
-ukify build \
-  --linux "${target}/usr/lib/modules/${kver}/vmlinuz" \
-  --initrd "${target}/usr/lib/modules/${kver}/initramfs.img" \
-  --uname="${kver}" \
-  --cmdline "${cmdline}" \
-  --os-release "@${target}/usr/lib/os-release" \
+# Build the UKI using bootc container ukify
+# This computes the composefs digest, reads kargs from kargs.d, and invokes ukify
+#
+# WORKAROUND: SELinux must be permissive for sealed UKI boot
+# See https://github.com/bootc-dev/bootc/issues/1826
+bootc container ukify --rootfs "${target}" \
+  --karg enforcing=0 \
+  -- \
   --signtool sbsign \
   --secureboot-private-key "${secrets}/secureboot_key" \
   --secureboot-certificate "${secrets}/secureboot_cert" \
--- a/crates/lib/src/cli.rs
+++ b/crates/lib/src/cli.rs
@@ -392,6 +392,29 @@ pub(crate) enum ContainerOpts {
         /// Identifier for image; if not provided, the running image will be used.
         image: Option<String>,
     },
+    /// Build a Unified Kernel Image (UKI) using ukify.
+    ///
+    /// This command computes the necessary arguments from the container image
+    /// (kernel, initrd, cmdline, os-release) and invokes ukify with them.
+    /// Any additional arguments after `--` are passed through to ukify unchanged.
+    ///
+    /// Example:
+    ///   bootc container ukify --rootfs /target -- --output /output/uki.efi
+    Ukify {
+        /// Operate on the provided rootfs.
+        #[clap(long, default_value = "/")]
+        rootfs: Utf8PathBuf,
+
+        /// Additional kernel arguments to append to the cmdline.
+        /// Can be specified multiple times.
+        /// This is a temporary workaround and will be removed.
+        #[clap(long = "karg", hide = true)]
+        kargs: Vec<String>,
+
+        /// Additional arguments to pass to ukify (after `--`).
+        #[clap(last = true)]
+        args: Vec<OsString>,
+    },
 }
 
 /// Subcommands which operate on images.
@@ -1598,6 +1621,11 @@ async fn run_from_opt(opt: Opt) -> Result<()> {
 
                 Ok(())
             }
+            ContainerOpts::Ukify {
+                rootfs,
+                kargs,
+                args,
+            } => crate::ukify::build_ukify(&rootfs, &kargs, &args),
         },
         Opt::Completion { shell } => {
             use clap_complete::aot::generate;
--- a/crates/lib/src/kernel.rs
+++ b/crates/lib/src/kernel.rs
@@ -7,6 +7,7 @@
 use std::path::Path;
 
 use anyhow::Result;
+use camino::Utf8PathBuf;
 use cap_std_ext::cap_std::fs::Dir;
 use cap_std_ext::dirext::CapStdExtDirExt;
 use serde::Serialize;
@@ -25,6 +26,28 @@ pub(crate) struct Kernel {
     pub(crate) unified: bool,
 }
 
+/// Internal-only kernel wrapper with extra information (paths to
+/// vmlinuz, initramfs) that are useful but we don't want to leak out
+/// via serialization to inspection.
+///
+/// `Kernel` implements `From<KernelInternal>` so we can just `.into()`
+/// to get the "public" form where needed.
+pub(crate) struct KernelInternal {
+    pub(crate) kernel: Kernel,
+    /// Path to vmlinuz for traditional kernels.
+    /// This is `None` for UKI images.
+    pub(crate) vmlinuz: Option<Utf8PathBuf>,
+    /// Path to initramfs.img for traditional kernels.
+    /// This is `None` for UKI images.
+    pub(crate) initramfs: Option<Utf8PathBuf>,
+}
+
+impl From<KernelInternal> for Kernel {
+    fn from(kernel_internal: KernelInternal) -> Self {
+        kernel_internal.kernel
+    }
+}
+
 /// Find the kernel in a container image root directory.
 ///
 /// This function first attempts to find a UKI in `/boot/EFI/Linux/*.efi`.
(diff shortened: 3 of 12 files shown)


crates/lib/src/ukify.rs
```diff
@@ -0,0 +1,150 @@
...
+/// Build a UKI from the given rootfs.
+///
+/// This function:
+/// 1. Verifies that ukify is available
+/// 2. Finds the kernel in the rootfs
+/// 3. Computes the composefs digest
+/// 4. Reads kernel arguments from kargs.d
+/// 5. Invokes ukify with computed arguments plus any pass-through args
+#[context("Building UKI")]
+pub(crate) fn build_ukify(rootfs: &Utf8Path, args: Vec<OsString>) -> Result<()> {
```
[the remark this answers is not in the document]
jeckersb: Sure why not

contrib/packaging/usr-extras/lib/bootc/kargs.d/20-selinux-permissive.toml
```diff
@@ -0,0 +1,3 @@
+# WORKAROUND: SELinux must be permissive for sealed UKI boot
+# See https://github.com/bootc-dev/bootc/issues/1826
+kargs = ["enforcing=0"]
```
cgwalters: But this will now be set for the ostree tests too...I don't think we want that.

crates/lib/src/ukify.rs
```diff
@@ -0,0 +1,150 @@
...
+            kernel.version
+        );
+    }
+
+    let kver = &kernel.version;
+
+    // Compute paths for kernel and initrd
+    let modules_dir = format!("usr/lib/modules/{kver}");
+    let vmlinuz_path = rootfs.join(&modules_dir).join("vmlinuz");
+    let initramfs_path = rootfs.join(&modules_dir).join("initramfs.img");
```
cgwalters: Hmm *another* copy of this code is pretty unfortunate, let's have `find_kernel` return it?

docs/src/man/bootc-container-ukify.8.md
```diff
@@ -0,0 +1,42 @@
...
+Build a Unified Kernel Image (UKI) using ukify
+
+This command computes the necessary arguments from the container image
+(kernel, initrd, cmdline, os-release) and invokes ukify with them.
+Any additional arguments are passed through to ukify unchanged.
+
+# OPTIONS
+
+<!-- BEGIN GENERATED OPTIONS -->
+**ARGS**
```
cgwalters, line 20: I think it'd be clearer if we required `--` for this, so it's always clear what tool handles what argument.

crates/lib/src/ukify.rs
```diff
@@ -0,0 +1,150 @@
...
+    // Compute the composefs digest
+    let composefs_digest = compute_composefs_digest(rootfs, None)?;
+
+    // Get kernel arguments from kargs.d
+    let kargs = crate::bootc_kargs::get_kargs_in_root(&root, std::env::consts::ARCH)?;
+    let kargs_str: Vec<&str> = kargs.iter_str().collect();
+
+    // Build the cmdline: composefs digest + kargs from kargs.d
+    let cmdline = if kargs_str.is_empty() {
+        format!("composefs={composefs_digest}")
```
cgwalters: It'd be nice to have this be a `const` somewhere but I guess it's not that hard to grep for

crates/lib/src/ukify.rs
```diff
@@ -0,0 +1,150 @@
...
+    } else {
+        format!("composefs={composefs_digest} {}", kargs_str.join(" "))
```
cgwalters: This isn't using the nicer apis we have to add/change kargs?

crates/lib/src/ukify.rs
```diff
@@ -0,0 +1,150 @@
...
+    };
+
+    // Path to os-release (ukify expects @path format for reading from file)
+    let os_release_path = rootfs.join("usr/lib/os-release");
+    let os_release_arg = format!("@{os_release_path}");
```
cgwalters: My recommendation is to use `.cwd_dir(rootfs.try_clone()?)` below, then paths we provide can just be relative to that.

docs/src/man/bootc-container-ukify.8.md
```diff
@@ -0,0 +1,42 @@
...
+
+    Default: /
+
+<!-- END GENERATED OPTIONS -->
+
+# EXAMPLES
+
+    bootc container ukify --rootfs /target --output /output/uki.efi
+
+# SEE ALSO
```
cgwalters, line 36: also `systemd-ukify`

contrib/packaging/seal-uki
```diff
@@ -29,12 +21,9 @@ fi
...
 
-ukify build \
-  --linux "${target}/usr/lib/modules/${kver}/vmlinuz" \
-  --initrd "${target}/usr/lib/modules/${kver}/initramfs.img" \
-  --uname="${kver}" \
-  --cmdline "${cmdline}" \
-  --os-release "@${target}/usr/lib/os-release" \
+# Build the UKI using bootc container ukify
+# This computes the composefs digest, reads kargs from kargs.d, and invokes ukify
+bootc container ukify --rootfs "${target}" \
```
cgwalters, line 29: This definitely looks nicer!

cgwalters (changes requested):

crates/lib/src/ukify.rs
jeckersb: Yes good catch :+1:

contrib/packaging/usr-extras/lib/bootc/kargs.d/20-selinux-permissive.toml
cgwalters: So how about for now let's keep this in `seal-uki` ?

jeckersb: This will mean that `container ukify` will need to take arbitrary cmdline args for now.  I guess we'll just mark it as hidden, don't mention in the docs, and print a big scary warning that it will eventually be removed when we fix the enforcing thing?

crates/lib/src/ukify.rs
jeckersb: :arrow_right: https://github.com/bootc-dev/bootc/blob/main/crates/lib/src/composefs_consts.rs#L4

crates/lib/src/ukify.rs
jeckersb: :+1:  except with `current_dir` instead of `cwd_dir` since `rootfs` is already `&Utf8Path` and `current_dir` takes `AsRef<Path>`.

crates/lib/src/ukify.rs
jeckersb: I updated `Kernel` to keep track of `modules_dir` and then vmlinuz and initramfs are derived from that.  One TODO here is that means modules_dir ends up being serialized with `container inspect --json` (but not in human-readable).  So from here we either (a) hide serialization of modules_dir, or (b) add modules_dir to human-readable output.

docs/src/man/bootc-container-ukify.8.md
jeckersb: Done

crates/lib/src/ukify.rs
jeckersb: Updated to use the const

crates/lib/src/ukify.rs
jeckersb: Updated to use `Cmdline::extend`

docs/src/man/bootc-container-ukify.8.md
jeckersb: Done (except just `ukify` since that's the manpage)

contrib/packaging/usr-extras/lib/bootc/kargs.d/20-selinux-permissive.toml
jeckersb: Leaving this unresolved until we agree that the current approach of hidden `--karg` with scary warning is ok for now.

crates/lib/src/ukify.rs
cgwalters: I was thinking we have a different internal only struct with more data that has `impl Into` the type we serialize for now

jeckersb: Ok `find_kernel` now returns a `KernelInternal` wrapper that also tracks the paths for vmlinuz and initramfs.  The ergonomics aren't great in a few places but it works.  Let me know what you think.

contrib/packaging/seal-uki
```diff
@@ -29,12 +21,14 @@ fi
...
-ukify build \
-  --linux "${target}/usr/lib/modules/${kver}/vmlinuz" \
-  --initrd "${target}/usr/lib/modules/${kver}/initramfs.img" \
-  --uname="${kver}" \
-  --cmdline "${cmdline}" \
-  --os-release "@${target}/usr/lib/os-release" \
+# Build the UKI using bootc container ukify
+# This computes the composefs digest, reads kargs from kargs.d, and invokes ukify
+#
+# WORKAROUND: SELinux must be permissive for sealed UKI boot
```
cgwalters, line 27: Super minor can be followup, imo a nicer way to do this would be like
```
ukifyargs=()
# We use sbsign by default here
ukifyargs+(--signtool sbsign  ...)
# See https://github.com/bootc-dev/bootc/issues/1826
ukifyargs+=(--karg enforcing=0)
bootc container ukify ... -- ${ukifyargs[@]}
```

or so then we can associate clearly comments with args.

crates/lib/src/cli.rs
```diff
@@ -392,6 +392,29 @@ pub(crate) enum ContainerOpts {
...
+    /// Example:
+    ///   bootc container ukify --rootfs /target -- --output /output/uki.efi
+    Ukify {
+        /// Operate on the provided rootfs.
+        #[clap(long, default_value = "/")]
+        rootfs: Utf8PathBuf,
+
+        /// Additional kernel arguments to append to the cmdline.
+        /// Can be specified multiple times.
+        /// This is a temporary workaround and will be removed.
```
cgwalters, line 410: We might as well make it `--enforcing0` or something then?

That said maybe there are sufficient use cases for dynamically extending the kargs (apart from a "static" kargs.d that we could just support it, it doesn't seem too onerous).

crates/lib/src/kernel.rs
```diff
@@ -25,35 +26,67 @@ pub(crate) struct Kernel {
...
+/// Internal-only kernel wrapper with extra information (paths to
+/// vmlinuz, initramfs) that are useful but we don't want to leak out
+/// via serialization to inspection.
+///
+/// `Kernel` implements `From<KernelInternal>` so we can just `.into()`
+/// to get the "public" form where needed.
+pub(crate) struct KernelInternal {
+    pub(crate) kernel: Kernel,
+    /// Path to vmlinuz for traditional kernels.
+    /// This is `None` for UKI images.
```
cgwalters, line 38: How about making this an `enum { Uki(Utf8PathBuf), Vmlinuz { path: Utf8PathBuf, initramfs: ... }`

crates/lib/src/cli.rs
jeckersb: Yeah but that starts to stray from the whole "the image is the definitive source of truth" expectation.  I mean I guess in this case the image *is* still the source of truth, it's just that you'd have to introspect the UKI to see what the actual cmdline is (and I'm almost positive that it's not going to match what `bootc container inspect` reports).

cgwalters: > guess in this case the image is still the source of truth, it's just that you'd have to introspect the UKI to see what 

Yes 100% the kernel commandline is in the UKI which is part of the image.

jeckersb: Spun off https://github.com/bootc-dev/bootc/issues/1965 to track surfacing possible divergence between kargs.d and UKI.

PR #1962, opened 2026-01-29
[RFC] ostree-ext: import container layers directly from overlay filesystem

giuseppe:
When pulling container images from local containers-storage, detect overlay filesystem layers and import them directly to OSTree instead of processing tar streams. This allows to use reflinks on file systems that support them.

Assisted-by: Claude Opus 4.5

This is just a PoC at the moment, I've not fully reviewed the generated code yet.

@cgwalters This takes a different approach that we discussed, but I believe it is easier to maintain in the containers library, as it is much easier to add a new API to the image/storage library to expose low level objects (i.e. layer diff when available, file objects by their checksum...) that will replace the manual access to the overlay diff that is implemented now.

Another benefit, IMO, is that it moves us away from tarballs which ultimately I'd like to see it happening in the containers library as well.

I've manually tested the change and that reflinks are created to the files in the containers storage.

### DIFF
--- a/crates/ostree-ext/src/container/unencapsulate.rs
+++ b/crates/ostree-ext/src/container/unencapsulate.rs
@@ -34,6 +34,8 @@
 use crate::container::store::LayerProgress;
 
 use super::*;
+use anyhow::Context as _;
+use camino::{Utf8Path, Utf8PathBuf};
 use containers_image_proxy::{ImageProxy, OpenedImage};
 use fn_error_context::context;
 use futures_util::{Future, FutureExt};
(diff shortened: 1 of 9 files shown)


cgwalters:
Awesome, thanks for kickstarting this! So I have a giant amount of code generated in https://github.com/cgwalters/cstor-rs that is a lot more fleshed-out "read (and maybe write to) containers-storage" - but it also needs some cleanup, review, testing etc.

Backing up a second...here's what I'm thinking re https://github.com/bootc-dev/bootc/issues/20

Basically let's try to go to a world where the flow is:

- Pull into containers-storage: using podman/skopeo (probably podman since the bollard apis are probably just fine for this, but we could make it pluggable too)
- Reflink/hardlink from there into composefs-rs
- If ostree is enabled, reflink/hardlink from composefs-rs into ostree

The thing that is really appealing to me about this is we're centralizing more on composefs.

giuseppe:
> * 
> * Reflink/hardlink from there into composefs-rs

I have missed this one.  Where is the composefs-rs store?

cgwalters:
> I have missed this one. Where is the composefs-rs store?

It's initialized by default only in the composefs path, but...it would make sense probably to do so by default now even with ostree. It's in `/composefs`.

giuseppe:
I guess I've hit https://github.com/ostreedev/ostree/issues/2712 as it fails to import the whiteout and it seems there is no way to filter it without changing the source directory

cgwalters:
> Reflink/hardlink from there into composefs-rs

:arrow_right: https://github.com/containers/composefs-rs/pull/218

crates/ostree-ext/src/container/unencapsulate.rs
```diff
@@ -185,6 +187,108 @@ pub async fn unencapsulate(repo: &ostree::Repo, imgref: &OstreeImageReference) -
...
+    let composefs_blob_path = storage_root.join(format!(
+        "overlay/{}/composefs-data/composefs.blob",
+        layer_id
+    ));
+    if composefs_blob_path.exists() {
+        return Ok(None);
+    }
+
+    // Construct the layer diff path: $STORAGE/overlay/$LAYER_ID/diff
+    let layer_diff_path = storage_root.join(format!("overlay/{}/diff", layer_id));
```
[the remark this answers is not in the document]
cgwalters: We're effectively trusting the source storage so this is not a critical problem at all, but https://github.com/containers/composefs-rs/pull/218 comprehensively uses cap_std `Dir` for this reason.

crates/ostree-ext/src/container/unencapsulate.rs
```diff
@@ -185,6 +187,108 @@ pub async fn unencapsulate(repo: &ostree::Repo, imgref: &OstreeImageReference) -
...
+    compressed_diff_digest: Option<String>,
+}
+
+/// Look up the overlay directory ID from layers.json given a digest
+fn lookup_layer_id_from_digest(storage_root: &Utf8Path, digest: &str) -> Result<Option<String>> {
+    let layers_json_path = storage_root.join("overlay-layers/layers.json");
+
+    let file = match std::fs::File::open(&layers_json_path) {
+        Ok(f) => f,
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
```
cgwalters, line 228: I maintain this cap-std-ext crate which has https://docs.rs/cap-std-ext/latest/cap_std_ext/dirext/trait.CapStdExtDirExt.html#tymethod.open_optional for this reason.

crates/ostree-ext/src/container/unencapsulate.rs
```diff
@@ -185,6 +187,108 @@ pub async fn unencapsulate(repo: &ostree::Repo, imgref: &OstreeImageReference) -
...
+    };
+
+    let reader = std::io::BufReader::new(file);
+    let layers: Vec<OverlayLayerEntry> =
+        serde_json::from_reader(reader).context("Failed to parse layers.json")?;
+
+    // Search for matching digest in both diff-digest and compressed-diff-digest
+    for layer in &layers {
+        let matches = layer.diff_digest.as_deref() == Some(digest)
+            || layer.compressed_diff_digest.as_deref() == Some(digest);
```
cgwalters, line 241: Hmm this is interesting, I need to understand the distinction here, will look.

crates/ostree-ext/src/filesystem/filter.rs
```diff
@@ -0,0 +1,313 @@
...
+    src_path: &Utf8Path,
+    config: &FilesystemFilterConfig,
+) -> Result<crate::tar::WriteTarResult> {
+    let cancellable = gio::Cancellable::NONE;
+
+    // Create the root MutableTree
+    let root_mtree = ostree::MutableTree::new();
+    let mut filtered_stats = BTreeMap::new();
+
+    // Create a commit modifier - no SELinux here, it's applied during merge
```
cgwalters, line 104: We really should set the selinux labeling policy here. Which is messy! Because we need to mount (or compute/access) the *merged* root for just `/etc/selinux`.

But not doing so causes effective double disk space usage temporarily, it's very ugly.

Again the SELinux labeling thing is one of the most awesome parts about composefs - we just stick all that in the metadata.

crates/ostree-ext/src/filesystem/filter.rs
```diff
@@ -0,0 +1,313 @@
...
+    let entries =
+        std::fs::read_dir(src_path).with_context(|| format!("Reading directory: {}", src_path))?;
+
+    for entry in entries {
+        let entry = entry.context("Reading directory entry")?;
+        let name = entry.file_name();
+        let name_str = name.to_string_lossy();
+        let file_type = entry.file_type().context("Getting file type")?;
+
+        // Only process directories at toplevel
```
cgwalters, line 123: You know what's funny:

```
$ podman run --rm -ti quay.io/fedora/fedora cat /.profile
kiwi_align='1048576'
kiwi_boot_timeout=''
kiwi_bootkernel=''
kiwi_bootloader='grub2'
```

Yes, we build the Fedora base image with this "kiwi" tool that likes to dump its metadata as a file into `/`...soooooo ugly

crates/ostree-ext/src/filesystem/filter.rs
```diff
@@ -0,0 +1,313 @@
...
+            repo.write_dfd_to_mtree(
+                src_fd.as_raw_fd(),
+                &file_name_str,
+                mtree,
+                Some(modifier),
+                cancellable,
+            )
+            .with_context(|| format!("Importing file: {}", file_path.display()))?;
+        } else if is_overlay_whiteout(&metadata) {
+            // Convert overlay whiteout to OCI format (.wh.<filename>)
```
cgwalters, line 267: Hmm but we're not handling `.wh..wh..opq` here?

giuseppe:
I’m closing this PR for now since the long term plan seems to be using the composefs backend. I’ll put my efforts to contributing there rather than this PR (at least for now)

PR #1963, opened 2026-01-29
Update to latest composefs-rs

cgwalters:
To unblock reverse dependency testing.

(We still have the forked cfsctl, that's the only thing)

### DIFF
--- a/crates/lib/src/cfsctl.rs
+++ b/crates/lib/src/cfsctl.rs
@@ -383,7 +383,7 @@ where
             }
         }
         Command::GC => {
-            repo.gc()?;
+            repo.gc(&[])?;
         }
     }
     Ok(())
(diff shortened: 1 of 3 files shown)


jmarrero (approved):
lgtm

crates/lib/src/cfsctl.rs
```diff
@@ -383,7 +383,7 @@ where
             }
         }
         Command::GC => {
-            repo.gc()?;
+            repo.gc(&[])?;
```
jmarrero, line 386: why this change is needed?

cgwalters: It came in with https://github.com/containers/composefs-rs/pull/200 and if you dig in there you'll notice I asked the same question - and IMO (part of) the answer is https://github.com/containers/composefs-rs/pull/21

PR #1978, opened 2026-02-04
Bootuuid option

jbtrystram:
install/bootloader: make --write-uuid optional
    
    Allow skiping the boot uuid stamp file writing in bootupd. When creating
    a golden image, we expect the UUIDS to be randomized at first boot, so
    always writing the stamp is odd.
    At least for Fedora CoreOS the stamp being already there at first boot
    short-circuit the partitions uuid randomization process. So let's make
    it optionnal. It will default to `false`, preserving existing behaviour
    when not specified.
    
    ref https://github.com/coreos/fedora-coreos-config/pull/3898
    
    Signed-off-by: jbtrystram <jbtrystram@redhat.com>

---

    install/config: wire up `skip_boot_uuid_stamp`
    
    Wiring-up this knob to the install config to allow it's use in
    bootc-image-builder.
    
    Signed-off-by: jbtrystram <jbtrystram@redhat.com>

### DIFF
--- a/crates/lib/src/install.rs
+++ b/crates/lib/src/install.rs
@@ -1517,7 +1517,7 @@ async fn verify_target_fetch(
 
 /// Preparation for an install; validates and prepares some (thereafter immutable) global state.
 async fn prepare_install(
-    config_opts: InstallConfigOpts,
+    mut config_opts: InstallConfigOpts,
     source_opts: InstallSourceOpts,
     target_opts: InstallTargetOpts,
     mut composefs_options: InstallComposefsOpts,
@@ -1642,8 +1642,17 @@ async fn prepare_install(
     }
 
     let install_config = config::load_config()?;
-    if install_config.is_some() {
+    if let Some(ref config) = install_config {
         tracing::debug!("Loaded install configuration");
+        // Merge config file values into config_opts (CLI takes precedence)
+        // Only apply config file value if CLI didn't explicitly set it
+        if !config_opts.bootupd_skip_boot_uuid {
+            config_opts.bootupd_skip_boot_uuid = config
+                .bootupd
+                .as_ref()
+                .and_then(|b| b.skip_boot_uuid)
+                .unwrap_or(false);
+        }
     } else {
         tracing::debug!("No install configuration found");
     }
--- a/docs/src/man/bootc-install-config.5.md
+++ b/docs/src/man/bootc-install-config.5.md
@@ -54,6 +54,14 @@ Configuration options for the ostree repository. There is one valid field:
   Boot Loader Spec entries, except for the default entry. This is useful for configuring
   arguments that should only apply to non-default deployments.
 
+# bootupd
+
+Configuration options for bootupd, responsible of setting up the bootloader.
+There is only one valid field:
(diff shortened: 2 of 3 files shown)


jbtrystram:
Oh snap it took the branch name as the PR title instead of the commit title, because it's two commits. 
Sorry, I cannot edit that

docs/src/man/bootc-install-config.5.md
```diff
@@ -33,6 +33,9 @@ The `install` section supports these subfields:
 - `boot-mount-spec`: A string specifying the /boot filesystem mount specification.
    If not provided and /boot is a separate mount, its UUID will be used.
    An empty string signals to omit boot mount kargs entirely.
+- `skip-boot-uuid-stamp`: A boolean that controls whether to skip writing partition UUIDs
```
cgwalters: This is actually specific to bootupd and does not apply with e.g. sdboot and DPS in general. I think it's also meaningless with s390x. I wonder if we should have a dedicated `bootupd` section with this as the first option.

Now...then the next question is if we change things so that we *error* if bootupd isn't used. That would force cases like the FCOS image builds to conditionalize the install config per-arch/layout, but it does feel more right to me instead of silently ignoring it here, which is less compatible if we ever aded any other knobs

crates/lib/src/install.rs
```diff
@@ -368,6 +368,11 @@ pub(crate) struct InstallConfigOpts {
     /// The stateroot name to use. Defaults to `default`.
     #[clap(long)]
     pub(crate) stateroot: Option<String>,
+
+    /// Don't pass --write-uuid to bootupd during bootloader installation.
+    #[clap(long)]
+    #[serde(default)]
+    pub(crate) skip_boot_uuid_stamp: bool,
```
cgwalters: Also per below how about `--bootupd-skip-boot-uuid`

docs/src/man/bootc-install-config.5.md
jbtrystram: Another approach suggested by @jlebon was to hide that logic behind the `boot-mount-spec=""` special case, since this is effectively signaling that the user want to own that process. But the kernel mount spec and a bootloader mount spec are still technically different stages, so I am not sure it's a better approach.

Now about having it fail when not applicable.. I am not really opposed to it. It make the pipeline slightly more complex, but we already have different manifests for each arch so I guess that isn't too much more work.

Final question: you said "bootupd", by that you mean grub ? Isn't bootupd support sd-boot at some point ?

cgwalters: > Final question: you said "bootupd", by that you mean grub ? Isn't bootupd support sd-boot at some point ?

Yes well the main argument is to support the *shim* + sdboot case actually but the uuid is really IMO a legacy for GRUB systems and newer things should use the DPS which we enforce with the sealeduki-composefs case.

(I have a WIP to *allow* DPS with ostree+grub but it requires newer Fedora)

jbtrystram: > (I have a WIP to _allow_ DPS with ostree+grub but it requires newer Fedora)

Nice ! Did you patch grub to support looking at the part type label ?

PR #1979, opened 2026-02-04
MAINTAINERS: Add Preethi as representative

jmarrero:
Preethi represents bootc in public forums like Kubecon, RH Summit and our community meetings.


cgwalters:
FTR I have no problem force merging this but we still need to figure out why CI broke - I'm looking at it

PR #1982, opened 2026-02-05
test: Fix btrfs FS on CS10 image issue

henrywang:
Test `centos-10, composefs-sealeduki-sdboot` failed (https://github.com/bootc-dev/bootc/actions/runs/21662811846/job/62588162722?pr=1974).

MR https://gitlab.com/redhat/centos-stream/src/kernel/centos-stream-10/-/merge_requests/2009 already added "CONFIG_BTRFS_FS=m" into `config`.

Update grub to use `CONFIG_BTRFS_FS=y` for btrfs support.

### DIFF
--- a/contrib/packaging/configure-rootfs
+++ b/contrib/packaging/configure-rootfs
@@ -26,7 +26,8 @@ else
   fs=
   case "${VARIANT}" in
     composefs*)
-      btrfs=$(grep -qEe '^CONFIG_BTRFS_FS' /usr/lib/modules/${kver}/config && echo btrfs || true)
+      # CONFIG_BTRFS_FS=m has been added into CS10 kernel config
+      btrfs=$(grep -qEe '^CONFIG_BTRFS_FS=y' /usr/lib/modules/${kver}/config && echo btrfs || true)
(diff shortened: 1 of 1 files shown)


contrib/packaging/configure-rootfs
```diff
@@ -26,7 +26,8 @@ else
   fs=
   case "${VARIANT}" in
     composefs*)
-      btrfs=$(grep -qEe '^CONFIG_BTRFS_FS' /usr/lib/modules/${kver}/config && echo btrfs || true)
+      # CONFIG_BTRFS_FS=m has been added into CS10 kernel config
```
cgwalters, line 29: Well the thing is, the real problem is we assume that if the kernel has it enabled, `btrfs-progs` is in the base container image as well.

So...I think a different fix here would be to check for the *combination* of the two.

---

Actually hmm....actually though the other even better thing here is to ensure we honor the default fs type in the container image, I thought we were doing that.

henrywang:
The failure `CI / test-integration (centos-10, ostree) (pull_request)` has been fixed by PR #1983 . We can manual merge both PRs.

Issue #1989, opened 2026-02-07
Is systemd-boot supported or not?


chlorodose:
Install with `bootc install to-disk --bootloader systemd <disk>` will report `error: Installing to disk: bootupd is required for ostree-based installs`. I tried with `fedora-bootc:rawhide` and my own image with `dnf swap bootupd systemd-boot`.

Dose that means only grub is supported without `--composefs-backend`?  I'm really confused because this error only report after parting the disks(bad options should report just after the command is started right?) and the [docs](https://bootc-dev.github.io/bootc/bootloaders.html) seems to say that systemd-boot is always supported

Johan-Liebert1:
It's currently only supported for composefs backend, not for ostree. The docs did not mention this fact

PR #1990, opened 2026-02-07
cli: Only allow some arguments if composefs-backend is true

Johan-Liebert1:
Currently the cli accepts `--insecure`, `--bootloader`, `--uki-addon` for ostree installs as well, which is not quite right as in the best case scenario they won't do anything and in the worst case scenario they might leave the install in undefined state

Also, update the bootloader docs to specify systemd-boot is only
available for composefs backend

Xref: https://github.com/bootc-dev/bootc/issues/1989

### DIFF
--- a/crates/lib/src/install.rs
+++ b/crates/lib/src/install.rs
@@ -383,18 +383,18 @@ pub(crate) struct InstallComposefsOpts {
     pub(crate) composefs_backend: bool,
 
     /// Make fs-verity validation optional in case the filesystem doesn't support it
-    #[clap(long, default_value_t)]
+    #[clap(long, default_value_t, requires = "composefs_backend")]
     #[serde(default)]
     pub(crate) insecure: bool,
 
     /// The bootloader to use.
-    #[clap(long)]
+    #[clap(long, requires = "composefs_backend")]
     #[serde(default)]
     pub(crate) bootloader: Option<Bootloader>,
 
     /// Name of the UKI addons to install without the ".efi.addon" suffix.
     /// This option can be provided multiple times if multiple addons are to be installed.
-    #[clap(long)]
+    #[clap(long, requires = "composefs_backend")]
     #[serde(default)]
     pub(crate) uki_addon: Option<Vec<String>>,
 }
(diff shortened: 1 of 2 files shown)


crates/lib/src/install.rs
```diff
@@ -383,18 +383,18 @@ pub(crate) struct InstallComposefsOpts {
     pub(crate) composefs_backend: bool,
 
     /// Make fs-verity validation optional in case the filesystem doesn't support it
-    #[clap(long, default_value_t)]
+    #[clap(long, default_value_t, requires = "composefs_backend")]
     #[serde(default)]
     pub(crate) insecure: bool,
```
cgwalters, line 388: While we're here...(not part of this PR but as followup) I would like to propose renaming this to something like `--allow-missing-fsverity` or something.

docs/src/bootloaders.md
```diff
@@ -13,6 +13,8 @@ Currently, `bootc` only runs `bootupd` during the installation process. It does
 
 ## systemd-boot
 
+NOTE: systemd-boot is only supported for Composefs Backend and not for Ostree
```
cgwalters, line 16: How about just

> systemd-boot is only supported by the composefs backend, and not the (default) ostree backend.

docs/src/bootloaders.md
```diff
@@ -13,6 +13,8 @@ Currently, `bootc` only runs `bootupd` during the installation process. It does
...
+
 If bootupd is not present in the input container image, then systemd-boot will be used
 by default (except on s390x).
```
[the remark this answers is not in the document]
cgwalters: Yeah, but not new with this PR

crates/lib/src/install.rs
```diff
@@ -1928,8 +1914,6 @@ fn installation_complete() {
 #[context("Installing to disk")]
 #[cfg(feature = "install-to-disk")]
 pub(crate) async fn install_to_disk(mut opts: InstallToDiskOpts) -> Result<()> {
-    opts.validate()?;
```
cgwalters, line 1931: I'm fine with this, however I think we could easily change this to be `anyhow::ensure!()` instead - yes it's double checking but probably worth doing.

crates/lib/src/install.rs
Johan-Liebert1: Sounds good. I'll update this when finishing up https://github.com/bootc-dev/bootc/pull/1913#discussion_r2786891300

crates/lib/src/install.rs
Johan-Liebert1: In agreement. The error that clap provides when passing, say `--insecure`, without the composefs backend flag is a bit misleading as it says

```
error: the following required arguments were not provided:
  --composefs-backend
```

imo, we'd want it so say something like `--insecure is only available for --composefs-backend` or something like that. What do you think?

crates/lib/src/install.rs
Johan-Liebert1: Done here https://github.com/bootc-dev/bootc/pull/2004

PR #1995, opened 2026-02-11
upgrade: Add pre-flight disk space check

shi2wei3:
Extends PR #1245's approach to all bootc upgrade modes that download layers (default, --apply, --download-only). Moves check_disk_space() to deploy.rs for reuse by both install and upgrade operations.

This prevents wasted bandwidth and provides immediate feedback when insufficient disk space is available, matching the install behavior.

Related: BIFROST-1088

Assisted-by: AI

### DIFF
--- a/crates/lib/src/deploy.rs
+++ b/crates/lib/src/deploy.rs
@@ -4,6 +4,7 @@
 
 use std::collections::HashSet;
 use std::io::{BufRead, Write};
+use std::os::fd::AsFd;
 use std::process::Command;
 
 use anyhow::{Context, Result, anyhow};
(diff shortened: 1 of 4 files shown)


shi2wei3:
@cgwalters @ckyrouac Would you mind help me to review it, since this MR is heavily based on yours.

crates/lib/src/deploy.rs
```diff
@@ -361,6 +362,31 @@ pub(crate) async fn prune_container_store(sysroot: &Storage) -> Result<()> {
...
+pub(crate) fn check_disk_space(
+    repo_fd: impl AsFd,
+    image_meta: &PreparedImportMeta,
+    imgref: &ImageReference,
+) -> Result<()> {
+    let stat = rustix::fs::fstatvfs(repo_fd)?;
+    let bytes_avail: u64 = stat.f_bsize * stat.f_bavail;
+    tracing::trace!("bytes_avail: {bytes_avail}");
+
+    if image_meta.bytes_to_fetch > bytes_avail {
```
cgwalters: Note that ostree has a slightly stricter check and has a configurable `min-free-space` in the repository: the rationale is that we basically never want to get close to filling up a disk with an OS update.

I'm OK with this as is, but how about just a 
```
// TODO consider also syncing with ostree min-free-space
```

crates/lib/src/deploy.rs
```diff
@@ -658,6 +684,9 @@ pub(crate) async fn pull(
             Ok(existing)
         }
         PreparedPullResult::Ready(prepared_image_meta) => {
+            // Check disk space before attempting to pull
+            check_disk_space(repo.dfd_borrow(), &prepared_image_meta, imgref)?;
```
cgwalters: One thing of course here is this doesn't apply to the composefs backend.

Hmm, I wonder if container-libs has a similar check? If we get to https://github.com/bootc-dev/bootc/issues/20 then the check would need to be there.

cgwalters (commented):
Looks sane to me. BTW I think it'd be relatively easy to have an integration test for this that makes a fake image in an `oci:` directory that claims to have a very large layer or so and verify we fail before we try to fetch the layers

crates/lib/src/deploy.rs
shi2wei3: Implemented now.

crates/lib/src/deploy.rs
shi2wei3: I didn't find any with AI's help, so added separated fn for composefs now.

tmt/tests/booted/test-upgrade-preflight-disk-check.nu
```diff
@@ -0,0 +1,66 @@
...
+
+    # OCI image manifest pointing to the real config + one fake large layer
+    let manifest_content = $'{"schemaVersion":2,"mediaType":"application/vnd.oci.image.manifest.v1+json","config":{"mediaType":"application/vnd.oci.image.config.v1+json","digest":"sha256:($config_digest)","size":($config_size)},"layers":[{"mediaType":"application/vnd.oci.image.layer.v1.tar+gzip","digest":"sha256:($fake_layer_digest)","size":($fake_layer_size)}]}'
+    let manifest_digest = $manifest_content | hash sha256
+    let manifest_size = ($manifest_content | str length)
+
+    # Write manifest blob
+    $manifest_content | save $"($td)/blobs/sha256/($manifest_digest)"
+
+    # OCI layout marker
```
cgwalters, line 48: This is all 100% fine however...we actually do have a crate for this in https://docs.rs/ocidir/latest/ocidir/ and just noting if our tests were in Rust it'd be trivial to use there.

tmt/tests/booted/test-upgrade-preflight-disk-check.nu
```diff
@@ -0,0 +1,66 @@
...
+    let config_size = ($config_content | str length)
+
+    # Write config blob
+    mkdir $"($td)/blobs/sha256"
+    $config_content | save $"($td)/blobs/sha256/($config_digest)"
+
+    # Fake layer: a digest that points to a non-existent blob is fine because
+    # the preflight check reads the declared size from the manifest only.
+    let fake_layer_digest = "0000000000000000000000000000000000000000000000000000000000000000"
+    let fake_layer_size = 999_999_999_999_999  # ~999 TiB — will never fit on disk
```
cgwalters, line 38: At least, "never" in our CI tests and it'd be pretty wild to use something that large for a root volume.

PR #2000, opened 2026-02-11
Blockdev cleanup

ckyrouac:
This was a lot of back and forth with claude. Pushing this for some feedback.

---

Prep for supporting installing to devices with multiple parents (e.g. RAID, LVM), see #1911. The core change is refactoring bootc-blockdev to use `lsblk` instead of `sfdisk` to resolve the device tree. See the individual commits for more details.

### DIFF
--- a/crates/blockdev/src/blockdev.rs
+++ b/crates/blockdev/src/blockdev.rs
@@ -1,14 +1,11 @@
-use std::collections::HashMap;
 use std::env;
 use std::path::Path;
 use std::process::{Command, Stdio};
-use std::sync::OnceLock;
 
 use anyhow::{Context, Result, anyhow};
 use camino::{Utf8Path, Utf8PathBuf};
 use cap_std_ext::cap_std::fs::Dir;
 use fn_error_context::context;
-use regex::Regex;
 use serde::Deserialize;
 
 use bootc_utils::CommandRunExt;
@@ -56,10 +53,15 @@ pub struct Device {
 }
 
 impl Device {
-    #[allow(dead_code)]
     // RHEL8's lsblk doesn't have PATH, so we do it
     pub fn path(&self) -> String {
         self.path.clone().unwrap_or(format!("/dev/{}", &self.name))
+    }
+
+    /// Alias for path() for compatibility
+    #[allow(dead_code)]
+    pub fn node(&self) -> String {
+        self.path()
     }
 
     #[allow(dead_code)]
@@ -254,122 +256,6 @@ pub fn list_dev_by_dir(dir: &Dir) -> Result<Device> {
     list_dev(&Utf8PathBuf::from(&fsinfo.source))
 }
 
-#[derive(Debug, Deserialize)]
-struct SfDiskOutput {
-    partitiontable: PartitionTable,
-}
-
-#[derive(Debug, Deserialize)]
-#[allow(dead_code)]
-pub struct Partition {
-    pub node: String,
-    pub start: u64,
-    pub size: u64,
-    #[serde(rename = "type")]
-    pub parttype: String,
-    pub uuid: Option<String>,
-    pub name: Option<String>,
-    pub bootable: Option<bool>,
-}
-
-#[derive(Debug, Deserialize, PartialEq, Eq)]
-#[serde(rename_all = "kebab-case")]
-pub enum PartitionType {
-    Dos,
-    Gpt,
-    Unknown(String),
-}
-
-#[derive(Debug, Deserialize)]
-#[allow(dead_code)]
-pub struct PartitionTable {
-    pub label: PartitionType,
-    pub id: String,
-    pub device: String,
-    // We're not using these fields
-    // pub unit: String,
-    // pub firstlba: u64,
-    // pub lastlba: u64,
-    // pub sectorsize: u64,
-    pub partitions: Vec<Partition>,
-}
-
-impl PartitionTable {
-    /// Find the partition with the given device name
-    #[allow(dead_code)]
-    pub fn find<'a>(&'a self, devname: &str) -> Option<&'a Partition> {
-        self.partitions.iter().find(|p| p.node.as_str() == devname)
-    }
-
-    pub fn path(&self) -> &Utf8Path {
-        self.device.as_str().into()
-    }
-
-    // Find the partition with the given offset (starting at 1)
-    #[allow(dead_code)]
-    pub fn find_partno(&self, partno: u32) -> Result<&Partition> {
-        let r = self
-            .partitions
-            .get(partno.checked_sub(1).expect("1 based partition offset") as usize)
-            .ok_or_else(|| anyhow::anyhow!("Missing partition for index {partno}"))?;
-        Ok(r)
-    }
-
-    /// Find the partition with the given type UUID (case-insensitive).
-    ///
-    /// Partition type UUIDs are compared case-insensitively per the GPT specification,
-    /// as different tools may report them in different cases.
-    pub fn find_partition_of_type(&self, uuid: &str) -> Option<&Partition> {
-        self.partitions.iter().find(|p| p.parttype_matches(uuid))
-    }
-
-    /// Find the partition with bootable is 'true'.
-    pub fn find_partition_of_bootable(&self) -> Option<&Partition> {
-        self.partitions.iter().find(|p| p.is_bootable())
-    }
-
-    /// Find the esp partition.
-    pub fn find_partition_of_esp(&self) -> Result<Option<&Partition>> {
-        match &self.label {
-            PartitionType::Dos => Ok(self.partitions.iter().find(|b| {
-                u8::from_str_radix(&b.parttype, 16)
-                    .map(|pt| ESP_ID_MBR.contains(&pt))
-                    .unwrap_or(false)
-            })),
-            PartitionType::Gpt => Ok(self.find_partition_of_type(ESP)),
-            _ => Err(anyhow::anyhow!("Unsupported partition table type")),
-        }
-    }
-}
-
-impl Partition {
-    #[allow(dead_code)]
-    pub fn path(&self) -> &Utf8Path {
-        self.node.as_str().into()
-    }
-
-    /// Check if this partition's type matches the given UUID (case-insensitive).
-    ///
-    /// Partition type UUIDs are compared case-insensitively per the GPT specification,
-    /// as different tools may report them in different cases.
-    pub fn parttype_matches(&self, uuid: &str) -> bool {
-        self.parttype.eq_ignore_ascii_case(uuid)
-    }
-
-    /// Check this partition's bootable property.
-    pub fn is_bootable(&self) -> bool {
-        self.bootable.unwrap_or(false)
-    }
-}
-
-#[context("Listing partitions of {dev}")]
-pub fn partitions_of(dev: &Utf8Path) -> Result<PartitionTable> {
-    let o: SfDiskOutput = Command::new("sfdisk")
-        .args(["-J", dev.as_str()])
-        .run_and_parse_json()?;
-    Ok(o.partitiontable)
-}
-
 pub struct LoopbackDevice {
     pub dev: Option<Utf8PathBuf>,
     // Handle to the cleanup helper process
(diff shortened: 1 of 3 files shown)


crates/blockdev/src/blockdev.rs
```diff
@@ -53,12 +53,43 @@ pub struct Device {
...
+    pub fn find_device_by_partno(&self, partno: u32) -> Result<&Device> {
+        self.children
+            .as_ref()
+            .ok_or_else(|| anyhow!("Device has no children"))?
+            .iter()
+            .find(|child| child.partn == Some(partno))
+            .ok_or_else(|| anyhow!("Missing partition for index {partno}"))
+    }
+
+    /// Find a child partition by partition type GUID.
```
cgwalters: Note for MBR it's a [partition type](https://en.wikipedia.org/wiki/Partition_type). I'd just `s/ GUID//` - or to be more accurate we could explain here that GPT is GUID, MBR is partition type code.

crates/blockdev/src/blockdev.rs
```diff
@@ -519,6 +550,13 @@ mod test {
             first_child.partuuid.as_deref().unwrap(),
             "3979e399-262f-4666-aabc-7ab5d3add2f0"
         );
+        // Verify find_device_by_partno works
```
cgwalters, line 502: This looks sane, but do we have a test JSON that's MBR? Would be good to add here if not

crates/blockdev/src/blockdev.rs
```diff
@@ -139,10 +150,48 @@ pub fn list_dev(dev: &Utf8Path) -> Result<Device> {
...
+        .arg(dev)
+        .log_debug()
+        .run_and_parse_json()?;
+
+    // The first device in inverse output is the target device itself,
+    // its children (in inverse mode) are the parent devices
+    let target = output.blockdevices.first_mut();
+
+    match target {
+        Some(target) => {
```
cgwalters: Minor style but how about:

```
let r = match output.blockdevices.as_slice() {
  [target, ...] => { ...; parents },
  [] => Vec::new();
}
Ok(r)
```

crates/lib/src/bootloader.rs
```diff
@@ -24,21 +23,6 @@ const BOOTUPD_UPDATES: &str = "usr/lib/bootupd/updates";
 // from: https://github.com/systemd/systemd/blob/26b2085d54ebbfca8637362eafcb4a8e3faf832f/man/systemd-boot.xml#L392
 const SYSTEMD_KEY_DIR: &str = "loader/keys";
 
-#[allow(dead_code)]
```
cgwalters, line 27: 👍 nice to see this duplicated code go away

crates/lib/src/bootloader.rs
```diff
@@ -133,16 +122,20 @@ pub(crate) fn install_via_bootupd(
         let mut bwrap_args = vec!["bootupctl"];
         bwrap_args.extend(bootupd_args);
 
+        // Collect partition paths first so they live long enough
+        let partition_paths: Vec<String> =
+            device.children.iter().flatten().map(|p| p.path()).collect();
```
cgwalters, line 127: We could have a `Vec<&str>` i think if we just moved this call up closer to the function entrypoint

crates/blockdev/src/blockdev.rs
```diff
@@ -1,22 +1,15 @@
...
 use cap_std_ext::cap_std::fs::Dir;
 use fn_error_context::context;
-use regex::Regex;
 use serde::Deserialize;
 
 use bootc_utils::CommandRunExt;
 
-/// EFI System Partition (ESP) on MBR
-/// Refer to <https://en.wikipedia.org/wiki/Partition_type>
-pub const ESP_ID_MBR: &[u8] = &[0x06, 0xEF];
```
cgwalters, line 17: Wait but are we sure we're covering the ESP-on-MBR in other places?

crates/blockdev/src/blockdev.rs
```diff
@@ -1,22 +1,15 @@
...
-
 /// EFI System Partition (ESP) for UEFI boot on GPT
 pub const ESP: &str = "c12a7328-f81f-11d2-ba4b-00a0c93ec93b";
```
cgwalters, line 20: While we're here I think we probably also want to have this one point to the DPS module's copy?

crates/blockdev/src/blockdev.rs
```diff
@@ -196,127 +189,26 @@ fn list_parents(dev: &Utf8Path) -> Result<Vec<Device>> {
             let mut parents: Vec<Device> = target.children.take().unwrap_or_default();
             for parent in parents.iter_mut() {
                 parent.backfill_missing()?;
+                // Recursively convert children to parents for the full chain
+                convert_inverse_children_to_parents(parent)?;
```
cgwalters: I think this is a bugfix that should be squashed into the prior commit conceptually

crates/blockdev/src/blockdev.rs
```diff
@@ -142,75 +142,104 @@ pub fn list_dev_by_dir(dir: &Dir) -> Result<Device> {
     list_dev(&Utf8PathBuf::from(&fsinfo.source))
 }
 
+/// List a device and build a complete tree with both parents and children populated.
```
cgwalters: I think ditto, this feels like a rewrite of a prior commit

crates/blockdev/src/blockdev.rs
```diff
@@ -142,75 +142,104 @@ pub fn list_dev_by_dir(dir: &Dir) -> Result<Device> {
...
+            populate_parents(child, Some(parent_ref.clone()))?;
         }
-        dev.parents = Some(children);
     }
+
     Ok(())
 }
 
+/// Find a device in the tree by its path.
+fn find_device_in_tree(root: &Device, target_path: &Utf8Path) -> Option<Device> {
```
cgwalters: This one looks trivial to generate some unit tests for too

crates/blockdev/src/blockdev.rs
```diff
@@ -120,10 +120,32 @@ impl Device {
         Ok(())
     }
 
+    // The "partn" column was added in util-linux 2.39, which is newer than
+    // what CentOS 9 / RHEL 9 ship (2.37).
```
cgwalters: Ohhhh yes I recall getting burned by a variant of this too 😢

util-linux basically doesn't rebase by default in RHEL so at this point in RHEL9 it's quite old.

But they will patch things if asked, so we could ask...

crates/blockdev/src/blockdev.rs
```diff
@@ -184,6 +184,37 @@ impl Device {
         }
     }
 
+    /// Walk the parent chain to find the root (whole disk) device.
+    ///
+    /// Returns the root device with its children (partitions) populated.
+    /// If this device is already a root device, returns a clone of `self`.
+    /// Fails if the device has multiple parents at any level.
+    pub fn root_disk(&self) -> Result<Device> {
```
cgwalters, line 211: But this function is unused in this commit, which seems surprising?

Probably could also have unit tests.

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -535,7 +533,7 @@ pub(crate) fn setup_composefs_bls_boot(
             cmdline.add_or_modify(&param);
 
             // Locate ESP partition device
-            let root_dev = bootc_blockdev::list_dev_by_dir(&storage.physical_root)?;
+            let root_dev = bootc_blockdev::list_dev_by_dir(&storage.physical_root)?.root_disk()?;
```
cgwalters, line 548: (Unrelated to this line of code but I'd argue to squash this with previous)

crates/lib/src/bootloader.rs
```diff
@@ -45,7 +45,7 @@ pub(crate) fn mount_esp_part(root: &Dir, root_path: &Utf8Path, is_ostree: bool)
         root
     };
 
-    let dev = bootc_blockdev::list_dev_by_dir(physical_root)?;
+    let dev = bootc_blockdev::list_dev_by_dir(physical_root)?.root_disk()?;
     if let Some(esp_dev) = dev.find_partition_of_type(bootc_blockdev::ESP) {
```
cgwalters, line 49: But in the end state, don't we want to handle multiple devices? it seems surprising after this series that we end up with calling a fallible `.root_disk()` in situations like this.

I think *this* particular case really wants to use a shared `find_esp_from_filesystem()` or so right? (Also I note here this is using `bootc_blockdev::ESP` but we need to handle MBR too so we really do need to have a find_esp function which handles either.

cgwalters:
Backing up at a high level what I think would make sense for a series is:

1. Pure bugfixes/cleanups like the "add missing ptnum for rhel9"
2. Consolidating a shared "find_esp_from_filesystem" and "find_esp_in_device" functions
3. Rewriting blockdev.rs (adding new APIs where needed), port things to use the new API?

That said it's possible that the 2. and 3. are too entangled.

crates/blockdev/src/blockdev.rs
ckyrouac: 👍 to unit tests

yea I need to clean up the commits some more. I tried to have claude help but it was a little aggressive on how many commits it created. For reference this function is used in future commits, e.g. 

https://github.com/bootc-dev/bootc/pull/2000/changes#diff-1604878878a88c4dcbd7f51444b69fc01a58e3ab48ade4e10b7a9ed05fa867f8R536

https://github.com/bootc-dev/bootc/pull/2000/changes#diff-1604878878a88c4dcbd7f51444b69fc01a58e3ab48ade4e10b7a9ed05fa867f8R1068

crates/lib/src/bootloader.rs
ckyrouac: root_disk() is sort of a stop-gap function. My current plan for multi device support is to gradually add support depending on the bootloader. So the followup commit to this will add multi device support for uefi and bios installations, but systemd-boot and zipl will continue to only support a single root disk.

ckyrouac:
> That said it's possible that the 2. and 3. are too entangled.

They definitely felt entangled as I was iterating on this but I should be able to mostly isolate them.

ckyrouac:
Updated with cleaned up commits and addressed review comments. Isolating 2 and 3 from https://github.com/bootc-dev/bootc/pull/2000#issuecomment-3887044075 was getting tricky so they are both in one commit.

crates/blockdev/src/blockdev.rs
```diff
@@ -62,6 +67,67 @@ impl Device {
...
+        match self.pttype.as_deref() {
+            Some("dos") => children
+                .iter()
+                .find(|child| {
+                    child
+                        .parttype
+                        .as_ref()
+                        .and_then(|pt| {
+                            let pt = pt.strip_prefix("0x").unwrap_or(pt);
+                            u8::from_str_radix(pt, 16).ok()
```
cgwalters, line 100: This is a little ugly...I think a cleaner way would be a custom type like `enum PartitionType { MBR(u16) GPT(uuid)} ` or so that we deserialize instead of a string.

But nonblocking.

crates/blockdev/src/blockdev.rs
```diff
@@ -62,6 +67,67 @@ impl Device {
...
+            // to GPT UUID matching which will simply not match MBR hex types.
+            Some("gpt") | None => self
+                .find_partition_of_type(ESP)
+                .ok_or_else(|| anyhow!("ESP not found in GPT partition table")),
+            Some(other) => Err(anyhow!("Unsupported partition table type: {other}")),
+        }
+    }
+
+    /// Find a child partition by partition number (1-indexed).
+    pub fn find_device_by_partno(&self, partno: u32) -> Result<&Device> {
```
cgwalters, line 115: `find_` makes me think this should return `Option`.

If we're returning a `Result` I'd call it something more like `require_child_by_partno` here "require" is emphasizing that it should exist.

crates/blockdev/src/blockdev.rs
```diff
@@ -103,6 +169,67 @@ impl Device {
...
+            .run_and_parse_json()?;
+
+        let device = output
+            .blockdevices
+            .into_iter()
+            .next()
+            .ok_or_else(|| anyhow!("no device output from lsblk --inverse for {path}"))?;
+
+        match device.children {
+            Some(mut children) if !children.is_empty() => {
```
cgwalters, line 196: `if let Some(children) = device.children.as_mut()`

cgwalters (approved):
I think this is OK but...it seems like it wouldn't be significantly harder to just take the next step and change how we find the ESP to "require unique ESP on any parent disk".

crates/blockdev/src/blockdev.rs
```diff
@@ -103,6 +171,67 @@ impl Device {
...
+            return list_dev(Utf8Path::new(&self.path()));
+        };
+        let mut current = parents;
+        loop {
+            anyhow::ensure!(
+                current.len() == 1,
+                "Device {} has multiple parents; cannot determine root disk",
+                self.path()
+            );
+            let mut parent = current.into_iter().next().unwrap();
```
cgwalters, line 223: We can drop the `unwrap()` here if we do like

```
match current {
  [] => 
  [parent] =>
  parents => anyhow::bail!("Device has {} parents", parents.len());
```

or so.

crates/blockdev/src/blockdev.rs
```diff
@@ -48,20 +48,88 @@ pub struct Device {
...
+            .iter()
+            .find(|child| child.partn == Some(partno))
+            .ok_or_else(|| anyhow!("Missing partition for index {partno}"))
+    }
+
+    /// Re-query this device's information from lsblk, updating all fields.
+    /// This is useful after partitioning when the device's children have changed.
+    pub fn refresh(&mut self) -> Result<()> {
+        let path = self.path();
+        let new_device = list_dev(Utf8Path::new(&path))?;
```
cgwalters, line 128: here and elsewhere it should also work to just `list_dev(path.into())`

crates/blockdev/src/blockdev.rs
```diff
@@ -121,120 +250,10 @@ pub fn list_dev(dev: &Utf8Path) -> Result<Device> {
         .ok_or_else(|| anyhow!("no device output from lsblk for {dev}"))
 }
 
-#[derive(Debug, Deserialize)]
-struct SfDiskOutput {
```
cgwalters, line 125: Also definitely nice to have this go away!

cgwalters (approved):

PR #2002, opened 2026-02-11
Update ADOPTERS.md

mohan-shash:
Adding CIQ, rocky linux to the adopters list as per the recent announcement.


cgwalters:
Ah mid-air collision with https://github.com/bootc-dev/bootc/pull/2003 - since that one came from one of those team members I think it seems more reasonable to merge that one?

mohan-shash:
absolutely... i will close this.

PR #1986, opened 2026-02-06
ci: Update for Fedora 44 branching from rawhide

jeckersb:
Fedora 44 has branched from rawhide, so update the CI matrix
accordingly:

- Point fedora-44 images to :44 instead of :rawhide
- Add fedora-45 as the new rawhide
- Add fedora-44 to integration test matrix (BZ 2429501 is fixed)
- Add fedora-44 to Packit COPR and TMT test targets
- Re-enable rawhide Packit TMT tests (BZ 2429501 is fixed)

Closes: #1985
Assisted-by: OpenCode (claude-opus-4-6)
Signed-off-by: John Eckersberg <jeckersb@redhat.com>

PR #2004, opened 2026-02-12
Composefs: handle fs-verity disabled installs/updates, fix rollback bug

Johan-Liebert1:
- Fix unqueuing rollback for composefs backend
We were simply checking to booted system's verity and simply setting the
corresponding boot entry as the secondary boot entry, even if rollback
was already queued. Update the code to actually consider the bootloader
entries as the source of truth, similar to what ostree does

- Add test for rollback

- composefs/tests: Add tests for filesystems
Add filesystems ext4 and xfs in github CI matrix so that we test systems
with and without fs-verity enabled

- cli: Change `insecure` param to `allow_missing_fsverity`
`allow_missing_fsverity` conveys the intent in a much better way than
just `insecure`

- ukify: Accept `allow-missing-verity` param

- composefs: Disable auto fsverity enforcement on unsupported fs
On filesystems that do not support fsverity, we now do not auto enforce
fsverity. On filesystems that do support it, we look at the
`--allow-missing-verity` option, if passed in by the user, and if true,
fs-verity enforcement is disabled and vice-versa.
In case of UKIs, we take a different approach of looking at the UKI
cmdline beforehand and checking if the `composefs=` parameter has `?` or
not. This approach, though valid, fails in a few cases, viz,
    - The cmdline is in a UKI addon
    - The target image is not the one we're currently running in

Fixes: https://github.com/bootc-dev/bootc/issues/2010

### DIFF
--- a/crates/lib/src/install.rs
+++ b/crates/lib/src/install.rs
@@ -201,7 +201,7 @@ use crate::store::Storage;
 use crate::task::Task;
 use crate::utils::sigpolicy_from_opt;
 use bootc_kernel_cmdline::{INITRD_ARG_PREFIX, ROOTFLAGS, bytes, utf8};
-use bootc_mount::{Filesystem, inspect_filesystem};
+use bootc_mount::Filesystem;
 use composefs::fsverity::FsVerityHashValue;
 
 /// The toplevel boot directory
@@ -2353,22 +2353,6 @@ pub(crate) async fn install_to_filesystem(
         target_path
     );
 
-    let fs_inspect = inspect_filesystem(&opts.filesystem_opts.root_path)?;
-
-    // Gather global state, destructuring the provided options.
-    // IMPORTANT: We might re-execute the current process in this function (for SELinux among other things)
-    // IMPORTANT: and hence anything that is done before MUST BE IDEMPOTENT.
-    // IMPORTANT: In practice, we should only be gathering information before this point,
-    // IMPORTANT: and not performing any mutations at all.
-    let state = prepare_install(
-        opts.config_opts,
-        opts.source_opts,
-        opts.target_opts,
-        opts.composefs_opts,
-        Some(fs_inspect.fstype.as_str().try_into()?),
-    )
-    .await?;
-
     // And the last bit of state here is the fsopts, which we also destructure now.
     let mut fsopts = opts.filesystem_opts;
 
@@ -2390,6 +2374,7 @@ pub(crate) async fn install_to_filesystem(
     }
 
     let target_root_path = fsopts.root_path.clone();
+
     // Get a file descriptor for the root path /target
     let target_rootfs_fd =
         Dir::open_ambient_dir(&target_root_path, cap_std::ambient_authority())
@@ -2410,11 +2395,6 @@ pub(crate) async fn install_to_filesystem(
         if !st.is_dir() {
             anyhow::bail!("Not a directory: {root_path}");
         }
-    }
-
-    // Check to see if this happens to be the real host root
-    if !fsopts.acknowledge_destructive {
-        warn_on_host_root(&target_rootfs_fd)?;
     }
 
     // If we're installing to an ostree root, then find the physical root from
@@ -2446,6 +2426,28 @@ pub(crate) async fn install_to_filesystem(
         target_rootfs_fd.try_clone()?
     };
 
+    // Gather data about the root filesystem
+    let inspect = bootc_mount::inspect_filesystem(&fsopts.root_path)?;
+
+    // Gather global state, destructuring the provided options.
+    // IMPORTANT: We might re-execute the current process in this function (for SELinux among other things)
+    // IMPORTANT: and hence anything that is done before MUST BE IDEMPOTENT.
+    // IMPORTANT: In practice, we should only be gathering information before this point,
+    // IMPORTANT: and not performing any mutations at all.
+    let state = prepare_install(
+        opts.config_opts,
+        opts.source_opts,
+        opts.target_opts,
+        opts.composefs_opts,
+        Some(inspect.fstype.as_str().try_into()?),
+    )
+    .await?;
+
+    // Check to see if this happens to be the real host root
+    if !fsopts.acknowledge_destructive {
+        warn_on_host_root(&target_rootfs_fd)?;
+    }
+
     match fsopts.replace {
         Some(ReplaceMode::Wipe) => {
             let rootfs_fd = rootfs_fd.try_clone()?;
@@ -2458,9 +2460,6 @@ pub(crate) async fn install_to_filesystem(
         }
         None => require_empty_rootdir(&rootfs_fd)?,
     }
-
-    // Gather data about the root filesystem
-    let inspect = bootc_mount::inspect_filesystem(&fsopts.root_path)?;
 
     // We support overriding the mount specification for root (i.e. LABEL vs UUID versus
     // raw paths).


Johan-Liebert1:
Requires https://github.com/bootc-dev/bcvk/pull/205

crates/initramfs/src/lib.rs
```diff
@@ -262,9 +262,13 @@ fn open_root_fs(path: &Path) -> Result<OwnedFd> {
 /// * name     - Name of the EROFS image to be mounted
 /// * insecure - Whether fsverity is optional or not
 #[context("Mounting composefs image")]
-pub fn mount_composefs_image(sysroot: &OwnedFd, name: &str, insecure: bool) -> Result<OwnedFd> {
+pub fn mount_composefs_image(
+    sysroot: &OwnedFd,
+    name: &str,
+    allow_missing_fsverity: bool,
+) -> Result<OwnedFd> {
```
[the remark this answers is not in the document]
Johan-Liebert1: nice catch. Done

Johan-Liebert1:
/gemini Where is this "However, a critical path traversal vulnerability has been identified in the UKI/addon installation logic, which could allow a malicious container image to overwrite arbitrary files on the host system during installation"?

Johan-Liebert1:
> However, a critical path traversal vulnerability has been identified in the UKI/addon installation logic, which could allow a malicious container image to overwrite arbitrary files on the host system during installation

I'm not really sure where Gemini found this

cgwalters:
Same comment on the matrix effect here https://github.com/bootc-dev/bootc/pull/1986#pullrequestreview-3764643471

I can try to take a look at having some of the tests become postsubmits.

So this blocks on https://github.com/bootc-dev/bcvk/pull/205 ?

But honestly...why don't we just change bootc by default to not hard require fsverity unless one is doing a sealed UKI?

cgwalters:
> /gemini Where is this "However, a critical path traversal vulnerability has been identified in the UKI/addon installation logic, which could allow a malicious container image to overwrite arbitrary files on the host system during installation"?

In general asking LLMs to explain themselves often fails; but I'm not too surprised at this hallucination, what we're doing is pretty special and it's probably hard for the model to reconcile that we're effectively doing privileged containers here.

Johan-Liebert1:
> I can try to take a look at having some of the tests become postsubmits.

The tests themselves, or some matrix combinations? Is the concern here that CI is taking too long, or do we not want to have so many runners at the same time? If the former is the concern, then we could (probably, in theory) separate out the container + disk image creation from the actual matrix. But I guess the actual stuff that's slow IS the testing itself

> So this blocks on https://github.com/bootc-dev/bcvk/pull/205 ?

yes

> But honestly...why don't we just change bootc by default to not hard require fsverity unless one is doing a sealed UKI?

So instead of `allow-missing-verity` we have something like `enforce-verity` as an option? Feels a bit counter intuitive to me. People could, in theory, use the same container for secure and inscure installs by putting the actual cmdline inside of an addon

Dockerfile
```diff
@@ -186,8 +187,15 @@ RUN --network=none --mount=type=tmpfs,target=/run --mount=type=tmpfs,target=/tmp
     --mount=type=bind,from=packaging,src=/,target=/run/packaging \
     --mount=type=bind,from=base-penultimate,src=/,target=/run/target <<EORUN
 set -xeuo pipefail
+
+allow_missing_verity=false
+
+if [[ filesystem == "xfs" ]]; then
```
Johan-Liebert1: Ah, this is completely wrong. I was wondering how the sealed xfs test passed

cgwalters:
> So instead of allow-missing-verity we have something like enforce-verity as an option? Feels a bit counter intuitive to me. People could, in theory, use the same container for secure and inscure installs by putting the actual cmdline inside of an addon

If the image is not sealed, I think we just default to enabling fsverity if it's available. WDYT?

(I have some WIP patches for composefs-rs that would make it easier to transition a repository from no-fsverity -> fsverity)

Johan-Liebert1:
> If the image is not sealed, I think we just default to enabling fsverity if it's available. WDYT?

did you mean disable, because `If the image is not sealed, I think we just default to enabling fsverity if it's available` is what we do right now? Maybe we can just look at the filesystem and enable/disable depending upon if the fs supports it or not. The issue arises with remote container pulls (if we're not pulling the same container image that we're running in), then we wouldn't know what to do as the user might be expecting fsverity but get the opposite. This issue only arises with UKI as we don't control the cmdline in that case

Johan-Liebert1:
@cgwalters Made the changes about enabling/disabling fs-verity automatically based on UKI/Filesystems. I'm not 100% sure about https://github.com/bootc-dev/bootc/pull/2004/changes/bdfd2d34c681c6344c056f27051b6b30e68f31f5 commit. We are not to make any mutations BEFORE the `prepare_install` function, and I made sure this change isn't doing that, but worth a closer look

crates/lib/src/install/config.rs
```diff
@@ -32,6 +32,25 @@ impl std::fmt::Display for Filesystem {
...
+            "ext4" => Ok(Self::Ext4),
+            "btrfs" => Ok(Self::Btrfs),
+            other => anyhow::bail!("Unknown filesystem: {}", other),
+        }
+    }
+}
+
+impl Filesystem {
+    pub(crate) fn supports_fsverity(&self) -> bool {
+        matches!(self, Self::Ext4 | Self::Btrfs)
```
cgwalters, line 50: I think this is OK for now but in the medium term I'd prefer we just try to do the ioctl to enable it and handle `ENOTSUPP`

crates/lib/src/install.rs
```diff
@@ -1576,12 +1578,15 @@ async fn prepare_install(
     };
     tracing::debug!("Target image reference: {target_imgref}");
 
-    let composefs_required = if let Some(root) = target_rootfs.as_ref() {
-        crate::kernel::find_kernel(root)?
-            .map(|k| k.kernel.unified)
-            .unwrap_or(false)
+    let (composefs_required, kernel) = if let Some(root) = target_rootfs.as_ref() {
+        let kernel = crate::kernel::find_kernel(root)?;
```
cgwalters, line 1582: It feels cleaner to just unconditionally look for the kernel.

crates/lib/src/install.rs
```diff
@@ -1656,6 +1661,59 @@ async fn prepare_install(
...
+            .and_then(|c| c.filesystem_root())
+            .and_then(|r| r.fstype))
+        .ok_or_else(|| anyhow::anyhow!("No root filesystem specified"))?;
+
+    let mut is_uki = false;
+
+    // For composefs backend, automatically disable fs-verity hard requirement if the
+    // filesystem doesn't support it
+    //
+    // If we have a sealed UKI on our hands, then we can assume that user wanted fs-verity so
```
cgwalters, line 1676: I guess to me "sealed UKI" really also means "signed"...and when one takes it out, also Secure Boot being enabled.

How about just adjusting this comment to say "If we have a UKI...".

crates/lib/src/install.rs
```diff
@@ -1656,6 +1661,59 @@ async fn prepare_install(
...
+        composefs_options.allow_missing_verity = !root_filesystem.supports_fsverity();
+
+        tracing::debug!(
+            "Missing fsverity {}",
+            if composefs_options.allow_missing_verity {
+                "allowed"
+            } else {
+                "not allowed"
+            }
+        );
```
cgwalters, line 1714: This feels very "manual". tracing supports structured logging https://docs.rs/tracing/latest/tracing/index.html#recording-fields, how about
```
tracing::debug!("composefs install", allow_missing_verity = composefs_options.allow_missing_verity, uki = is_uki)
```
or so?

crates/lib/src/install.rs
```diff
@@ -1656,6 +1661,59 @@ async fn prepare_install(
...
+
+    let mut is_uki = false;
+
+    // For composefs backend, automatically disable fs-verity hard requirement if the
+    // filesystem doesn't support it
+    //
+    // If we have a sealed UKI on our hands, then we can assume that user wanted fs-verity so
+    // we hard require it in that particular case
+    //
+    // NOTE: This isn't really 100% accurate 100% of the time as the cmdline can be in an addon
```
cgwalters, line 1679: If the cmdline is an add-on then it's still a UKI right?

crates/lib/src/kernel.rs
```diff
@@ -31,8 +34,11 @@ pub(crate) struct Kernel {
 /// UKI kernels only have the single PE binary, whereas
 /// traditional "vmlinuz" kernels have distinct kernel and
 /// initramfs.
-pub(crate) enum KernelPath {
-    Uki(Utf8PathBuf),
+pub(crate) enum KernelType {
+    Uki {
+        path: Utf8PathBuf,
+        allow_missing_fsverity: bool,
```
cgwalters, line 40: Might be cleaner to have the full kernel command line here? Or just the composefs option.

I can see `bootc container inspect` wanting to output the target composefs digest.

crates/lib/src/install.rs
```diff
@@ -1656,6 +1661,59 @@ async fn prepare_install(
...
+            }
+
+            crate::kernel::KernelType::Vmlinuz { .. } => {}
+        },
+
+        None => {}
+    }
+
+    // If `--allow-missing-verity` is already passed via CLI, don't modify
+    if composefs_options.composefs_backend && !composefs_options.allow_missing_verity && !is_uki {
```
cgwalters, line 1704: It's a bit confusing because passing `--allow-missing-verity` on the CLI can only lead to failures at boot time if the target kernel *is* sealed right?

At least...if Secure Boot is enabled.

So that's really it right? The only use case for `--allow-missing-verity` is installing a sealed UKI (without `?` in `composefs=`) on a system without fsverity enabled.

But because we're not *changing* the UKI it's still going to fail at *runtime* in that scenario no?

Wouldn't these use cases just need to disable Secure Boot anyways?

---

So...I'm kind of questioning if we need the CLI option at all. The *real* fix here seems to me to be honoring the `?` in the embedded cmdline in the UKI right?

And on that topic...aren't we then missing:

- support for e.g. `bootc container ukify --allow-missing-verity` that changes the embedded cmdline
- Changing our build infrastructure to pass that if we're targeting "non-UKI composefs backend"

?

crates/lib/src/kernel.rs
```diff
@@ -67,12 +73,48 @@ pub(crate) fn find_kernel(root: &Dir) -> Result<Option<KernelInternal>> {
     // First, try to find a UKI
     if let Some(uki_path) = find_uki_path(root)? {
         let version = uki_path.file_stem().unwrap_or(uki_path.as_str()).to_owned();
+
+        let uki = root.read(&uki_path).context("Reading UKI")?;
+
+        // Best effort to check for composefs=?verity in the UKI cmdline
+        let cmdline = composefs_boot::uki::get_section(&uki, ".cmdline");
```
cgwalters, line 80: I think it'd be cleaner to have `get_section_optional` that returns `Result<Option<>>`

crates/lib/src/kernel.rs
```diff
@@ -67,12 +73,48 @@ pub(crate) fn find_kernel(root: &Dir) -> Result<Option<KernelInternal>> {
...
+
+        let allow_missing_fsverity = match cmdline {
+            Some(Ok(cmdline)) => {
```
cgwalters, line 83: Then we wouldn't need this complicated `match`

crates/lib/src/install.rs
Johan-Liebert1: We can't do that if we are not installing the image we're running in, right?

crates/lib/src/install.rs
Johan-Liebert1: Okay, yeah this is definitely nicer. Also, a lot of journal logging is missing for the composefs route. I'll create a tracker for that

crates/lib/src/install.rs
Johan-Liebert1: yes, but right now we've kinda "paused" the UKI Addons development, and we don't collect the addons here to check for cmdline in them. Someone could just as easily use the same UKI image for fs-verity enabled/disabled installations using addons

crates/lib/src/install.rs
Johan-Liebert1: > So...I'm kind of questioning if we need the CLI option at all. The real fix here seems to me to be honoring the ? in the embedded cmdline in the UKI right?

I think you're right. We probably don't need the CLI option now. My thinking for leaving it in was for the case when we pull a remote image to install, and find conflicting fsverity options in the filesystems (say xfs), vs in the cmdline (say without the `?`). Currently in this case we just have a warning, but we should error out. 

Regarding this topic, we had a discussion where we were talking about a small subset of users who'd want the switch to composefs backend but not want to have the (albeit very minute) overhead of fs-verity. So, this option would be useful for them? Again though, this would only be for Type1 bootloaders


> support for e.g. bootc container ukify --allow-missing-verity that changes the embedded cmdline

I've added this [here](https://github.com/bootc-dev/bootc/pull/2004/changes/1481c9b417dd7daf9c0526c22e6a4b5a6d864565)

crates/lib/src/kernel.rs
Johan-Liebert1: I remember doing something similar here https://github.com/composefs/composefs-rs/pull/198, but reverted the change

crates/lib/src/kernel.rs
Johan-Liebert1: > Might be cleaner to have the full kernel command line here?

Updated to have the full cmdline (haven't pushed changes yet)

cgwalters:
FTR we had a live chat on this and takeaways:

- For the "unsealed UKI" case we should also not sign the UKI (and we should also disable secure boot) because a signed UKI using `?` is basically not providing much security. We'll also rename the CI check to be "unsealed-uki"
- Something else I forget but will try to recover from transcripts

crates/lib/src/install.rs
cgwalters: Yeah you're right.

Hmmm...maybe this is an argument that we should include *labels* in the image metadata for this state. `containers.bootc=sealed-uki` or so?

PR #2005, opened 2026-02-12
Blockdev bugfixes

ckyrouac:
A couple small fixes that are prep for multi device support. Related to #1911 and #2000.

### DIFF
--- a/crates/blockdev/src/blockdev.rs
+++ b/crates/blockdev/src/blockdev.rs
@@ -33,6 +33,8 @@ pub struct Device {
     pub partlabel: Option<String>,
     pub parttype: Option<String>,
     pub partuuid: Option<String>,
+    /// Partition number (1-indexed). None for whole disk devices.
+    pub partn: Option<u32>,
     pub children: Option<Vec<Device>>,
     pub size: u64,
     #[serde(rename = "maj:min")]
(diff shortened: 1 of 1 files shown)


crates/blockdev/src/blockdev.rs
```diff
@@ -82,10 +84,32 @@ impl Device {
...
+    // The "partn" column was added in util-linux 2.39, which is newer than
+    // what CentOS 9 / RHEL 9 ship (2.37).
+    fn backfill_partn(&mut self) -> Result<()> {
+        let Some(majmin) = self.maj_min.as_deref() else {
+            return Ok(());
+        };
+        let sysfs_partn_path = format!("/sys/dev/block/{majmin}/partition");
+        if Utf8Path::new(&sysfs_partn_path).try_exists()? {
+            let partn = std::fs::read_to_string(&sysfs_partn_path)
+                .with_context(|| format!("Reading {sysfs_partn_path}"))?;
```
cgwalters: Minor but I like using https://docs.rs/cap-std-ext/latest/cap_std_ext/dirext/trait.CapStdExtDirExt.html#tymethod.open_optional for things like this.

Though I guess what we should also do (for the many cases that operate on an ambient `/`) is have an extension trait wrapper somewhere like exists in composefs-rs https://github.com/containers/composefs-rs/blob/fc1d4e27ef054b9ececeeb58c985c77727ceaff4/crates/composefs/src/util.rs#L119

crates/blockdev/src/blockdev.rs
```diff
@@ -82,10 +84,32 @@ impl Device {
         Ok(())
     }
 
+    // The "partn" column was added in util-linux 2.39, which is newer than
+    // what CentOS 9 / RHEL 9 ship (2.37).
+    fn backfill_partn(&mut self) -> Result<()> {
+        let Some(majmin) = self.maj_min.as_deref() else {
```
cgwalters: shouldn't this be `.partn`?

ckyrouac:
refactored this to use a shared function to backfill both `start` and `partn` which should make the code's intentions more obvious.

crates/blockdev/src/blockdev.rs
ckyrouac: this was refactored in the latest push, but `majmin` is part of the path to locate the partition sysfs partition.

PR #2006, opened 2026-02-12
store: Centralize composefs directory creation with mode 0700

jeckersb:
The install-time composefs directory creation in repo.rs used
create_dir_all() which relies on the process umask for permissions,
potentially creating /sysroot/composefs with overly permissive modes
and leaking information.

Centralize the directory creation into a new ensure_composefs_dir()
helper in store/mod.rs that explicitly sets mode 0700. Both the
install-time path (repo.rs) and the runtime lazy-init path
(Storage::get_ensure_composefs) now use this single helper.

Also removes #[allow(dead_code)] from COMPOSEFS_MODE since it is now
actively used, and adds a unit test verifying the directory permissions
and idempotency.

### DIFF
--- a/crates/lib/src/bootc_composefs/repo.rs
+++ b/crates/lib/src/bootc_composefs/repo.rs
@@ -26,9 +26,7 @@ pub(crate) async fn initialize_composefs_repository(
 ) -> Result<(String, impl FsVerityHashValue)> {
     let rootfs_dir = &root_setup.physical_root;
 
-    rootfs_dir
-        .create_dir_all("composefs")
-        .context("Creating dir composefs")?;
+    crate::store::ensure_composefs_dir(rootfs_dir)?;
 
     let repo = open_composefs_repo(rootfs_dir)?;
 
--- a/crates/lib/src/store/mod.rs
+++ b/crates/lib/src/store/mod.rs
@@ -24,7 +24,9 @@ use anyhow::{Context, Result};
 use bootc_mount::tempmount::TempMount;
 use camino::Utf8PathBuf;
 use cap_std_ext::cap_std;
-use cap_std_ext::cap_std::fs::{Dir, DirBuilder, DirBuilderExt as _};
+use cap_std_ext::cap_std::fs::{
+    Dir, DirBuilder, DirBuilderExt as _, Permissions, PermissionsExt as _,
+};
 use cap_std_ext::dirext::CapStdExtDirExt;
 use fn_error_context::context;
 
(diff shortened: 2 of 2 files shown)


crates/lib/src/store/mod.rs
```diff
@@ -430,3 +441,37 @@ impl Storage {
...
+        let assert_mode = || -> Result<()> {
+            let perms = td.metadata(COMPOSEFS)?.permissions();
+            let mode = Mode::from_raw_mode(perms.mode());
+            // Owner should have full access
+            assert!(mode.contains(Mode::RUSR));
+            assert!(mode.contains(Mode::WUSR));
+            assert!(mode.contains(Mode::XUSR));
+            // Group and other should have no access
+            assert!(!mode.intersects(Mode::RGRP | Mode::WGRP | Mode::XGRP));
+            assert!(!mode.intersects(Mode::ROTH | Mode::WOTH | Mode::XOTH));
```
cgwalters: What is this doing over just directly comparing it to 0700 here again?

crates/lib/src/bootc_composefs/repo.rs
```diff
@@ -26,9 +26,7 @@ pub(crate) async fn initialize_composefs_repository(
 ) -> Result<(String, impl FsVerityHashValue)> {
     let rootfs_dir = &root_setup.physical_root;
 
-    rootfs_dir
-        .create_dir_all("composefs")
-        .context("Creating dir composefs")?;
+    crate::store::ensure_composefs_dir(rootfs_dir)?;
```
cgwalters, line 29: I think we should also consider upgrades, my vote would be to just automatically change the mode if it doesn't match.

crates/lib/src/store/mod.rs
jeckersb: What Gemini write below is better.  Basically this is the full `st_mode` which has other bits besides the permissions, and the initial AI-generated test was doing `& 0o777` which was masking everything except the permissions bits.  But (imo) that's non-obvious so I told it to be a bit more verbose for the sake of clarity.  The Gemini recommendation of `& Mode::PERMS` is a lot more concise while also being more explicit.

jeckersb: Well except apparently `Mode::PERMS` is a hallucination and doesn't actually exist...

crates/lib/src/store/mod.rs
```diff
@@ -51,8 +53,29 @@ pub const SYSROOT: &str = "sysroot";
...
+
+/// Ensure the composefs directory exists in the given physical root
+/// with the correct permissions (mode 0700).
+pub(crate) fn ensure_composefs_dir(physical_root: &Dir) -> Result<()> {
+    let mut db = DirBuilder::new();
+    db.mode(COMPOSEFS_MODE.as_raw_mode());
+    physical_root
+        .ensure_dir_with(COMPOSEFS, &db)
+        .context("Creating composefs directory")?;
+    // Always update permissions, in case the directory pre-existed
```
jeckersb, line 69: This always updates now to handle the upgrade case.

crates/lib/src/store/mod.rs
jeckersb: Latest version just defines `PERMS` as a const with a comment explaining, then uses that.

crates/lib/src/store/mod.rs
```diff
@@ -51,8 +53,29 @@ pub const SYSROOT: &str = "sysroot";
...
+    // with incorrect mode (e.g. from an older version of bootc).
+    physical_root
```
cgwalters, line 71: Minor nit but I'd prefer if we only called `chmod()` if we had work to do. It may not matter...but I think it might still dirty the inode even in the no-op case, causing write traffic.

Also important: At some point in the future we will need to properly handle "physically" readonly rootfs (think ISO), so we don't want to introduce new places where we unconditionally try writes.

crates/lib/src/store/mod.rs
```diff
@@ -430,3 +451,62 @@ impl Storage {
...
+
+        // Calling again should be a no-op (ensure is idempotent)
+        ensure_composefs_dir(&td)?;
+        assert_mode()?;
+
+        Ok(())
+    }
+
+    #[test]
+    fn test_ensure_composefs_dir_fixes_existing() -> Result<()> {
```
cgwalters, line 488: The unit test is fine and cheap, but I think this one also wants to be integration tested.

crates/lib/src/store/mod.rs
cgwalters: These things can be followups

PR #2014, opened 2026-02-19
tmt: Temporarily disable unified storage test

jeckersb:
There's something weird going on with selinux labeling, see discussion
at:

https://github.com/bootc-dev/bootc/pull/1986

Disable the test for now while we troubleshoot, because it's blocking
other stuff that needs to merge.

### DIFF
--- a/tmt/tests/booted/test-switch-to-unified.nu
+++ b/tmt/tests/booted/test-switch-to-unified.nu
@@ -17,6 +17,10 @@ let st = bootc status --json | from json
 let booted = $st.status.booted.image
 
 def main [] {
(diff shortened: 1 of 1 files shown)


jeckersb:
I'm very unsure if adding `enabled: false` is enough for it to skip the test and allow the whole job to complete, but I guess we'll see...

jeckersb:
> I'm very unsure if adding `enabled: false` is enough for it to skip the test and allow the whole job to complete, but I guess we'll see...

Yeah didn't work, new version just early-returns from main in the test instead of disabling the test.

ckyrouac (approved):

Issue #1957, opened 2026-01-26
Add `bootc container export --format=tar`


cgwalters:
This relates a bit to https://github.com/bootc-dev/bootc/issues/898 - but it's for people who want to use bootc-oriented containers with e.g. the [`liveimg` verb of Anaconda](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/automatically_installing_rhel/kickstart-commands-and-options-reference_rhel-installer#liveimg_kickstart-commands-for-installation-program-configuration-and-flow-control) and similar things.

What I'm thinking here is we have `bootc container export [--experimental]` that is usable like this:

```
podman run --rm --mount=type=image,source=<image>,target=/target <image> bootc container export /target > output.tar` or so.
```

## What does this do beyond just `tar -cf - /target`?

- SELinux labeling as if it was a booted host
- Possibly adjusting (for legacy compatibility) by writing the kernel into `/boot` in the tar stream instead of `/usr/lib/modules`

And maybe more things - but those two are important.

PR #2018, opened 2026-02-23
container: Add export --format=tar command

cgwalters:
Some people want to use container build tools, but for compatibility with older systems export a tar format of the OS state e.g. Anaconda liveimg expects this.

Basically this is only *slightly* more than just `tar cf`; we need to handle SELinux labeling and move the kernel.

Ref: #1957

Assisted-by: OpenCode (Sonnet 4.5)

### DIFF
--- a/crates/tests-integration/Cargo.toml
+++ b/crates/tests-integration/Cargo.toml
@@ -26,10 +26,20 @@ xshell = { workspace = true }
 bootc-kernel-cmdline = { path = "../kernel_cmdline", version = "0.0.0" }
 
 # Crate-specific dependencies
+# bcvk-qemu: QEMU/virtiofsd management from the bcvk project.
+# This is a git dependency — not published to crates.io.
+# When updating, also check the bcvk-qemu Cargo.toml for its own
+# dependency versions (cap-std-ext, etc.) to avoid conflicts.
+bcvk-qemu = { git = "https://github.com/bootc-dev/bcvk", rev = "d4be4933e09796f355076ad6daa60d570e64cec8" }
+data-encoding = "2.9"
+indicatif = { workspace = true }
 libtest-mimic = "0.8.0"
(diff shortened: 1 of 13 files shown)


crates/xtask/src/anaconda.rs
```diff
@@ -0,0 +1,947 @@
...
+        .stdout(Stdio::null())
+        .stderr(Stdio::null())
+        .spawn()
+        .context("Spawning virtiofsd")?;
+
+    let _virtiofsd = VirtiofsProcess {
+        child: virtiofsd_child,
+    };
+
+    // Wait for virtiofsd socket to appear
```
cgwalters: It's obviously not great to end up with another qemu+virtiofs wrapper here, so I put up https://github.com/bootc-dev/bcvk/pull/215

crates/tests-integration/Cargo.toml
```diff
@@ -26,10 +26,21 @@ xshell = { workspace = true }
 bootc-kernel-cmdline = { path = "../kernel_cmdline", version = "0.0.0" }
 
 # Crate-specific dependencies
+# bcvk-qemu: QEMU/virtiofsd management from the bcvk project.
+# This is a git dependency — not published to crates.io.
+# When updating, also check the bcvk-qemu Cargo.toml for its own
+# dependency versions (cap-std-ext, etc.) to avoid conflicts.
+# TODO: Update rev after https://github.com/bootc-dev/bcvk/pull/217 merges
+bcvk-qemu = { git = "https://github.com/cgwalters/bcvk-fork", branch = "iso-boot-mode" }
```
ckyrouac: looks like this can be updated now

ckyrouac:
lgtm other than the bcvk-qemu dep update

cgwalters:
Ah right this also needs https://github.com/bootc-dev/bcvk/pull/220

PR #2019, opened 2026-02-24
Makefile: Remove build dependency from install target

cgwalters:
The Makefile's CARGO_FEATURES auto-detection based on /usr/lib/os-release
can produce different results in the install environment vs build
environment. This causes 'make install' to rebuild the binary without
the intended features (e.g., rhsm).

Pass CARGO_FEATURES explicitly to ensure the install step doesn't
trigger an unwanted rebuild with different features.


cgwalters:
Already trying this downstream in https://gitlab.com/redhat/centos-stream/rpms/bootc/-/merge_requests/81

Issue #2020, opened 2026-02-24
Mislabeled selinux context with unified storage


jeckersb:
Some previous discussion in https://github.com/bootc-dev/bootc/pull/1986 but breaking this off into its own issue.  In a nutshell we started hitting this in CI (which prompted https://github.com/bootc-dev/bootc/pull/2014):

```
[root@localhost tmp.pCWKBpLqYd]# RUST_LOG=trace bootc image cmd build -t localhost/bootc-unified-derived .
TRACE starting bootc
DEBUG argv0="bootc"
TRACE Verified uid 0 with CAP_SYS_ADMIN
DEBUG Re-executing current process for _ostree_unshared
TRACE starting bootc
DEBUG argv0="bootc"
TRACE Verified uid 0 with CAP_SYS_ADMIN
DEBUG Already in a mount namespace
DEBUG Current security context is unconfined_u:system_r:install_t:s0-s0:c0.c1023
DEBUG We have install_t
DEBUG Target . is a mountpoint, remounting rw
TRACE loading sepolicy from booted ostree deployment
TRACE sepolicy in get_ensure_imgstore: Some(SePolicy { inner: TypedObjectRef { inner: 0x556f1723b150, type: OstreeSePolicy } })
TRACE Opening container image store
TRACE exec: REGISTRY_AUTH_FILE="/proc/self/fd/26" "podman" "--root" "/run/bootc/storage" "--runroot" "/proc/self/fd/3" "image" "build" "-t" "localhost/bootc-unified-derived" "."
STEP 1/2: FROM localhost/bootc
STEP 2/2: RUN echo 'unified-storage-test-marker' > /usr/share/unified-storage-test.txt
error: Subprocess failed: ExitStatus(unix_wait_status(32512))
/bin/sh: error while loading shared libraries: /lib64/libc.so.6: cannot apply additional memory protection after relocation: Permission denied
Error: building at STEP "RUN echo 'unified-storage-test-marker' > /usr/share/unified-storage-test.txt": while running runtime: exit status 127
```

Definitely a problem with the labels because if I just force-relabel the whole bootc-owned store to match the host containers-storage then it works as expected:

```
[root@localhost tmp.pCWKBpLqYd]# mount -o remount,rw /sysroot
[root@localhost tmp.pCWKBpLqYd]# semanage fcontext -a -e /var/lib/containers/storage /sysroot/ostree/bootc/storage
[...]

PR #2021, opened 2026-02-24
podstorage: Relabel storage after podman write operations

jeckersb:
Files written by podman into the bootc-owned container storage get
incorrect SELinux labels because neither the on-disk path
(/sysroot/ostree/bootc/storage) nor the bind-mount path
(/run/bootc/storage) has file context rules in the system SELinux
policy. The existing ensure_labeled() mechanism only ran once at
storage creation time via a stamp file, so all files created by
subsequent podman operations (image layers, metadata, etc.) were
mislabeled, causing container execution failures like:

  /bin/sh: error while loading shared libraries: /lib64/libc.so.6:
  cannot apply additional memory protection after relocation:
  Permission denied

Fix this by storing the SELinux policy in CStorage and recursively
relabeling all files as if they lived under /var/lib/containers/storage
after every podman write operation (pull, pull_from_host_storage, and
imgcmd_entrypoint for build/pull commands).

Also re-enable the unified storage integration test that was disabled
in c26ae5aa while this was being investigated.

Closes: #2020

### DIFF
--- a/crates/lib/src/image.rs
+++ b/crates/lib/src/image.rs
@@ -220,7 +220,16 @@ pub(crate) async fn imgcmd_entrypoint(
     let mut cmd = storage.new_image_cmd()?;
     cmd.arg(arg);
     cmd.args(args);
-    cmd.run_capture_stderr()
+    cmd.run_capture_stderr()?;
+    // Relabel after podman operations that write new files into storage.
+    // Commands like `build` and `pull` write image layers with incorrect
+    // SELinux labels because our storage path has no file context rules
+    // in the system SELinux policy.
+    match arg {
+        "build" | "pull" => storage.relabel()?,
+        _ => {}
+    }
+    Ok(())
 }
 
 /// Re-pull the currently booted image into the bootc-owned container storage.
--- a/crates/lib/src/podstorage.rs
+++ b/crates/lib/src/podstorage.rs
@@ -56,6 +56,13 @@ pub(crate) struct CStorage {
     #[allow(dead_code)]
     /// Our runtime state
     run: Dir,
+    /// The SELinux policy used to relabel files after podman operations.
+    /// Files written by podman into our storage get incorrect SELinux labels
+    /// because the storage path (/sysroot/ostree/bootc/storage, bind-mounted
+    /// to /run/bootc/storage) has no file context rules in the system policy.
+    /// We use this policy to relabel files as if they lived under
+    /// /var/lib/containers/storage after every write operation.
+    sepolicy: Option<ostree::SePolicy>,
     /// Disallow using this across multiple threads concurrently; while we
     /// have internal locking in podman, in the future we may change how
     /// things work here. And we don't have a use case right now for
@@ -234,6 +241,28 @@ impl CStorage {
         Ok(())
     }
 
+    /// Relabel the storage directory to have correct SELinux labels.
+    ///
+    /// Podman writes files into our storage with incorrect SELinux labels
+    /// because the actual storage path has no file context rules in the
+    /// system SELinux policy. This method recursively relabels all files
+    /// as if they were under /var/lib/containers/storage, which is the
+    /// path the SELinux policy does have rules for.
+    #[context("Relabeling imgstorage after write")]
+    pub(crate) fn relabel(&self) -> Result<()> {
+        let Some(sepolicy) = self.sepolicy.as_ref() else {
+            return Ok(());
+        };
+
+        crate::lsm::relabel_recurse(
+            &self.storage_root,
+            ".",
+            Some(Utf8Path::new("/var/lib/containers/storage")),
+            sepolicy,
+        )
+        .context("labeling storage root")
+    }
+
     #[context("Creating imgstorage")]
     pub(crate) fn create(
         sysroot: &Dir,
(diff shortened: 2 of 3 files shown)


jeckersb:
Draft and intentionally no DCO since I need to disappear for a bit and haven't thoroughly reviewed this, but a cursory glance looks promising and it seems to fix things.

cgwalters:
> Fix this by storing the SELinux policy in CStorage and recursively
relabeling all files as if they lived under /var/lib/containers/storage
after every podman write operation (pull, pull_from_host_storage, and
imgcmd_entrypoint for build/pull commands).

Ugh. I would *really* like to avoid such a big hammer. It's inefficient and also racy (as written today): if the bootc process dies (SIGINT, force power off) between pull and relabel, then nothing fixes it until the next pull.

My understanding is that if we have the labels set up correctly on the *base* directories, subdirectories (and files) should inherit that.

Unless this relates to the runtime label/domain of the `podman` process we fork? Perhaps we're not doing a domain transition when we should be?

Can you or someone looking at this try to more precisely figure out why this seems to be working at install time but not day 2? What are the *precise* subdirectories/files that are incorrectly labeled in the latter?

cgwalters:
As an aside: in a future where podman is backed by composefs, we won't need special labeling for the podman storage and it can go away...that will be a nice day.

cgwalters:
To say this another way: AFAIK the way things work for `/var/lib/containers` is the same: the directory is initialized once with the right labels, and thereafter there's no recursive relabeling required by `podman pull`. So why would one be required here?

It just smells to me like there's a bug with the labeling setup.

BTW different thing: it'd be interesting to add stuff like this to `bootc internals fsck`, I think that'd be a good place to add force relabeling of our storage.

jeckersb:
> To say this another way: AFAIK the way things work for `/var/lib/containers` is the same: the directory is initialized once with the right labels, and thereafter there's no recursive relabeling required by `podman pull`. So why would one be required here?
> 
> It just smells to me like there's a bug with the labeling setup.
> 
> BTW different thing: it'd be interesting to add stuff like this to `bootc internals fsck`, I think that'd be a good place to add force relabeling of our storage.

I touched on it in my [other comment](https://github.com/bootc-dev/bootc/issues/2020#issuecomment-3953793131), but if the policy contained rules for the bootc containers-storage the same as it does the podman container-storage they would function the same; the files would just get the correct labels by default via the policy, right?  Disclaimer being that I haven't reviewed exactly how the policy works for /var/lib/containers...

jeckersb:
```
[root@localhost ~]# semanage fcontext -l | grep /var/lib/containers
/var/lib/containers(/.*)?                          all files          system_u:object_r:container_var_lib_t:s0
/var/lib/containers/atomic(/.*)?                   all files          <<None>>
/var/lib/containers/home(/.*)?                     all files          system_u:object_r:openshift_var_lib_t:s0
/var/lib/containers/overlay(/.*)?                  all files          system_u:object_r:container_ro_file_t:s0
/var/lib/containers/overlay-images(/.*)?           all files          system_u:object_r:container_ro_file_t:s0
/var/lib/containers/overlay-layers(/.*)?           all files          system_u:object_r:container_ro_file_t:s0
/var/lib/containers/overlay2(/.*)?                 all files          system_u:object_r:container_ro_file_t:s0
/var/lib/containers/overlay2-images(/.*)?          all files          system_u:object_r:container_ro_file_t:s0
/var/lib/containers/overlay2-layers(/.*)?          all files          system_u:object_r:container_ro_file_t:s0
/var/lib/containers/storage/artifacts(/.*)?        all files          system_u:object_r:container_ro_file_t:s0
/var/lib/containers/storage/overlay(/.*)?          all files          system_u:object_r:container_ro_file_t:s0
/var/lib/containers/storage/overlay-images(/.*)?   all files          system_u:object_r:container_ro_file_t:s0
/var/lib/containers/storage/overlay-layers(/.*)?   all files          system_u:object_r:container_ro_file_t:s0
/var/lib/containers/storage/overlay2(/.*)?         all files          system_u:object_r:container_ro_file_t:s0
/var/lib/containers/storage/overlay2-images(/.*)?  all files          system_u:object_r:container_ro_file_t:s0
/var/lib/containers/storage/overlay2-layers(/.*)?  all files          system_u:object_r:container_ro_file_t:s0
/var/lib/containers/storage/volumes/[^/]*/.*       all files          system_u:object_r:container_file_t:s0
```

jeckersb:
And those ultimately come from https://github.com/containers/container-selinux/blob/521cf8c56e8df7cd6809571ed9e796b495ceaa71/container.fc#L112-L127

jeckersb:
On a positive note, the tests actually passed with this across the matrix (ignoring the docs flake), so once we nail down a decent fix with the labeling nothing else should be waiting to further block this.

cgwalters:
> I touched on it in my https://github.com/bootc-dev/bootc/issues/2020#issuecomment-3953793131, but if the policy contained rules for the bootc containers-storage the same as it does the podman container-storage they would function the same; the files would just get the correct labels by default via the policy, right?

Our relabeling logic relabels the files underneath the boot-owned cstor *as if* they actually were at `/var/lib/containers` - so it *should* result in those same labels.

jeckersb:
> > I touched on it in my [#2020 (comment)](https://github.com/bootc-dev/bootc/issues/2020#issuecomment-3953793131), but if the policy contained rules for the bootc containers-storage the same as it does the podman container-storage they would function the same; the files would just get the correct labels by default via the policy, right?
> 
> Our relabeling logic relabels the files underneath the boot-owned cstor _as if_ they actually were at `/var/lib/containers` - so it _should_ result in those same labels.

Yeah but that's a one-shot operation. Once you run podman again it adds new unlabeled files until you run the relabel again at some future point.

But the podman store at `/var/lib/containers` has the correct labels in the system policy, so they are labeled correctly at creation time.  We don't have the policy for our containers-storage instance so they are not labeled correctly and then we have to go fix them after the fact.

Or to put it another way, it's the reason this works:

```
[root@localhost ~]# touch /etc/some_totally_unique_configname
[root@localhost ~]# ls -lZ /etc/some_totally_unique_configname
-rw-r--r--. 1 root root unconfined_u:object_r:etc_t:s0 0 Feb 24 21:44 /etc/some_totally_unique_configname
```

I don't have to do anything for that to become `etc_t`, because:

```
[root@localhost ~]# semanage fcontext -l | grep '/etc/\.\* '
/etc/.*                                            all files          system_u:object_r:etc_t:s0
```

jeckersb:
Closed in favor of #2025

PR #2023, opened 2026-02-25
cli/cfsctl: Use proxy from composefs-rs

Johan-Liebert1:
- Bump composefs-rs
Also, update the git URL to point to
https://github.com/composefs/composefs-rs

- cli/cfsctl: Use proxy from composefs-rs
Now that we have cfsctl as a library, we simply use that to proxy all of
our `bootc internal cfs` commands. Cleans up the codebase a bit, DRYs
it and keeps things in sync

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -201,7 +201,7 @@ dependencies = [
  "bootc-internal-utils",
  "bootc-mount",
  "camino",
- "cap-std-ext",
+ "cap-std-ext 5.0.0",
  "fn-error-context",
  "indoc",
  "libc",
--- a/tmt/tests/booted/readonly/030-test-composefs.nu
+++ b/tmt/tests/booted/readonly/030-test-composefs.nu
@@ -39,8 +39,12 @@ if $is_composefs {
     # When not on composefs, run the full test including initialization
     bootc internals test-composefs
     bootc internals cfs --help
-    bootc internals cfs oci pull docker://busybox busybox
-    test -L /sysroot/composefs/streams/refs/busybox
+
+    # We use a separate `/sysroot` as we need rw access to the repo which
+    # we can't get from `bootc internals cfs ...`
+    mkdir /var/tmp/sysroot/composefs
+    bootc internals cfs --insecure --repo /var/tmp/sysroot/composefs oci pull docker://busybox busybox
+    test -L /var/tmp/sysroot/composefs/streams/refs/oci/busybox
 }
 
 tap ok


cgwalters (approved):
Nice!

Johan-Liebert1:
Ok wow

```
error: Installing to disk: Setting up composefs boot: Failed to mount composefs image: Mounting image '3ba387b6e47dacf983937fd9e3824cef20ebdce824199215dfa803fbe9a6e6b7379539fad23444667aeba209739b395a51345ba1f57dfc17e50a81da7fbf0074': Creating filesystem mount: Invalid argument (os error 22)
```

Johan-Liebert1:
Looks like an issue with creating loop device

```
write(2, "[crates/composefs/src/mountcompa"..., 69[crates/composefs/src/mountcompat.rs:138:5] "make_erofs_mountable" = ) = 69
write(2, "\"", 1")                       = 1
write(2, "make_erofs_mountable", 20make_erofs_mountable)    = 20
write(2, "\"", 1")                       = 1
write(2, "\n", 1
)                       = 1
openat(AT_FDCWD, "/dev/loop-control", O_RDWR|O_CLOEXEC) = 12
ioctl(12, LOOP_CTL_GET_FREE)            = 1
openat(AT_FDCWD, "/dev/loop1", O_RDWR|O_CLOEXEC) = 13
ioctl(13, _IOC(_IOC_WRITE, 0x4c, 0xa, 0x130), 0x7ffe9710f9a8) = -1 EINVAL (Invalid argument)
```

Precisely the issue seems to be here https://github.com/composefs/composefs-rs/blob/022c9c62216615af471acaa78f486203a53f06e3/crates/composefs-ioctls/src/loop_device.rs#L154

Using raw `ioctl` fixes the issue

```rust
        if libc::ioctl(loop_dev.as_raw_fd(), 0x4C0A, &config) < 0 {
            println!("IOCTL RAW FAILED");
        }
```

which is really weird to me as I don't see any difference between the outputs for both produced by strace

Johan-Liebert1:
We regressed with this commit https://github.com/composefs/composefs-rs/commit/984ce030b69549a2459d655efb6266f9f5854685

But I don't see the issue immediately

Johan-Liebert1:
Okay issues seems to be that `_IOC(_IOC_WRITE, 0x4c, 0xa, 0x130)` resolves to `0x41304c0a` and not `0x4c0a` which is what `LOOP_CONFIGURE` should be

Johan-Liebert1:
Fix here https://github.com/composefs/composefs-rs/pull/240
Replaces the _IO macros with hardcoded opcodes

crates/lib/src/store/mod.rs
```diff
@@ -426,20 +427,35 @@ impl Storage {
     ///
     /// This lazily opens the composefs repository, creating the directory if needed
     /// and bootstrapping verity settings from the ostree configuration.
-    pub(crate) fn get_ensure_composefs(&self) -> Result<Arc<ComposefsRepository>> {
+    pub(crate) fn get_ensure_composefs(
```
cgwalters: I think this PR is trying to solve the same problem ("ro sysroot") in two different ways. 

The simplest and IMO best way is what you've already done in the test code below: change the cfsctl invocation to operate on something in `/var/tmp`.

Changing this core storage code is I believe unnecessary and conceptually wrong because this particular module is really about the bootc-owned storage only.

Johan-Liebert1: Yes, in hindsight this makes sense. Reverted the change in `get_ensure_composefs`

cgwalters:
BTW once this lands we should also be able to bump the bootc commit in composefs-rs and restore the revdep check.

Issue #1826, opened 2025-12-03
composefs working examples currently set enforcing=0


jeckersb:
I haven't dug into why exactly, but clearly we don't want this so I'm creating an issue to not forget about it!

cgwalters:
https://github.com/containers/composefs-rs/issues/183

cgwalters:
In the immediate term let's link the selinux=0 there to this issue

gucci-on-fleek:
I was able to (kinda) get SELinux to work. In the container build, I ran

```shell
sed -i 's/enforcing/permissive/' /etc/selinux/config
semanage fcontext --add --seuser system_u --type lib_t '/sysroot/composefs(/.*)?'
```

and I added the following unit to the image:

```desktop
[Unit]
Description=Fix SELinux
DefaultDependencies=false
ConditionFirstBoot=true

Before=sysinit.target
Before=dbus-broker.service
After=local-fs.target

[Service]
RuntimeDirectory=%N-sysroot
Type=oneshot
ExecStart=restorecon -vRF /etc /var
ExecStart=mount --options=rw --source=/dev/gpt-auto-root --target=%t/%N-sysroot
ExecStart=chcon --recursive --user=system_u --role=object_r --type=lib_t %t/%N-sysroot/composefs
ExecStart=umount %t/%N-sysroot
ExecStart=sed -i 's/permissive/enforcing/' /etc/selinux/config
ExecStart=setenforce Enforcing

[Install]
WantedBy=sysinit.target
```

After that, everything seems to work correctly—I tested it with confined users plus `unconfined_t` disabled, and I didn't see any suspicious AVC messages. I haven't done any extensive testing yet though, so there's a good chance that something is still broken.

 `lib_t` isn't really the right label for these files, but it's readable by every domain, and only writable by the installer domains, so it's an okay-ish choice.

I also tried to investigate _why_ this is happening, and I couldn't figure it out. Some random notes:

- `/lib64/libc.so.6` was the source of most of the denials.
- When I added `strace` to the `ExecStart=` for the programs listed in the AVC denial messages, the problem went away (tested with `firewalld.service` and `sshd.service`).
- The problem consistently happened every time that I rebooted. And the first time that I restarted each service (using `systemctl`) after rebooting, I got the same AVC denials, but the second and subsequent times I restarted them, the problem went away.
- Inspecting `/proc/<pid>/root`, `/proc/<pid>/mounts`, and `/proc/<pid>/map_files/` didn't show anything unusual.

PR #2025, opened 2026-02-26
install: Defer SELinux relabeling of container storage until after image pulls

jeckersb:
Previously, the bootc-owned containers-storage instance was SELinux-relabeled
immediately upon creation in CStorage::create(), before any logically bound
images (LBIs) were pulled into it. When bootc install runs inside a VM with
SELinux disabled (e.g. bcvk ephemeral mode passes selinux=0 on the kernel
command line), all files created during the install are unlabeled. The early
relabel only covered the empty storage scaffold, while LBI content pulled
afterward remained unlabeled. The .bootc_labeled stamp file then prevented
re-relabeling.

Fix this by splitting the behavior in CStorage::create(): when opening
pre-existing storage (the booted/runtime case), ensure_labeled() is called
immediately as before. When creating fresh storage (the install case),
labeling is deferred. The install and completion code paths now call
ensure_labeled() explicitly after all image pulls are complete, ensuring
all content is covered by the relabel pass.

The SELinux policy is stored in CStorage at creation time so that
ensure_labeled() can be called without needing the policy from the caller.

This also reverts the test disablement from PR#2014, re-enabling
test-switch-to-unified.nu.

### DIFF
--- a/crates/lib/src/podstorage.rs
+++ b/crates/lib/src/podstorage.rs
@@ -41,6 +41,10 @@ const STORAGE_RUN_FD: i32 = 3;
 
 const LABELED: &str = ".bootc_labeled";
 
+/// The system path to the canonical containers-storage instance,
+/// used as the SELinux label reference path.
+const SYS_CSTOR_PATH: &str = "/var/lib/containers/storage";
+
 /// The path to the image storage, relative to the bootc root directory.
 pub(crate) const SUBPATH: &str = "storage";
 /// The path to the "runroot" with transient runtime state; this is
@@ -56,6 +60,8 @@ pub(crate) struct CStorage {
     #[allow(dead_code)]
     /// Our runtime state
     run: Dir,
+    /// The SELinux policy used for labeling the storage.
+    sepolicy: Option<ostree::SePolicy>,
     /// Disallow using this across multiple threads concurrently; while we
     /// have internal locking in podman, in the future we may change how
     /// things work here. And we don't have a use case right now for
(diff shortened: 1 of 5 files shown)


crates/lib/src/podstorage.rs
```diff
@@ -207,11 +220,20 @@ impl CStorage {
         Ok(())
     }
 
+    /// Finalize the container storage by ensuring all content is properly
+    /// SELinux-labeled. This must be called after all image pulls are complete
+    /// and before the `CStorage` is dropped.
```
cgwalters: Only for the install flow. When we're booted, then actually we want to relabel *before* we do pulls I'd say - for the same reason as my initial comment - doing the relabeling after for the booted case is racy (though we could fix it if we ran a systemd service on early boot to fix it up).

So...my vote would be to drop this `Drop` and finished variable, and keep the code I'll tag below...

crates/lib/src/podstorage.rs
```diff
@@ -265,23 +287,22 @@ impl CStorage {
...
                 .run_capture_stderr()
                 .context("Initializing images")?;
-            Self::ensure_labeled(&storage_root, sepolicy)?;
             drop(storage_root);
             sysroot
                 .rename(&tmp, sysroot, subpath)
                 .context("Renaming tmpdir")?;
             tracing::debug!("Created image store");
-        } else {
-            // the storage already exists, make sure it has selinux labels
```
cgwalters, line 275: this code would be kept

crates/lib/src/podstorage.rs
```diff
@@ -265,23 +267,30 @@ impl CStorage {
                 .arg("images")
                 .run_capture_stderr()
                 .context("Initializing images")?;
-            Self::ensure_labeled(&storage_root, sepolicy)?;
```
cgwalters, line 268: I think this is OK I'll just note that the code went to some effort to make "create and label" as an *atomic* unit by doing the "Create tempdir then rename" pattern. But it doesn't really matter, we'll just relabel anyways if interrupted.

jeckersb:
Ah install test failed with:

```
---- Install and verify selinux state ----
No SELinux label found for: "bootc/storage/.bootc_labeled"
```

Will need to fix that.

cgwalters:
brb editing the wikipedia page on irony to add this as an entry

crates/lib/src/podstorage.rs
```diff
@@ -211,25 +217,35 @@ impl CStorage {
...
-            Some(Utf8Path::new("/var/lib/containers/storage")),
+            Some(Utf8Path::new(SYS_CSTOR_PATH)),
             sepolicy,
         )
         .context("labeling storage root")?;
 
-        root.create(LABELED)?;
+        self.storage_root.create(LABELED)?;
+
+        // Label the stamp file itself to match the storage directory context
```
cgwalters, line 240: Per discussion I think this is fine, but...I think it would also work to just relabel all content under the installation root (except the ostree deployment).

For the booted case, this shouldn't be necessary - we'll just inherit a default label.

jeckersb: :+1: I'm going to do that in a follow up which should also hopefully fix #1826

PR #2026, opened 2026-02-27
Follow up PR to #2004

Johan-Liebert1:
### xtask: Add `seal-state` and `boot-type` options
seal-state: Required to switch between secure/insecure firmware options
boot-type: Required to send kargs to only bls installs

### composefs/tests: More flexibility for insecure UKI testing
Update the CI matrix to include `seal_state` and `boot_type`. This does
not increase our matrix, but only rearranges it to be a bit more
meaningful. Earlier even when testing "insecure UKI", it still showed up
as "composefs-sealeduki-sdboot" which is incorrect.

This also allows us flexibility to, in future, test grub + UKI which is
disabled currently.

Update Justfile and the Dockerfile to make use of these new arguments.
Now we only sign the UKI, if `seal_state == sealed`, and in the Justfile
we disallow combinations that don't make sense, like BLS boot + sealed,
allowing missing verity (xfs) + sealed, etc.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -180,6 +180,7 @@ jobs:
             filesystem: xfs
           - seal_state: "unsealed"
             filesystem: ext4
+            boot_type: uki # we still want to test ext4 unsealed bls
           - bootloader: grub
             boot_type: "uki"
 
--- a/crates/tests-integration/src/container.rs
+++ b/crates/tests-integration/src/container.rs
@@ -50,9 +50,12 @@ pub(crate) fn test_bootc_container_inspect() -> Result<()> {
         .expect("kernel.unified should be present")
         .as_bool()
         .expect("kernel.unified should be a boolean");
+
+    let is_uki = std::env::var("BOOTC_boot_type").is_ok_and(|var| var == "uki");
+
     if let Some(variant) = std::env::var("BOOTC_variant").ok() {
-        match variant.as_str() {
-            v @ "ostree" | v @ "composefs" => {
+        match (variant.as_str(), is_uki) {
+            (v @ "ostree", _) | (v @ "composefs", false) => {
                 assert!(!unified, "Expected unified=false for variant {v}");
                 // For traditional kernels, version should look like a uname (contains digits)
                 assert!(
@@ -60,7 +63,7 @@ pub(crate) fn test_bootc_container_inspect() -> Result<()> {
                     "version should contain version numbers for traditional kernel: {version}"
                 );
             }
-            "composefs-sealeduki-sdboot" => {
+            ("composefs", true) => {
                 assert!(unified, "Expected unified=true for UKI variant");
                 // For UKI, version is the filename without .efi extension (should not end with .efi)
                 assert!(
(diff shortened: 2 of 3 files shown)


Johan-Liebert1:
Something's wrong with unsealed UKI case. Need to look into it

Johan-Liebert1:
Also, we don't need unsealed + ext4 as the xfs case covers it. I'll remove that from matrix

cgwalters:
Perhaps split out the first two commits separate from the larger test changes?

Johan-Liebert1:
Split out the first two commits to https://github.com/bootc-dev/bootc/pull/2032

Johan-Liebert1:
Okay, so this needs https://github.com/bootc-dev/bootc/pull/2032 to prevent conflicts. Test failures should also be fixed by rebasing afterwards. I made a mistake here while trying to get the composefs cmdline from the UKI which is fixed in #2032

.github/workflows/ci.yml
```diff
@@ -161,22 +161,36 @@ jobs:
...
-          - test_os: centos-9
-            variant: composefs-grub
+            variant: composefs
+          - seal_state: "sealed"
+            boot_type: bls
+          - seal_state: "sealed"
+            bootloader: grub
+          - seal_state: "sealed"
+            filesystem: xfs
+          - seal_state: "unsealed"
```
Johan-Liebert1, line 181: this should only skip the uki case. We still want to test unsealed bls ext4

Johan-Liebert1: Done

.github/workflows/ci.yml
```diff
@@ -161,22 +161,36 @@ jobs:
       matrix:
         # No fedora-44 due to https://bugzilla.redhat.com/show_bug.cgi?id=2429501
         test_os: [fedora-43, centos-9, centos-10]
-        variant: [ostree, composefs-sealeduki-sdboot, composefs-sdboot, composefs-grub]
+        variant: [ostree, composefs]
         filesystem: ["ext4", "xfs"]
```
cgwalters, line 165: as an aside we should probably have btrfs in this list too

crates/tests-integration/src/container.rs
```diff
@@ -50,17 +50,20 @@ pub(crate) fn test_bootc_container_inspect() -> Result<()> {
         .expect("kernel.unified should be present")
         .as_bool()
         .expect("kernel.unified should be a boolean");
+
+    let is_uki = std::env::var("BOOTC_boot_type").is_ok_and(|var| var == "uki");
+
     if let Some(variant) = std::env::var("BOOTC_variant").ok() {
-        match variant.as_str() {
-            v @ "ostree" | v @ "composefs" => {
+        match (variant.as_str(), is_uki) {
```
cgwalters, line 57: Nonblocking but for followup: we could also validate the sealed/unsealed state here right?

.github/workflows/ci.yml
Johan-Liebert1: yes, but that will add a bunch more combinations to the matrix

crates/tests-integration/src/container.rs
Johan-Liebert1: yes, we can

PR #2028, opened 2026-02-27
install/flat: Use composefs pipeline for rootfs

ericcurtin:
--flat install implementation will go through the composefs
repository. This gives several advantages:

- SELinux labeling is handled by composefs-rs (selabel), which applies
  labels from the image's file_contexts rather than inheriting the
  running container's labels.
- The kernel installation flow is reused from composefs-rs
  (get_boot_resources / UsrLibModulesVmlinuz).
- The composefs repo is preserved at /sysroot/composefs, making it
  easy to convert to immutable/bootc mode later. Users who don't want
  the metadata overhead can rm -rf /sysroot/composefs.
- write_to_path supports reflink copies on btrfs/XFS, sharing blocks
  with the composefs object store.
- --source-imgref now works with --flat (the in_host_mountns guard is
  removed since we pull via the image reference, not via cp /).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,10 +44,11 @@ clap_mangen = { version = "0.2.20" }
 # If adding/removing crates here, also update docs/Dockerfile.mdbook and docs/src/internals.md.
 #
 # To develop against a local composefs-rs checkout, add a [patch] section at the end of this file:
-#   [patch."https://github.com/containers/composefs-rs"]
+#   [patch."https://github.com/composefs/composefs-rs"]
 #   composefs = { path = "/home/user/src/composefs-rs/crates/composefs" }
 #   composefs-boot = { path = "/home/user/src/composefs-rs/crates/composefs-boot" }
 #   composefs-oci = { path = "/home/user/src/composefs-rs/crates/composefs-oci" }
+#   cfsctl = { path = "/home/user/src/composefs-rs/crates/cfsctl" }
 # The Justfile will auto-detect these and bind-mount them into container builds.
 composefs = { git = "https://github.com/composefs/composefs-rs", rev = "b928c6bd6c051e111d3efc3d25cdaf9159182ed0", package = "composefs", features = ["rhel9"] }
 cfsctl = { git = "https://github.com/composefs/composefs-rs", rev = "b928c6bd6c051e111d3efc3d25cdaf9159182ed0", package = "cfsctl", features = ["rhel9"] }
--- a/crates/lib/src/bootc_composefs/boot.rs
+++ b/crates/lib/src/bootc_composefs/boot.rs
@@ -730,7 +730,7 @@ pub(crate) fn setup_composefs_bls_boot(
     let (config_path, booted_bls) = if is_upgrade {
         let boot_dir = Dir::open_ambient_dir(&entry_paths.config_path, ambient_authority())?;
 
-        let mut booted_bls = get_booted_bls(&boot_dir)?;
+        let mut booted_bls: BLSConfig = get_booted_bls(&boot_dir)?;
         booted_bls.sort_key = Some(secondary_sort_key(&os_id));
 
         let staged_path = loader_path.join(STAGED_BOOT_LOADER_ENTRIES);
--- a/crates/lib/src/install.rs
+++ b/crates/lib/src/install.rs
@@ -173,6 +173,10 @@ use cap_std_ext::cap_tempfile::TempDir;
 use cap_std_ext::cmdext::CapStdExtCommandExt;
 use cap_std_ext::prelude::CapStdExtDirExt;
 use clap::ValueEnum;
+use composefs::fs::write_to_path;
+use composefs_boot::bootloader::{BootEntry, get_boot_resources};
+use composefs_boot::selabel;
+use composefs_oci::image::create_filesystem as create_composefs_filesystem;
 use fn_error_context::context;
 use ostree::gio;
 use ostree_ext::ostree;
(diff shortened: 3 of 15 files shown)


ericcurtin:
Not complete, wanted to see what shape the builds are in

ericcurtin:
This is admittedly quite basic, it's just copy stuff in the right place, but it's a start

ericcurtin:
What I am thinking the next step here is:

For /usr, /bin, /lib, etc.: do a Rust-written equivalent of `rsync -aHAX --delete
  --one-file-system` instead of `cp`, so extraneous files are removed
For /etc: apply the same 3-way merge logic from finalize.rs
For /var, /home, /root: do nothing

cgwalters:
> regular (non-bootc) Linux system

Request: Please don't use "regular" (or worse "normal") in that way, it implies other things are "irregular" (or worse "abnormal"). In this case I'd say "mutable" (or just non-bootc) as you did.

---

Anyways on the implementation side, I think it'd be really nice if we actually *did* create a composefs repo but then from there we did a reflink based copy out.

There are several huge advantages to that, among them we'd be reusing the selinux labeling support in composefs(-rs) for example, and the kernel installation flow.

Further it makes it *way* easier to then convert the image to immutable/bootc style later on. And anyone who doesn't want the extra metadata overhead can just `rm -rf /sysroot/composefs` or so.

I think we should not regenerate the initramfs client side by default.

ericcurtin:
> > regular (non-bootc) Linux system

Apologies, you've mentioned this before and I completely understand the logic.

> 
> Request: Please don't use "regular" (or worse "normal") in that way, it implies other things are "irregular" (or worse "abnormal"). In this case I'd say "mutable" (or just non-bootc) as you did.
> 
> Anyways on the implementation side, I think it'd be really nice if we actually _did_ create a composefs repo but then from there we did a reflink based copy out.
> 
> There are several huge advantages to that, among them we'd be reusing the selinux labeling support in composefs(-rs) for example, and the kernel installation flow.
> 
> Further it makes it _way_ easier to then convert the image to immutable/bootc style later on. And anyone who doesn't want the extra metadata overhead can just `rm -rf /sysroot/composefs` or so.
> 
> I think we should not regenerate the initramfs client side by default.

All fair points.

crates/blockdev/src/blockdev.rs
```diff
@@ -1,3 +1,11 @@
+//! Block device utilities for bootc.
+//!
+//! This module provides utilities for querying and managing block devices,
+//! partitions, and filesystems during bootc installation and operation.
+
+// Allow unsafe code for prctl FFI declaration
+#![allow(unsafe_code)]
```
cgwalters: This should be an *especially* large warning flag.

I'll probably do something like https://github.com/composefs/composefs-rs/pull/230 here too soon.

crates/blockdev/src/blockdev.rs
```diff
@@ -10,6 +18,14 @@ use serde::Deserialize;
 
 use bootc_utils::CommandRunExt;
 
+// prctl constants and extern declaration for parent death signal
+// PR_SET_PDEATHSIG = 1
+const PR_SET_PDEATHSIG: libc::c_int = 1;
```
cgwalters: This is exposed via rustix already https://docs.rs/rustix/latest/rustix/process/fn.parent_process_death_signal.html

crates/blockdev/src/blockdev.rs
```diff
@@ -26,27 +42,41 @@ struct DevicesOutput {
 
 #[allow(dead_code)]
 #[derive(Debug, Clone, Deserialize)]
+/// Information about a block device.
```
cgwalters: This seems fine, but it should be a prep commit in a distinct PR

crates/blockdev/src/blockdev.rs
```diff
@@ -384,17 +417,26 @@ impl Drop for LoopbackDevice {
...
 pub async fn run_loopback_cleanup_helper(device_path: &str) -> Result<()> {
     // Check if we're running as a cleanup helper
     if std::env::var("BOOTC_LOOPBACK_CLEANUP_HELPER").is_err() {
         anyhow::bail!("This function should only be called as a cleanup helper");
     }
 
     // Set up death signal notification - we want to be notified when parent dies
-    rustix::process::set_parent_process_death_signal(Some(rustix::process::Signal::TERM))
-        .context("Failed to set parent death signal")?;
+    // Use prctl directly via libc since rustix removed this API
```
cgwalters: I don't think that's true and even if it was should be clearly done as a distinct prep commit (+PR)

ADOPTERS.md
```diff
@@ -7,11 +7,11 @@
 
 | Type | Name | Since | Website | Use-Case |
 |:-|:-|:-|:-|:-|
-Vendor | Red Hat | 2024 | https://redhat.com | Image Based Linux 
-Vendor | HeliumOS | 2024 | https://www.heliumos.org/ | An atomic desktop operating system for your devices
-Vendor | AlmaLinux (Atomic SIG) | 2025 | [atomic-desktop](https://github.com/AlmaLinux/atomic-desktop), [atomic-workstation](https://github.com/AlmaLinux/atomic-workstation) | Atomic AlmaLinux desktop respins
-Vendor | Caligra | 2025 | [workbench](https://caligra.com/workbench/) | An OS designed to accelerate knowledge work
-Vendor | CIQ | 2026 | https://ciq.com | Rocky Linux from CIQ (RLC) - Image Based Linux - Standard and Cloud variants
+| Vendor | Red Hat | 2024 | https://redhat.com | Image Based Linux |
```
cgwalters: Same thing while the markdown formatting may need improvement (honestly we for sure need a markdown synax check run via `just check` or so) that again needs to be a clear distinct change with its own PR.

PR #2029, opened 2026-02-27
test: Add tests for missing /boot and --bootloader=none

henrywang:
Add two new integration tests:
- test-35-install-no-boot-dir: Tests install to-filesystem when /boot directory is absent
- test-36-install-bootloader-none: Tests install with --bootloader=none option to skip bootloader management

### DIFF
--- a/tmt/tests/booted/tap.nu
+++ b/tmt/tests/booted/tap.nu
@@ -19,61 +19,15 @@ export def is_composefs [] {
     $st.status.booted.composefs? != null
 }
 
-# Get the target image for install tests based on the running OS
-# This ensures the target image matches the host OS to avoid version mismatches
-# (e.g., XFS features created by newer mkfs.xfs not recognized by older grub2)
-export def get_target_image [] {
-    # Parse os-release to get ID and VERSION_ID
-    let os = open /usr/lib/os-release
-        | lines
-        | filter {|l| $l != "" and not ($l | str starts-with "#") }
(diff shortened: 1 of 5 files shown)


tmt/tests/booted/tap.nu
```diff
@@ -18,3 +18,62 @@ export def is_composefs [] {
     let st = bootc status --json | from json
     $st.status.booted.composefs? != null
 }
+
+# Get the target image for install tests based on the running OS
```
cgwalters: Why can't we just use the image from `bootc status`?

(Of course with unified storage enabled it becomes suddenly simple and easy to actually run as a container as well via podman...getting closer hopefully)

tmt/tests/booted/test-install-bootloader-none.nu
```diff
@@ -0,0 +1,31 @@
...
+let target_image = (tap get_target_image)
+
+def main [] {
+    tap begin "install with --bootloader=none"
+
+    truncate -s 10G disk.img
+
+    setenforce 0
+
+    systemd-run -p MountFlags=slave -qdPG -- /bin/sh -c $"
```
cgwalters: We're getting a lot of copy-pasta of this at this point, seems worth lowering into a helper in tap.nu

henrywang: Yes, totally agree.

PR #2035, opened 2026-03-03
Enable enforcing SELinux for sealed images

jeckersb:
install: Label composefs objects as /usr for SELinux

The composefs object store currently has no matching SELinux policy
for its storage path, so its contents end up labeled as default_t
which causes AVC denials at runtime. Explicitly label the composefs
directory tree as /usr (giving objects usr_t) after the composefs
install completes.

Future work should increase the granularity of SELinux behavior here,
ideally adding composefs-specific types and policy instead of re-using
usr_t.

Closes: #1826

---

install: Add final SELinux relabel of the physical root filesystem

Perform a full SELinux relabel pass over the physical root filesystem as
the very last step before filesystem finalization. This ensures all files
on the physical root have SELinux labels.

Files that are already labeled (e.g. ostree deployment contents,
composefs objects) are skipped since the relabel only acts on unlabeled
files.

---

Remove enforcing=0 for sealed composefs

### DIFF
--- a/contrib/packaging/seal-uki
+++ b/contrib/packaging/seal-uki
@@ -45,10 +45,6 @@ fi
 # Baseline container ukify options
 containerukifyargs=(--rootfs "${target}")
 
-# WORKAROUND: SELinux must be permissive for sealed UKI boot
-# See https://github.com/bootc-dev/bootc/issues/1826
-containerukifyargs+=(--karg enforcing=0)
-
 missing_verity=()
 
 if [[ $allow_missing_verity == "true" ]]; then
--- a/crates/xtask/src/tmt.rs
+++ b/crates/xtask/src/tmt.rs
@@ -31,10 +31,8 @@ const ENV_BOOTC_UPGRADE_IMAGE: &str = "BOOTC_upgrade_image";
 // Distro identifiers
 const DISTRO_CENTOS_9: &str = "centos-9";
 
-const COMPOSEFS_KERNEL_ARGS: [&str; 1] = ["--karg=enforcing=0"];
-
 // Import the argument types from xtask.rs
-use crate::{BootType, Bootloader, RunTmtArgs, SealState, TmtProvisionArgs};
+use crate::{Bootloader, RunTmtArgs, SealState, TmtProvisionArgs};
 
 /// Generate a random alphanumeric suffix for VM names
 fn generate_random_suffix() -> String {
@@ -488,11 +486,6 @@ pub(crate) fn run_tmt(sh: &Shell, args: &RunTmtArgs) -> Result<()> {
                 let filesystem = args.filesystem.as_deref().unwrap_or("ext4");
                 opts.push(format!("--filesystem={}", filesystem));
                 opts.push("--composefs-backend".into());
-
-                // UKI install fails with extra args
-                if args.boot_type == BootType::Bls {
-                    opts.extend(COMPOSEFS_KERNEL_ARGS.map(|x| x.into()));
-                }
             }
 
             if let Some(b) = &args.bootloader {
--- a/docs/src/experimental-composefs.md
+++ b/docs/src/experimental-composefs.md
@@ -159,7 +159,6 @@ The composefs backend is experimental; on-disk formats are subject to change.
 ### Deployment blockers
 
 - [Garbage collection](https://github.com/bootc-dev/bootc/pull/2040): In progress
-- [SELinux enforcement](https://github.com/bootc-dev/bootc/issues/1826): Currently requires `enforcing=0`
 - Extended install APIs: Ability to cleanly implement anaconda %post and osbuild post mutations and general post-install pre-reboot; right now some tools just mount the deployment directory (note this one also relates to [APIs in general](https://github.com/bootc-dev/bootc/issues/522))
 - [OCI registry install](https://github.com/bootc-dev/bootc/issues/1703): Installing from registry can fail due to config mismatch (suggestion: just clean reject v2s2)
 - [composefs-rs repository finalization](https://github.com/bootc-dev/bootc/issues/1320)


jeckersb:
Assuming this actually works properly (seems to work for me locally but we'll see what the matrix says), we should be able to revert the install parts of #2025, keeping the booted system part as well as the test re-enablement.

crates/lib/src/install.rs
```diff
@@ -1970,6 +1970,32 @@ async fn install_to_filesystem_impl(
         ostree_install(state, rootfs, cleanup).await?;
     }
 
+    // As the very last step before filesystem finalization, do a full SELinux
+    // relabel of the physical root filesystem.  We skip ostree/deploy because
```
cgwalters: I *think* (but there's $details here) we actually need to precisely skip deployment roots not just all of `ostree/deploy`.

However...can't we also drop the current ostree-only relabeling step with this? That would really prove things out.

jeckersb:
Looks like CI mostly worked, except:

```
error: Installing to filesystem: Final SELinux relabeling of physical root: No label found in policy 'Some("99a67cf6f12f746505cc0fdc62ddf3ac8fca59ed19f2eb227c17d0667aa68043")' for /sys/fs/cgroup/io.prio.class)
```

And yes I realized later in the evening after I had walked away from this that I forgot to remove the `enforcing=0` bit here :roll_eyes:

crates/lib/src/install.rs
jeckersb: Is there a difference in the labeling between ostree-with-composefs vs ostree-without-composefs?  In the composefs case I don't think it matters too much how accurate the labels are since the source of truth ends up in the erofs (but does it populate the erofs from the underlying files?  I admit I haven't looked too much into that before now...).

But in the non-composefs case (do we even use that anywhere?) the labels *are* the final labels.

cgwalters:
> /sys/fs/cgroup/io.prio.class

The relabeling code should probably be changed to use https://docs.rs/cap-std-ext/latest/cap_std_ext/dirext/trait.CapStdExtDirExt.html#tymethod.walk with noxdev

crates/lib/src/install.rs
cgwalters: > Is there a difference in the labeling between ostree-with-composefs vs ostree-without-composefs? In the composefs case I don't think it matters too much how accurate the labels are since the source of truth ends up in the erofs (but does it populate the erofs from the underlying files? I admit I haven't looked too much into that before now...).

There's not a difference AFAIK no, but a problem is that in the ostree code path we still sometimes for legacy reasons operate directly on a deployment root without the EROFS (running semodule e.g.) and so the labels need to be correct on the underlying files still too.

---

Anyways though what we're trying to solve here is the labeling of everything else; ostree should be getting labels of the deployment root correct.

jeckersb:
Doesn't seem enough for the files to get just any label, because `default_t` on the underlying storage is getting denied:

```
[  178.773356] systemd[1]: Starting ssh-host-keys-migration.service - Update OpenSSH host key permissions...
[  178.774513] audit: type=1400 audit(1772558678.864:100): avc:  denied  { read } for  pid=838 comm="9" path="/39/556cc67f965f16554cb7d0f115a486ad86971eaf7ea804daa445ca3488596f28601a2762ef3115f2bc2ef2faae14e50b1f46f72499a869d561d8209c092b5f" dev="vda3" ino=1064524 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:default_t:s0 tclass=file permissive=0
[  178.777454] systemd[1]: Starting sshd-keygen@ecdsa.service - OpenSSH ecdsa Server Key Generation...
[  178.779754] audit: type=1400 audit(1772558678.868:101): avc:  denied  { read } for  pid=839 comm="9" path="/39/556cc67f965f16554cb7d0f115a486ad86971eaf7ea804daa445ca3488596f28601a2762ef3115f2bc2ef2faae14e50b1f46f72499a869d561d8209c092b5f" dev="vda3" ino=1064524 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:default_t:s0 tclass=file permissive=0
[  178.781967] systemd[1]: Starting sshd-keygen@ed25519.service - OpenSSH ed25519 Server Key Generation...
[  178.785205] audit: type=1400 audit(1772558678.872:102): avc:  denied  { read } for  pid=840 comm="9" path="/39/556cc67f965f16554cb7d0f115a486ad86971eaf7ea804daa445ca3488596f28601a2762ef3115f2bc2ef2faae14e50b1f46f72499a869d561d8209c092b5f" dev="vda3" ino=1064524 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:default_t:s0 tclass=file permissive=0
[  178.787198] systemd[1]: Starting sshd-keygen@rsa.service - OpenSSH rsa Server Key Generation...
[  178.790744] audit: type=1400 audit(1772558678.878:103): avc:  denied  { read } for  pid=841 comm="9" path="/39/556cc67f965f16554cb7d0f115a486ad86971eaf7ea804daa445ca3488596f28601a2762ef3115f2bc2ef2faae14e50b1f46f72499a869d561d8209c092b5f" dev="vda3" ino=1064524 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:default_t:s0 tclass=file permissive=0
[  178.791656] systemd[1]: sshd-keygen@ecdsa.service: Main process exited, code=exited, status=127/n/a
[  178.797668] systemd[1]: sshd-keygen@ecdsa.service: Failed with result 'exit-code'.
```

cgwalters:
Ah right I think that's because this code needs to pass an override path of `/usr` like the ostree one does.

jeckersb:
Circling back to [this comment](https://github.com/composefs/composefs-rs/issues/183#issuecomment-3891363176) it feels like the most correct approach here would be to label all of the composefs objects with some custom type along with some policy glue (via https://fedoraproject.org/wiki/SELinux/IndependentPolicy ?) to make it function similarly to `usr_t`.  I guess initially we could use `usr_t` directly although that feels incorrect somehow.

cgwalters:
I agree it'd be better to have a dedicated type for composefs objects. In theory we could ship custom policy addition to add it now without blocking on upstream. There's a variety of examples of this, e.g. https://github.com/cockpit-project/cockpit/tree/main/selinux

I personally would say it seems way easier to just use `usr_t` and avoid that complexity, but I have no opposition to trying to add it.

jeckersb:
> > /sys/fs/cgroup/io.prio.class
> 
> The relabeling code should probably be changed to use https://docs.rs/cap-std-ext/latest/cap_std_ext/dirext/trait.CapStdExtDirExt.html#tymethod.walk with noxdev

xref for when I inevitably find myself back here in the future looking for it :arrow_right: https://github.com/bootc-dev/bootc/pull/2045

jeckersb:
Ok so coming back to this, I don't think we need to special-case the ostree bits at all, since the bulk relabel will only add a label where the file was previously `unlabeled_t`.  And since the ostree data is already labeled, it's a no-op in there.

The one thing we *do* need to do though is to explicitly label the composefs data, because that *is* unlabeled and there's no policy for it, so it ends up as `default_t` which doesn't work as noted above.

Next pass at this I'm going to:

1. rebase to pick up #2045
2. drop the part where we skip ostree since it's unnecessary
3. add code to explicitly re-label the composefs data as `/usr` for the time being; we can go more into setting up our own custom types and policy in a followup
4. turn off enforcing=0

That should get us at least closer to the desired end-state.  Progress bit by bit :smile:

cgwalters (approved):

PR #2036, opened 2026-03-03
docs/composefs: Note semantics with Secure Boot disabled

cgwalters:
Came up in chat.

### DIFF
--- a/docs/src/experimental-composefs.md
+++ b/docs/src/experimental-composefs.md
@@ -45,6 +45,22 @@ Sealed images also require:
 - Secure Boot support in the target system firmware
 - A filesystem with fsverity support (e.g., ext4, btrfs) for the root partition
 
+#### Using without Secure Boot
+
+You can use a sealed UKI without Secure Boot enabled. The composefs and mounting
+code is fully orthogonal to Secure Boot - the fsverity digest of the root filesystem
+and all of its contents will still be validated at runtime, which does provide
+an increased level of integrity.
+
+However: nothing validates that root digest itself, meaning any locally running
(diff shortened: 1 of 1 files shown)


jeckersb:
On a related note (maybe tag onto this, maybe a followup?) what if a user very specifically wants the image to fail when run with secure boot disabled?

Probably more of a followup because I could see that being a somewhat common case where for compliance reasons you'd want to bake "allow_insecure_boot = false" into the config in your image and we would check that very early in the boot process and panic if necessary.  It would give image authors reasonable (not absolute) guarantees that their image couldn't accidentally be deployed without secureboot.  More guardrails than anything, if you wanted to be *really* sure you'd have to pre-measure and check PCR values I guess.

cgwalters:
One can already today add a systemd unit with `ConditionSecurity=!uefi-secureboot` that performs any desired action, whether in the initramfs or real root.

I think you raise a valid point, but it's probably best addressed in a different doc; could be a midstream/downstream OS focused one, but could be tips&tricks here?

> if you wanted to be really sure you'd have to pre-measure and check PCR values I guess.

Yes this gets into "attestation" in the general case really.

That said I did see https://mjg59.dreamwidth.org/54203.html go by a while ago and there's probably some cool stuff possible there even for standalone desktop/single-node-home-server cases.

mrguitar:
Thanks for adding this. I've never once spoken with a user who likes/enjoys adding secure boot keys. We should assume X% won't mess with it and defaulting to a degraded security state is the right answer IMO. I think if users want something more strict as @jeckersb mentions they should opt into it, not the default. $0.02

PR #2040, opened 2026-03-04
Composefs gc

Johan-Liebert1:
### composefs/gc: Introduce `dry_run` command

With accordance to the new GC API, Introduce a dry_run parameter. Also,
remove custom object GC code and call the composefs-rs implementation

### composefs/gc: Fix orphan detection and staged deployment cleanup

Update "roots" to include all images/oci-configs we want to keep
especially the staged deployment if there is any

### composefs: Use boot entries as the source of truth

Up until now we were using the state directory as the source of truth
for what all deployments we have. This doesn't work in the case a GC
operation deletes the bootloader entries, but fails before deleting the
state directory. This would leave the output of `bootc status` command
in an undefined state. Even worse, we use kernel args, hence bootloader
entries to check for soft reboot capabilities, which would just brick as
we would have the state dir (so we'd assume the deployment is active),
but won't have the actual bootloader entry

### composefs/gc: GC bootloader binaries

Instead of deleting bootloader binaries in `delete_deployment` function,
add logic to GC them along with everything else. So our source or truth
becomes the bootloader entries which is also what we use for getting the
status.

The flow now becomes when deleting a deployment:
- Delete the bootloader entries (.conf files and user.cfg for grub)
- GC the bootloader binaries + images + state dir + composefs objects

This way the GC remains idempotent and we handle in-between deletion
failures gracefully.

Also, this gives us the opportunity to retrieve the deleted bootloader
entries using the data (binaries) from the /boot or /esp in the future

refactor: Add helper methods

Add `require_esp` method to Storage struct
Add `list_all_deployments` method to Host struct

### composefs/uki: Install all UKIs in EFI/Linux/bootc

We were making a distinction based on the bootloader and installing UKIs
in EFI/Linux for Grub and EFI/Linux/bootc for sd-boot. IMO it's better
if we use the same directory for both bootloaders

### composefs/gc: Call GC function at the end of update

Call the garbage collection function at the end of `do_update` function
which is the function that performs an update/switch. Make sure we also
take into account the staged entries here, as we wouldn't want to delete
the staged entry we just created by upgrading.

Add a helper method to get staged grub UKI entries

### DIFF
--- a/crates/lib/src/bootc_composefs/gc.rs
+++ b/crates/lib/src/bootc_composefs/gc.rs
@@ -7,15 +7,15 @@
 use anyhow::{Context, Result};
 use cap_std_ext::{cap_std::fs::Dir, dirext::CapStdExtDirExt};
 use composefs::repository::GcResult;
-use composefs_boot::bootloader::{EFI_ADDON_DIR_EXT, EFI_EXT};
+use composefs_boot::bootloader::EFI_EXT;
 
 use crate::{
     bootc_composefs::{
-        boot::{BOOTC_UKI_DIR, BootType, get_type1_dir_name},
+        boot::{BOOTC_UKI_DIR, BootType, get_type1_dir_name, get_uki_addon_dir_name, get_uki_name},
         delete::{delete_image, delete_staged, delete_state_dir},
         status::{get_composefs_status, get_imginfo, list_bootloader_entries},
     },
-    composefs_consts::{STATE_DIR_RELATIVE, TYPE1_BOOT_DIR_PREFIX},
+    composefs_consts::{STATE_DIR_RELATIVE, TYPE1_BOOT_DIR_PREFIX, UKI_NAME_PREFIX},
     store::{BootedComposefs, Storage},
 };
 
@@ -89,8 +89,12 @@ fn collect_uki_binaries(boot_dir: &Dir, boot_binaries: &mut Vec<BootBinary>) -> 
         let entry = entry?;
         let name = entry.file_name()?;
 
+        let Some(verity) = name.strip_prefix(UKI_NAME_PREFIX) else {
+            continue;
+        };
+
         if name.ends_with(EFI_EXT) {
-            boot_binaries.push((BootType::Uki, name));
+            boot_binaries.push((BootType::Uki, verity.into()));
         }
     }
 
@@ -151,7 +155,7 @@ fn delete_uki(storage: &Storage, uki_id: &str, dry_run: bool) -> Result<()> {
         let entry_name = entry.file_name()?;
 
         // The actual UKI PE binary
-        if entry_name == format!("{}{}", uki_id, EFI_EXT) {
+        if entry_name == get_uki_name(uki_id) {
             tracing::debug!("Deleting UKI: {}", entry_name);
 
             if dry_run {
@@ -159,7 +163,7 @@ fn delete_uki(storage: &Storage, uki_id: &str, dry_run: bool) -> Result<()> {
             }
 
             entry.remove_file().context("Deleting UKI")?;
-        } else if entry_name == format!("{}{}", uki_id, EFI_ADDON_DIR_EXT) {
+        } else if entry_name == get_uki_addon_dir_name(uki_id) {
             // Addons dir
             tracing::debug!("Deleting UKI addons directory: {}", entry_name);
 
--- a/crates/lib/src/bootc_composefs/boot.rs
+++ b/crates/lib/src/bootc_composefs/boot.rs
@@ -91,7 +91,6 @@ use rustix::{mount::MountFlags, path::Arg};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
-use crate::parsers::bls_config::{BLSConfig, BLSConfigType};
 use crate::task::Task;
 use crate::{
     bootc_composefs::repo::get_imgref,
@@ -109,6 +108,10 @@ use crate::{
     bootc_composefs::status::get_container_manifest_and_config, bootc_kargs::compute_new_kargs,
 };
 use crate::{bootc_composefs::status::get_sorted_grub_uki_boot_entries, install::PostFetchState};
+use crate::{
+    composefs_consts::UKI_NAME_PREFIX,
+    parsers::bls_config::{BLSConfig, BLSConfigType},
+};
 use crate::{
     composefs_consts::{
         BOOT_LOADER_ENTRIES, COMPOSEFS_CMDLINE, ORIGIN_KEY_BOOT, ORIGIN_KEY_BOOT_DIGEST,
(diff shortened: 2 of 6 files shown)


crates/lib/src/cli.rs
```diff
@@ -1901,6 +1908,17 @@ async fn run_from_opt(opt: Opt) -> Result<()> {
                     }
                 }
             }
+            InternalsOpts::ComposefsGC { dry_run } => {
+                let storage = &get_storage().await?;
+
+                match storage.kind()? {
+                    BootedStorageKind::Ostree(..) => todo!(),
```
Johan-Liebert1: we can probably call the ostree impl here?

Johan-Liebert1: Just bailed from here for now, as the command *IS* called `composefs-gc`

crates/lib/src/bootc_composefs/gc.rs
```diff
@@ -108,46 +59,137 @@ fn list_state_dirs(sysroot: &Dir) -> Result<Vec<String>> {
...
+            .open_dir(&dir_path)
+            .with_context(|| format!("Opening {dir_path}"))?;
+
+        // Scan inside this directory for kernel and initrd files
+        for file_entry in verity_dir.entries_utf8()? {
+            let file_entry = file_entry?;
+            let file_name = file_entry.file_name()?;
+
+            // Look for kernel and initrd files
+            if file_name.starts_with(VMLINUZ) {
```
cgwalters: For type1 i think we should actually use the `linux` and `initrd` keys from the config.

crates/lib/src/bootc_composefs/gc.rs
```diff
@@ -108,46 +59,137 @@ fn list_state_dirs(sysroot: &Dir) -> Result<Vec<String>> {
...
+        let entry = entry?;
+        let dir_path = entry.file_name()?;
+
+        if !entry.file_type()?.is_dir() {
+            continue;
+        }
+
+        // Check if the directory name looks like a verity hash
+        // 128 hex chars for SHA512, 64 hex chars for SHA256
+        // TODO: There might be a better way to do this
```
cgwalters: Yes definitely. Right now ostree prefixes all of its type1 entries with `ostree-` as a way to identify entries it owns.

I think it'd make sense to do something similar here (`composefs-`?)

Alternatively, we could have a magic comment in the file.

I don't have a strong opinion, but either would be better than a match on digest length.

crates/lib/src/bootc_composefs/gc.rs
```diff
@@ -216,12 +311,55 @@ pub(crate) async fn composefs_gc(storage: &Storage, booted_cfs: &BootedComposefs
...
+        booted_cfs.repo.gc_dry_run(&additional_roots)?
+    } else {
+        booted_cfs.repo.gc(&additional_roots)?
+    };
+
+    if dry_run {
+        println!("Dry run (no files deleted)");
+    }
+
+    println!(
```
cgwalters: I think we need to impose a rule (via clippy lint even) that there's no usage of `(e)println!` in library code, only the CLI.

In this case basically we should move the printing to the CLI.

crates/lib/src/bootc_composefs/gc.rs
Johan-Liebert1: We do that someplace else, but the main issue is the conf files being deleted, then say the process is aborted. In that case we have these orphan bootloader binaries that we need to keep track of

cgwalters: Yes, so for that ostree always puts its binaries in `/boot/ostree` and performs GC there. I think we can easily do the same like `/boot/composefs`?

crates/lib/src/bootc_composefs/gc.rs
```diff
@@ -54,6 +59,143 @@ fn list_state_dirs(sysroot: &Dir) -> Result<Vec<String>> {
...
+
+    Ok(())
+}
+
+#[fn_error_context::context("Deleting kernel and initrd")]
+fn delete_kernel_initrd(storage: &Storage, entry_to_delete: &str, dry_run: bool) -> Result<()> {
+    let boot_dir = storage.require_boot_dir()?;
+
+    let path = readlink(format!("/proc/self/fd/{}", boot_dir.as_raw_fd()), [])
+        .context("Getting kernel path")?;
```
cgwalters: I don't think we need to do this just for a logging message.

crates/lib/src/bootc_composefs/gc.rs
Johan-Liebert1: > I think it'd make sense to do something similar here (composefs-?)

this sounds good to me.

crates/lib/src/bootc_composefs/gc.rs
Johan-Liebert1: But we can't as systemd-boot binaries are stored in the ESP. We do have the `boot_dir` property in `Storage` which points to `/sysroot/boot` or the ESP depending on the bootloader

crates/lib/src/bootc_composefs/gc.rs
Johan-Liebert1: Okay. Not sure how easy/hard it would be for such a clippy rule as we have `println` throughout

crates/lib/src/bootc_composefs/gc.rs
Johan-Liebert1: Right. We can just print the digest and that should be fine

crates/lib/src/bootc_composefs/gc.rs
Johan-Liebert1: I settled on `bootc_composefs-` as the prefix

tmt/tests/booted/test-composefs-gc.nu
```diff
@@ -135,6 +139,14 @@ def third_boot [] {
 }
 
 def fourth_boot [] {
+    let bootloader = (bootc status --json | from json).status.booted.composefs.bootloader
+
+    if ($bootloader | str downcase) == "systemd" {
+        # TODO: Some concrete API for this would be great
```
cgwalters, line 145: Valid but why do we need to mount it in this test?

crates/lib/src/bootc_composefs/gc.rs
cgwalters: > But we can't as systemd-boot binaries are stored in the ESP. 

Do you mean UKIs? Or do you mean systemd-boot itself? We should never GC that.

For UKIs we can apply the same filename prefixing I think since it shouldn't appear to the user.

tmt/tests/booted/test-composefs-gc.nu
Johan-Liebert1: Yes, because for systemd-boot we store everything in the ESP. The BLS .conf files + the Kernel + initrd


> Do you mean UKIs? Or do you mean systemd-boot itself? We should never GC that.

I don't mean the systemd-boot binary, and yes we shouldn't touch that. It's for the same reason as as my comment above 

I can't reply to this in the convo below... Thanks GitHub, very cool

Johan-Liebert1: > For UKIs we can apply the same filename prefixing I think since it shouldn't appear to the user

Hmm.. yeah, I should've done that along with the BLS renaming. I'll update this

PR #2042, opened 2026-03-04
Move blockdev code from bootupd into blockdev crate

ckyrouac:
Consolidating the blockdev code that was scattered across bootupd and bootc into the blockdev crate. This is prep and cleanup for multi device parent support in bootupd. This needs to be merged before the bootupd changes can be merged.  Then, after the multi device support changes to bootupd are merged, I'll open another PR against bootc to integrate with the multi device bootupd changes.

### DIFF
--- a/crates/blockdev/src/blockdev.rs
+++ b/crates/blockdev/src/blockdev.rs
@@ -279,34 +279,48 @@ impl Device {
         }
     }
 
-    /// Walk the parent chain to find the root (whole disk) device.
+    /// Walk the parent chain to find all root (whole disk) devices,
+    /// and fail if more than one root is found.
     ///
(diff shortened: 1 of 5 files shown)


crates/blockdev/src/blockdev.rs
```diff
@@ -284,7 +284,7 @@ impl Device {
     /// Returns the root device with its children (partitions) populated.
     /// If this device is already a root device, returns a clone of `self`.
     /// Fails if the device has multiple parents at any level.
-    pub fn root_disk(&self) -> Result<Device> {
+    pub fn find_single_root(&self) -> Result<Device> {
```
cgwalters: Can't this code use `find_all_roots()` now and require one result?

Issue #2038, opened 2026-03-03
Install/aleph: use `target-imgref` reference


jbtrystram:
When creating a coreOS disk image with `osbuild` I get the following `.boot-aleph.json`: 

```json
{
  "image": "/tmp/tmpb29j6pi3/image",
  "kernel": "6.18.12-200.fc43.x86_64",
  "selinux": "disabled",
  "timestamp": null,
  "version": "43.20260301.20.dev1"
}
```

Which isn't really helpful. I suppose that's due to the way osbuild copy the image around before passing it to bootc install. 
However in the logs I see: `Installing image: docker://ostree-image-signed:docker://quay.io/fedora/fedora-coreos:testing-devel`, which seems to originate from : 

https://github.com/bootc-dev/bootc/blob/864b7f063f6c4f89e1e8d7a6a6972d4b53581ab9/crates/lib/src/install.rs#L1617

But target-imgref is the image to fetch for subsequent updates. So is that print output really correct ?

Also, let's add the input image labels as it often contains useful info injected by the build system (e.g. konflux).

For reference, here is what the coreOS aleph looks like now : 
```json
{
    "container-image": {
        "image-digest": "sha256:bb0e7c76a263f1308892f984e7ebfaa785ebf1f2ae2954188b8c58222512b872",
        "image-labels": {
            "com.coreos.inputhash": "fe9883169714c593d98058606e886b9747710ed15ab1b9cdbd7fa538fb435b3c",
            "com.coreos.osname": "fedora-coreos",
            "com.coreos.stream": "testing-devel",
            "containers.bootc": "1",
            "io.buildah.version": "1.42.2",
            "org.opencontainers.image.description": "Fedora CoreOS testing-devel",
            "org.opencontainers.image.revision": "233fe18749c7d2749581e4307c4cac60967acde4",
            "org.opencontainers.image.source": "git@github.com:jbtrystram/fedora-coreos-config.git",
            "org.opencontainers.image.title": "Fedora CoreOS testing-devel",
            "org.opencontainers.image.version": "43.20260301.20.dev0",
            "ostree.bootable": "1",
            "ostree.commit": "b89ba7137667d369dd5ff646141e3bfd8aa645a4bae6103d8d7cbb031aa5a2e0",
            "ostree.final-diffid": "sha256:12787d84fa137cd5649a9005efe98ec9d05ea46245fdc50aecb7dd007f2035b1"
        },
[...]

PR #2043, opened 2026-03-04
install/aleph: include the image labels in aleph

jbtrystram:
Include the container labels in the aleph file, since they often contain useful information about the image provenance, such as the source commit the image was build from.

Also we skip serializing the source image reference if it start with `/tmp` since this is a good signal it was source from a local copy of an image, e.g. in an osbuild environnement.

Whith this, a build of Fedora CoreOS through osbuild goes from:

```
{
  "image": "/tmp/tmpb29j6pi3/image",
  "kernel": "6.18.12-200.fc43.x86_64",
  "selinux": "disabled",
  "timestamp": null,
  "version": "43.20260301.20.dev1"
}
```

to
```
{
  "digest": "sha256:07bf537cc4e4d208eb0b978f76e5046e55529ce6192b982d8c1a41fa1d61b95a",
  "kernel": "6.18.13-200.fc43.x86_64",
  "labels": {
    "com.coreos.inputhash": "fe9883169714c593d98058606e886b9747710ed15ab1b9cdbd7fa538fb435b3c",
    "com.coreos.osname": "fedora-coreos",
    "com.coreos.stream": "testing-devel",
    "containers.bootc": "1",
    "io.buildah.version": "1.42.2",
    "org.opencontainers.image.description": "Fedora CoreOS testing-devel",
    "org.opencontainers.image.revision": "233fe18749c7d2749581e4307c4cac60967acde4",
    "org.opencontainers.image.source": "git@github.com:jbtrystram/fedora-coreos-config.git",
    "org.opencontainers.image.title": "Fedora CoreOS testing-devel",
    "org.opencontainers.image.version": "43.20260301.20.dev1",
    "ostree.bootable": "1",
    "ostree.commit": "89635f7cba9de932fc60d71a6bded65ad0db06a35c9d016da03ca7ade9ba4736",
    "ostree.final-diffid": "sha256:12787d84fa137cd5649a9005efe98ec9d05ea46245fdc50aecb7dd007f2035b1"
  },
  "selinux": "disabled",
  "target-image": "ostree-image-signed:docker://quay.io/fedora/fedora-coreos:testing-devel",
  "timestamp": null,
  "version": "43.20260301.20.dev1"
}
```
which is way more useful.

See https://github.com/bootc-dev/bootc/issues/2038

Assisted-by: OpenCode(Opus 4.6)

### DIFF
--- a/crates/lib/src/install/aleph.rs
+++ b/crates/lib/src/install/aleph.rs
@@ -1,3 +1,5 @@
+use std::collections::BTreeMap;
+
 use anyhow::{Context as _, Result};
 use canon_json::CanonJsonSerialize as _;
 use cap_std_ext::{cap_std::fs::Dir, dirext::CapStdExtDirExt as _};
@@ -15,9 +17,19 @@ pub(crate) const BOOTC_ALEPH_PATH: &str = ".bootc-aleph.json";
 /// be used to trace things like the specific version of `mkfs.ext4` or
 /// kernel version that was used.
 #[derive(Debug, Serialize)]
+#[serde(rename_all = "kebab-case")]
 pub(crate) struct InstallAleph {
     /// Digested pull spec for installed image
-    pub(crate) image: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub(crate) image: Option<String>,
+    /// The manifest digest of the installed image
+    pub(crate) digest: String,
+    /// The target image reference, used for subsequent updates
+    #[serde(rename = "target-image")]
+    pub(crate) target_image: String,
+    /// The OCI image labels from the installed image
+    #[serde(skip_serializing_if = "BTreeMap::is_empty")]
+    pub(crate) labels: BTreeMap<String, String>,
     /// The version number
     pub(crate) version: Option<String>,
     /// The timestamp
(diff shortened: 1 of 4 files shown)


crates/lib/src/install/aleph.rs
```diff
@@ -17,7 +19,16 @@ pub(crate) const BOOTC_ALEPH_PATH: &str = ".bootc-aleph.json";
 #[derive(Debug, Serialize)]
 pub(crate) struct InstallAleph {
     /// Digested pull spec for installed image
-    pub(crate) image: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub(crate) image: Option<String>,
+    /// The manifest digest of the installed image
+    pub(crate) digest: String,
+    /// The target image reference, used for subsequent updates
+    #[serde(rename = "target-image")]
```
cgwalters, line 28: I think we can just `rename-all = kebab-case` on the struct

cgwalters:
Missing DCO

jbtrystram:
> Could also use a test in the tmt readonly tests.

Added a test and signed-off the commit

cgwalters:
@jbtrystram failing tests

jbtrystram:
> @jbtrystram failing tests

Thanks for the ping @cgwalters . Fixed. But the remaining test failures highlight that in the composefs path, bootc does not write an aleph file. Is it an oversignt or intended ? 
cc @Johan-Liebert1

Johan-Liebert1:
>  But the remaining test failures highlight that in the composefs path, bootc does not write an aleph file

Had a chat about this. We don't write one currently, but we should. For now I think we can put `fixme_skip_if_compoesfs: true` in the test to skip it for composefs backend?

cgwalters:
The more I think about the more I feel the aleph file is a legacy and we should actually be aiming to drop it.

Like, among the CoreOS team, has anyone ever used it for anything that we know of recently?

Maybe it's better to just intentionally not write it with composefs.

jbtrystram:
Ok so the two upgrade tests are failing because the base disk image is built with the published bootc base image, which don't contain the changes in this PR. So that base image will have an incomplete aleph in regard to the test added here. That will fix itself once this PR merges and a new base image will be published with theses changes. 

Not sure how to handle that.

cgwalters:
Yes this is an interesting corner case. Any time we introduce a change which only happens at install time and add a readonly test, the upgrade tests will by design not see that change.

I think the simplest thing here is - skip those tests if `$env.BOOTC_test_upgrade_image` is set.

(Although of course it's quite subtle because eventually of course we will have this change in the source base images right now...but per previous discussion I'd like to eventually have the upgrade tests be *fixed* to a version by default)

jmarrero:
s390x has a known issue: https://github.com/fedora-copr/copr/issues/4219 this can be merged now. ~~@ckyrouac can you force this in?~~ I guess these are non-blocking.

PR #2044, opened 2026-03-04
Update composefs, unified storage

### LINKED ISSUE #1977: `bootc container ukify` fails with Operation not supported
```bash
+ bootc container ukify --rootfs /run/target --karg enforcing=0 -- --signtool sbsign --secureboot-private-key /run/secrets/secureboot_key --secureboot-certificate /run/secrets/secureboot_cert
 --measure --json pretty --output /out/6.12.0-192.el10.x86_64.efi
error: Building UKI: Operation not supported (os error 95)
Error: building at STEP "RUN --network=none --mount=type=tmpfs,target=/run --mount=type=tmpfs,target=/tmp --mount=type=secret,id=secureboot_key --mount=type=secret,id=secureboot_cert --mount=
type=bind,from=packaging,src=/,target=/run/packaging --mount=type=bind,from=base-penultimate,src=/,target=/run/target <<EORUN": while running runtime: exit status 1
```

This occurs when building the sealed variant for testing with `just variant=composefs-sealeduki-sdboot test-tmt`

cgwalters:
The latest composefs-rs stores manifest and config objects and
the manifest becomes a GC root, so we can use that instead of `.imginfo`
sidecar files.

The flow now is:
 bootloader entry -> deployment -> origin file
   -> manifest digest -> manifest -> [config | objects]

For backward compatibility, fall back to the legacy .imginfo
file if the .origin does not contain a manifest_digest key.

Drop the really old hacky fallback that did network fetches.

Note the manifest becomes part of the GC root.

Closes: https://github.com/bootc-dev/bootc/issues/1977

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,13 +46,18 @@ clap_mangen = { version = "0.3.0" }
 # The Justfile will auto-detect these and bind-mount them into container builds.
 cfsctl = { git = "https://github.com/composefs/composefs-rs", rev = "2203e8f", package = "cfsctl", features = ["rhel9"] }
 fn-error-context = "0.2.1"
+futures-util = "0.3"
 hex = "0.4.3"
+http-body-util = "0.1"
+hyper = { version = "1", features = ["client", "http1"] }
+hyper-util = { version = "0.1", features = ["tokio"] }
 indicatif = "0.18.0"
 indoc = "2.0.5"
 libc = "0.2.154"
 log = "0.4.21"
 openssl = "0.10.72"
 owo-colors = { version = "4" }
+percent-encoding = "2"
 regex = "1.10.4"
 # For the same rationale as https://github.com/coreos/rpm-ostree/commit/27f3f4b77a15f6026f7e1da260408d42ccb657b3
 rustix = { "version" = "1", features = ["use-libc", "thread", "net", "fs", "system", "process", "mount"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,7 @@ clap_mangen = { version = "0.3.0" }
 #   [patch."https://github.com/composefs/composefs-rs"]
 #   cfsctl = { path = "/path/to/composefs-rs/crates/cfsctl" }
 # The Justfile will auto-detect these and bind-mount them into container builds.
-cfsctl = { git = "https://github.com/composefs/composefs-rs", rev = "2203e8f", package = "cfsctl", features = ["rhel9"] }
+cfsctl = { git = "https://github.com/cgwalters/composefs-rs", rev = "c04bc6fb16", package = "cfsctl" }
 fn-error-context = "0.2.1"
 futures-util = "0.3"
 hex = "0.4.3"
@@ -104,7 +104,7 @@ bins = ["skopeo", "podman", "ostree", "zstd", "setpriv", "systemctl", "chcon"]
 unsafe_code = "deny"
 # Absolutely must handle errors
 unused_must_use = "forbid"
-missing_docs = "deny"
+missing_docs = "allow"
 missing_debug_implementations = "deny"
 # Feel free to comment this one out locally during development of a patch.
 dead_code = "deny"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,7 @@ clap_mangen = { version = "0.3.0" }
 #   [patch."https://github.com/composefs/composefs-rs"]
 #   cfsctl = { path = "/path/to/composefs-rs/crates/cfsctl" }
 # The Justfile will auto-detect these and bind-mount them into container builds.
-cfsctl = { git = "https://github.com/cgwalters/composefs-rs", rev = "c04bc6fb16", package = "cfsctl" }
+cfsctl = { git = "https://github.com/composefs/composefs-rs", rev = "54d248f7a7", package = "cfsctl" }
 fn-error-context = "0.2.1"
 futures-util = "0.3"
 hex = "0.4.3"
--- a/crates/lib/src/bootc_composefs/boot.rs
+++ b/crates/lib/src/bootc_composefs/boot.rs
@@ -61,10 +61,10 @@
 //! 1. **Primary**: New/upgraded deployment (default boot target)
 //! 2. **Secondary**: Currently booted deployment (rollback option)
 
-use std::ffi::OsStr;
 use std::fs::create_dir_all;
 use std::io::Write;
 use std::path::Path;
+use std::sync::Arc;
 
 use anyhow::{Context, Result, anyhow, bail};
 use bootc_kernel_cmdline::utf8::{Cmdline, Parameter};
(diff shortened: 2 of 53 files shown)


cgwalters:
OK, this is passing tests now and I think is a notable cleanup.

I think this may be one of the last ~breaking changes for the composefs installation layout.

cgwalters:
Actually thinking about this more...it feels like we should really be creating tags in the cfs repo instead of using this additional root stuff. That's effectively what ostree does with `ostree/` refs. Will look at a PoC

Basically we create `localhost/bootc-<manifestsha>` tags that are "owned" by bootc, and our GC is:

- walk over all bootloader entires, gather their origin images and load the manifest sha digest
- load all `localhost/bootc-` prefixed tags in the cfs repo
- Remove all unreferenced tags
- Just run `cfsctl gc` effectively

In particular a key thing is that this works if the composefs repo holds other data such as app images.

crates/lib/src/bootc_composefs/state.rs
```diff
@@ -44,6 +44,27 @@ use crate::{
...
+#[context("Reading origin for deployment {deployment_id}")]
+pub(crate) fn read_origin(sysroot: &Dir, deployment_id: &str) -> Result<Option<tini::Ini>> {
+    let depl_state_path = std::path::PathBuf::from(STATE_DIR_RELATIVE).join(deployment_id);
+
+    let Some(state_dir) = sysroot.open_dir_optional(&depl_state_path)? else {
+        return Ok(None);
+    };
+
+    let origin_filename = format!("{deployment_id}.origin");
+    let Some(origin_contents) = state_dir.read_to_string_optional(&origin_filename)? else {
```
Johan-Liebert1, line 59: The `.origin` file should always exist right?

Cargo.toml
```diff
@@ -114,3 +114,6 @@ todo = "deny"
 needless_borrow = "allow"
 needless_borrows_for_generic_args = "allow"
 
+[patch."https://github.com/composefs/composefs-rs"]
```
Johan-Liebert1: I don't think we need this

crates/lib/src/bootc_composefs/gc.rs
```diff
@@ -259,73 +248,133 @@ pub(crate) async fn composefs_gc(
...
         .iter()
-        .filter(|i| !bootloader_entries.contains(i))
-        .chain(bootloader_entries.iter().filter(|b| !images.contains(b)))
-        .collect::<Vec<_>>();
+        .filter(|s| !bootloader_entries.contains(s))
+        .collect();
 
-    tracing::debug!("img_bootloader_diff: {img_bootloader_diff:#?}");
+    // Bootloader entries without a state dir are from interrupted cleanups.
+    let orphaned_boot_entries: Vec<_> = bootloader_entries
```
Johan-Liebert1, line 306: this will always be an empty vec iiuc? Since we always delete the bootloader entries first

Cargo.toml
cgwalters: We do it's for https://github.com/composefs/composefs-rs/pull/263

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -1250,25 +1198,36 @@ fn get_secureboot_keys(fs: &Dir, p: &str) -> Result<Option<SecurebootKeys>> {
...
+        repo.set_insecure();
+    }
+
+    let repo = Arc::new(repo);
+
+    // Generate the bootable EROFS image (idempotent).
+    let id = composefs_oci::generate_boot_image(&repo, &pull_result.manifest_digest)
+        .context("Generating bootable EROFS image")?;
+
+    // Get boot entries from the OCI filesystem (untransformed).
```
Johan-Liebert1: comment seems out of place here. Maybe it was intended for the next line? Also, I don't think we "transform" the boot entries in any way

crates/lib/src/bootc_composefs/digest.rs
```diff
@@ -29,9 +29,11 @@ pub(crate) fn new_temp_composefs_repo() -> Result<(TempDir, Arc<ComposefsReposit
 
     td_dir.create_dir("repo")?;
     let repo_dir = td_dir.open_dir("repo")?;
-    let mut repo = ComposefsRepository::open_path(&repo_dir, ".").context("Init cfs repo")?;
+    let (mut repo, _created) =
+        ComposefsRepository::init_path(&repo_dir, ".", Algorithm::SHA512, false)
+            .context("Init cfs repo")?;
     // We don't need to hard require verity on the *host* system, we're just computing a checksum here
-    repo.set_insecure(true);
+    repo.set_insecure();
```
Johan-Liebert1, line 37: a bit confusing as I'd assume passing `enable_verity = false` in the `init_path` should've set this automatically

crates/lib/src/bootc_composefs/repo.rs
```diff
@@ -47,107 +60,246 @@ pub(crate) async fn initialize_composefs_repository(
...
-/// docker-daemon:some-image-id
-pub(crate) fn get_imgref(transport: &str, image: &str) -> String {
-    let img = image.strip_prefix(":").unwrap_or(&image);
-    let transport = transport.strip_suffix(":").unwrap_or(&transport);
-
-    if transport == "registry" || transport == "docker://" {
+/// The `spec::ImageReference` stores transport as a string (e.g. "registry:",
+/// "containers-storage:"). This parses that into a proper typed reference
+/// that renders correctly for skopeo (e.g. "docker://quay.io/some-image").
+pub(crate) fn get_imgref(transport: &str, image: &str) -> Result<containers_image_proxy::ImageReference> {
```
Johan-Liebert1: we should be able to use `containers_image_proxy::Transport`

cgwalters:
This now requires https://github.com/composefs/composefs-rs/pull/218

cgwalters:
We were discussing this and @Johan-Liebert1 was suggesting we don't yet turn it on by default, but wire up to the existing opt-in flags.

cgwalters:
Awesome! This one is finally passing all the tests.

We need to merge https://github.com/composefs/composefs-rs/pull/285 and then rebase this, and it could use a bit more cleanup, but then should be good to go.

cgwalters:
OK! Now that we have a release out, I think we can get this one in! Obviously a huge change...

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -594,16 +546,21 @@ pub(crate) fn setup_composefs_bls_boot(
                 Utf8PathBuf::from("/sysroot"),
                 esp_dev.path(),
                 cmdline,
-                fs,
                 bootloader,
             )
         }
     };
 
+    // Remove "root=" from kernel cmdline as systemd-auto-gpt-generator should use DPS
```
Johan-Liebert1: This was removed here https://github.com/bootc-dev/bootc/commit/103a8cdd094ddd1df01c59f1c6948bfcd76708d1

crates/lib/src/bootc_composefs/state.rs
```diff
@@ -245,12 +264,13 @@ pub(crate) async fn write_composefs_state(
...
-        path_relative_to(state_path.as_std_path(), actual_var_path.as_std_path())
-            .context("Getting var symlink path")?,
-        state_path.join("var"),
-    )
-    .context("Failed to create symlink for /var")?;
+    let var_target = path_relative_to(state_path.as_std_path(), actual_var_path.as_std_path())
+        .context("Getting var symlink path")?;
+    match symlink(&var_target, state_path.join("var")) {
+        Ok(()) => {}
+        Err(e) if e.kind() == std::io::ErrorKind::AlreadyExists => {}
```
Johan-Liebert1: The case where var symlnk would already exist would be if the new pulled repository and hence deployment has the same fs-verity as one of the deployments already present. I think this case should be handled much earlier than populating the state directory. imo, this should be an error if var symlink exists

Johan-Liebert1 (commented):
Overall looks okay. A few comments

crates/lib/src/bootc_composefs/boot.rs
cgwalters: This was a rebase artifact/mistake

cgwalters: Well spotted!

crates/lib/src/bootc_composefs/state.rs
cgwalters: After live discussion, we decided for now to make `bootc switch` just error out when the image verity is identical for now (and we'd probably want to implement `bootc switch --in-place` or so that would be explicitly documented to not create a new bootloader entry, and would instead *require* that the image digest is identical)

The other option is to implement what ostree does with a "deployment counter" that creates distinct `/etc` as well, but it's not clear that's worth it.

Johan-Liebert1 (approved):

cgwalters:
Man, it was just *one* job that flaked on koji. https://github.com/bootc-dev/infra/issues/143#issuecomment-4380214380

Issue #2034, opened 2026-03-02
Add `bootc usr-overlay --read-only`


evan-goode:
Similar to `ostree admin unlock --transient`, `bootc usr-overlay` should have a `--read-only`/`--readonly`/`--transient` flag to create a read-only overlay instead of a writable overlay.

I would personally prefer `--read-only` over `--transient`. To me, "transient" sounds more like "disappears upon reboot" than "read-only". But I'm sure I'm missing some context from OSTree history.

We'd like to use this for `dnf --transient`, both dnf4 and dnf5. Currently, DNF 4 calls `ostree admin unlock --transient` which does not work with the composefs backend. `dnf5 --transient` is not yet implemented: https://github.com/rpm-software-management/dnf5/issues/2358.

I'd be happy to work on this myself.

cgwalters:
`--read-only` SGTM!

Issue #1891, opened 2026-01-06
Gate on `update-generated`


jeckersb:
Spun off from https://github.com/bootc-dev/bootc/pull/1835#issuecomment-3715986839

We should ensure that `just update-generated` is a no-op.  In other words, if the PR changes something that necessitates updating the generated content, then it should be included as part of the PR.

jeckersb:
My first thought is just to inline this as part of the `just validate` flow, although that would entail building the `xtask` crate as part of that flow (I don't think it happens currently) so it might slow things a bit.  It obviously could be split into its own target (`validate-generated` or whatever) as well.

cgwalters:
Fine by me in having this in `just validate`.

One concern Chris raised is that right now we deploy the website from git main - which means if we require updating the man pages right away then the website will show unreleased changes.

In the medium term though...I think we need to do at least one of two things:

- Add explicit "Added in version <version>" to each option (like systemd does it e.g.)
- Add versioning to the website (like Rust crates do it)

The latter I think could be done with a bit of CI work, there has to be some prior art for it too.

cgwalters:
I was looking at things related to this and an issue right now is `just validate` runs xtask code in Rust, and so for consistency we do that as part of the main container build.

But there are things we want to validate (like the website) that aren't copied into that build intentionally.

Like in theory the scope of "validate" is everything stored in git. Maybe it's simplest to just require a host Rust toolchain and run the xtask stuff there, but in the general case it would require building the xtask code twice.

A different option I guess is that since validate conceptually has no "output" other than stdout/stderr (it doesn't result in binaries/containers) we could have it be done in a `podman run` using the devenv toolchain, though I think an annoying thing there is that `--mount=type=cache` isn't exposed to that, but we could probably hack it with a volume.

PR #2046, opened 2026-03-04
Add `usroverlay --readonly`

evan-goode:
Resolves https://github.com/bootc-dev/bootc/issues/2034.

### DIFF
--- a/tmt/tests/booted/test-usroverlay.nu
+++ b/tmt/tests/booted/test-usroverlay.nu
@@ -8,33 +8,46 @@ use std assert
 use tap.nu
 use bootc_testlib.nu
 
+def usr_is_writable []: nothing -> bool {
+    (do -i { /bin/test -w /usr } | complete | get exit_code) == 0
+}
+
 # Status should initially report no overlay in JSON
-let status_json_before = bootc status --json | from json
-assert ($status_json_before.status.usrOverlay? == null)
+let status_before = bootc status --json | from json
+assert ($status_before.status.usrOverlay? == null)
 
 # We should start out in a non-writable state on each boot
-let is_writable = (do -i { /bin/test -w /usr } | complete | get exit_code) == 0
-assert (not $is_writable)
+assert (not (usr_is_writable))
 
 def initial_run [] {
     bootc usroverlay
-    let is_writable = (do -i { /bin/test -w /usr } | complete | get exit_code) == 0
-    assert ($is_writable)
+    assert (usr_is_writable)
 
     # After `usroverlay`, status JSON should report a transient read/write overlay
-    let status_json_after = bootc status --json | from json
-    let overlay = $status_json_after.status.usrOverlay
+    let status_after = bootc status --json | from json
+    let overlay = $status_after.status.usrOverlay
     assert ($overlay.accessMode == "readWrite")
     assert ($overlay.persistence == "transient")
 
     bootc_testlib reboot
 }
 
-# The second boot; verify we're in the derived image
 def second_boot [] {
-    # After reboot, usr overlay should be gone
+    # After reboot, /usr overlay should be gone
     let status_after_reboot = bootc status --json | from json
     assert ($status_after_reboot.status.usrOverlay? == null)
+    # And /usr should not be writable
+    assert (not (usr_is_writable))
+
+    # Mount a read-only /usr overlay
+    bootc usroverlay --read-only
+    assert (not (usr_is_writable))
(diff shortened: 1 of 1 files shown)


docs/src/man/bootc-usr-overlay.8.md
```diff
@@ -1,16 +1,16 @@
 # NAME
 
-bootc-usr-overlay - Adds a transient writable overlayfs on `/usr` that
-will be discarded on reboot
+bootc-usr-overlay - Adds a transient overlayfs on `/usr` that will be discarded
+on reboot
 
 # SYNOPSIS
 
 **bootc usr-overlay** \[*OPTIONS...*\]
```
cgwalters, line 8: `just update-generated` should resync the new options

docs/src/man/bootc.8.md
```diff
@@ -30,7 +30,7 @@ pulled and `bootc upgrade`.
 | **bootc rollback** | Change the bootloader entry ordering; the deployment under `rollback` will be queued for the next boot, and the current will become rollback.  If there is a `staged` entry (an unapplied, queued upgrade) then it will be discarded |
 | **bootc edit** | Apply full changes to the host specification |
 | **bootc status** | Display status |
-| **bootc usr-overlay** | Add a transient writable overlayfs on `/usr` |
+| **bootc usr-overlay** | Add a transient overlayfs on `/usr` |
```
ckyrouac, line 33: could we have a more descriptive doc explaining the purpose of --readonly?

docs/src/man/bootc-usr-overlay.8.md
evan-goode: Done in a separate commit

docs/src/man/bootc.8.md
evan-goode: I added more info under the `--read-only` flag, wdyt?

tmt/plans/integration.fmf
```diff
@@ -182,8 +182,15 @@ execute:
     test:
       - /tmt/tests/tests/test-34-user-agent
 
+/plan-35-upgrade-preflight-disk-check:
```
ckyrouac, line 192: not a huge deal but I think these tmt/* changes belong in the previous commit that adds the test code

tmt/plans/integration.fmf
```diff
@@ -182,8 +182,15 @@ execute:
...
+  summary: Verify pre-flight disk space check rejects images with inflated layer sizes
+  discover:
+    how: fmf
+    test:
+      - /tmt/tests/tests/test-35-upgrade-preflight-disk-check
```
ckyrouac, line 197: I think this might have been a bad rebase? I believe this should point to the new `test-usroverlay.nu` file from the previous commit

ckyrouac (changes requested):

tmt/plans/integration.fmf
evan-goode: Hmm, just tried `just update-generated` again over f8eef4a485f0f082d8d7a77fa323e3af27ae65ae and got an identical diff. Maybe `update-generated` was not run after https://github.com/bootc-dev/bootc/pull/2022 or https://github.com/bootc-dev/bootc/pull/1995?

jeckersb: Yeah we have https://github.com/bootc-dev/bootc/issues/1891 to hopefully gate on `update-generated` so we don't continue getting into this state where things are out of date.  Known problem unfortunately :frowning_face:

tmt/plans/integration.fmf
evan-goode: Still want me to squash these? Maybe it's better to keep them separate if the latter commit has unrelated changes from `update-generated`.

ckyrouac: I think you might need to revert the update-generated commit, rebase against origin/main, then re-run update-generated. After that, the update-generated changes should make more sense and keeping them in a separate commit is ok

evan-goode: Oh gotcha. Rebased and I just put the `update-generated` changes in with the doc commit.

cgwalters:
Ah annoying, this one hit flakes enough that the artifacts from the `package-` job expired.

Hmmm...you know what we should really do is cut over to merge queues. Our per-PR gating job could be more minimal, but we could do more expensive CI on PR groups.

If we'd been doing that this one would almost certainly have merged OK.

PR #2048, opened 2026-03-05
bwrap: Bind-mount host /dev instead of creating fresh devtmpfs

ckyrouac:
Replace `--dev /dev` with `--dev-bind /dev /dev` in the bwrap container setup so that lsblk inside the sandbox can properly enumerate partition children of block devices (e.g. loop devices).

The previous approach created a minimal devtmpfs that lacked complete device information, causing ESP partition discovery to fail inside the bwrap sandbox. With a full bind-mount of host /dev, the per-device bind_device() mechanism is no longer needed and is removed.

Additionally, bind-mount /run/udev into the sandbox when it exists so that lsblk and libblkid can read the udev database for partition type GUIDs and other device properties. Without this, tools that query device metadata (e.g. PARTTYPE) would get incomplete results even with /dev properly mounted.

Assisted-by: Claude Code (Opus 4)

### DIFF
--- a/crates/utils/src/bwrap.rs
+++ b/crates/utils/src/bwrap.rs
@@ -17,8 +17,6 @@ pub struct BwrapCmd<'a> {
     chroot_path: Cow<'a, Utf8Path>,
     /// Bind mounts in format (source, target)
     bind_mounts: Vec<(&'a str, &'a str)>,
-    /// Device nodes to bind into the container
-    devices: Vec<&'a str>,
     /// Environment variables to set
     env_vars: Vec<(&'a str, &'a str)>,
 }
@@ -31,7 +29,6 @@ impl<'a> BwrapCmd<'a> {
         Self {
             chroot_path: Cow::Owned(Utf8PathBuf::from(&fd_path)),
             bind_mounts: Vec::new(),
-            devices: Vec::new(),
             env_vars: Vec::new(),
         }
     }
@@ -41,7 +38,6 @@ impl<'a> BwrapCmd<'a> {
         Self {
             chroot_path: Cow::Borrowed(path),
             bind_mounts: Vec::new(),
-            devices: Vec::new(),
             env_vars: Vec::new(),
         }
     }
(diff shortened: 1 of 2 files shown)


crates/utils/src/bwrap.rs
```diff
@@ -79,19 +69,20 @@ impl<'a> BwrapCmd<'a> {
...
         cmd.args(["--proc", "/proc"]);
-        cmd.args(["--dev", "/dev"]);
+        cmd.args(["--dev-bind", "/dev", "/dev"]);
         cmd.args(["--bind", "/sys", "/sys"]);
 
+        // Bind the udev database so that lsblk/libblkid inside the sandbox
+        // can read partition type GUIDs and other device properties.
+        if Utf8Path::new("/run/udev").try_exists().unwrap_or(false) {
+            cmd.args(["--bind", "/run/udev", "/run/udev"]);
+        }
```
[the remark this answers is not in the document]
cgwalters: Agreed on this one

cgwalters (approved):
Gemini had a nit, but we can also fix it after if you want as part of other things

cgwalters:
> No SELinux label found for: "deploy/default/deploy/1fdbcca17fd068e449e98ec7129838c0520be8d8b2536bad13d4565e5a59cb84.0/run/udev"

Hmm that's concerning that looks like we're leaking content into the target deployment root. Ah yes I think we're missing `--tmpfs /run`

crates/utils/src/bwrap.rs
```diff
@@ -79,19 +69,21 @@ impl<'a> BwrapCmd<'a> {
...
         cmd.args(["--proc", "/proc"]);
-        cmd.args(["--dev", "/dev"]);
+        cmd.args(["--dev-bind", "/dev", "/dev"]);
         cmd.args(["--bind", "/sys", "/sys"]);
 
+        // Bind the udev database so that lsblk/libblkid inside the sandbox
+        // can read partition type GUIDs and other device properties.
+        if Utf8Path::new("/run/udev").try_exists()? {
+            cmd.args(["--bind", "/run/udev", "/run/udev"]);
+            cmd.args(["--tmpfs", "/run"]);
```
cgwalters: Actually on reflection I think it's probably better to just `--bind /run /run` - there's just too many things there that we may want; same as what we do when in podman.

(But even if we didn't take that path, we would need to *always* pass the `--tmpfs /run` unconditionally)

ckyrouac: good idea - updated

PR #2047, opened 2026-03-05
status: Render cached update info in human-readable output

cgwalters:
After running `bootc upgrade --check`, the registry metadata for a
newer image is cached in ostree commit metadata. The `bootc status`
command already reads this into the `cachedUpdate` field and exposes
it in JSON/YAML output, but the human-readable output never displayed
it. This meant users had to parse structured output or re-run
`upgrade --check` to see available updates.

Render the cached update inline with each deployment entry, showing
version, timestamp, and digest when the cached digest differs from
the currently deployed image.

Relates: https://issues.redhat.com/browse/RHEL-139384

PR #2049, opened 2026-03-07
tests: Add composefs-upgrade test for sealed UKI builds 

cgwalters:
The goal is ensuring we have upgrade coverage also for sealed UKIs;
most of the other update code paths (because tmt doesn't make it easy
to have a registry) do on-machine synthetic updates.

Assisted-by: OpenCode (claude-opus-4)

### DIFF
--- a/tmt/tests/booted/test-image-upgrade-reboot.nu
+++ b/tmt/tests/booted/test-image-upgrade-reboot.nu
@@ -11,6 +11,10 @@
 # bootc switch <into that image> --apply
 # Verify we boot into the new image
 #
+# For composefs builds, it additionally verifies that composefs is
+# still active after upgrade.  For sealed UKI builds, it checks that
+# both the original and upgrade UKIs exist on the ESP.
+#
 use std assert
 use tap.nu
 
@@ -21,6 +25,7 @@ journalctl --list-boots
 
 let st = bootc status --json | from json
 let booted = $st.status.booted.image
+let is_composefs = (tap is_composefs)
 
 # Parse the kernel commandline into a list.
 # This is not a proper parser, but good enough
@@ -50,6 +55,12 @@ RUN touch /usr/share/testing-bootc-upgrade-apply
         podman build -t $imgsrc .
     }
 
+    # For composefs, save state so we can verify it's preserved after upgrade.
+    if $is_composefs {
+        "true" | save /var/was-composefs
+        $st.status.booted.composefs.verity | save /var/original-verity
+    }
+
     # Now, switch into the new image
     print $"Applying ($imgsrc)"
     bootc switch --transport containers-storage ($imgsrc)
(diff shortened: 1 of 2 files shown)


tmt/tests/booted/test-composefs-upgrade.nu
```diff
@@ -0,0 +1,111 @@
...
+        print $"upgrade verity: ($upgrade_verity)"
+
+        # The two verities must differ since the upgrade image has different content
+        assert ($original_verity != $upgrade_verity) "upgrade should produce a different verity digest"
+
+        # There should be two .efi UKI files on the ESP: one for the booted
+        # deployment (upgrade) and one for the rollback (original)
+        let efi_files = (glob $"($boot_dir)/*.efi")
+        print $"EFI files: ($efi_files)"
+        assert ((($efi_files | length) >= 2)) $"expected at least 2 UKIs on ESP, found ($efi_files | length)"
```
henrywang: The triple parentheses are redundant.

henrywang:
The failure should be fixed in PR #2047. And need a sign-off.

tmt/tests/booted/test-composefs-upgrade.nu
```diff
@@ -0,0 +1,111 @@
+# number: 39
```
Johan-Liebert1: Could we not use the `test-image-upgrade` test here? Since it already has 

```nu
    if ($imgsrc | str ends-with "-local") {
        bootc image copy-to-storage

        # A simple derived container that adds a file
        "FROM localhost/bootc
RUN touch /usr/share/testing-bootc-upgrade-apply
" | save Dockerfile
         # Build it
        podman build -t $imgsrc .
    }
```

It won't rebuild if we have the upgrade image built already

I think for now this is really good to have test, but imo we should have the same test (code) for all cases, so we don't end up missing something while special casing certain tests

Johan-Liebert1:
https://github.com/bootc-dev/bootc/pull/2085 should fix the xfs test failures

tmt/tests/booted/test-image-upgrade-reboot.nu
```diff
@@ -63,7 +74,47 @@ def second_boot [] {
...
+        # For UKI boot type, verify both the original and upgrade UKIs exist on the ESP
+        if ($composefs_info.bootType | str downcase) == "uki" {
+            let bootloader = ($composefs_info.bootloader | str downcase)
+
+            let boot_dir = if $bootloader == "systemd" {
+                mkdir /var/tmp/efi
+                mount /dev/disk/by-partlabel/EFI-SYSTEM /var/tmp/efi
+                "/var/tmp/efi/EFI/Linux/bootc"
+            } else {
+                "/sysroot/boot/EFI/Linux/bootc"
```
Johan-Liebert1: I don't think we need to check the bootloader here as the ESP should be the same regardless

PR #2060, opened 2026-03-12
Add an "upgrade from previous" test

cgwalters:
We have tests that do upgrades, but they start from the *new* bootc. Add a `just`+CI workflow that starts from a stable shipped image (we just need to inject tmt deps + nu).

The readonly test then gains a helper which optionally performs an upgrade to the new target.

IOW the flow is

- deploy stock image with unmodified bootc etc
- upgrade
- run readonly tests

It could of course make sense to run *all* of the tests this way as an optional thing, nothing blocks that, but it would be *another* entry in our matrix and we're going to need to figure out how to wrangle that matrix size.

Also do to that we'd need to abstract over TMT_REBOOT_COUNT.

Assisted-by: OpenCode (Claude claude-opus-4-6)

### DIFF
--- a/tmt/tests/booted/readonly/000-test-selinux-enforcing.nu
+++ b/tmt/tests/booted/readonly/000-test-selinux-enforcing.nu
@@ -1,0 +1,21 @@
+use std assert
+use tap.nu
+
+tap begin "verify SELinux is enforcing"
+
+# Composefs upgrade source images boot with enforcing=0 because the
+# base image's SELinux policy doesn't yet cover composefs file contexts.
+# Skip this check in that case.
+let upgrade_image = $env.BOOTC_test_upgrade_image? | default ""
+let is_composefs = (tap is_composefs)
+if $upgrade_image != "" and $is_composefs {
+    print "# skip: composefs upgrade boots with enforcing=0 (base image SELinux policy gap)"
+    tap ok
+    exit 0
+}
+
+let enforce = (open /sys/fs/selinux/enforce | str trim)
+assert equal $enforce "1" "SELinux should be in enforcing mode"
+print "SELinux is enforcing"
+
+tap ok


.github/workflows/ci.yml
```diff
@@ -261,6 +262,50 @@ jobs:
...
 
+  # Test the upgrade path: boot from published base image, upgrade to locally-built image,
+  # then run readonly tests to verify the upgrade worked.
+  # Excluded: centos-9 (lacks systemd.extra-unit.* support needed for --bind-storage-ro)
+  test-upgrade:
+    needs: package
+    strategy:
+      fail-fast: false
+      matrix:
+        test_os: [fedora-43, centos-10]
```
Johan-Liebert1, line 273: we'd want ostree and composefs in the matrix as well. Right now we're only testing the ostree case

tmt/tests/booted/readonly/031-test-upgrade-check.nu
```diff
@@ -0,0 +1,41 @@
...
+
+print "Running bootc upgrade --check to verify upgrade machinery works..."
+let result = do -i { bootc upgrade --check } | complete
+if $result.exit_code == 0 {
+    print "bootc upgrade --check succeeded"
+} else {
+    print $"bootc upgrade --check failed: ($result.stderr)"
+    # Known failure: composefs upgrades from bootc <= 1.13 have
+    # incompatible on-disk state (see #2074).
+    let old_bootc_le_1_13 = ($old_ver.0 < 1) or (($old_ver.0 == 1) and ($old_ver.1 <= 13))
```
cgwalters, line 33: OK cool, this successfully reproduces the problem from https://github.com/bootc-dev/bootc/issues/2074

tmt/tests/Dockerfile.upgrade-source
```diff
@@ -0,0 +1,32 @@
...
+cd /run/provision
+# Install nu and tmt dependencies using the same script as the main build,
+# but with SKIP_CONFIGS=1 since we don't need LBIs or test kargs in the
+# upgrade source image (those come from the image we upgrade into).
+SKIP_CONFIGS=1 ./provision-derived.sh
+# Ensure a root filesystem type is configured so `bootc install to-disk`
+# works for base images that don't ship a default (e.g. Fedora).
+# For the ostree variant, configure-rootfs will default to xfs.
+./configure-rootfs "${variant}"
+# For composefs, rebuild the initramfs to include the bootc dracut module.
```
cgwalters, line 22: I'm going to fix this separately too

cgwalters:
OK this one should be good to go now!

jeckersb:
Needs `cargo fmt`

Justfile
```diff
@@ -141,6 +143,29 @@ test-composefs bootloader filesystem boot_type seal_state *ARGS:
...
+    #!/bin/bash
+    set -xeuo pipefail
+    composefs_args=()
+    if [[ "{{variant}}" = composefs ]]; then
+        composefs_args=(--composefs-backend \
+            --bootloader={{bootloader}} \
+            --filesystem={{filesystem}} \
+            --seal-state={{seal_state}} \
+            --boot-type={{boot_type}} \
+            --karg=enforcing=0)
```
Johan-Liebert1, line 161: We don't need this after https://github.com/bootc-dev/bootc/pull/2035 right?

cgwalters: This is building the *from* image which at this time still has an older bootc. But yes once the update propagates we can update the baseline.

That said, what I do really want to do here is have another variant of this which starts from a *pinned* old OS version (think e.g. rhel 9.4) - of course we can't even do a composefs based install of that to start (and bcvk might not even work) but we can prove it out.

PR #2063, opened 2026-03-13
ci: Add manual crates release action and update crates versions to match bootc version on release

ckyrouac:
See two commits for details. Automatically update the two internal crates versions in the bootc release workflow, and add a manual crates release trigger.

### DIFF
--- a/.github/workflows/crates-release.yml
+++ b/.github/workflows/crates-release.yml
@@ -16,7 +16,7 @@ jobs:
     - run: |
         # Publish crates if their current version is not already on crates.io.
         # Order matters: dependencies must be published first.
-        CRATES="bootc-internal-utils bootc-internal-blockdev"
+        CRATES="bootc-internal-utils bootc-internal-mount bootc-internal-blockdev"
 
         for crate in $CRATES; do
           VERSION=$(cargo read-manifest -p "$crate" | jq -r '.version')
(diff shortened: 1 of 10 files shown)


cgwalters:
I had an agent run on this and it noticed at least that we need to publish -mount. Can you reconcile with https://github.com/bootc-dev/bootc/compare/main...cgwalters:bootc:crate-publish?expand=1

.github/workflows/crates-release.yml
```diff
@@ -3,6 +3,16 @@ name: Publish to crates.io
 on:
   push:
     tags: ['v*']  # Triggers when pushing tags starting with 'v'
+  workflow_dispatch:
+    inputs:
+      bootc-internal-utils:
```
cgwalters: Hmm I think we could probably easily change this to "publish if not published" i.e. reconcile to desired state instead of being imperative.

i.e. drop these inputs then...

.github/workflows/crates-release.yml
```diff
@@ -13,7 +23,26 @@ jobs:
...
+          fi
+          # Default to all crates if none selected
+          if [ -z "$CRATES" ]; then
+            CRATES="$ALL_CRATES"
+          fi
+        fi
+
+        for crate in $CRATES; do
           echo "Publishing $crate..."
           cargo publish -p "$crate"
```
cgwalters: Ideally there'd be a `cargo publish --idempotent`, I don't see it offhand but I bet writing it (check if crate already published) would be as much code as the above

ckyrouac: updated to be idempotent

ckyrouac:
> I had an agent run on this and it noticed at least that we need to publish -mount. Can you reconcile with https://github.com/bootc-dev/bootc/compare/main...cgwalters:bootc:crate-publish?expand=1

Added the bootc-internal-mount release. Claude wanted to clean up some stuff to prepare for release which seemed reasonable.

PR #2071, opened 2026-03-16
initramfs: Inherit SELinux label on transient root tmpfs

### LINKED ISSUE #1992: Bad selinux tag for root dir when using transient root
When enables `root.transient`, the root dir's selinux tag changes to `system_u:object_r:tmpfs_t:s0`, while i think the right type should be `root_t`. which causes many things to crash like systemd-networkd will refuse to start cause it can't watch on `/`.

andrewdunndev:
## Summary

When `root.transient` is enabled, the overlay root at `/` gets SELinux label
`tmpfs_t` instead of `root_t`, causing services like systemd-networkd to crash.

This PR fixes the composefs backend by reading the SELinux label from the lower
(composefs) filesystem and setting it as `rootcontext` on the tmpfs used for
the overlay upper layer.

## What changed

- New `mount_tmpfs_for_overlay()` that reads `security.selinux` xattr from the
  base filesystem and sets `rootcontext` on the tmpfs before overlay creation
- Uses the two-call `fgetxattr` pattern (query size → allocate → read) for
  correct handling of arbitrarily long MLS/MCS contexts
- Uses `CString` (not `str`) to avoid requiring UTF-8 labels
- Gracefully handles SELinux-disabled systems (fgetxattr error → no context set)
- `overlay_transient()` passes the base fd through to inherit the correct label
- Removed the now-unused `mount_tmpfs()` wrapper

## Design decisions

- **`rootcontext=` over `context=`**: `rootcontext` sets only the root inode's
  label, letting new files inherit labels from SELinux policy. `context=` would
  force a uniform label on all inodes in the tmpfs, which is too broad.
  (systemd uses `context=` in nspawn and has had issues with it — see
  systemd/systemd#12292.)

- **Read from lower layer, not hardcode**: The label is read from the composefs
  base filesystem rather than hardcoding `root_t`. This works regardless of the
  SELinux policy in use.

- **Silent fallback**: If SELinux is not enabled or the xattr read fails, no
  context is set and behavior is unchanged from before this patch.

## Related issues

- **ostree backend**: The ostree backend (`otcore-prepare-root.c`) has the same
  issue but uses `/run` as the tmpfs backing store. That fix would go through
  `lcfs_mount_options_s` in libcomposefs or the ostree repo. Happy to file a
  follow-up issue on ostreedev/ostree.

- **systemd**: `systemd.volatile=overlay` has the same unfixed bug
  (systemd/systemd#12292). Dan Walsh's fix attempt (systemd/systemd#26141) has
  been stalled since 2023.

## Testing

This runs in the initramfs before the real root is mounted, so testing requires
a VM with SELinux enforcing and `root.transient` enabled. Verification:

```bash
# Before fix: / has tmpfs_t
ls -Zd /
# system_u:object_r:tmpfs_t:s0 /

# After fix: / has root_t
ls -Zd /
# system_u:object_r:root_t:s0 /
```

Closes #1992

Signed-off-by: Andrew Dunn <andrew@dunn.dev>

### DIFF
--- a/crates/initramfs/src/lib.rs
+++ b/crates/initramfs/src/lib.rs
@@ -1,7 +1,7 @@
 //! Mount helpers for bootc-initramfs
 
 use std::{
-    ffi::OsString,
+    ffi::{CString, OsString},
     fmt::Debug,
     io::ErrorKind,
     os::fd::{AsFd, AsRawFd, OwnedFd},
@@ -9,6 +9,8 @@ use std::{
 };
 
 use anyhow::{Context, Result};
+use cap_std_ext::cap_std::fs::Dir;
+use cap_std_ext::dirext::CapStdExtDirExt;
 use clap::Parser;
 use rustix::{
     fs::{CWD, Mode, OFlags, major, minor, mkdirat, openat, stat, symlink},
@@ -19,6 +21,7 @@ use rustix::{
     },
     path,
 };
+
 use serde::Deserialize;
 
 use cfsctl::composefs;
@@ -202,9 +205,21 @@ fn bind_mount(fd: impl AsFd, path: &str) -> Result<OwnedFd> {
     Ok(res?)
 }
 
-#[context("Mounting tmpfs")]
-fn mount_tmpfs() -> Result<OwnedFd> {
+/// Mount a tmpfs, inheriting the SELinux label from the base filesystem
+/// if provided. See <https://github.com/containers/bootc/issues/1992>.
+#[context("Mounting tmpfs for overlay")]
+fn mount_tmpfs_for_overlay(base: Option<impl AsFd>) -> Result<OwnedFd> {
     let tmpfs = FsHandle::open("tmpfs")?;
+
+    if let Some(base_fd) = base {
+        let base_dir = Dir::reopen_dir(&base_fd.as_fd())?;
+        if let Some(label) = base_dir.getxattr(".", "security.selinux")? {
+            if let Ok(cstr) = CString::new(label) {
+                fsconfig_set_string(tmpfs.as_fd(), "rootcontext", &cstr)?;
+            }
+        }
+    }
+
     fsconfig_create(tmpfs.as_fd())?;
     Ok(fsmount(
         tmpfs.as_fd(),
@@ -239,16 +254,20 @@ fn overlay_state(
     mount_at_wrapper(fs, base, ".").context("Moving mount")
 }
 
-/// Mounts a transient overlayfs with passed in fd as the lowerdir
+/// Mounts a transient overlayfs with passed in fd as the lowerdir.
+///
+/// The tmpfs used for the overlay upper layer inherits the SELinux label
+/// from the base filesystem to prevent label mismatches (see #1992).
 #[context("Mounting transient overlayfs")]
 pub fn overlay_transient(
     base: impl AsFd,
     mode: Option<rustix::fs::Mode>,
     mount_attr_flags: Option<MountAttrFlags>,
 ) -> Result<()> {
+    let tmpfs = mount_tmpfs_for_overlay(Some(&base))?;
     overlay_state(
         base,
-        prepare_mount(mount_tmpfs()?)?,
+        prepare_mount(tmpfs)?,
         "transient",
         mode,
         mount_attr_flags,
--- a/crates/initramfs/Cargo.toml
+++ b/crates/initramfs/Cargo.toml
@@ -7,6 +7,7 @@ publish = false
 
 [dependencies]
 anyhow.workspace = true
+cap-std-ext.workspace = true
 clap = { workspace = true, features = ["std", "help", "usage", "derive"] }
 libc.workspace = true
 rustix.workspace = true


crates/initramfs/src/lib.rs
```diff
@@ -204,7 +207,58 @@ fn bind_mount(fd: impl AsFd, path: &str) -> Result<OwnedFd> {
...
+                    buf.resize(4096, 0);
+                    match rustix::fs::fgetxattr(base_fd.as_fd(), SELINUX_XATTR, &mut buf) {
+                        Ok(len) => Some(buf[..len].to_vec()),
+                        Err(_) => None,
+                    }
+                }
+                // SELinux not enabled, or filesystem doesn't support xattrs.
+                Err(_) => None,
+            }
+        };
```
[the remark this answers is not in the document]
andrewdunndev: Good suggestion — incorporated. Switched to the two-call `fgetxattr` pattern (query size with empty buffer, then allocate and read). This eliminates the hardcoded buffer sizes and handles arbitrarily long MLS/MCS contexts correctly. Force-pushed with the fix.

gursewak1997 (dismissed):
lgtm

crates/initramfs/src/lib.rs
```diff
@@ -202,9 +205,51 @@ fn bind_mount(fd: impl AsFd, path: &str) -> Result<OwnedFd> {
...
+/// filesystem and setting it as `rootcontext` on the tmpfs, the overlay root
+/// gets the correct label (typically `root_t`).
+///
+/// See: <https://github.com/containers/bootc/issues/1992>
+#[context("Mounting tmpfs for overlay")]
+fn mount_tmpfs_for_overlay(base: Option<impl AsFd>) -> Result<OwnedFd> {
     let tmpfs = FsHandle::open("tmpfs")?;
+
+    if let Some(base_fd) = base {
+        // Read the SELinux label from the base filesystem root using the
```
cgwalters: This could probably use https://docs.rs/cap-std-ext/latest/cap_std_ext/dirext/trait.CapStdExtDirExt.html#tymethod.getxattr

crates/initramfs/src/lib.rs
```diff
@@ -202,9 +205,51 @@ fn bind_mount(fd: impl AsFd, path: &str) -> Result<OwnedFd> {
...
+        })();
+
+        if let Some(mut label) = context {
+            // The kernel may return a null-terminated byte string; strip it
+            // before constructing a CString (which adds its own terminator).
+            if label.last() == Some(&0) {
+                label.pop();
+            }
+            // Construct a CString for fsconfig_set_string. This avoids
+            // requiring the label to be valid UTF-8 (per maintainer feedback).
```
cgwalters: To me this is a perfect example of a "very obvious" LLM generated (I'm sure) comment that's just unnecessary. I'll try to update our AGENTS.md and such to try to suppress some of this, I have it in my personal one.

crates/initramfs/src/lib.rs
```diff
@@ -19,6 +19,9 @@ use rustix::{
     },
     path,
 };
+
+/// The SELinux extended attribute name.
+const SELINUX_XATTR: &str = "security.selinux";
```
cgwalters: We also have this const elsewhere

crates/initramfs/src/lib.rs
```diff
@@ -202,9 +205,51 @@ fn bind_mount(fd: impl AsFd, path: &str) -> Result<OwnedFd> {
...
+            buf.truncate(len);
+            Some(buf)
+        })();
+
+        if let Some(mut label) = context {
+            // The kernel may return a null-terminated byte string; strip it
+            // before constructing a CString (which adds its own terminator).
+            if label.last() == Some(&0) {
+                label.pop();
+            }
```
cgwalters: Yeah, here be some 🐉 - we messed something up like this in ostree IIRC

cgwalters (commented):
I think this is all OK, however...and this is not something you need to fix as part of this PR, but we are missing testing transient root at all I believe as part of gating CI here by default.

It's tricky because it would be *another* entry in our ever-growing matrix, I'll try to write something up for this.

crates/initramfs/src/lib.rs
andrewdunndev: Switched to cap-std-ext::getxattr via Dir::reopen_dir. Force-pushed.

crates/initramfs/src/lib.rs
andrewdunndev: Noted, will keep replies terse going forward (and we adjusted our harness, you're detecting LLM assist correctly).

crates/initramfs/src/lib.rs
andrewdunndev: Removed. Using the string literal inline since the existing consts in lsm.rs aren't reachable from initramfs.

crates/initramfs/src/lib.rs
andrewdunndev: cap-std-ext::getxattr handles the buffer management now, so the manual null-terminator stripping is gone.

andrewdunndev:
Happy to help wire up a transient-root test or file an issue to track.

cgwalters:
Thanks for this! We had live chat and agreed to not block on extending the test suite for this yet, but will circle back on that.

Issue #2078, opened 2026-03-18
ostree-container: Add diffid as metadata to refs


cgwalters:
(This issue would be obviated by https://github.com/bootc-dev/bootc/issues/20 )

Today the ostree-container backend stores layers indexed by their manifest digest, but it would have made more sense to index by diffid.

But we can compatibly handle this today: before initiating a fetch, iterate over all stored ostree-container manifests, and build a mapping from blob -> diffid. If we go to fetch a layer, and it has a different blob but the same diffid, just reuse that existing layer. I think that should be it.

This should be relatively easy to integration test by constructing an oci: storage that has the same diffid, one layer compressed with gzip and another with zstd for example.

PR #2080, opened 2026-03-19
ostree-ext: Store container blob refs by diff_id as well

alexlarsson:
When importing layers we now generate and look for ostree refs by the layer diff_id, as well as the (for compat with earlier versions) digest layer id. This means that when importing images that have the same layer (same diff_id), but different layer id, for example due to recompression of the layers, we will still reuse the previously downloaded layers.

Additionally we now copy both refs when copying images, and when pruning unreferences blobs, we look for both types of refs.

This also adds some tests for the new functionallity.

Fixes: https://github.com/bootc-dev/bootc/issues/2078

### DIFF
--- a/crates/ostree-ext/src/container/store.rs
+++ b/crates/ostree-ext/src/container/store.rs
@@ -15,7 +15,7 @@
 //!
 //! Layers and images are stored using these ref prefixes (defined as constants in this module):
 //!
-//! - `ostree/container/blob`: Individual OCI layers stored as commits
+//! - `ostree/container/blob`: Individual OCI layers stored as commits (both by layer id and by diff id)
 //! - `ostree/container/image`: Merge commits for complete images
 //! - [`BASE_IMAGE_PREFIX`] (`ostree/container/baseimage`): Protected base images (public)
 //!
@@ -82,7 +82,7 @@
 //!
 //! Unreferenced layers are automatically pruned after imports via [`gc_image_layers`]:
 //!
-//! 1. Collect all layer digests referenced by stored images and deployments
+//! 1. Collect all layer digests (both blob digests and diff_ids) referenced by stored images and deployments
 //! 2. List all layer refs under `ostree/container/blob/`
 //! 3. Remove refs for layers not in the referenced set
 //!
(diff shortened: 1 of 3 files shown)


crates/ostree-ext/src/container/store.rs
```diff
@@ -1015,6 +1056,8 @@ impl ImageImporter {
                     if write_refs {
                         repo.transaction_set_ref(None, &target_ref, Some(commit.as_str()));
                         tracing::debug!("Wrote {} => {}", target_ref, commit);
+                        repo.transaction_set_ref(None, &target_compat_ref, Some(commit.as_str()));
```
cgwalters, line 1059: Did you see the suggestion in https://github.com/bootc-dev/bootc/issues/2078 instead?

It has the cost of reading all metadata, which I wouldn't want to do for a generic image store, but in practice we're only storing one container image here.

A disadvantage of writing additional refs in this way is that older versions of bootc won't understand them and won't GC them.

And actually you know what, we already read all the manifests in GC anyways.

I guess it doesn't matter too much.

alexlarsson: I did see that, but way to late, when this was essentially done. However, given that the refs are the same format they will be GCed by older bootc versions.

I can take a look at your approach as well, and we'll see how that looks.

crates/ostree-ext/src/container/store.rs
```diff
@@ -440,13 +461,29 @@ impl PreparedImport {
 // Given a manifest, compute its ostree ref name and cached ostree commit
 pub(crate) fn query_layer(
     repo: &ostree::Repo,
-    layer: oci_image::Descriptor,
+    layer: &oci_image::Descriptor,
+    manifest: &oci_image::ImageManifest,
+    config: &oci_image::ImageConfiguration,
```
cgwalters, line 466: Style nit I think it's better to put more "global" arguments first.

crates/ostree-ext/src/container/store.rs
cgwalters: > However, given that the refs are the same format they will be GCed by older bootc versions.

Yeah, I meant that in an upgrade -> downgrade scenario the old version would prune them as unreachable - and in that scenario I think we won't write the uncompressed ref again?

I guess that doesn't matter too much though.

alexlarsson: New version in https://github.com/bootc-dev/bootc/pull/2081
I think I prefer that, as it keeps the on-disk format. As you say, its slightly less efficient on pull, but that is not really a practical problem for the bootc usecase.

cgwalters:
Closing in favor of https://github.com/bootc-dev/bootc/pull/2081

PR #2082, opened 2026-03-19
tmt: Disable rhel-ci AVC check for expected SELinux denials

cgwalters:
We have a probe for `install_t` which causes an AVC denial. At some point we'll need to fix that, but for now
set AVC_ERROR=+no_avc_check in the integration plan environment to disable the check since it's gating in RHEL now.


henrywang:
Thanks for fixing this.

PR #2084, opened 2026-03-23
composefs: add pre-flight disk space check for native backend

shi2wei3:
The previous commit (7196079e) added a pre-flight disk space check for the ostree and ostree+unified-storage backend paths. This commit extends the same protection to the native composefs backend.

- Rename check_disk_space_composefs -> check_disk_space_unified for the ostree unified-storage path (uses PreparedImportMeta), to clarify that this variant is not used by the native composefs backend.

- Add check_disk_space_composefs for the native composefs backend: sums layer sizes from the raw ImageManifest and calls fstatvfs on the composefs objects directory to verify available space before pulling.

- Call check_disk_space_composefs in do_upgrade() (covers bootc upgrade and bootc switch on the native composefs backend).

- Call check_disk_space_composefs in install_to_filesystem_impl() before initialize_composefs_repository() (covers bootc install --composefs-backend). Note: ensure_composefs_dir() is called first to create the objects directory if it does not yet exist on the fresh install target.

- Enable the existing pre-flight disk check tmt test for the composefs backend (remove fixme_skip_if_composefs).

Assisted-by: AI

### DIFF
--- a/crates/lib/src/bootc_composefs/update.rs
+++ b/crates/lib/src/bootc_composefs/update.rs
@@ -255,6 +255,13 @@ pub(crate) async fn do_upgrade(
 ) -> Result<()> {
     start_finalize_stated_svc()?;
 
+    // Pre-flight disk space check before pulling any data.
+    crate::deploy::check_disk_space_composefs(
+        &booted_cfs.repo,
+        &img_manifest_config.manifest,
+        imgref,
+    )?;
+
     let (repo, entries, id, fs) = pull_composefs_repo(
         &imgref.transport,
         &imgref.image,
--- a/crates/lib/src/deploy.rs
+++ b/crates/lib/src/deploy.rs
@@ -411,13 +411,29 @@ pub(crate) fn check_disk_space_ostree(
     )
 }
 
-/// Verify there is sufficient disk space to pull an image into the composefs store.
-pub(crate) fn check_disk_space_composefs(
+/// Verify there is sufficient disk space to pull an image into the composefs store
+/// via the ostree unified-storage path (uses `PreparedImportMeta`).
+pub(crate) fn check_disk_space_unified(
     cfs: &crate::store::ComposefsRepository,
     image_meta: &PreparedImportMeta,
     imgref: &ImageReference,
 ) -> Result<()> {
     check_disk_space_inner(cfs.objects_dir()?, image_meta.bytes_to_fetch, 0, imgref)
+}
+
+/// Verify there is sufficient disk space to pull an image into the composefs store
+/// for the native composefs backend (uses a raw `ImageManifest`).
+pub(crate) fn check_disk_space_composefs(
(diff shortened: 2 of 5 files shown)


crates/lib/src/bootc_composefs/update.rs
```diff
@@ -255,6 +255,17 @@ pub(crate) async fn do_upgrade(
 ) -> Result<()> {
     start_finalize_stated_svc()?;
 
+    // Pre-flight disk space check before pulling any data.
+    {
+        let rootfs_dir = Dir::open_ambient_dir("/sysroot", ambient_authority())?;
```
Johan-Liebert1: we already have "/sysroot" opened in `storage.physical_root`

crates/lib/src/bootc_composefs/update.rs
```diff
@@ -255,6 +255,17 @@ pub(crate) async fn do_upgrade(
...
+        let cfs_repo = crate::bootc_composefs::repo::open_composefs_repo(&rootfs_dir)?;
```
Johan-Liebert1: Also have this in `booted_cfs`

crates/lib/src/deploy.rs
```diff
@@ -411,15 +411,31 @@ pub(crate) fn check_disk_space_ostree(
...
 }
 
+/// Verify there is sufficient disk space to pull an image into the composefs store
+/// for the native composefs backend (uses a raw `ImageManifest`).
+pub(crate) fn check_disk_space_composefs(
+    cfs: &crate::store::ComposefsRepository,
+    manifest: &ostree_ext::oci_spec::image::ImageManifest,
+    imgref: &ImageReference,
+) -> Result<()> {
+    let bytes_to_fetch: u64 = manifest
```
Johan-Liebert1, line 431: This is great. I wonder if there is a way to get the uncompressed layer size.

crates/lib/src/bootc_composefs/update.rs
shi2wei3: Updated.

crates/lib/src/bootc_composefs/update.rs
shi2wei3: Updated.

crates/lib/src/deploy.rs
shi2wei3: Yeah, the actual disk usage will be larger than what we check, I didn't find a way to do that w/o download the layer, use a heuristic compression ratio multiplier is also fragile.

Johan-Liebert1: Pretty unfortunate that we can't get the uncompressed size from manifest/config

cgwalters: > Pretty unfortunate that we can't get the uncompressed size from manifest/config

https://github.com/opencontainers/image-spec/issues/1246

Sigh. But it's still tempting to Just Do It in a few build tools like podman...

Issue #1637, opened 2025-09-22
ostree backend: Fix duplicate objects without /ostree in container


jlebon:
Edited by @cgwalters 

Goal: Ensure that when a container image doesn't have `/ostree` we don't end up with duplicate ostree objects (one labeled in a default way and one with the final label).

Mitigations:

 - payload-link
 - Change the commit process to be two phase: save layers but do it in way that we avoid duplicate objects
 - https://github.com/bootc-dev/bootc/issues/20

Hard constraint: Must not break rpm-ostree (or ostree) when used directly

Original:

---

We've now moved to client-side SELinux labeling. I think we should still though support server-side labeling, I guess through `ostree container commit`/`bootc build commit`? I think this today though conflicts with wanting to move away from `/ostree` in the container image, but it could be implemented differently of course.

The main argument is simply reproducibility. Notably, https://github.com/coreos/fedora-coreos-tracker/issues/2030 happened which is a great example of why doing this server-side would be better for those that want to opt in (like FCOS/RHCOS).

cgwalters:
This also came up in https://github.com/ostreedev/ostree/pull/3523#issuecomment-3272115593

> I think this today though conflicts with wanting to move away from /ostree in the container image, but it could be implemented differently of course.

Yes though the other obvious implementation is to just support having `security.selinux` xattrs in the tar stream. This would need some validation that it doesn't break/conflict with container runtimes though.

jlebon:
Yes OK, definitely quite surprising I think that even just encapsulated commits still get relabeled (https://github.com/coreos/fedora-coreos-tracker/issues/2030#issuecomment-3329691926).

And in the new buildah path, while we do generate an OSTree commit as part of `build-chunked-oci`, I tend not to think of it that way and I'm not sure if it's where relabeling should happen. E.g. what I initially had in mind for this issue is something applicable to both `FROM scratch` flows _and_ simple derivations.

cgwalters:
We had a discussion about this one today and I guess it is a soft blocker for dropping `/ostree` from the container for the ostree backend.

I edited the [top issue](https://github.com/bootc-dev/bootc/issues/1637#issue-3442487696) here.

jlebon:
And in turn this is a blocker for moving over to chunkah which doesn't understand `/ostree` (nor wants to :) ).

> Change the commit process to be two phase: save layers but do it in way that we avoid duplicate objects

[...]

PR #2088, opened 2026-03-23
ostree-ext/store: Relabel layer commits to avoid duplicate objects

jlebon:
When importing a non-ostree container (i.e. without /ostree), layer commits are created without SELinux labels. The merge commit then applies the correct labels but the layer commits still reference the unlabeled version of every object. We do get reflinks, but it still effectively doubles the number of commit objects in the repo associated with that import.

Fix this by relabeling each layer commit after the merge commit is written. Each layer is checked out, recommitted with the same SELinux policy used for the merge commit, and its ref updated to the new commit. Since the correctly-labeled objects already exist in the repo from the merge commit, most writes are no-ops. A subsequent prune removes the orphaned pre-relabel objects.

Closes: https://github.com/bootc-dev/bootc/issues/1637

[Assisted-by](https://gist.github.com/jlebon/dd287bd01218a59bb41d000c9e7d3897): OpenCode (Claude Opus 4.6)

### DIFF
--- a/crates/ostree-ext/ci/priv-integration.sh
+++ b/crates/ostree-ext/ci/priv-integration.sh
@@ -135,6 +135,78 @@ echo "ok deploy derived container from local dir"
 ostree container image remove --repo "${repo}" "${derived_img_dir}"
 rm -rf /var/tmp/derived.dir
 
+# Test: non-ostree container import with SELinux relabeling
+# Converts the FCOS image to a plain (non-ostree) image using chunkah,
+# then deploys it and verifies the relabeling optimization ran.
+# See https://github.com/bootc-dev/bootc/issues/1637
+
+# Clean state
+ostree --repo="${repo}" refs ostree/container/image --delete
+ostree container image prune-images --full --sysroot="${sysroot}"
+
+# Convert FCOS to non-ostree image using chunkah
+# --prune /sysroot/  strips the ostree deployment data
+# --label KEY-       removes ostree-specific labels
+# See also https://github.com/coreos/chunkah?tab=readme-ov-file#compatibility-with-bootable-bootc-images
+nonostree_archive=/var/tmp/nonostree.ociarchive
+chunkah_config="$(podman inspect ${image})"
+systemd-run -dP --wait podman run --rm \
+    --mount=type=image,src=${image},dst=/chunkah \
+    -v /var/tmp:/output:z \
+    -e CHUNKAH_CONFIG_STR="${chunkah_config}" \
+    quay.io/coreos/chunkah build \
+    --prune /sysroot/ \
+    --label ostree.commit- \
+    --label ostree.final-diffid- \
+    -o /output/nonostree.ociarchive
+
+# Deploy the non-ostree image with debug logging to capture relabeling messages
+RUST_LOG=ostree_ext=debug ostree container image deploy \
+    --sysroot "${sysroot}" \
+    --stateroot "${stateroot}" \
+    --imgref ostree-unverified-image:oci-archive:${nonostree_archive} 2>deploy-nonostree.txt
+
+# Verify relabeling occurred (N > 0 layers were relabeled)
+if ! grep -qE 'relabeled [1-9][0-9]* layer commits' deploy-nonostree.txt; then
+    echo "Relabeling did not occur or relabeled 0 layers" 1>&2
+    cat deploy-nonostree.txt
+    exit 1
+fi
+
+# Verify orphaned pre-relabel objects were pruned
+if ! grep -qE 'pruned [1-9][0-9]* orphaned objects after relabeling' deploy-nonostree.txt; then
+    echo "Post-relabel prune did not remove any objects" 1>&2
+    cat deploy-nonostree.txt
+    exit 1
+fi
+
+# Verify that layer and merge commit share the same file objects after relabeling.
+# Find the layer containing the bootc binary via chunkah's manifest annotation.
+ostree container image metadata --repo "${repo}" oci-archive:${nonostree_archive} > nonostree-manifest.json
+layer_digest=$(jq -r '.layers[] | select(.annotations["org.chunkah.component"] | test("rpm/bootc")) | .digest' nonostree-manifest.json | head -1)
+layer_ref="ostree/container/blob/$(echo ${layer_digest} | sed 's/:/_3A_/')"
+
+# Get the checksum of /usr/bin/bootc from the layer commit
+layer_bootc_csum=$(ostree --repo="${repo}" ls -RC "${layer_ref}" /usr/bin/bootc | awk '{print $5}')
+
+# Get the checksum of /usr/bin/bootc from the merge commit
+img_ref=$(ostree --repo="${repo}" refs ostree/container/image | head -1)
+merge_bootc_csum=$(ostree --repo="${repo}" ls -RC "ostree/container/image/${img_ref}" /usr/bin/bootc | awk '{print $5}')
+
+# Sanity check: ostree checksums are 64 hex chars
+test ${#layer_bootc_csum} = 64
+
+# If relabeling worked, both should have the same SELinux-labeled objects
+test "${layer_bootc_csum}" = "${merge_bootc_csum}"
+echo "ok layer and merge commit share objects after relabeling"
+
+# Cleanup
+ostree admin --sysroot="${sysroot}" undeploy 0
+ostree container image prune-images --full --sysroot="${sysroot}"
+rm -f ${nonostree_archive}
+
+echo "ok non-ostree container import with SELinux relabeling"
+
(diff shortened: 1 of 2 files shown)


crates/ostree-ext/src/container/store.rs
```diff
@@ -1301,19 +1310,110 @@ impl ImageImporter {
...
 
         if !disable_gc {
             let n: u32 = gc_image_layers_impl(repo, cancellable)?;
             tracing::debug!("pruned {n} layers");
         }
 
+        // Prune orphaned objects from the old (pre-relabel) layer commits.
+        // Unfortunately, there's no way to make this _not_ a global prune.
+        if n_relabeled_layers > 0 {
+            let (_, n_pruned, _) = repo.prune(ostree::RepoPruneFlags::REFS_ONLY, 0, cancellable)?;
```
jlebon: I initially had work to tie this to `disable_gc` and then also turn that on for the bootc top-level calls which already do `deploy::prune()`. That would also save on the duplicate `gc_image_layers` calls and duplicate repo pruning. But I didn't like that it makes us log something like:

> Pruned images: 0 (layers: 0, objsize: 1.7GB)

Which is confusing to end-users. Anyway, the actual cost of pruning the repo twice is minimal compared to the larger cost of relabeling all the layers.

crates/ostree-ext/ci/priv-integration.sh
```diff
@@ -135,6 +135,75 @@ echo "ok deploy derived container from local dir"
...
+ostree container image metadata --repo "${repo}" oci-archive:${nonostree_archive} > nonostree-manifest.json
+layer_digest=$(jq -r '.layers[] | select(.annotations["org.chunkah.component"] | test("rpm/bootc")) | .digest' nonostree-manifest.json | head -1)
+layer_ref="ostree/container/blob/$(echo ${layer_digest} | sed 's/:/_3A_/')"
+
+# Get the checksum of /usr/bin/bootc from the layer commit
+layer_bootc_csum=$(ostree --repo="${repo}" ls -RC "${layer_ref}" /usr/bin/bootc | awk '{print $4}')
+
+# Get the checksum of /usr/bin/bootc from the merge commit
+img_ref=$(ostree --repo="${repo}" refs ostree/container/image | head -1)
+merge_bootc_csum=$(ostree --repo="${repo}" ls -RC "ostree/container/image/${img_ref}" /usr/bin/bootc | awk '{print $4}')
```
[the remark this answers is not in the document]
jlebon: Gemini is right that the field number is wrong, but wrong that the correct one is $1. Fixed this now and added a check to make sure it's right.

crates/ostree-ext/src/container/store.rs
cgwalters: > But I didn't like that it makes us log something like: Pruned images: 0 (layers: 0, objsize: 1.7GB)

Hmm...but don't we log that sometimes anyways? And wait though the "disable GC" here is about the refs, not the objects.

I definitely don't like doing multiple object GCs.

If this is about not printing "pruned 0"...well, let's just change to skip a user visible output in that case?

Although "objsize" here means we pruned a lot of objects. Are you *sure* this isn't just a side effect of earlier experimentation?

In a nutshell we have roughly ~4 cases I think:

- Fetching an image where there's no drift i.e. layer labels == merge labels
- Fetching an image with policy modifications (issue under discussion) and we necessarily end up with some garbage objects

Either one of those two cases but also where we had *pre-existing* garbage before the operation even started.

> Anyway, the actual cost of pruning the repo twice is minimal compared to the larger cost of relabeling all the layers.

Interesting...did you measure that? It's not obvious to me that'd be the case. I'm not saying you need to anything rigorous just it'd be interesting to see timings from your setup. Are we timing the length of time to relabel a layer at all?

Cargo.lock
cgwalters: BTW this is is an unrelated change, I think we need to change our CI to error out if `Cargo.lock` isn't up to date

crates/ostree-ext/src/container/store.rs
```diff
@@ -1301,19 +1310,110 @@ impl ImageImporter {
...
     }
 
+    /// Relabel layer commits with the given commit modifier (which carries
+    /// an SELinux policy).  Each layer is checked out, recommitted with the
+    /// policy applied, and its ref is updated to point at the new commit.
+    /// Returns the number of layers that were actually relabeled (i.e. whose
+    /// commit changed).
+    fn relabel_layers(
+        repo: &ostree::Repo,
+        layer_commits: &[(String, String)],
```
cgwalters: Nonblocking but I think this is a case where a `struct Something` would be a lot more self-explanatory than a tuple.

crates/ostree-ext/src/container/store.rs
```diff
@@ -1301,19 +1310,110 @@ impl ImageImporter {
...
+        let checkout_opts = ostree::RepoCheckoutAtOptions {
+            mode: checkout_mode,
+            no_copy_fallback: true,
+            force_copy_zerosized: true,
+            ..Default::default()
+        };
+
+        for (layer_ref, old_commit) in layer_commits {
+            let td = cap_std_ext::cap_tempfile::TempDir::new_in(&repo_tmp)?;
+            repo.checkout_at(
```
cgwalters, line 1392: Not a blocker but note: The composefs labeling logic always operates on an in-memory VFS tree for this stuff and it's obviously WAY more efficient.

A case where we have the right fix in the next-gen backend, and definitely some risk in trying to do bigger changes to this ostree backend.

That said it may be relatively simple to have "recursive relabel of OstreeMutableTree" which would avoid all the hardlink + unlink traffic.

(Or build a bridge from ostree commit -> composefs generic_tree -> ostree commit)

crates/ostree-ext/src/container/store.rs
jlebon: OK wait, I messed up the logic here. We're not actually using the active SELinux policy during initial layer labeling, which would make this look a lot better.

> And wait though the "disable GC" here is about the refs, not the objects.

Right yeah, I'm saying it would fix both the pre-existing issue of calling `gc_image_layers` twice and the new issue of duplicate pruning.

> Interesting...did you measure that? It's not obvious to me that'd be the case. I'm not saying you need to anything rigorous just it'd be interesting to see timings from your setup. Are we timing the length of time to relabel a layer at all?

Just from counting while debug output flows by, I count around e.g. 6 seconds to relabel all the layers, vs 0.5-1s to prune the repo. Obviously depends on the size of the repo in the general case.

crates/ostree-ext/src/container/store.rs
jlebon: Sure. I tend to cut over once we've got three members.

crates/ostree-ext/src/container/store.rs
jlebon: Hmm yeah, I was following the checkout and commit pattern that exists for the merge commit, but doing this at the mutable tree level would indeed make this a lot cheaper.

jlebon: I tried the `ostree_repo_write_directory_to_mtree` approach, passing in the commit root, but that actually has way _way_ worse performance because we're not hitting the dfd local path in libostree, but the GIO path.

I didn't investigate the composefs path.

Anyway, relabeling the layers is now down to 2.3s locally with the latest changes, which is still unfortunate but much better.

jlebon:
OK, updated this now. See individual commits.

jlebon:
I think the CI failures here are all infra flakes.

jeckersb:
> I think the CI failures here are all infra flakes.

Yeah seems so, I restarted the failed jobs, hopefully that's enough to sort it all out.

jeckersb:
Oh right the artifacts are expired so I think the whole thing needs re-run :sob:

crates/ostree-ext/src/container/store.rs
```diff
@@ -1311,11 +1327,32 @@ impl ImageImporter {
...
+        } else {
+            0
+        };
+
         txn.commit(cancellable)?;
 
         if !disable_gc {
             let n: u32 = gc_image_layers_impl(repo, cancellable)?;
             tracing::debug!("pruned {n} layers");
+            // Prune orphaned objects from the old (pre-relabel) layer commits.
```
cgwalters, line 1350: Hmm. I am not sure we need to do this here. In practice we aren't, right? Because you changed it so most callers set disable_gc, I wonder if there are any cases where we are?

jlebon: Yeah, IIRC basically all the `ostree container` paths that hit here wouldn't disable GC.

jlebon: Yeah, IIRC basically all the `ostree container` paths that hit here wouldn't disable GC.

Issue #2089, opened 2026-03-24
install to-disk --block-setup tpm2-luks hangs: libdevmapper udev cookie semaphore deadlock in container IPC namespace


andrewdunndev:
### Summary

`bootc install to-disk --block-setup tpm2-luks` hangs indefinitely at `cryptsetup luksOpen`. The root cause is a SysV semaphore deadlock between libdevmapper and udevd across the container's isolated IPC namespace.

This affects all `tpm2-luks` installs regardless of TPM2 hardware, PCR configuration, or token state. The hang occurs on any LUKS device-mapper activation inside bootc's container environment when the container uses a separate IPC namespace (the podman default).

Users who happened to pass `--ipc=host` to podman would not have hit this bug, which may explain why some users reported TPM2-related issues (#421, #476, #477, #561) without mentioning the install hang -- they may have bypassed the semaphore deadlock without realizing it.

### Root cause

libdevmapper uses SysV semaphores ("udev cookies") to synchronize device-mapper operations with udevd. When bootc runs inside a container (the standard `podman run --privileged --pid=host` invocation), the container has an isolated IPC namespace by default. udevd runs on the host in the host's IPC namespace.

The sequence:

1. `cryptsetup luksFormat` creates a LUKS volume. libdevmapper creates a SysV semaphore in the container's IPC namespace.
2. `cryptsetup luksOpen` tries to activate a dm-crypt mapping. libdevmapper attempts to acquire the udev cookie semaphore and waits for udevd to signal completion.
3. udevd (on the host) cannot see the semaphore because it is in a different IPC namespace. The semaphore is never released.
4. `luksOpen` blocks forever on `semop()`.

Kernel stack trace of the hanging process:

```
[<0>] __do_semtimedop+0x3a8/0xd50
[<0>] do_semtimedop+0x15e/0x1a0
[<0>] do_syscall_64+0x7e/0x6b0
[<0>] entry_SYSCALL_64_after_hwframe+0x76/0x7e
```

SysV semaphore visible inside the container's IPC namespace:

```
$ nsenter -t <cryptsetup_pid> --ipc ipcs -s
key        semid      owner      perms      nsems
0x0d4d3b80 0          root       600        1
```

### Reproduction

```bash
# This hangs (default IPC namespace):
podman run --rm --privileged --pid=host \
    --security-opt label=type:unconfined_t \
    -v /dev:/dev \
    -v /var/lib/containers:/var/lib/containers \
[...]

Issue #421, opened 2024-03-22
install to-disk with LUKS + TPM


jmpolom:
Does `bootc install to-disk --block-setup tpm2-luks /dev/diskX` actually work? I tried this in a qemu virtual machine with emulated TPM (via swtpm) and while it installed successfully, upon rebooting the system into the freshly installed system the `systemd-cryptsetup` units failed to decrypt the LUKS volume. Has this actually been tested or otherwise known to work? I will try on real hardware but this has me concerned this feature is not really in a functional state.

jmpolom:
I am able to unlock the LUKS volume via TPM on a live system (FCOS) booted on the same VM I used to perform the `bootc install to-disk --block-setup tpm2-luks` from. It looks like the TPM2 binding on the volume is valid.

jmpolom:
Looks like this was observed with bootc 0.1.7 on an image created from the [treefile configuration at this point](https://github.com/jmpolom/ostree-base-image/tree/d314b2ef81497b99bb48c6f11a7b40aa371d7790) that resulted in [this container image](https://github.com/jmpolom/ostree-base-image/pkgs/container/fedora-ostree-base/194748193)

jmpolom:
I believe this may be a bug with the systemd package in my build (254-10). Will add more info later if I can get a reasonable level of confidence that is the case.

cgwalters:
Thanks for filing this, indeed we have a CI gap on this here. 

(The integration of `bootc install` needs to be at least partly owned by the particular OS you're using; we will try to be generic where we can of course).

jmpolom:
Regardless though, there needs to be a failover way to unlock any LUKS volumes bound to tokens at install time. Right now bootc actually creates a temp password and then discards it. At a very basic level this could just be output to the console as a very crude method to provide a way in if the TPM method isn't working. This is a basic issue with the to-disk install path. There really should be some user configurable options though.

Unless the thought is bootc will only handle the bottom half of the logic (installing to a pre-made filesystem) and to-disk install path gets removed??

> The integration of bootc install needs to be at least partly owned by the particular OS you're using

This was tested with Fedora 39 which has plenty of ostree based releases so I don't think that is necessarily the issue. Are there specific additional integrations needed (link to docs please)?

[...]

PR #2090, opened 2026-03-24
fix(install): set DM_DISABLE_UDEV=1 to prevent dm semaphore deadlock in container IPC namespace

andrewdunndev:
## Summary

Prevent the libdevmapper udev cookie semaphore deadlock that causes `cryptsetup luksOpen`/`luksClose` to hang inside containers with isolated IPC namespaces.

## Root cause

When `bootc install` runs inside a container (the standard invocation), the container has its own IPC namespace by default. libdevmapper creates System V semaphores ("udev cookies") to synchronize with udevd, but udevd runs in the host IPC namespace and can't see the container's semaphores. The `semop()` call blocks forever.

Full analysis: #2089

## Fix

Two changes, as suggested in review:

1. **`--ipc=host` in documented invocations** (primary fix): Add `--ipc=host` to the `podman run` commands in `bootc-install.md`. This shares the host IPC namespace so libdevmapper's semaphores reach udevd. We already pass `--pid=host` and propagate `/run`, so `--ipc=host` is consistent.

2. **`DM_DISABLE_UDEV=1` in `global_init()`** (defense-in-depth): Set the env var early in process initialization, alongside the existing `HOME` workaround. This tells libdevmapper to skip udev synchronization entirely, catching cases where IPC sharing is not configured.

Fixes: #2089
Related: #421

### DIFF
--- a/crates/lib/src/cli.rs
+++ b/crates/lib/src/cli.rs
@@ -5,6 +5,7 @@
 use std::ffi::{CString, OsStr, OsString};
 use std::fs::File;
 use std::io::{BufWriter, Seek};
+use std::os::fd::AsFd;
 use std::os::unix::process::CommandExt;
 use std::process::Command;
 
@@ -1514,6 +1515,23 @@ async fn usroverlay(access_mode: FilesystemOverlayAccessMode) -> Result<()> {
 /// in the standard `main` function.
 #[allow(unsafe_code)]
 pub fn global_init() -> Result<()> {
+    // Join the host IPC namespace if we're in an isolated one. Inside a
+    // container with a separate IPC namespace (the podman/docker default),
+    // udevd on the host cannot see the container's semaphores, causing
+    // cryptsetup operations to deadlock on semop(). The primary fix is to
+    // run the install container with --ipc=host; this is defense-in-depth
+    // for cases where the caller forgets that flag.
+    let ns_pid1 = std::fs::read_link("/proc/1/ns/ipc").context("reading /proc/1/ns/ipc")?;
+    let ns_self = std::fs::read_link("/proc/self/ns/ipc").context("reading /proc/self/ns/ipc")?;
+    if ns_pid1 != ns_self {
+        let pid1ipcns = std::fs::File::open("/proc/1/ns/ipc").context("open pid1 ipcns")?;
+        rustix::thread::move_into_link_name_space(
+            pid1ipcns.as_fd(),
+            Some(rustix::thread::LinkNameSpaceType::InterProcessCommunication),
+        )
+        .context("setns(ipc)")?;
+        tracing::debug!("Joined pid1 IPC namespace");
+    }
     // In some cases we re-exec with a temporary binary,
     // so ensure that the syslog identifier is set.
     ostree::glib::set_prgname(bootc_utils::NAME.into());
--- a/docs/src/bootc-install.md
+++ b/docs/src/bootc-install.md
@@ -60,7 +60,7 @@ to an existing system and install your container image. Failure to run
 Here's an example of using `bootc install` (root/elevated permission required):
 
 ```bash
-podman run --rm --privileged --pid=host -v /var/lib/containers:/var/lib/containers -v /dev:/dev --security-opt label=type:unconfined_t <image> bootc install to-disk /path/to/disk
+podman run --rm --privileged --pid=host --ipc=host -v /var/lib/containers:/var/lib/containers -v /dev:/dev --security-opt label=type:unconfined_t <image> bootc install to-disk /path/to/disk
 ```
 
 Note that while `--privileged` is used, this command will not perform any
@@ -68,7 +68,7 @@ destructive action on the host system.  Among other things, `--privileged`
 makes sure that all host devices are mounted into container. `/path/to/disk` is
 the host's block device where `<image>` will be installed on.
 
-The `--pid=host --security-opt label=type:unconfined_t` today
+The `--pid=host --ipc=host --security-opt label=type:unconfined_t` today
 make it more convenient for bootc to perform some privileged
 operations; in the future these requirements may be dropped.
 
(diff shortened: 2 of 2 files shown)


crates/lib/src/install/baseline.rs
```diff
@@ -149,6 +149,29 @@ pub(crate) fn wipefs(dev: &Utf8Path) -> Result<()> {
...
+/// consumers.
+#[allow(unsafe_code)]
+fn disable_dm_udev_sync() {
+    // SAFETY: bootc uses a single-threaded tokio runtime (new_current_thread).
+    // No other threads can race on environment reads. The variable is only
+    // inherited by child processes (cryptsetup, systemd-cryptenroll).
+    unsafe {
+        std::env::set_var("DM_DISABLE_UDEV", "1");
+    }
+}
```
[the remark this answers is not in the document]
andrewdunndev: Considered this approach. The reason for the process-wide env var is that `luksClose` in install.rs (line ~2122, `Task::new_and_run("Closing root LUKS device", "cryptsetup", ["close", luksdev])`) also needs it, and that call is outside baseline.rs. The per-command approach would require changes in two files and refactoring the `new_and_run` convenience method to expose `cmd.env()`. Went with `set_var` to minimize the blast radius of the PR to one file. Happy to switch to per-command `.env()` if the maintainer prefers it.

andrewdunndev:
Reworked per your feedback. Added `--ipc=host` to the documented podman invocations as the primary fix, rebased on main, force-pushed.

I kept `DM_DISABLE_UDEV=1` in `global_init()` as defense-in-depth for cases where someone runs the install without `--ipc=host`. But re-reading your review, I'm not sure that's what you wanted -- should I drop the setenv entirely and make this purely the docs change?

crates/lib/src/cli.rs
```diff
@@ -1493,6 +1493,15 @@ pub fn global_init() -> Result<()> {
...
         }
     }
+    // Disable libdevmapper's udev synchronization. Inside a container with an
+    // isolated IPC namespace (the podman/docker default), udevd on the host
+    // cannot see the container's semaphores, causing cryptsetup luksOpen and
+    // luksClose to deadlock on semop(). This is a defense-in-depth measure;
+    // the primary fix is to run the install container with --ipc=host.
+    // SAFETY: Called early in main() before any threads are spawned.
+    unsafe {
+        std::env::set_var("DM_DISABLE_UDEV", "1");
```
cgwalters: While a deadlock is not great, I think since we have a primary fix already I'd rather just rely on that. I would guess that setting this means we're entering much less well tested logic in devmapper. IOW if we *are* in the right namespace, we should definitely keep the udev synchronization.

What we could do instead here is 
```
    let pid1ipcns = std::fs::File::open("/proc/1/ns/ipc").context("open pid1 icons")?;
    rustix::thread::move_into_link_name_space(
        pid1ipcns.as_fd(),
        Some(rustix::thread::LinkNameSpaceType::Ipc),
    )
    .context("setns")?;
```

We should do that really first thing here.

Though it'd be better to detect when we're already in the namespace, and no-op then.

andrewdunndev: Done. Dropped `DM_DISABLE_UDEV` entirely and replaced with `setns()` into `/proc/1/ns/ipc`, comparing against `/proc/self/ns/ipc` to no-op when already in the right namespace. Docs updated with `--ipc=host` as the primary fix.

We took "really first thing here" literally and put it before everything else in `global_init()`. If it'd be better after the prgname/thread setup, happy to move it, quick change.

crates/lib/src/cli.rs
```diff
@@ -1514,6 +1515,25 @@ async fn usroverlay(access_mode: FilesystemOverlayAccessMode) -> Result<()> {
...
+    // cryptsetup operations to deadlock on semop(). The primary fix is to
+    // run the install container with --ipc=host; this is defense-in-depth
+    // for cases where the caller forgets that flag.
+    if let Ok(ns_pid1) = std::fs::read_link("/proc/1/ns/ipc") {
+        if let Ok(ns_self) = std::fs::read_link("/proc/self/ns/ipc") {
+            if ns_pid1 != ns_self {
+                let pid1ipcns = std::fs::File::open("/proc/1/ns/ipc").context("open pid1 ipcns")?;
+                rustix::thread::move_into_link_name_space(
+                    pid1ipcns.as_fd(),
+                    Some(rustix::thread::LinkNameSpaceType::Ipc),
```
cgwalters: This doesn't compile, the enum member has a longer name

crates/lib/src/cli.rs
```diff
@@ -1514,6 +1515,25 @@ async fn usroverlay(access_mode: FilesystemOverlayAccessMode) -> Result<()> {
 /// in the standard `main` function.
 #[allow(unsafe_code)]
 pub fn global_init() -> Result<()> {
+    // Join the host IPC namespace if we're in an isolated one. Inside a
+    // container with a separate IPC namespace (the podman/docker default),
+    // udevd on the host cannot see the container's semaphores, causing
+    // cryptsetup operations to deadlock on semop(). The primary fix is to
+    // run the install container with --ipc=host; this is defense-in-depth
+    // for cases where the caller forgets that flag.
+    if let Ok(ns_pid1) = std::fs::read_link("/proc/1/ns/ipc") {
```
cgwalters: As a general rule avoid "error swallowing" - let's just propagate the error if we can't open. It shouldn't happen and if it does we want to know.

crates/lib/src/cli.rs
andrewdunndev: Fixed both. The enum member is `InterProcessCommunication` (not `Ipc`), and the namespace reads now propagate errors via `?` with `.context()` instead of swallowing them with `if let Ok(...)`.

andrewdunndev:
CI failures are all quay.io/registry 502 Bad Gateway errors during container image pulls, not code issues. Should pass on re-run.

cgwalters (approved):

cgwalters:
Followup to regression from this in https://github.com/bootc-dev/bootc/pull/2117

Issue #1270, opened 2025-04-11
system-reinstall-bootc Add indication of the system "working" during the layer download and deploy steps


arburka:
The process of running the `system-reinstall-boot` command takes an indeterminant amount of time based on image size and complexity during the layer download and deployment steps. During these times, the user is not given much feedback about the system's current status.

The user should receive feedback that the system is still working/responding while these actions are taking place. A simple(?) way to accomplish this is the CLI "spinner" used when connecting to insights (attached), but whatever aligns with your current patterns would suffice.  

The amount and granularity of the information can be explored iteratively based on feedback if needed.

arburka:
<img width="420" alt="Image" src="https://github.com/user-attachments/assets/ea35446f-261a-417a-b71e-be1207727b87" />

PR #2091, opened 2026-03-24
system-reinstall-bootc: add progress indication during install steps

judavi:
Add progress indication to the long-running steps of system-reinstall-bootc to address #1270.

This also fixes a latent bug raised in review: `bootc_has_clean` was calling `podman run <image>` which could implicitly pull the image if it wasn't already cached locally, with no clear indication to the user.

The flow is now three clearly distinct phases:

1. Pull phase: pull_if_not_present explicitly pulls the image first, with a message so the user knows what is happening. podman pull streams its own progress output.
2. Capability check phase:  `bootc_has_clean` is called after the image is guaranteed to be local, with a spinner labeled "Checking image capabilities...".
3. Install phase : reinstall_command is now a pure command builder that takes has_clean: bool as a parameter and has no I/O side effects

image

### DIFF
--- a/crates/system-reinstall-bootc/src/main.rs
+++ b/crates/system-reinstall-bootc/src/main.rs
@@ -5,6 +5,7 @@ use bootc_utils::CommandRunExt;
 use clap::Parser;
 use fn_error_context::context;
 use rustix::process::getuid;
+use std::time::Duration;
 
 mod btrfs;
 mod config;
@@ -54,10 +55,23 @@ fn run() -> Result<()> {
 
     podman::ensure_podman_installed()?;
 
-    //pull image early so it can be inspected, e.g. to check for cloud-init
+    // Pull phase: explicitly pull the image before any other operations that use it.
+    // This ensures no implicit pulls happen in later steps (e.g. capability check).
     podman::pull_if_not_present(&opts.image)?;
 
     println!();
+
+    // Capability check phase: run after the image is guaranteed to be present locally.
+    let spinner = indicatif::ProgressBar::new_spinner();
+    spinner.set_style(
+        indicatif::ProgressStyle::default_spinner()
+            .template("{spinner} {msg}")
+            .expect("Failed to parse spinner template"),
+    );
+    spinner.set_message("Checking image capabilities...");
+    spinner.enable_steady_tick(Duration::from_millis(150));
+    let has_clean = podman::bootc_has_clean(&opts.image)?;
+    spinner.finish_and_clear();
 
     let ssh_key_file = tempfile::NamedTempFile::new()?;
     let ssh_key_file_path = ssh_key_file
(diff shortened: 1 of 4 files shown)


crates/tests-integration/Cargo.toml
```diff
@@ -30,13 +30,13 @@ bootc-kernel-cmdline = { path = "../kernel_cmdline", version = "0.0.0" }
...
 # dependency versions (cap-std-ext, etc.) to avoid conflicts.
-bcvk-qemu = { git = "https://github.com/bootc-dev/bcvk", rev = "5c42e91250ab87502b7beeb2c7ea94a54ee8f5d9" }
+bcvk-qemu = { git = "https://github.com/bootc-dev/bcvk", rev = "3b4dc84453d7ebac9e544d15931f0a9cd3d26215" }
 data-encoding = "2.9"
 indicatif = { workspace = true }
 libtest-mimic = "0.8.0"
 oci-spec = "0.9.0"
 rand = "0.10"
-rexpect = "0.7"
+rexpect = "0.6"
```
cgwalters: Hmm, this looks like a likely unintentional change?

crates/system-reinstall-bootc/src/main.rs
```diff
@@ -85,6 +85,9 @@ fn run() -> Result<()> {
 
     prompt::temporary_developer_protection_prompt()?;
 
+    println!("Starting bootc installation. This may take several minutes...");
```
cgwalters, line 103: This part makes sense

crates/system-reinstall-bootc/src/podman.rs
```diff
@@ -80,7 +81,17 @@ pub(crate) fn reinstall_command(opts: &ReinstallOpts, ssh_key_file: &str) -> Res
...
-    if bootc_has_clean(&opts.image)? {
+    let spinner = indicatif::ProgressBar::new_spinner();
+    spinner.set_style(
+        indicatif::ProgressStyle::default_spinner()
+            .template("{spinner} {msg}")
+            .unwrap(),
+    );
+    spinner.set_message("Checking image capabilities...");
+    spinner.enable_steady_tick(Duration::from_millis(150));
+    let has_clean = bootc_has_clean(&opts.image)?;
```
cgwalters: Hmm wait wait...did we end up pulling the image here? I think we did...if so that seems like the real bug. I think we should break out a clearly distinct pull phase.

cgwalters (commented):
Sorry for the delay in review and thank you so much for your contribution!

crates/tests-integration/Cargo.toml
judavi: That's correct. Thank you! I'm not that used to use Codespaces for development

crates/system-reinstall-bootc/src/podman.rs
judavi: You're right `bootc_has_clean` was calling `podman run <image>` which would have implicitly pulled the image if it wasn't already cached locally. Fixed in the latest push: `bootc_has_clean` is now `pub(crate)` and called explicitly from `run()` in `main.rs`, after `pull_if_not_present` completes. `reinstall_command` is now a pure command builder that takes `has_clean: bool` as a parameter with no I/O side effects. The flow is now three clearly distinct phases: pull → capability check (spinner) → install

judavi:
Thanks @jmarrero ! Commits are squashed in one now. 
Btw It was a nice experience meeting the team during the Kubecon :)

cgwalters:
@jmarrero this one is blocked on your requested changes

PR #2094, opened 2026-03-24
cli: Add tag-aware upgrade operations

gursewak1997:
Implements `bootc upgrade --tag <TAG>` to upgrade to a different tag
of the current image without typing the full registry path.

Implementation:
- Added `ImageReference::with_tag()` method in spec.rs
- Uses OCI Reference API for registry and containers-storage transports
- Falls back to string manipulation for oci: and other complex transports
- Properly strips digests when replacing tags
- Works with all transport types

Testing:
- Unit tests cover registry, containers-storage, digest handling, and edge cases
- Integration test (test-upgrade-tag.nu) uses containers-storage to verify
  full tag-switching workflow across reboots without registry dependency

### DIFF
--- a/crates/lib/src/cli.rs
+++ b/crates/lib/src/cli.rs
@@ -123,6 +123,13 @@ pub(crate) struct UpgradeOpts {
     #[clap(long, conflicts_with_all = ["check", "download_only"])]
     pub(crate) from_downloaded: bool,
 
+    /// Upgrade to a different tag of the currently booted image.
+    ///
+    /// This derives the target image by replacing the tag portion of the current
+    /// booted image reference.
+    #[clap(long)]
+    pub(crate) tag: Option<String>,
+
     #[clap(flatten)]
     pub(crate) progress: ProgressOptions,
 }
@@ -1047,7 +1054,19 @@ async fn upgrade(
     let repo = &booted_ostree.repo();
 
     let host = crate::status::get_status(booted_ostree)?.1;
-    let imgref = host.spec.image.as_ref();
+    let current_image = host.spec.image.as_ref();
+
+    // Handle --tag: derive target from current image + new tag
+    let derived_image = if let Some(ref tag) = opts.tag {
+        let image = current_image.ok_or_else(|| {
+            anyhow::anyhow!("--tag requires a booted image with a specified source")
+        })?;
+        Some(image.with_tag(tag)?)
+    } else {
+        None
+    };
+
+    let imgref = derived_image.as_ref().or(current_image);
     let prog: ProgressWriter = opts.progress.try_into()?;
 
     // If there's no specified image, let's be nice and check if the booted system is using rpm-ostree
@@ -1063,7 +1082,9 @@ async fn upgrade(
         }
     }
 
-    let spec = RequiredHostSpec::from_spec(&host.spec)?;
+    let imgref = imgref.ok_or_else(|| anyhow::anyhow!("No image source specified"))?;
+    // Use the derived image reference (if --tag was specified) instead of the spec's image
+    let spec = RequiredHostSpec { image: imgref };
     let booted_image = host
         .status
         .booted
@@ -1071,7 +1092,6 @@ async fn upgrade(
         .map(|b| b.query_image(repo))
         .transpose()?
         .flatten();
-    let imgref = imgref.ok_or_else(|| anyhow::anyhow!("No image source specified"))?;
     // Find the currently queued digest, if any before we pull
     let staged = host.status.staged.as_ref();
     let staged_image = staged.as_ref().and_then(|s| s.image.as_ref());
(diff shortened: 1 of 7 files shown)


crates/lib/src/cli.rs
```diff
@@ -123,6 +125,17 @@ pub(crate) struct UpgradeOpts {
...
+    /// Upgrade to a different tag of the currently booted image.
+    ///
+    /// This derives the target image by replacing the tag portion of the current
+    /// booted image reference. Only works when booted from a registry image.
+    #[clap(long)]
+    pub(crate) tag: Option<String>,
+
+    /// List available tags for the currently booted image repository.
+    #[clap(long, conflicts_with_all = ["apply", "download_only", "from_downloaded"])]
+    pub(crate) list_tags: bool,
```
cgwalters: This one feels like it should be a separate operation.

crates/lib/src/cli.rs
```diff
@@ -1220,6 +1253,70 @@ async fn upgrade(
     Ok(())
 }
 
+/// Derive a new image reference by replacing the tag of the current image
+fn derive_image_with_tag(current: &ImageReference, new_tag: &str) -> Result<ImageReference> {
+    // Only works for registry transport
```
cgwalters: I think we can easily support non-registry transports here.

crates/lib/src/cli.rs
```diff
@@ -1220,6 +1253,70 @@ async fn upgrade(
...
+        );
+    }
+
+    // Parse the current image reference
+    let reference: oci_spec::distribution::Reference = current.image.parse()?;
+
+    // Build new image reference with the new tag (stripping any digest)
+    let registry = reference.registry();
+    let repository = reference.repository();
+    let new_image = format!("{}/{}:{}", registry, repository, new_tag);
```
cgwalters: There's an API for this https://docs.rs/oci-spec/latest/oci_spec/distribution/struct.Reference.html#method.with_tag

crates/lib/src/cli.rs
```diff
@@ -1220,6 +1253,70 @@ async fn upgrade(
...
+
+    // Construct repository name: registry + repository
+    let registry = reference.registry();
+    let repository = reference.repository();
+    let repo_name = format!("{}/{}", registry, repository);
+
+    let repo = format!("docker://{}", repo_name);
+
+    // Use skopeo to list tags
+    let output = tokio::process::Command::new("skopeo")
```
cgwalters: One problem is this won't use the ostree/bootc pull secret by default; we have a whole containers-image-proxy abstraction that we need to use in general for this.

Though the next problem here is that code doesn't expose an abstraction for this...but should.

crates/lib/src/cli.rs
```diff
@@ -1220,6 +1253,70 @@ async fn upgrade(
...
+        .output()
+        .await
+        .context("Failed to execute skopeo")?;
+
+    if !output.status.success() {
+        let stderr = String::from_utf8_lossy(&output.stderr);
+        anyhow::bail!("Failed to list tags: {}", stderr);
+    }
+
+    let stdout = String::from_utf8_lossy(&output.stdout);
```
cgwalters: For things like this it would be a lot cleaner to use `skopeo inspect` which outputs JSON. I think it's almost always wrong to use `from_utf8_lossy`.

Perhaps this might be best done with `upgrade --check`? We could also cache the latest available tags there.

crates/lib/src/cli.rs
```diff
@@ -2203,6 +2300,78 @@ mod tests {
...
+            Opt::Upgrade(opts) => {
+                assert_eq!(opts.tag, Some("v1.1".to_string()));
+                assert!(opts.check);
+            }
+            _ => panic!("Expected Upgrade variant"),
+        }
+    }
+
+    #[test]
+    fn test_derive_image_with_tag() {
```
cgwalters: Nothing wrong with unit tests (although this is basically reimplementing the `with_tag` API I mentioned) - but we really want an integration (tmt) test for this. Right now our default integration tests don't involve a registry (that's a different issue) but I think we can do the same with `containers-storage:` - basically bootc switch to one tag and then switch to a different one etc.

crates/lib/src/cli.rs
gursewak1997: I've switched to skopeo inspect with JSON parsing (parsing the RepoTags field) instead of from_utf8_lossy. Also integrated it with `upgrade --check` as an optional `--show-tags` flag.
Caching the tags during check is a good idea. I can add that if you think it's worth the added complexity.

crates/lib/src/cli.rs
gursewak1997: Added

crates/lib/src/cli.rs
```diff
@@ -1220,6 +1263,87 @@ async fn upgrade(
     Ok(())
 }
 
+/// Derive a new image reference by replacing the tag of the current image
+fn derive_image_with_tag(current: &ImageReference, new_tag: &str) -> Result<ImageReference> {
```
cgwalters: Hmm I think this should probably be a method on `ImageReference` instead of being all the way at the top in the CLI code.

That said I am pretty sure for `containers-storage` we can also create a `Reference` instead of reimplementing parsing.

The complicated one is `oci` because it's filesystem paths and `:` creates ambiguity; there's some complex code IIRC in containers/container-libs around this in Go which we may need to replicate. I think this came up elsewhere.

crates/lib/src/cli.rs
cgwalters: You marked this resolved but I don't think it is right? IOW if the image is behind a pull secret then this will fail. Look at the containers-image-proxy code that we use. We need to pass the ostree auth file to skopeo here.

cgwalters: Bigger picture can we just drop this list-tags functionality for now and instead suggest to users to directly run `skopeo inspect` on their own?

I know it's less ergonomic and this whole thing is just about ergonomics, but hopefully the `--tag` is enough.

My worry is basically again:

- We must handle authfiles and such using the ostree/bootc one
- This feels like a kind of scope creep into general "inspect the remote container" reference but users can already just run `skopeo` with their own options if that's what they want.

Maybe we could add `skopeo inspect --booted` for example? But that still doesn't solve the authfile discrepancy.

(I know the authfile thing is also quite annoying...see also https://github.com/containers/image/pull/2600 )

gursewak1997: I see. Dropped the `list-tags` functionality for now

PR #2097, opened 2026-03-26
container-export: Handle metacopy=off overlay setups

cgwalters:
This fixes a bug in `container export` on overlayfs setups with `metacopy=off` (a common setup). In that scenario we will see a single mountpoint, but different device+inode.

Change hardlink tracking to account for device+inode.

### DIFF
--- a/crates/lib/src/container_export.rs
+++ b/crates/lib/src/container_export.rs
@@ -134,17 +134,12 @@ fn export_filesystem_walk<W: Write>(
     root_dir: &cap_std_ext::cap_std::fs::Dir,
     sepolicy: Option<&ostree::SePolicy>,
 ) -> Result<()> {
-    use cap_std_ext::cap_primitives::fs::MetadataExt;
     use std::path::Path;
 
-    // Get the device number of the root directory - we should never see a different device
-    // since we use noxdev() which prevents crossing mount points
-    let root_meta = root_dir.dir_metadata()?;
-    let expected_dev = root_meta.dev();
-
-    // Track hardlinks: maps inode -> first path seen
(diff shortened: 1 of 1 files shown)


crates/lib/src/container_export.rs
```diff
@@ -250,34 +244,21 @@ fn add_file_to_tar_from_walk<W: Write>(
...
-    hardlinks: &mut HashMap<u64, std::path::PathBuf>,
+    hardlinks: &mut HashMap<(u64, u64), std::path::PathBuf>,
 ) -> Result<()> {
     use cap_std_ext::cap_primitives::fs::{MetadataExt, PermissionsExt};
     use std::path::Path;
 
     let filename_path = Path::new(filename);
     let metadata = dir.metadata(filename_path)?;
 
-    // Skip files on different devices (e.g., bind mounts in containers like /etc/hosts).
```
jeckersb, line 262: Does this mean that now we would start including the bind-mounted /etc/hosts when we export?

hone: I believe it does.

cgwalters: As documented, this expects the image as a separate mount point:

```
podman run --rm \
    --mount=type=image,source=quay.io/example/example,target=/run/target \
    quay.io/example/example \
    bootc container export --kernel-in-boot /run/target > example.tar
```

And in that case, there won't be any container runtime injected content like `/etc/hosts`.

cgwalters: (IOW I messed this up originally, the previous code was never necessary)

Issue #1229, opened 2025-03-25
install: Add `--karg-delete`


cgwalters:
In some relatively rare use cases, one might want to *remove* a kernel argument shipped in the container image.

This would be the same logic in the `ostree admin deploy --karg-delete` path.

ref https://github.com/openshift/assisted-installer/pull/1076#discussion_r2012829469

PR #2105, opened 2026-03-29
install: Add `--karg-delete`

hone:
Following up from the contribfest session. This adds `--karg-delete` to install.

In some relatively rare use cases, one might want to remove a kernel argument shipped in the container image.

Fixes: #1229

### DIFF
--- a/tmt/tests/booted/test-install-karg-delete.nu
+++ b/tmt/tests/booted/test-install-karg-delete.nu
@@ -2,6 +2,8 @@
 # tmt:
 #   summary: Test bootc install --karg-delete
 #   duration: 30m
+# extra:
+#   fixme_skip_if_composefs: true
 #
 use std assert
 use tap.nu
@@ -13,9 +15,6 @@ let target_image = (tap get_target_image)
 mkdir /var/mnt
 truncate -s 10G disk.img
 mkfs.ext4 disk.img
-if (tap is_composefs) {
-  tune2fs -O verity disk.img
-}
 mount -o loop disk.img /var/mnt
 
 # Mask off the bootupd state to reproduce https://github.com/bootc-dev/bootc/issues/1778
@@ -27,14 +26,8 @@ setenforce 0
 mkdir /etc/bootc/install
 { install: { kargs: ["foo=bar"] } } | to toml | save /etc/bootc/install/99-test.toml
 
-let base_args = $"bootc install to-filesystem --disable-selinux --bootloader none --source-imgref ($target_image) --karg-delete localtestkarg --karg-delete foo"
-let install_cmd = if (tap is_composefs) {
-    $"($base_args) --composefs-backend /var/mnt"
-} else {
-    $"($base_args) /var/mnt"
-}
+tap run_install $"bootc install to-filesystem --disable-selinux --bootloader none --source-imgref ($target_image) --karg-delete localtestkarg --karg-delete foo /var/mnt"
 
-tap run_install $install_cmd
 
 # Verify the karg is gone from the bootloader entries
 let entries = (glob /var/mnt/boot/loader/entries/*.conf
--- a/tmt/plans/integration.fmf
+++ b/tmt/plans/integration.fmf
@@ -244,4 +244,5 @@ execute:
     how: fmf
     test:
       - /tmt/tests/tests/test-40-install-karg-delete
+  extra-fixme_skip_if_composefs: true
 # END GENERATED PLANS


hone:
I wasn't sure about a few things when implementing this.

1. What is the preferred order of precedence for how `--karg` and `--karg-delete` interface. kargs can come from root_setup, install_config, and the CLI. In order to let you delete kargs and allow you to replace it with a CLI flag, I deleted the kargs before applying the CLI flags.
2. Do all install CLI flags also get applied to config? I modeled a configuration option based off of `kargs`.
3. I wasn't sure which tests to add, so I added some unit tests, integration, and tmt tests.

hone:
For my tmt test, I used `fixme_skip_if_composefs: true`. It didn't look like the `composefs` boot supported the CLI flags.

henrywang:
Hi @hone, please skip Packit s390x rpm build issue. The build is successful, but status can't be updated in PR. I already ask for help in Packit channel.

tmt/tests/booted/test-install-karg-delete.nu
```diff
@@ -0,0 +1,47 @@
+# number: 40
+# tmt:
+#   summary: Test bootc install --karg-delete
+#   duration: 30m
+# extra:
+#   fixme_skip_if_composefs: true
```
Johan-Liebert1, line 6: This should work for composefs backend as well since the config gathering in the install path is pretty much the same for ostree and composefs. Were you getting some sort of error when testing this for composefs backend?

hone: Just that if you use `--to-filesystem` you need to setup the entire boot partition table, so you can't use `--bootloader none`.

I was looking through [`setup_composefs_boot`](https://github.com/bootc-dev/bootc/blob/main/crates/lib/src/bootc_composefs/boot.rs#L1235) and it didn't look like it handled the CLI karg/-delete args I was working on in [`install_container`](https://github.com/bootc-dev/bootc/blob/main/crates/lib/src/install.rs#L1020).

crates/lib/src/install.rs
```diff
@@ -1142,6 +1153,19 @@ async fn install_container(
...
+    if let Some(install_config_karg_deletes) = install_config_karg_deletes {
+        for karg_delete in install_config_karg_deletes {
+            karg_deletes.push(karg_delete);
+        }
+    }
+    if let Some(deletes) = state.config_opts.karg_delete.as_ref() {
+        for karg_delete in deletes {
+            karg_deletes.push(karg_delete);
+        }
+    }
```
cgwalters, line 1166: Minor style nit I think this could all be one combinator chain via `.chain().flatten().collect()`

crates/lib/src/install.rs
```diff
@@ -1224,6 +1248,18 @@ async fn install_container(
     Ok((deployment, aleph))
 }
 
+pub(crate) fn delete_kargs(existing: &mut Cmdline, deletes: &Vec<&str>) {
+    for delete in deletes {
+        if let Some(param) = utf8::Parameter::parse(&delete) {
```
cgwalters, line 1253: Hmm when would we get None from parsing here? Just when it's an empty string right? We should probably have filtered those out or errored if I'm right

tmt/tests/booted/test-install-karg-delete.nu
cgwalters: We can only do this with type1 BLS though to be clear with composefs, not UKIs. We have to error out on any local karg manipulations there.

crates/lib/src/install.rs
jeckersb: Empty strings or strings with only whitespace will return `None`, anything else should successful parse.

cgwalters:
> What is the preferred order of precedence for how --karg and --karg-delete interface. kargs can come from root_setup, install_config, and the CLI. In order to let you delete kargs and allow you to replace it with a CLI flag, I deleted the kargs before applying the CLI flags.

I think as implemented the CLI should always win, right? We should probably error with `--karg=foo` and `--karg-delete=foo` though.

> Do all install CLI flags also get applied to config? I modeled a configuration option based off of kargs.

I'm not totally parsing this - it's more that the CLI and config all fold into the bootloader ocnfig.

> I wasn't sure which tests to add, so I added some unit tests, integration, and tmt tests.

It's confusing, what you did is fine, but a lot of the install tests were already duplicated. It's a complex problem though, I'll write up a separate tracker issue for this.

PR #2106, opened 2026-03-30
test: skip OSCI gating AVC test

henrywang:
Skip in TMT test to avoid duplicate with downstream


cgwalters:
The test failures here are really strange, I don't understand how this PR could have triggered them. Unless there's some kind of race somehow, like we installed with an older image?

henrywang:
> The test failures here are really strange, I don't understand how this PR could have triggered them. Unless there's some kind of race somehow, like we installed with an older image?

The failure is in the prepare phase - tmt's essential-requires step is trying to install audit, policycoreutils on the VM. These are dependencies of the `check: how: avc` plugin.

Tricky thing is even though the check is `enabled: false`, tmt still installs the plugin's essential requirements.

crates/xtask/src/tmt.rs
```diff
@@ -1070,6 +1070,18 @@ pub(crate) fn update_integration() -> Result<()> {
     tests_content.push_str("# THIS IS GENERATED CODE - DO NOT EDIT\n");
     tests_content.push_str("# Generated by: cargo xtask tmt\n");
     tests_content.push_str("\n");
+    // bootc probes for SELinux mac_admin capability by attempting chcon with
+    // an intentionally invalid label, which generates expected AVC denials.
+    // Report as informational only in OSCI gating test
+    tests_content
```
cgwalters, line 1076: For followup, we should split this out to a separate file via include_str! or so

Issue #2102, opened 2026-03-28
[ComposeFS-Native] Garbage collection sometimes removes all files from /Linux in EFI partition


hecknt:
I have no idea how to recreate this bug, but 2 times now after an update, my system has been unable to fetch the boot entries. I discovered that this is because there is **nothing** in my EFI partition's `/Linux` folder.

<img width="377" height="224" alt="Image" src="https://github.com/user-attachments/assets/295dcb7e-bcfd-456c-b30b-d21412879529" />

This leaves my system entirely unbootable, as there is no kernel to boot into.

Johan-Liebert1:
Seems like a bug with the GC code. Is this reproducible? Even if it's not, it would be helpful if you could detail out the upgrade process. Like how many times you upgraded/switched  to a different image

Johan-Liebert1:
Also, are you using UKIs or simple BLS entries?

hecknt:
I'm using simple BLS entries, no UKIs in sight. My upgrade process is (mostly) automatic, done by [uupd](https://github.com/ublue-os/uupd) (which just runs bootc upgrade). I think this happens when it runs multiple upgrades in one boot?

I haven't tried to reproduce this in a test environment yet, I'll do so tonight. I've been busy with other stuff.

hecknt:
I did some testing and it seems like the `/EFI/Linux` directory gets cleared when i run `bootc upgrade`, and then is repopulated with the new kernel from the next deployment after it finishes. Personally I feel like it shouldn't run GC until **after** the upgrade is finished, maybe? That feels a lot more sane, I think.

PR #2108, opened 2026-03-31
BLS shared entries GC fix

Johan-Liebert1:
### boot/config: Add method to get boot artifact name

Add a method in BLSConfig and Grub Menuconfig to get the boot artifact
name, i.e. get the name of the UKI or the name of the directory
containing the Kernel + Initrd. The names are stripped of all our custom
prefixes and suffixes, so basically they return the verity digest part
of the name.

This is useful for GC-ing Kernel + Initrd that are shared among multiple
deployments since we can't rely on the composefs= parameter in the
options as the cmdline verity digest might be different than the verity
digest of the shared Kernel + Initrd.

Tests written by Claude Code (Opus)

---

### composefs/boot: Change the way we find sharable vmlinuz + initrd

Instead of looking in the ".origin" files and trying to match the
boot_digest to digests in other origin files, we now simply re-compute
the sha256sum for vmlinuz + initrd for all boot entries present.

This fixes the bug that arises after mutiple upgrades where the original
deployment that created the boot entry has been garbage collected, so we
end up linking to another deployment that does have the same boot
digest, but the verity digest doesn't match the verity digest used for
the name of the directory where we store the kernel + initrd pair.

---

### composefs/gc: Fix shared entry GC even when they're in use

This fixes a bug where a shared Type1 entry would get GCd even when it's
in use due to the original image that created it being deleted. Combined
with the fact that we were comparing the fsverity digest in the options
field of the BLS config (which will be different than the name of the
directory containing the vmlinuz + initrd pair).

Now, we compare against the directory name when GC-ing boot binaries

Fixes: https://github.com/bootc-dev/bootc/issues/2102

---

test: Add test for GC-ing shared Type1 entries

Add a test to make sure we do not GC shared Type1 entries when they're
still referenced

### DIFF
--- a/hack/provision-derived.sh
+++ b/hack/provision-derived.sh
@@ -17,6 +17,9 @@ mkdir -p ~/.config/nushell
 echo '$env.config = { show_banner: false, }' > ~/.config/nushell/config.nu
 touch ~/.config/nushell/env.nu
 
+# We don't want openh264
+rm -f "/etc/yum.repos.d/fedora-cisco-openh264.repo"
+
 . /usr/lib/os-release
 case "${ID}-${VERSION_ID}" in
     "centos-9")
--- a/tmt/tests/booted/test-composefs-gc.nu
+++ b/tmt/tests/booted/test-composefs-gc.nu
@@ -53,9 +53,6 @@ def second_boot [] {
 
     let path = cat /var/large-file-marker-objpath
 
-    echo "\$path"
-    echo $path
-
     assert ($path | path exists)
 
     # Create another image with a different initrd so we can test kernel + initrd cleanup
@@ -107,13 +104,11 @@ def third_boot [] {
     let boot_dir = if ($bootloader | str downcase) == "systemd" {
         # TODO: Some concrete API for this would be great
         mkdir /var/tmp/efi
-        mount /dev/vda2 /var/tmp/efi
+        mount /dev/disk/by-partlabel/EFI-SYSTEM /var/tmp/efi
         "/var/tmp/efi/EFI/Linux"
     } else {
         "/sysroot/boot"
     }
-
-    print $"bootdir ($boot_dir)"
 
     assert ($"($boot_dir)/($dir_prefix)($booted_verity)" | path exists)
 
@@ -125,11 +120,27 @@ def third_boot [] {
 
     echo $"($boot_dir)/($dir_prefix)(cat /var/first-verity)" | save /var/to-be-deleted-kernel
 
+    # Switching and rebooting here won't delete the old kernel because we still
+    # have it as the rollback deployment
+    echo "
+        FROM localhost/bootc-derived-initrd
+        RUN echo 'another file' > /usr/share/another-one
+    " | podman build -t localhost/bootc-prefinal . -f -
+
+
+    bootc switch --transport containers-storage localhost/bootc-prefinal
+
+    tmt-reboot
+}
+
+def fourth_boot [] {
+    assert equal $booted.image.image "localhost/bootc-prefinal"
+
     # Now we create a new image derived from the current kernel + initrd
     # Switching to this and rebooting should remove the old kernel + initrd
     echo "
         FROM localhost/bootc-derived-initrd
-        RUN echo 'another file' > /usr/share/another-one
+        RUN echo 'another file 1' > /usr/share/another-one-1
     " | podman build -t localhost/bootc-final . -f -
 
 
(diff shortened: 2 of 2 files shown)


crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -355,52 +375,35 @@ pub(crate) fn compute_boot_digest_uki(uki: &[u8]) -> Result<String> {
...
-            }
-
-            // No SHASum recorded in origin file
-            // `symlink_to` is already none, but being explicit here
-            None => symlink_to = None,
-        };
+        if entry_digest == digest {
+            return Ok(Some(dir_name));
+        }
     }
```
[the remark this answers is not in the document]
Johan-Liebert1: This would be fixed by the GC, but the GC would need to be run manually. Maybe we update the error message, or should we run the GC before and after the upgrade? @cgwalters I believe ostree does a GC before the upgrade operation?

cgwalters:
Currently btw ostree never shares entries between deployments, but it makes sense to do so for karg differences!

crates/lib/src/parsers/bls_config.rs
```diff
@@ -205,6 +206,59 @@ impl BLSConfig {
         }
     }
 
+    /// Returns name of UKI in case of EFI config
+    /// Returns name of the directory containing Kernel + Initrd in case of Non-EFI config
+    ///
+    /// The names are stripped of our custom prefix and suffixes, so this basically returns
```
cgwalters: I don't like the "basically" here. Either it's a verity digest or it's not.

crates/lib/src/parsers/bls_config.rs
```diff
@@ -205,6 +206,59 @@ impl BLSConfig {
...
+    /// the verity digest part of the name
+    pub(crate) fn boot_artifact_name(&self) -> Result<&str> {
```
cgwalters, line 214: So in that vein...and yes this is not a new problem but I'd like to start using a formal type for string digests. https://docs.rs/oci-spec/latest/oci_spec/image/struct.Digest.html is one candidate.

crates/lib/src/parsers/bls_config.rs
```diff
@@ -585,4 +639,128 @@ mod tests {
         assert!(config_final < config_rc1);
         Ok(())
     }
+
+    #[test]
+    fn test_boot_artifact_name_efi_success() -> Result<()> {
```
cgwalters, line 640: Let's rework these to be table-driven tests, it really helps reduce the amount of code.

crates/lib/src/bootc_composefs/boot.rs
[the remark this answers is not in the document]
cgwalters: Hmm well one thing here is we should ensure we atomically generate the dir by using the "create tempdir + rename_exchange" model.

But yes we should also be robust to partially written dirs - and sure we could GC those before upgrade too, but it seems like it'd also work to ensure we ignore corrupted/partial ones instead of erroring out.

commits
```diff
@@ -0,0 +1,67 @@
+commit f074ee3087538faa1aca0e9a6d906d1d1bfbc25a
```
cgwalters: This looks like an accidental add?

crates/lib/src/parsers/bls_config.rs
cgwalters: I put up https://github.com/composefs/composefs-rs/pull/278 related to this

commits
Johan-Liebert1: Oops, my bad. Removed

cgwalters:
> error: Status: Getting composefs deployment status: Getting composefs deployment status: Listing bootloader entries: Boot directory missing expected prefix 'bootc_composefs-': 76de4528c48f28077ac9322d91b57885efe0d1942239f6a5eabcdfdd18d1c9dd7718f332d2645af6fa5bf4b15ef6e862510fe29863153f5b6f15c2d8f9582c9e

Ah hah, the upgrade test looks like it's correctly failing here. We should probably try to accomodate the previous layout?

crates/lib/src/parsers/bls_config.rs
Johan-Liebert1: I tried to add this in this PR, but it didn't really fit here. I'll open a separate PR for this

crates/lib/src/parsers/bls_config.rs
Johan-Liebert1: I think for this we can use the `Sha512HashValue` and its 256 counterpart from composefs as that's what fits here

Johan-Liebert1:
Tests on centos10 seem to be failing due to issues with mirrors

crates/lib/src/parsers/bls_config.rs
Johan-Liebert1: So throwing Claude code at this, the reduction in loc isn't really that much. The diff just turns out the be -167 lines

```
 crates/lib/src/parsers/bls_config.rs      | 547 ++++++++++++++++++++++++++++++++++++++++++++++++++++-----------------------------------------------------------------------------------------
 crates/lib/src/parsers/grub_menuconfig.rs | 414 ++++++++++++++++++++++++++++++++++++++++++++++++++--------------------------------------------------------
 2 files changed, 397 insertions(+), 564 deletions(-)
```

Johan-Liebert1: Mainly because `cargo fmt` ends up putting stuff on separate lines anyway, and the test cases are less readable now imo

PR #2109, opened 2026-03-31
ostree-ext: reuse prior packing with exclusive components

### LINKED ISSUE #2057: ostree-ext: Packing with prior build fails if build has exclusive components
When building a chunked image, one is supposed to be able to take a prior build's layer plan into account, minimizing layer changes with an update. However, if the image to be built has exclusive components (which should each be placed in a layer of their own), using a prior build's layer plan fails, making ostree-ext fall back to computing a new layer plan. (I did some local testing with `rpm-ostree compose build-chunked-oci` that seems to confirm this behavior.)

### Steps to reproduce

1. Build a bootc container image with some files marked as belonging to an exclusive component, for example [using the `user.component` xattr](https://coreos.github.io/rpm-ostree/build-chunked-oci/#assigning-files-to-specific-layers).
2. Create a chunked version of this image with build-chunked-oci.
3. Build a modified version of the image from step (2), for example by installing some additional packages.
4. Set the environment variable `RUST_LOG=debug`.
5. Use build-chunked-oci to created a chunked version of the modified image from step (3), outputting it to the same image reference in local storage that the image from step (2) was stored in. This makes build-chunked-oci try to reuse the layer plan of the image from step (2); however, the debug output shows that it fails to reuse the layer plan and falls back to creating a new layer plan.

### Analysis

I believe the problem is that, in `ostree_ext::chunking::basic_packing_with_prior_build`, to use the prior build's layer plan, we require `bin_size` to be >= the number of layers of the prior build (excluding the ostree commit layer):
https://github.com/bootc-dev/bootc/blob/99da62b6f09ed42841396ac66298030f25daf439/crates/ostree-ext/src/chunking.rs#L640-L643

But on the other hand, the `bin_size` input comes from here: https://github.com/bootc-dev/bootc/blob/99da62b6f09ed42841396ac66298030f25daf439/crates/ostree-ext/src/chunking.rs#L381-L384
And this is essentially computed as the maximum layer count minus the number of layers that have already been reserved at that point in the function... but this takes place after we've already made a layer for each exclusive component earlier in the same function:
https://github.com/bootc-dev/bootc/blob/99da62b6f09ed42841396ac66298030f25daf439/crates/ostree-ext/src/chunking.rs#L343-L359
So, I think this will always fail if there are exclusive components.

I haven't wrapped my head around the full packing algorithm well enough yet to understand exactly which part of this needs to be fixed and how best to fix it, but it seems like the problem is that it doesn't take into account the likely presence of the same exclusive components in the prior build. Either that or maybe the requirement for the bin size is wrong?

officialasishkumar:
## Summary
Reuse the prior-build packing layout when exclusive components have already been split into dedicated layers.

## Changes
- drop prior-build bins that become empty after exclusive components are filtered out
- keep the trailing reserved "new packages" bin so added regular components still have a stable destination
- add a regression test covering prior-build reuse with exclusive components

## Testing
- `cargo fmt --check --all`
- `docker run --rm -v /home/ashish/asish_workspace/personal/bootc:/src -w /src fedora:42 bash -lc 'set -euo pipefail && dnf -y install rust cargo pkgconf-pkg-config glib2-devel ostree-devel openssl-devel systemd-devel libzstd-devel git gcc make && cargo test -p ostree-ext test_advanced_packing -- --nocapture && cargo test -p ostree-ext test_exclusive_chunks -- --nocapture'`

Fixes #2057

### DIFF
--- a/crates/ostree-ext/src/chunking.rs
+++ b/crates/ostree-ext/src/chunking.rs
@@ -637,11 +637,6 @@ fn basic_packing_with_prior_build<'a>(
         .collect();
     let mut curr_build = curr_build?;
 
-    if (bin_size.get() as usize) < curr_build.len() {
-        tracing::debug!("bin_size = {bin_size} is too small to be compatible with the prior build");
-        return Ok(None);
-    }
-
     // View the packages as unordered sets for lookups and differencing
     let prev_pkgs_set: BTreeSet<String> = curr_build
         .iter()
@@ -667,6 +662,21 @@ fn basic_packing_with_prior_build<'a>(
     let removed: BTreeSet<&String> = prev_pkgs_set.difference(&curr_pkgs_set).collect();
     for bin in curr_build.iter_mut() {
         bin.retain(|pkg| !removed.contains(pkg));
+    }
+
+    // Exclusive-component bins are already carved out by process_mapping(), so
+    // keep only the bins that still carry regular components plus the reserved
+    // "new packages" bin at the end of the prior layout.
+    let last_idx = curr_build.len().saturating_sub(1);
+    curr_build = curr_build
+        .into_iter()
+        .enumerate()
+        .filter_map(|(idx, bin)| (!bin.is_empty() || idx == last_idx).then_some(bin))
+        .collect();
+
+    if (bin_size.get() as usize) < curr_build.len() {
+        tracing::debug!("bin_size = {bin_size} is too small to be compatible with the prior build");
+        return Ok(None);
     }
 
     // Handle updated packages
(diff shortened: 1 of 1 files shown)


jmarrero:
This seems pretty good, the failures seem like flakes. Thanks @officialasishkumar you had something else you wanted to add?

officialasishkumar:
@jmarrero Nothing else to add from my side right now. I refreshed the branch onto current main, and I will keep an eye on the flakes and the updated checks.

officialasishkumar:
@jmarrero Thanks for the review! No, this PR covers everything I intended. Nothing else to add. Appreciate you taking the time to look through it!

cgwalters (approved):
Thanks for doing this! I had some agents crosscheck this too.

What I would say here though is the new main investment here is in https://github.com/jlebon/chunkah

PR #2115, opened 2026-04-01
composefs: Open deployments dir from physical root, not ambient path

cgwalters:
find_vmlinuz_initrd_duplicates() previously opened the absolute path /sysroot/state/deploy via STATE_DIR_ABS, which breaks in the case when we're doing an install outside of a container.

We hit this in https://github.com/bootc-dev/bootc/pull/1911

Fix by passing a Dir opened relative to the target's physical_root. During fresh install the state dir naturally does not exist yet, so open_dir_optional returns None and the check is skipped -- no special-case guard needed.

Also plumb physical_root through the setup_composefs_bls_boot match arms, and replace the hardcoded "/sysroot" in the Upgrade arm with storage.physical_root_path.

### DIFF
--- a/crates/lib/src/bootc_composefs/boot.rs
+++ b/crates/lib/src/bootc_composefs/boot.rs
@@ -117,7 +117,7 @@ use crate::{
 use crate::{
     composefs_consts::{
         BOOT_LOADER_ENTRIES, ORIGIN_KEY_BOOT, ORIGIN_KEY_BOOT_DIGEST, STAGED_BOOT_LOADER_ENTRIES,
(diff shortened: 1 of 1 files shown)


Johan-Liebert1:
I think this is going to conflict with https://github.com/bootc-dev/bootc/pull/2108

I believe this should be solved by https://github.com/bootc-dev/bootc/pull/2108/changes/cfe6941cbe6cc08f319acfc3276fe6c2aa4e7edf which is included in #2108

cgwalters:
Yep agreed closing!

PR #2116, opened 2026-04-02
install: Add a config knob for container sigpolicy

### LINKED ISSUE #2112: Add knob in install config for `--enforce-container-sigpolicy`
Let's surface this in the install config `toml` so we can pipe this through image-builder

jbtrystram:
Surface the `--enforce-container-sigpolicy` flag as a config file option.

Fixes https://github.com/bootc-dev/bootc/issues/2112

Assisted-by: OpenCode.ai <Opus 4.6>

### DIFF
--- a/crates/lib/src/install.rs
+++ b/crates/lib/src/install.rs
@@ -1536,7 +1536,7 @@ async fn verify_target_fetch(
 async fn prepare_install(
     mut config_opts: InstallConfigOpts,
     source_opts: InstallSourceOpts,
-    target_opts: InstallTargetOpts,
+    mut target_opts: InstallTargetOpts,
     mut composefs_options: InstallComposefsOpts,
     target_fs: Option<FilesystemEnum>,
 ) -> Result<Arc<State>> {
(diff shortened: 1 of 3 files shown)


cgwalters (dismissed):
Looks sane. That said I think it'd be helpful for us to

- consolidate our install tests into one thing
- add an install test case for this from a container image source config, not just a synthetic unit test

cgwalters:
For now just missing `cargo fmt`

jeckersb (dismissed):

jbtrystram:
the s390x builds here simply timed out : 
` !! Copr timeout => sending INT`
And the composeFS failure is a quay flake.

cgwalters (approved):
Looks sane. That said I think it'd be helpful for us to

- consolidate our install tests into one thing
- add an install test case for this from a container image source config, not just a synthetic unit test

cgwalters:
I'm still trying to debug the merge queue, it's possible this needs a rebase. I am not sure it was the s390x failure that broke the queue.

Johan-Liebert1:
@cgwalters Maybe we need to do this as our tests are quite time consuming https://github.com/orgs/community/discussions/47293#discussioncomment-8012426

cgwalters:
Yep the tiemout is now 4 hours

PR #2117, opened 2026-04-02
cli: Gate IPC namespace entrance on CAP_SYS_ADMIN

cgwalters:
Extract the IPC namespace joining logic into a join_host_ipc_namespace() helper that checks for CAP_SYS_ADMIN first, since setns() requires it. Without the capability we skip the operation entirely.

Fixes: #2090


cgwalters:
In theory we could try to move some other things we do at install time here...or...perhaps safer, consolidate this with a single "re-exec self" dance. I kind of lean to the latter, it has a lower blast radius. But I think this change will work fine for now.

Issue #2120, opened 2026-04-03
composefs: /boot not mounted after `to-existing-root` install (missing fstab entry)


henrywang:
In my test, install composefs image into package mode machine with `podman run --rm --tls-verify=false --privileged --pid=host -v /:/target -v /dev:/dev -v /var/lib/containers:/var/lib/containers --security-opt label=type:unconfined_t localhost/bootc:tmt bootc install to-existing-root --target-transport containers-storage --composefs-backend --allow-missing-verity` . 

After `bootc install to-existing-root --composefs-backend` with GRUB bootloader, the system boots successfully but `/boot` is not mounted. This causes `bootc status` (and likely `bootc upgrade`) to fail when trying to open `/boot/grub2/`.

```shell
   $ ls -la /boot/
   total 8
   drwxr-xr-x.  2 root root  27 Apr  2 23:58 .
   drwxr-xr-x. 13 root root 358 Apr  2 23:58 ..

   $ ls -la /boot/grub2/
   grub2 dir missing

   $ findmnt /boot
   (no output — /boot is not mounted)
```

Full log: https://artifacts.osci.redhat.com/testing-farm/b535151d-d764-4928-91e3-18718047540a/

The composefs install path in `install_to_filesystem_impl()` (`crates/lib/src/install.rs` ~line 1963-2012) does **not** write a `/boot` entry to `/etc/fstab`. The ostree path does this at line 1203-1212.

`to-disk` does not have this issue because `/boot` is just a directory on the root partition (/sysroot/boot) and /sysroot is always mounted, /sysroot/boot is accessible. No fstab entry needed — it's not a separate partition.

But for `to-disk` Tpm2Luks (separate /boot partition), that might have same issue.

PR #2121, opened 2026-04-03
composefs: add a systemd.mount-extra kernel argument to the BLS boot entry

henrywang:
The composefs install path was not ensuring /boot gets mounted after reboot on systems with a separate /boot partition (e.g. to-existing-root installs). This broke `bootc status` and `bootc upgrade`.

Instead of writing to /etc/fstab (which conflicts with transient etc per https://github.com/bootc-dev/bootc/issues/1388), add a systemd.mount-extra kernel argument to the BLS boot entry. This tells systemd to mount the boot partition without requiring any /etc state.

Closes: https://github.com/bootc-dev/bootc/issues/2120
Assisted-by: Claude Code (Opus 4.6)


cgwalters:
What's the OS version? I really want to get away from the `/etc/fstab` usage for this, see also https://github.com/bootc-dev/bootc/issues/1388

For `to-existing-root`...I would prefer that we take the approach of using a karg instead per that issue.

henrywang:
> What's the OS version?

It's Fedora 43

PR #2123, opened 2026-04-03
install: Add discoverable-partitions config option

cgwalters:
Right now we default to DPS for composefs + systemd-boot.

In Fedora 43+, GRUB has the `bli` module and supports this, so it *can* be used there.

Make this configurable (mainly intended for base image builders) so that those with new enough GRUB can flip it on by default.

We had a hacky thing here that removed the auto-injected `root=` arg if we detected composefs + sdboot; that can now instead flip on this flag, and then we ensure we don't inject it at all.


jmarrero:
looks like the CI failures here are the merge conflict.

jmarrero (dismissed):

Johan-Liebert1 (approved):

PR #2129, opened 2026-04-07
blockdev: Handle ESP discovery on Intel VROC RAID devices

ckyrouac:
Intel VROC creates a topology where NVMe disks have MD RAID arrays as children, with the actual partitions (including the ESP) nested under the RAID array: nvme0n1 → md126 → md126p1 (ESP).

Previously find_partition_of_esp_optional() only searched direct children, so it would miss the ESP in this layout. Now it recurses into children that carry their own partition table, which handles the firmware RAID case naturally.

Add a lsblk JSON fixture representing a dual-NVMe firmware RAID1 layout and a test validating that ESP discovery works through the RAID layer.

Assisted-by: Claude Code (Opus 4)


crates/blockdev/src/blockdev.rs
```diff
@@ -189,12 +193,25 @@ impl Device {
...
-            Some("gpt") | None => Ok(self.find_partition_of_type(ESP)),
-            Some(other) => Err(anyhow!("Unsupported partition table type: {other}")),
+            Some("gpt") | None => self.find_partition_of_type(ESP),
+            Some(other) => return Err(anyhow!("Unsupported partition table type: {other}")),
+        };
+        if direct.is_some() {
+            return Ok(direct);
+        }
+        // Recurse into children that carry their own partition table, such as
+        // firmware RAID arrays (disk → md array → partitions).
```
cgwalters, line 206: Optional followup: I think this works and is correct. However...just thinking about this more, what feels architecturally cleaner is a model where we always walk up to all parents, then gather all children from there, then check all ESPs we found - deduplicating here by UUID - since in the VROC case we see the same ESP via multiple paths.

cgwalters (approved):

cgwalters:
The c9s one failed with

```
[920](https://github.com/bootc-dev/bootc/actions/runs/24085638686/job/70303367001?pr=2129#step:10:3921)
                        content: [TRACE bootupd::efi] Unmounted
                        content: Installation complete!
                        content: umount: /var/mnt/esp_check: target is busy.
                        content: Error:   × Dual ESP test failed: {msg: External command had a non-zero exit code,
                        content:   │ debug: NonZeroExitCode { exit_code: 32, span: Span { start: 99519, end:
                        content:   │ 99525 } }, raw: NonZeroExitCode { exit_code: 32, span: Span { start:
                        content:   │ 99519, end: 99525 } }, exit_code: 32}
                        content:      ╭─[/var/tmp/tmt/bootc-tmt-jp3dinqf-tmt-plans-integration-plan-41-multi-device-esp/tmt/plans/integration/plan-41-multi-device-esp/discover/default-0/tests/tmt/tests/booted/test-multi-device-esp.nu:267:9]
                        content:  266 │         rm -f $disk1 $disk2
                        content:  267 │         error make {msg: $"Dual ESP test failed: ($e)"}
                        content:      ·         ─────┬────
                        content:      ·              ╰── originates from here
                        content:  268 │     }
                        content:      ╰────
                        content: 
                        content: Connection to localhost

```

I restarted it

PR #2134, opened 2026-04-09
sd-boot: Don't generate random seed for generic images

### LINKED ISSUE #2133: Add option to `bootc install to-disk` to not write a random-seed when installing systemd-boot (for bcvk)
Current support for systemd-boot uses `bootctl` to install it into the ESP and this tool will write a random-seed file by default in the ESP. This notably happens when using `bcvk` to create the image.

When producing disk images for re-use, we should tell `bootctl` to not write that file using the option: https://www.freedesktop.org/software/systemd/man/latest/bootctl.html#--random-seed=yes%7Cno

Hopefully this will not be a issue most of the time as:
- when using `bootc-install` for the re-installation/take-over use case, writing the random-seed file is fine and actually a feature.
- when we'll have support in bootupd for shim+systemd then it will not write this file.
- images produced with osbuild/image-builder will likely use bootupd support and thus won't be impacted.

Johan-Liebert1:
For installations with `--generic-image` we don't really want a random-seed file to be generated in the ESP as these images would most likely be reused

Closes: https://github.com/bootc-dev/bootc/issues/2133

### DIFF
--- a/crates/lib/src/bootloader.rs
+++ b/crates/lib/src/bootloader.rs
@@ -227,7 +227,7 @@ pub(crate) fn install_via_bootupd(
 pub(crate) fn install_systemd_boot(
     device: &bootc_blockdev::Device,
     _rootfs: &Utf8Path,
-    _configopts: &crate::install::InstallConfigOpts,
+    configopts: &crate::install::InstallConfigOpts,
     _deployment_path: Option<&str>,
     autoenroll: Option<SecurebootKeys>,
 ) -> Result<()> {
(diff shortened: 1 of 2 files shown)


cgwalters (dismissed):

Johan-Liebert1:
Restarted some failed checks

Johan-Liebert1:
Added this commit to hopefully fix C9S CI

Add `--allowerasing` for dnf to replace conflicting pkgs

To fix CI for C9S failing with

```
installed
  - cannot install both policycoreutils-3.6-2.1.el9.x86_64 from baseos and policycoreutils-3.6-7.el9.x86_64 from @System
  - cannot install both policycoreutils-3.6-3.el9.x86_64 from baseos and policycoreutils-3.6-7.el9.x86_64 from @System
  - cannot install both policycoreutils-3.6-4.el9.x86_64 from baseos and policycoreutils-3.6-7.el9.x86_64 from @System
  - cannot install both policycoreutils-3.6-5.el9.x86_64 from baseos and policycoreutils-3.6-7.el9.x86_64 from @System
  - cannot install both policycoreutils-3.6-6.el9.x86_64 from baseos and policycoreutils-3.6-7.el9.x86_64 from @System
```

cgwalters:
Yeah we need https://gitlab.com/fedora/bootc/base-images/-/merge_requests/317

cgwalters (approved):

PR #2135, opened 2026-04-09
system-reinstall-bootc: fallback image from /usr/lib/os-release

### LINKED ISSUE #1300: Add support for default image from /usr/lib/os-release or so
With `system-reinstall-bootc` one thing that might be nice for operating systems to support is basically having the user not provide a container image could default to a generic one (or one specific to the image type).

We've talked in the past for example about having a pre-built "cloud" bootc image and I can imagine that firing up e.g. c9s in AWS and just typing `system-reinstall-bootc` would default to that image. In this flow we also wouldn't do any of the SSH key injection because we can assume the target image has cloud-init. The goal here is to make this as close to "flip this system over to bootc" as possible.

I'm thinking we could have something like
`BOOTC_IMAGE=quay.io/centos-bootc/centos-bootc-cloud:stream9` in `/usr/lib/os-release` or so.

(Tangential but if we improve https://containers.github.io/bootc/booting-local-builds.html then I think we could even encourage a flow where the user tests things out that way, i.e. they just boot a cloud image and do s-r-b, then test some local builds, then move to git-ops etc.)

hone:
Add a fallback option when arguments are specified for an image. It works in this order:
  1. BOOTC_REINSTALL_CONFIG env var to a YAML file
  2. `--image` flag
  3. reads /etc/os-release for BOOTC_IMAGE
  4. reads /usr/lib/os-release for BOOTC_IMAGE

Fixes: #1300

### DIFF
--- a/crates/system-reinstall-bootc/src/main.rs
+++ b/crates/system-reinstall-bootc/src/main.rs
@@ -1,7 +1,7 @@
 //! The main entrypoint for the bootc system reinstallation CLI
 
 use anyhow::{Context, Result, ensure};
-use bootc_utils::CommandRunExt;
+use bootc_utils::{CommandRunExt, ResultExt};
 use clap::Parser;
 use fn_error_context::context;
 use rustix::process::getuid;
@@ -10,11 +10,14 @@ use std::time::Duration;
 mod btrfs;
 mod config;
 mod lvm;
+mod os_release;
 mod podman;
 mod prompt;
 pub(crate) mod users;
 
 const ROOT_KEY_MOUNT_POINT: &str = "/bootc_authorized_ssh_keys/root";
+const ETC_OS_RELEASE: &str = "/etc/os-release";
+const USR_LIB_OS_RELEASE: &str = "/usr/lib/os-release";
 
 /// Reinstall the system using the provided bootc container.
 ///
@@ -24,17 +27,52 @@ const ROOT_KEY_MOUNT_POINT: &str = "/bootc_authorized_ssh_keys/root";
 /// If the environment variable BOOTC_REINSTALL_CONFIG is set, it must be a YAML
 /// file with a single member `bootc_image` that specifies the image to install.
 /// This will take precedence over the CLI.
-#[derive(clap::Parser)]
 pub(crate) struct ReinstallOpts {
     /// The bootc image to install
     pub(crate) image: String,
+    // Note if we ever add any other options here,
+    pub(crate) composefs_backend: bool,
+}
+
+#[derive(clap::Parser)]
+pub(crate) struct ReinstallOptsArgs {
+    /// The bootc image to install
+    pub(crate) image: Option<String>,
     // Note if we ever add any other options here,
     #[arg(long)]
     pub(crate) composefs_backend: bool,
 }
 
+impl ReinstallOptsArgs {
+    pub(crate) fn build(self) -> Result<ReinstallOpts> {
+        let image = if let Some(image) = self.image {
+            image
+        } else {
+            [ETC_OS_RELEASE, USR_LIB_OS_RELEASE]
+                .iter()
+                .find_map(|path| {
+                    os_release::get_bootc_image_from_file(path)
+                        .log_err_default()
+                        .filter(|s| !s.is_empty())
+                })
+                .ok_or_else(|| {
+                    anyhow::anyhow!(
+                        "No image provided. Specify an image or set BOOTC_IMAGE in os-release."
+                    )
+                })?
+        };
+
+        Ok(ReinstallOpts {
+            image,
+            composefs_backend: self.composefs_backend,
+        })
+    }
+}
+
 #[context("run")]
 fn run() -> Result<()> {
+    let args = ReinstallOptsArgs::parse();
+
     // We historically supported an environment variable providing a config to override the image, so
     // keep supporting that. I'm considering deprecating that though.
     let opts = if let Some(config) = config::ReinstallConfig::load().context("loading config")? {
@@ -43,8 +81,8 @@ fn run() -> Result<()> {
             composefs_backend: config.composefs_backend,
         }
     } else {
-        // Otherwise an image is required.
-        ReinstallOpts::parse()
+        // Otherwise an image is specified via the CLI or fallback to the os-release
+        args.build()?
     };
 
     bootc_utils::initialize_tracing();
--- a/crates/system-reinstall-bootc/src/os_release.rs
+++ b/crates/system-reinstall-bootc/src/os_release.rs
@@ -1,0 +1,108 @@
+use std::fs::File;
+use std::io::{BufRead, BufReader};
+use std::path::Path;
+
+use anyhow::{Context, Result};
+
+/// Searches for the BOOTC_IMAGE key in a given os-release file.
+/// Follows standard os-release(5) quoting rules.
+fn parse_bootc_image_from_reader<R: BufRead>(reader: R) -> Result<Option<String>> {
+    let mut last_found = None;
+
+    for line in reader.lines() {
+        let line = line?;
+        let line = line.trim();
+
+        if line.is_empty() || line.starts_with('#') {
+            continue;
+        }
+
+        let Some((key, value)) = line.split_once('=') else {
+            continue;
+        };
+
+        if key.trim() == "BOOTC_IMAGE" {
+            if let Some(unquoted) = shlex::split(value).and_then(|mut values| values.pop()) {
+                last_found = Some(unquoted);
+            }
+        }
+    }
+
+    Ok(last_found)
+}
+
+/// Reads the provided os-release file and returns the BOOTC_IMAGE value if found.
+pub(crate) fn get_bootc_image_from_file<P: AsRef<Path>>(path: P) -> Result<Option<String>> {
+    let file = File::open(path.as_ref()).with_context(|| format!("Opening {:?}", path.as_ref()))?;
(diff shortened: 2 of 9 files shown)


tmt/plans/integration.fmf
```diff
@@ -25,12 +25,18 @@ prepare:
...
       - mkdir -p bootc && cp /var/share/test-artifacts/*.src.rpm bootc
       - cd bootc && rpm2cpio *.src.rpm | cpio -idmv && rm -f *-vendor.tar.zstd && zstd -d *.tar.zstd && tar -xvf *.tar -C . --strip-components=1 && ls -al
       - pwd && ls -al && cd bootc/hack && ./provision-packit.sh
     when: running_env != image_mode
+  - how: shell
+    order: 98
+    script:
+      - echo 'BOOTC_IMAGE=localhost/bootc' | tee -a /usr/lib/os-release
+      - pwd && ls -al && cd bootc/hack && ./provision-packit.sh
+    when: running_env != image_mode
```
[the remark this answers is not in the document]
hone: I thought it was easier to just build the image twice than try to do a cleanup step for testing the BOOTC_IMAGE path on reinstall.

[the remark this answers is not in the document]
hone: duplicate of above, closing this.

crates/system-reinstall-bootc/src/os_release.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    for line in reader.lines() {
+        let line = line?;
+        let line = line.trim();
+
+        if line.is_empty() || line.starts_with('#') {
+            continue;
+        }
+
+        if let Some((key, value)) = line.split_once('=') {
```
cgwalters: Style nit, also could be let/else here and avoid further nesting

crates/system-reinstall-bootc/src/os_release.rs
```diff
@@ -0,0 +1,118 @@
...
+            if key.trim() == "BOOTC_IMAGE" {
+                let value = value.trim();
+
+                if value.starts_with('"') && value.ends_with('"') && value.len() >= 2 {
+                    let unquoted = &value[1..value.len() - 1];
+                    let processed = unquoted
+                        .replace(r#"\""#, "\"")
+                        .replace(r#"\\"#, "\\")
+                        .replace(r#"\$"#, "$")
+                        .replace(r#"\`"#, "`");
```
cgwalters: I think we can use shlex to parse this, it's already in the depchain

crates/system-reinstall-bootc/src/main.rs
```diff
@@ -24,17 +27,53 @@ const ROOT_KEY_MOUNT_POINT: &str = "/bootc_authorized_ssh_keys/root";
...
+impl ReinstallOptsArgs {
+    pub(crate) fn build(self) -> Result<ReinstallOpts> {
+        let image = if let Some(image) = self.image {
+            image
+        } else {
+            [ETC_OS_RELEASE, USR_LIB_OS_RELEASE]
+                .iter()
+                .find_map(|path| {
+                    os_release::get_bootc_image_from_file(path)
+                        .ok()
```
cgwalters: Generally avoid "error swallowing" like this. There's two different cases:

- The file doens't exist - this is normal, and we have https://docs.rs/cap-std-ext/latest/cap_std_ext/dirext/trait.CapStdExtDirExt.html#tymethod.open_optional for this reason
- All other errors (permission denied, failure to parse the file, etc.). I think for this we'd want at least to use https://github.com/bootc-dev/bootc/blob/b1fe5d697855189f64ba87cff5cfb058f91792f6/crates/utils/src/result_ext.rs#L4 )

crates/system-reinstall-bootc/src/podman.rs
```diff
@@ -24,10 +24,29 @@ pub(crate) fn bootc_has_clean(image: &str) -> Result<bool> {
     Ok(stdout_str.contains("--cleanup"))
 }
 
+#[context("image_has_cloud_init")]
+pub(crate) fn image_has_cloud_init(image: &str) -> Result<bool> {
```
cgwalters: In https://github.com/bootc-dev/bcvk/pull/245 we're proposing to standardize labels on images that have Ignition, it could make sense to do the same for cloud-init.

Separate effort from this PR.

crates/system-reinstall-bootc/src/main.rs
```diff
@@ -110,20 +110,36 @@ fn run() -> Result<()> {
...
-        .path()
-        .to_str()
-        .ok_or_else(|| anyhow::anyhow!("unable to create authorized_key temp file"))?;
+    let mut _ssh_key_tempfile = None;
+    let mut ssh_key_file_path = None;
+
+    if podman::image_has_cloud_init(&opts.image)? {
+        let host_root_keys = std::path::Path::new("/root/.ssh/authorized_keys");
+        if host_root_keys.exists() {
+            println!("Detected cloud-init and host keys. Inheriting keys automatically.");
```
cgwalters: I think this change may make sense but it looks completely unrelated to a fallback image approach right?

This seems like a distinct PR?

cgwalters (changes requested):
Thanks so much for working on this!

crates/system-reinstall-bootc/src/main.rs
hone: From #1300:
> In this flow we also wouldn't do any of the SSH key injection because we can assume the target image has cloud-init.

This was my attempt to tackle this from the issue. I wasn't sure if that was required for this, but happy to split this into a separate PR if that's easier.

crates/system-reinstall-bootc/src/podman.rs
hone: Is there already an issue for that?

hone:
How concerned should I be about these packit failures?

jeckersb:
> How concerned should I be about these packit failures?

The s390x build ones at least can be ignored for now, there's a known issue with those.

jmarrero:
Thank you again for this. This needs another run of: `cargo fmt` which CI is complaining about and commits should be squashed. We tend to keep commits functionally grouped, if it's addressing reviews just add it to the main commit that touched that file.

hone:
> Thank you again for this. This needs another run of: `cargo fmt` which CI is complaining about and commits should be squashed. We tend to keep commits functionally grouped, if it's addressing reviews just add it to the main commit that touched that file.

@jmarrero happy to do that. Before I squash, should I split out the ssh authorized keys pieces or keep it in this PR?

jmarrero:
> > Thank you again for this. This needs another run of: `cargo fmt` which CI is complaining about and commits should be squashed. We tend to keep commits functionally grouped, if it's addressing reviews just add it to the main commit that touched that file.
> 
> @jmarrero happy to do that. Before I squash, should I split out the ssh authorized keys pieces or keep it in this PR?

Yes, I would move the ssh stuff to it's own PR where we will have some additional discussions on it. The background here is that we are not 100% sure yet if `we can assume the target image has cloud-init` and we might want to support checking for ignition or/and cloud init on the distributions we maintain. There is some uncertainty of assuming this still. When the issue was created I think were more certain about cloud-init vs ignition.

crates/system-reinstall-bootc/src/main.rs
hone: I've moved this to #2163.

hone:
> Yes, I would move the ssh stuff to it's own PR where we will have some additional discussions on it. The background here is that we are not 100% sure yet if `we can assume the target image has cloud-init` and we might want to support checking for ignition or/and cloud init on the distributions we maintain. There is some uncertainty of assuming this still. When the issue was created I think were more certain about cloud-init vs ignition.

Ok, I've squased the PR. I didn't squash 8a9b9d495058d4346e4503a3a4e61d2ae9830243 b/c it's not totally related to my PR, but something I needed to do to get CI to pass.

I've also moved the ssh `authorized_keys` + `cloud-init` commits into #2163 .

jmarrero (dismissed):
lgtm

cgwalters:
I think this just needs a rebase

hone:
@cgwalters rebased.

cgwalters (approved):
Thanks!

crates/system-reinstall-bootc/src/os_release.rs
```diff
@@ -0,0 +1,108 @@
+use std::fs::File;
+use std::io::{BufRead, BufReader};
+use std::path::Path;
+
+use anyhow::{Context, Result};
+
+/// Searches for the BOOTC_IMAGE key in a given os-release file.
+/// Follows standard os-release(5) quoting rules.
+fn parse_bootc_image_from_reader<R: BufRead>(reader: R) -> Result<Option<String>> {
```
Johan-Liebert1, line 9: We have a similar function in composefs-rs 

https://github.com/composefs/composefs-rs/blob/main/crates/composefs-boot/src/os_release.rs#L59

Johan-Liebert1 (commented):
Overall looks good. One comment though

crates/system-reinstall-bootc/src/os_release.rs
hone: @Johan-Liebert1 I did notice that, but `composefs` isn't a dependency on the `system-reinstall-bootc` crate. I wasn't sure if we wanted to add that in as a dependency for this or not.

Johan-Liebert1: I think it's okay to have it as a dependency. It will just be one crate `composefs_boot` if I'm not mistaken

PR #2145, opened 2026-04-15
provision-derived: Don't hardcode bootupd version

jeckersb:
This got bumped recently so the old version is no longer available.

### DIFF
--- a/hack/provision-derived.sh
+++ b/hack/provision-derived.sh
@@ -107,7 +107,23 @@ repo_gpgcheck=0
 enabled=1
 enabled_metadata=1
 REPOEOF
-dnf -y install bootupd-0.2.32.43.g38208d3
+
+# This unfortunately has "older" versions with higher NEVRA:
+#
+# # dnf --disablerepo=* --enablerepo=copr:copr.fedorainfracloud.org:group_CoreOS:continuous repoquery bootupd 2> /dev/null
+# bootupd-0:0.2.32.45.gb483a63-1.fc45.x86_64
+# bootupd-0:202501200321.0.2.25.65.ge296f82-1.fc42.src
+# bootupd-0:202501200321.0.2.25.65.ge296f82-1.fc42.x86_64
+# bootupd-0:202501210627.0.2.25.67.gefe41b6-1.fc42.src
+#
+# So we need to be more selective, but also be dynamic to grab newer
+# versions
+#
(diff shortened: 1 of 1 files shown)


hack/provision-derived.sh
```diff
@@ -107,7 +107,19 @@ repo_gpgcheck=0
...
+# # dnf --disablerepo=* --enablerepo=copr:copr.fedorainfracloud.org:group_CoreOS:continuous repoquery bootupd 2> /dev/null
+# bootupd-0:0.2.32.45.gb483a63-1.fc45.x86_64
+# bootupd-0:202501200321.0.2.25.65.ge296f82-1.fc42.src
+# bootupd-0:202501200321.0.2.25.65.ge296f82-1.fc42.x86_64
+# bootupd-0:202501210627.0.2.25.67.gefe41b6-1.fc42.src
+#
+# So we need to be more selective, but also be dynamic to grab newer
+# versions
+bootupd_nevra=$(dnf --disablerepo=* --enablerepo=copr:copr.fedorainfracloud.org:group_CoreOS:continuous repoquery --latest-limit 1 --arch "$(uname -m)" "bootupd-0.2.*")
+dnf -y install ${bootupd_nevra}
```
cgwalters: I think `dnf -y --disablerepo=* --enablerepo=<copr> distro-sync bootupd` would work

jeckersb: Doesn't seem to work for me:

```
[root@8fbeb7e8531f /]# dnf -y --disablerepo=* --enablerepo=copr:copr.fedorainfracloud.org:group_CoreOS:continuous distro-sync bootupd
Updating and loading repositories:
Repositories loaded.
Failed to resolve the transaction:
Packages for argument 'bootupd' available, but not installed.
```

Probably because of those malformed `fc42` packages that I mentioned in the comment in the diff?

jeckersb:
Oh and it looks like the subscription-manager plugin writes to stdout so that's breaking it too.

jeckersb:
Well on a somewhat-positive-since-it's-consistent note, we're seeing the same error on rawhide here that we're seeing over in the [fedora dist-git job](https://src.fedoraproject.org/rpms/bootc/pull-request/93#comment-319302), so that'll help with tracking that down.

cgwalters (approved):

PR #2147, opened 2026-04-16
Use --nosignature when installing dev rpms

jeckersb:
Newer fedora (rawhide at least) checks signatures by default and will
fail to install these since they're not signed


cgwalters (approved):

jeckersb:
:surfer: rebased to pick up #2145

PR #2153, opened 2026-04-21
install: Pass explicit filesystem type to mount after mkfs

### LINKED ISSUE #2148: xfs mount failure on rawhide
Seeing this in a few places, notably over in https://src.fedoraproject.org/rpms/bootc/pull-request/93 but also in our CI here like https://github.com/bootc-dev/bootc/pull/2145/checks?check_run_id=71548302273

```
Creating root filesystem (xfs) on device /dev/loop0p3 (size=10.2 GB)
> mkfs.xfs -m uuid=f36f7d0d-28b6-4271-8674-d1d76e54229f -L root /dev/loop0p3
mount: /run/bootc/mounts/rootfs: wrong fs type, bad option, bad superblock on /dev/loop0p3, missing codepage or helper program, or other error.
       dmesg(1) may have more information after failed mount system call.
error: Installing to disk: Creating rootfs: Failed to run command: Command {
    program: "mount",
    args: [
        "mount",
        "/dev/loop0p3",
        "/run/bootc/mounts/rootfs",
    ],
    create_pidfd: false,
}
```

jeckersb:
When mounting a freshly-created filesystem during install, pass
-t <fstype> to mount rather than relying on auto-detection.

util-linux 2.42 introduced an optimization (commit 8bdc2546d) where
libmount reads device properties from the udev database instead of
probing the device superblock with libblkid. When mkfs runs inside a
container, the udev database is stale for the filesystem type because
mkfs does not generate a kernel uevent — udev only re-probes devices
on kernel-generated events (such as partition table changes from
sfdisk), not on arbitrary block device writes. As a result, the udev
database has partition tags (PARTUUID, PARTLABEL) from sfdisk but
lacks ID_FS_TYPE. libmount's read_from_udev() returns success after
finding those partition tags, so the libblkid direct-probe fallback
is never called. Without a filesystem type, mount falls back to
iterating /etc/filesystems and /proc/filesystems, which fails for
filesystem modules (like xfs with CONFIG_XFS_FS=m) that are not yet
loaded.

Since bootc already knows the filesystem type (it just ran mkfs),
passing -t avoids the auto-detection path entirely.

Closes: #2148

### DIFF
--- a/crates/lib/src/install/baseline.rs
+++ b/crates/lib/src/install/baseline.rs
@@ -484,7 +484,8 @@ pub(crate) fn install_create_rootfs(
         }
     }
 
-    bootc_mount::mount(&rootdev_path, &physical_root_path)?;
+    let fstype = &root_filesystem.to_string();
+    bootc_mount::mount_typed(&rootdev_path, fstype, &physical_root_path)?;
     let target_rootfs = Dir::open_ambient_dir(&physical_root_path, cap_std::ambient_authority())?;
     crate::lsm::ensure_dir_labeled(&target_rootfs, "", Some("/".into()), 0o755.into(), sepolicy)?;
     let physical_root = Dir::open_ambient_dir(&physical_root_path, cap_std::ambient_authority())?;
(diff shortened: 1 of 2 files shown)


ckyrouac (approved):

travier:
Direct link to util-linux commit: https://github.com/util-linux/util-linux/commit/8bdc2546d

cgwalters (approved):

PR #2155, opened 2026-04-21
tests: Skip BLI detection test when discoverable-partitions config is absent

jmarrero:
The 052-test-bli-detection test asserts that Fedora 43+ systems boot
without a root= kernel argument (using DPS auto-discovery instead).
However, this only works when the install was performed via
bootc install to-disk, which creates partitions with DPS type GUIDs.

In Packit/gating CI, the system is installed via
bootc install to-existing-root on a Testing Farm VM whose partitions
were not created with DPS type GUIDs. The test was unconditionally
failing in this environment.

Fix this by checking for the BCVK_EXPORT environment variable, which
is set by xtask when running tests via bcvk (image-mode). When not
in image-mode, the test skips since DPS root discovery cannot work
on partitions that lack the proper type GUIDs.

Fixes: https://github.com/bootc-dev/bootc/pull/2123
Assisted-by: OpenCode (Claude Opus 4.6)
Signed-off-by: Joseph Marrero Corchado <jmarrero@redhat.com>

### DIFF
--- a/tmt/tests/booted/readonly/052-test-bli-detection.nu
+++ b/tmt/tests/booted/readonly/052-test-bli-detection.nu
@@ -29,6 +29,16 @@ if not ($os_id == "fedora" and $version_id >= 43) {
     exit 0
 }
 
+# DPS root discovery only works when bootc install to-disk created the
+# partitions with DPS type GUIDs.  In Packit/gating CI the system was
+# installed via to-existing-root on pre-existing partitions, so skip.
+# BCVK_EXPORT=1 is set by xtask when running via bcvk (image-mode).
+if ($env.BCVK_EXPORT? | default "" | is-empty) {
+    print "# skip: not running in image-mode (BCVK_EXPORT not set)"
+    tap ok
+    exit 0
+}
+
 print $"Running on ($os_id) ($version_id), checking DPS root discovery"
 
 let cmdline = (open /proc/cmdline)


jeckersb:
well this might eventually pass if we can stop getting gateway timeouts...

/me re-runs yet again

jmarrero:
idk... not looking good yet:

```
Running on fedora 45, checking DPS root discovery
Error:   × Fedora 43+ should use DPS auto-discovery (no root= in cmdline)
    ╭─[/var/ARTIFACTS/work-plan-01-readonlypgkxqhp6/tmt/plans/integration/plan-01-readonly/discover/default-0/tests/tmt/tests/booted/readonly/052-test-bli-detection.nu:37:8]
 36 │ 
 37 │ assert (not $has_root_karg) "Fedora 43+ should use DPS auto-discovery (no root= in cmdline)"
    ·        ──────────┬─────────
    ·                  ╰── It is not true.
 38 │ 
    ╰────
```

jmarrero:
@jeckersb this is starting to pass now, since this pass locally and in github ci and not in tmt, I am just skipping it on tmt.

jeckersb:
The tmt rawhide failures should be fixed by https://github.com/bootc-dev/bootc/pull/2153 but fortunately those rawhide jobs are non-voting so ideally we can get this PR merged and then I'll go rebase that one to pull in this fix and then we should be good?

Only other notable thing here is that the `test-coreos` test is failing but I'm not sure if that was already broken before all of this.

cgwalters:
Let's clearly mark the commit as

Fixes: https://github.com/bootc-dev/bootc/pull/2123

or so.

This one is definitely messy, with `to-filesystem` we could in theory also start converting partitions over to DPS, but it would make it more difficult to support rollbacks to prior state in the future.

Probably the simplest thing here is to pass through an env var into the tests. We already do `--context=running_env=image_mode` but I'm not sure if that makes it into the inner test environment. It looks like we unconditionally set `BCVK_EXPORT=1` right now, and we could check that in the test.

jeckersb (approved):

PR #2161, opened 2026-04-23
loader-entries: Add composefs support for set-options-for-source

### LINKED ISSUE #899: /etc/bootc/kargs.d
It would be useful to have `/etc/bootc/kargs.d` in addition to `/usr/lib/bootc/kargs.d` get merged in.

Example use case: If you have a mixed hardware cluster, you often will have a different set of console args related to the serial console, but otherwise all the nodes are the same. You could make custom images for each hardware class, but if there was an /etc/bootc/kargs.d, you could just drop that arg in there and the same image would work everywhere.

jmarrero:
The ostree path for set-options-for-source (merged in fb8c6682) stages a new deployment via ostree APIs. On composefs-booted systems this approach cannot work because the status detection logic treats any entry whose verity matches the booted deployment as "booted" rather than "staged", making same-image kargs-only changes invisible to finalization.

Instead, the composefs path directly modifies BLS entry files on /boot. This is architecturally correct because bootc already manages BLS entries directly on composefs systems, and the BLSConfig.extra HashMap already preserves x-options-source-* extension keys through parse/write roundtrips. No GVariant serialization, ostree version gating, or finalization service is needed.

Changes:
- Add composefs implementation in bootc_composefs/loader_entries.rs that reads the booted BLS entry, computes the merged options via the shared parsing functions, and writes the updated entry back atomically. When staged entries exist (e.g. from bootc upgrade), they are also updated so finalization does not overwrite the kargs changes.
- Refactor CLI dispatch to use match storage.kind() for routing between the ostree and composefs backends.
- Make SourceName, OPTIONS_SOURCE_KEY_PREFIX, extract_source_options_from_bls, and compute_merged_options pub(crate) for reuse by the composefs module.
- Bail with a clear error on UKI boot type since kargs are embedded in the PE binary.
- Add composefs-specific TMT test (test-43) with 4 phases / 3 reboots covering input validation, kargs persistence, source replacement, multiple sources, idempotency, and source removal.
- Add missing fixme_skip_if_composefs metadata to the existing ostree test file header so the TMT generator preserves the skip flag.
- 8 new unit tests for the composefs path, all existing tests unchanged.

Closes: https://github.com/bootc-dev/bootc/issues/899
Assisted-by: OpenCode (Claude Opus 4.6)

### DIFF
--- a/crates/lib/src/bootc_composefs/loader_entries.rs
+++ b/crates/lib/src/bootc_composefs/loader_entries.rs
@@ -1,0 +1,496 @@
+//! # Composefs backend for source-tracked kernel arguments
+//!
+//! This module implements `set-options-for-source` for composefs-booted systems.
+//! Unlike the ostree path (which stages a new deployment via ostree APIs), the
+//! composefs path directly modifies BLS entry files on /boot. This is both
+//! simpler and more efficient:
+//!
+//! - `BLSConfig.extra` already preserves `x-options-source-*` keys through
+//!   parse/write roundtrips
+//! - No GVariant serialization, no ostree version gating, no finalization needed
+//! - Kargs take effect on next boot without requiring a finalization service
+//!
+//! When a staged deployment exists (e.g. from `bootc upgrade`), the staged BLS
+//! entries are also updated so finalization doesn't overwrite the kargs changes.
+//!
+//! See <https://github.com/bootc-dev/bootc/issues/899>
+
+use anyhow::{Context, Result};
+use bootc_kernel_cmdline::utf8::CmdlineOwned;
+use cap_std_ext::cap_std::fs::Dir;
+use cap_std_ext::dirext::CapStdExtDirExt;
+use fn_error_context::context;
+use std::collections::BTreeMap;
+
+use super::state::get_booted_bls;
+use super::status::get_sorted_staged_type1_boot_entries;
+use crate::composefs_consts::TYPE1_ENT_PATH_STAGED;
+use crate::loader_entries::{OPTIONS_SOURCE_KEY_PREFIX, SourceName, compute_merged_options};
+use crate::parsers::bls_config::{BLSConfig, BLSConfigType, parse_bls_config};
+use crate::store::{BootedComposefs, Storage};
+
+/// Extract source options from a parsed `BLSConfig`'s `extra` HashMap.
+fn extract_source_options_from_extra(bls: &BLSConfig) -> BTreeMap<String, CmdlineOwned> {
+    let mut sources = BTreeMap::new();
+    for (key, value) in &bls.extra {
+        if let Some(name) = key.strip_prefix(OPTIONS_SOURCE_KEY_PREFIX) {
+            if !name.is_empty() && !value.is_empty() {
+                sources.insert(name.to_string(), CmdlineOwned::from(value.clone()));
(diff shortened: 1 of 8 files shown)


cgwalters:
I don't like the "Changes" section in commit messages like this. I did https://github.com/bootc-dev/infra/pull/171 to try to tweak that.

crates/lib/src/bootc_composefs/loader_entries.rs
```diff
@@ -0,0 +1,496 @@
...
+    new_options: Option<&str>,
+    source_options: &BTreeMap<String, CmdlineOwned>,
+) -> Result<()> {
+    // Update the options line
+    match &mut bls.cfg_type {
+        BLSConfigType::NonEFI { options, .. } => {
+            *options = Some(merged_options.clone());
+        }
+        _ => anyhow::bail!("BLS entry is not a NonEFI (BLS) type"),
+    }
```
cgwalters, line 72: Some overlap with `get_options_str` above

tmt/tests/booted/test-composefs-loader-entries-source.nu
```diff
@@ -0,0 +1,209 @@
+# number: 43
+# tmt:
+#   summary: Test bootc loader-entries set-options-for-source on composefs
```
cgwalters, line 3: But can't we have just one test that works on both backends?

crates/lib/src/bootc_composefs/loader_entries.rs
```diff
@@ -0,0 +1,496 @@
+//! # Composefs backend for source-tracked kernel arguments
+//!
+//! This module implements `set-options-for-source` for composefs-booted systems.
+//! Unlike the ostree path (which stages a new deployment via ostree APIs), the
```
cgwalters, line 4: That's not true, but we were missing docs for it. I took a stab at this in https://github.com/bootc-dev/bootc/pull/2168

cgwalters (changes requested):

crates/lib/src/bootc_composefs/loader_entries.rs
```diff
@@ -0,0 +1,545 @@
...
+    }
+
+    Ok(())
+}
+
+/// Read, update, and write back a BLS entry file in a directory.
+///
+/// Finds the entry matching the given version, parses it, applies the
+/// source kargs change, and writes it back atomically.
+fn update_bls_entry_in_dir(
```
Johan-Liebert1: We already have a function to achieve this `get_sorted_type1_boot_entries`, and I think this can be refactored with the rollback code

crates/lib/src/bootc_composefs/loader_entries.rs
```diff
@@ -0,0 +1,545 @@
+//! # Composefs backend for source-tracked kernel arguments
+//!
+//! This module implements `set-options-for-source` for composefs-booted systems.
+//! Unlike the ostree path (which stages a new deployment via ostree APIs), the
+//! composefs path directly modifies BLS entry files on /boot. This is both
```
Johan-Liebert1: I thought about this but it's not compatible with how ostree does things. I think users will expect a similar behaviour to ostree, i.e. expecting a rollback deployment w/o the new kargs.

PR #2164, opened 2026-04-24
install: Skip discoverable-partitions config for FCOS builds

### LINKED ISSUE #2160: `test-coreos` job is failing
Tentative diagnosis staring at the console during boot is that it's unable to find the `boot` disk by label and then hanging.  Will require more investigation.

cgwalters:
Currently Fedora CoreOS derivatives don't yet work with systemd-gpt-auto-generator, see https://github.com/coreos/fedora-coreos-tracker/issues/2137

It was briefly in an image which let our CI change to depend on it pass.

Fix this by passing the existing `SKIP_CONFIGS` variable into the coreos image build.

Closes: https://github.com/bootc-dev/bootc/issues/2160

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -165,6 +165,7 @@ RUN --network=none --mount=type=tmpfs,target=/run --mount=type=tmpfs,target=/tmp
 FROM base as base-penultimate
 ARG variant
 ARG bootloader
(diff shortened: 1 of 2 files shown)


cgwalters:
OK this is related, but not the real bug.

PR #2165, opened 2026-04-24
ci: Disable test-coreos

### LINKED ISSUE #2160: `test-coreos` job is failing

cgwalters:
This job recently fell over, and surprisingly it wasn't the DPS changes, it was https://github.com/ostreedev/ostree/pull/3571 which stopped mounting `/boot` in the initramfs.

Our flow here does an install without a separate `/boot` partition (which is not a normal FCOS path) and the combo of these things breaks.

The real fix here will be cutting over this job to use coreos-assembler, but it's a big lift.

Just disabling this job for now to unblock our CI - in practice we do have coverage testing bootc with FCOS, just later.

Closes: https://github.com/bootc-dev/bootc/issues/2160

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -311,6 +311,10 @@ jobs:
   # when run in the same job as test-integration).
   # Uses fedora-43 as it's the current stable Fedora release matching CoreOS.
   test-coreos:
+    # https://github.com/ostreedev/ostree/pull/3571 broke this because /boot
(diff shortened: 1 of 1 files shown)


jeckersb (approved):

jeckersb:
Good catch!

PR #2166, opened 2026-04-24
ci: Enable merge_group trigger for merge queue support

cgwalters:
Add merge_group to the CI workflow triggers so that GitHub merge queues can be used. This is a prerequisite for enabling merge queues on the repository.

Also drop the PR number from artifact names — artifacts are already scoped to a workflow run, so the matrix parameters alone provide sufficient uniqueness.

Ref: https://github.com/bootc-dev/infra/issues/143

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,7 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
+  merge_group:
   workflow_dispatch: {}
 
 env:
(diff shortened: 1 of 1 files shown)


jeckersb (approved):

jmarrero:
Maybe we should hold off for merge queue until the dust settles on the github regression that seems pretty bad. https://www.githubstatus.com/incidents/zsg1lk7w13cf

cgwalters:
> Maybe we should hold off for merge queue until the dust settles on the github regression that seems pretty bad. 

Yeah that came up in https://github.com/bootc-dev/infra/issues/143#issuecomment-4314586820

But...they say they fixed it too. I dunno. I'd still lean towards trying it. 

Among other things, in the situation we're in now I think it means we don't need to ensure each PR gets rebased on main to pick up the test-coreos job change, which alone is quite nice.

PR #2169, opened 2026-04-25
blockdev: Fall back to blkid -p when udev is unavailable

cgwalters:
See https://github.com/osbuild/osbuild/pull/2428

When running inside a container or sandbox without /run/udev bind-mounted (e.g. osbuild's bwrap), lsblk returns null for partition metadata fields like parttype (partition type GUID) and pttype (partition table type), because these come from the udev database.

This causes partition discovery to fail -- most visibly on ppc64le where the PReP partition can't be found by GUID, but also affecting ESP discovery on all architectures. On x86_64 UEFI, bootupd silently falls back to mounted-ESP detection, masking the problem; on ppc64le there is no fallback and the install fails hard.

Add a blkid -p fallback in backfill_missing() that probes partition metadata directly from the disk when udev is absent. The udev check tests for /run/udev/data (the actual database directory) rather than /run/udev, because bootc's ensure_mirrored_host_mount() creates an empty /run/udev directory on the bwrap tmpfs that is_same_as_host() then considers 'already mounted'.

Tested end-to-end via the BIB tmt test (plan-33-bib-build) with a patched bootupd that removes the mounted-ESP fallback entirely.

This follows the same approach coreos-installer uses (PRs #1511, sfdisk-based workaround originally added in PR #688.


Johan-Liebert1 (approved):
lgtm

PR #2171, opened 2026-04-27
composefs/uki: Remove stripping `.staged` suffix on finalize

Johan-Liebert1:
We were stripping `.staged` suffix from UKI/UKI Addons whenever unstaging the bootloader entries for the currently staged composefs deployment. Earlier we needed this as we were putting all EFI binaries in `ESP/EFI/Linux` all of which (staged or unstaged) would be picked up by the bootloader which is not what we want.

We now put all EFI binaries in `ESP/EFI/Linux/bootc` and use BLS .conf files to point to the binaries, which makes the `.staged` suffix for EFI binaries irrelevant.

### DIFF
--- a/crates/lib/src/bootc_composefs/finalize.rs
+++ b/crates/lib/src/bootc_composefs/finalize.rs
@@ -14,8 +14,7 @@ use cap_std_ext::dirext::CapStdExtDirExt;
 use cfsctl::composefs;
 use composefs::generic_tree::{Directory, Stat};
 use etc_merge::{compute_diff, merge, print_diff, traverse_etc};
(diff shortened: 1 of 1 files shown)


Johan-Liebert1:
ostree test on f43 failed with a flake

```
content: error: Installing to filesystem: Creating ostree deployment: Pulling: Importing: Unencapsulating base: failed to invoke method GetBlob: fetching blob: received unexpected HTTP status: 502 Bad Gateway
```
reran

cgwalters:
One random thing I was thinking we probably don't need to write staged bootloader entries until we start finalization

cgwalters (approved):

PR #2173, opened 2026-04-27
ci: Run full tests only on merge queue, add ci/tier-1 and ci/merge labels

cgwalters:
Now that we have a merge queue, change things so we only run cheaper jobs on PRs by default. It's possible to opt in to heavier CI via labels.

Part of https://github.com/bootc-dev/infra/issues/143

Assisted-by: OpenCode (Claude Sonnet 4.5)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,12 +10,14 @@ name: CI
 
 permissions:
   actions: read
+  contents: read
 
 on:
   push:
     branches: [main]
   pull_request:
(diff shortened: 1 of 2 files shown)


cgwalters:
OK I had an agent test this out in https://github.com/bootc-dev/ci-sandbox/ and it seems to work well!

ckyrouac (dismissed):
nice!

gursewak1997 (approved):

PR #2183, opened 2026-05-01
ci: Hard require that generated code is updated in CI

### LINKED ISSUE #1891: Gate on `update-generated`

cgwalters:
`just validate` now verifies that generated code is updated. We got burned badly by someone editing the generated code, and not noticing until the release PR which reverted the change.

While I was working on this, I wanted to preserve the property that e.g. ostree and other heavier dependencies aren't required on the host system - they're only part of the container build.

So using `podman build --output`, we can build generated docs etc as part of the container build, and copy them out to the host in the Justfile.

Assisted-by: OpenCode (Claude Sonnet 4.6)


jeckersb (approved):

cgwalters:
A good review from gemini there for sure.

PR #2187, opened 2026-05-05
ci: Split up network fetches and retry with Justfile targets

cgwalters:
Our CI is long and flakes are painful, especially with merge queues. Let's take a heavy hammer and refactor all network-fetched resources into a clearly separate stage, and then we have a fully enforced retry loop for those that has even longer timeouts in CI.

Assisted-by: OpenCode (claude-sonnet-4-6@default)


Johan-Liebert1 (approved):
lgtm

PR #2190, opened 2026-05-08
Parallel tests

Johan-Liebert1:


crates/xtask/src/tmt.rs
```diff
@@ -294,6 +296,204 @@ fn parse_plan_metadata(
...
+    fn new(plan_name: String, passed: bool, run_id: Option<String>) -> Self {
+        Self {
+            plan_name,
+            passed,
+            run_id,
+        }
+    }
+}
+
+fn run_plan(
```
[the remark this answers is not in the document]
Johan-Liebert1: Yeah this is an issue, but the final pass vs fail output is logged sequentially

.github/workflows/ci.yml
```diff
@@ -81,7 +81,7 @@ jobs:
     steps:
       - uses: actions/checkout@v6
       - name: Bootc Ubuntu Setup
-        uses: bootc-dev/actions/bootc-ubuntu-setup@main
+        uses: Johan-Liebert1/bootc-actions/bootc-ubuntu-setup@main
```
cgwalters: I think we could add an option to the action which fetches bcvk's binary built from git main or so

crates/xtask/src/tmt.rs
```diff
@@ -294,6 +296,204 @@ fn parse_plan_metadata(
...
+    plan: String,
+    vm_name: String,
+    image: String,
```
cgwalters: style nit prefer `&str`

cgwalters (commented):
Without any kind of deeper review, I think what we really want to do is upstream into https://github.com/teemtee/tmt support for bcvk.

I believe it already has support for concurrency (I mean I'd hope) and such - and that would make it a *lot* more sustainable for other projects to use tmt+bcvk (and we have many in the ecosystem that would make sense to do so)

.github/workflows/ci.yml
Johan-Liebert1: Yeah, that'd help

Johan-Liebert1:
One or two test failures, but not as good of a speedup as I'd had locally. GA seems to have 4cpu

cgwalters:
Yes, we can bump to larger runners though. I was experimenting with that previously in https://github.com/bootc-dev/ci-sandbox/issues/1 but it's been a while

Johan-Liebert1:
Alright, so composefs integration tests [now take ~30-40 mins](https://github.com/bootc-dev/bootc/actions/runs/25665271679/job/75339900454?pr=2190) instead [of ~1hr - 1hr30min](https://github.com/bootc-dev/bootc/actions/runs/25610547480/job/75181182355?pr=2191)

Tests are still quite inconsistent tough. Especially the GC one randomly takes over 1000s to complete, dunno why. Needs some investigation

Johan-Liebert1:
The ostree one takes around ~1hr instead of ~2h15min

Some tests are randomly taking a bit too long though 

```
/tmt/plans/integration/plan-32-multi-device-esp: PASSED (1140.54328774s)
```

I'll try to debug this

crates/xtask/src/tmt.rs
```diff
@@ -417,18 +684,77 @@ pub(crate) fn run_tmt(sh: &Shell, args: &RunTmtArgs) -> Result<()> {
...
 
+    let mut active_threads = 0;
+
+    let num_cpu = std::thread::available_parallelism()
+        .map(|c| c.get())
+        .unwrap_or(1);
+
+    println!("num_cpu: {num_cpu}");
+
+    let (vm_cpu, vm_mem) = (1, 1024);
```
cgwalters: We also need to cap max RAM usage.

Also on that topic, I think some of the tests may want more RAM, some less...bcvk has a concept of "instance size" that we could apply to the tests. I think tmt has similar, and maybe we could map between the two.

Johan-Liebert1: That's a good idea. Though in my testing, most tests were IO bound

Johan-Liebert1: On second thought, `ukify` keeps failing randomly with `Subprocess failed: ExitStatus(unix_wait_status(9))`. Could be due to it being oom killed.

Johan-Liebert1:
Interesting

```
content: error: 
    Building UKI: Computing composefs digest: 
        Reading container root: 
            Reading filtered filesystem: 
                Async reading filesystem from .: 
                    Scanning directory .: 
                        Scanning inode usr: Scanning directory usr: Scanning inode lib: Scanning directory lib: 
                            Getting file stats: Reading extended attributes: Operation not supported (os error 95)
```
I'll look into this one

Johan-Liebert1:
I think this one's ready for review now

crates/xtask/src/tmt.rs
```diff
@@ -564,6 +564,8 @@ fn run_plan(
     // Log disk usage after each test run to help diagnose "no space left on device" failures
     println!("Disk usage after plan {}:", plan);
     let _ = cmd!(sh, "df -h").run();
+    let _ = cmd!(sh, "du -h /var/* -d 1").run();
```
cgwalters: `df` is a cheap operation, this one is not; I don't think we should do it by default.

crates/xtask/src/tmt.rs
```diff
@@ -964,50 +1053,85 @@ pub(crate) fn run_tmt(sh: &Shell, args: &RunTmtArgs) -> Result<()> {
...
-                        eprintln!(
-                            "Warning: Failed to generate detailed report for {}: {:#}",
-                            plan, e
-                        );
-                    }
-                }
-            } else {
-                println!("Run ID not available - cannot generate detailed report");
-            }
+    let rerun_suffix = format!("{}-rerun", random_suffix);
```
cgwalters, line 1127: Shouldn't this be part of tmt?

I am really uncertain about doing this retry by default too at this level.

Ideally, we basically have retries around all network operations. Everything else - we really do want to drive flakes to zero.

crates/xtask/src/tmt.rs
```diff
@@ -407,13 +408,24 @@ fn run_plan(
 
     // Launch VM with bcvk
     let firmware_args_slice = firmware_args.as_slice();
+
+    let guard = match libvirt_lock.lock() {
```
cgwalters, line 473: >  error: failed to create directory '/run/user/1001/libvirt/qemu/run/swtpm': File exists

That's a bug in libvirt that we're just working around, right?

If we have to do a workaround, I think it'd be better in bcvk.

tmt/tests/booted/test-bib-build.nu
```diff
@@ -26,6 +26,8 @@ const BIB_IMAGE = "quay.io/centos-bootc/bootc-image-builder:latest"
 def main [] {
     tap begin "bootc-image-builder qcow2 build test"
 
+    print ">>>>>>>>>>>>>>>>>> the current working directory <<<<<<<<<<<<<<<<<" ($env.PWD)
```
cgwalters: This seems...verbose

crates/xtask/src/tmt.rs
```diff
@@ -39,6 +40,36 @@ const ENV_BOOTC_UPGRADE_IMAGE: &str = "BOOTC_upgrade_image";
 // Distro identifiers
 const DISTRO_CENTOS_9: &str = "centos-9";
 
+// Tests sorted by time taken (descending)
+const TESTS_SORTED_BY_TIME: [&str; 23] = [
```
cgwalters, line 44: Not a fan of duplicating all of our test names, this will get out of date quickly.

crates/xtask/src/tmt.rs
```diff
@@ -28,6 +28,10 @@ const FIELD_ADJUST: &str = "adjust";
 const FIELD_FIXME_SKIP_IF_COMPOSEFS: &str = "fixme_skip_if_composefs";
 const FIELD_FIXME_SKIP_IF_UKI: &str = "fixme_skip_if_uki";
 
+/// For tests that should only run for composefs systems
+/// Ex. composefs-gc
+const FIELD_SKIP_IF_OSTREE: &str = "skip_if_ostree";
```
cgwalters, line 34: Yep, I like it

crates/xtask/src/tmt.rs
```diff
@@ -299,14 +300,21 @@ fn parse_plan_metadata(
 struct RunPlanResult {
     plan_name: String,
     passed: bool,
+    time_taken: Option<Duration>,
```
cgwalters, line 393: I guess but isn't this part of tmt too?

crates/xtask/src/tmt.rs
```diff
@@ -438,16 +659,62 @@ pub(crate) fn run_tmt(sh: &Shell, args: &RunTmtArgs) -> Result<()> {
...
     let mut all_passed = true;
-    let mut test_results: Vec<(String, bool, Option<String>)> = Vec::new();
+    let mut test_results: Vec<RunPlanResult> = Vec::new();
 
     // Environment variables to pass to tmt (in addition to args.env)
     let mut tmt_env_vars = Vec::new();
 
+    let mut handles: Vec<JoinHandle<RunPlanResult>> = vec![];
+
+    let num_cpu = std::thread::available_parallelism()
```
cgwalters, line 1035: Doesn't tmt support parallelism? I think eventually the right thing to do here is to move the bcvk stuff as a first-class op in tmt.

But I don't object at all to doing it here in the interim.

cgwalters (changes requested):

crates/xtask/src/tmt.rs
Johan-Liebert1: Only added this for testing. I'll remove it

crates/xtask/src/tmt.rs
Johan-Liebert1: I'm not sure if this would be considered a bug. It only doesn't work sometimes so I believed it was due to a race condition

crates/xtask/src/tmt.rs
Johan-Liebert1: yeah... would you prefer adding a key `extra-time-taken` or something similar so we can do this sort using that key?

crates/xtask/src/tmt.rs
Johan-Liebert1: Sometimes I can see tests in CI failing because the VM never starts after rebooting. This should hopefully catch that. I'm not sure about how tmt manages VMs, but in the said case, we would have to create a new VM

crates/xtask/src/tmt.rs
Johan-Liebert1: I think for that tmt itself needs to provision the guests?

tmt/tests/booted/test-bib-build.nu
Johan-Liebert1: removed

crates/xtask/src/tmt.rs
Johan-Liebert1: Thinking a bit more about this, this doesn't need to track ALL of the plans. Only the ones that take the most time

crates/xtask/src/tmt.rs
Johan-Liebert1: We can get it from the results most probably, but I don't think there's a proper API to get elapsed time?

crates/xtask/src/tmt.rs
Johan-Liebert1: So, I think it's okay even if this diverges from what we actually have. We can definitely go with another approach

crates/xtask/src/tmt.rs
cgwalters: Yes.

crates/xtask/src/tmt.rs
Johan-Liebert1: what do you think about the rerun? Should we remove it from here? I couldn't find a proper `rerun` option in tmt

crates/xtask/src/tmt.rs
Johan-Liebert1: So, we cannot do this right now? Unless I'm missing something

PR #2191, opened 2026-05-08
bootloader: Fix a few things related to systemd-boot installs

cgwalters:
See: https://github.com/osbuild/image-builder-cli/issues/506

Right now most of our testing for composefs installs uses bcvk which uses `to-disk` in an ephemeral VM. That masks some issues, like the fact that `bootctl install` by default touches the ESP variables. Wire things up with `--generic-image` so it behaves the same as the bootupd backend.

Now, the next problem is that osbuild actually makes `/` readonly (which is great!) and in fact we should change our recommended `podman run` invocations to pass `--read-only`. But that reveals the next bug, which is that bootctl installs wants to write an `/etc/kernel/entry-token` which we basically don't use. Redirect those writes to a tempdir for now, though we should probably try to get upstream patches for that.

Also set SYSTEMD_RELAX_ESP_CHECKS=1 so bootctl skips the partition-type GUID check for the ESP - we don't want it to e.g. try to look up things in the udev database, which might not be available.

Assisted-by: OpenCode (claude-sonnet-4-6)

### DIFF
--- a/crates/lib/src/bootc_composefs/boot.rs
+++ b/crates/lib/src/bootc_composefs/boot.rs
@@ -195,10 +195,31 @@ fi
     )
 }
 
-/// Mount the ESP from the provided device
+/// Mount flags shared by all ESP mounts: non-executable, no setuid.
+const ESP_MOUNT_FLAGS: MountFlags =
+    MountFlags::from_bits_retain(MountFlags::NOEXEC.bits() | MountFlags::NOSUID.bits());
+
+/// FAT mount options: owner-only permissions on files (0600) and dirs (0700).
+const ESP_MOUNT_DATA: &std::ffi::CStr = c"fmask=0177,dmask=0077";
+
+/// Mount the ESP from the provided device into a temporary directory.
 pub fn mount_esp(device: &str) -> Result<TempMount> {
-    let flags = MountFlags::NOEXEC | MountFlags::NOSUID;
-    TempMount::mount_dev(device, "vfat", flags, Some(c"fmask=0177,dmask=0077"))
+    TempMount::mount_dev(device, "vfat", ESP_MOUNT_FLAGS, Some(ESP_MOUNT_DATA))
+}
+
+/// Mount the ESP from `device` at the given path and return a guard that
+/// synchronously unmounts (and flushes) it on drop.
+pub(crate) fn mount_esp_at(
+    device: &str,
+    path: std::path::PathBuf,
+) -> Result<bootc_mount::tempmount::MountGuard> {
+    bootc_mount::tempmount::MountGuard::mount(
+        device,
+        path,
+        "vfat",
+        ESP_MOUNT_FLAGS,
+        Some(ESP_MOUNT_DATA),
+    )
 }
 
 /// Filename release field for primary (new/upgraded) entry.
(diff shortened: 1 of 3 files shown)


crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -1145,6 +1145,108 @@ pub(crate) fn setup_composefs_uki_boot(
...
+            }
+        }
+        let esp_part = esp_part.ok_or_else(|| anyhow!("ESP partition not found"))?;
+
+        // Attach the detached composefs fsmount fd to a real tmpdir path so
+        // that mount(2) and bootctl --root can work with it.
+        let composefs = TempMount::mount_fd(composefs_mnt_fd)
+            .context("Attaching composefs image to temporary directory")?;
+
+        // Per BLS: prefer /efi when that directory exists, /boot otherwise.
```
supakeen: This comment isn't entirely clear I feel. The BLS prefers the ESP to be mounted to /boot if there's no XBOOTLDR. Otherwise it prefers the ESP mounted to /efi when there XBOOTLDR at `/boot`.

> It is recommended to mount $BOOT (either XBOOTLDR or the ESP) to /boot/. If both XBOOTLDR and the ESP are present, the ESP should be mounted to /efi/.

For *finding* the ESP the below is correct (though it omits that /boot/efi is also checked).

Bit of a nitpick so feel free to ignore.

crates/lib/src/bootloader.rs
```diff
@@ -218,66 +218,77 @@ pub(crate) fn install_via_bootupd(
...
+    let esp_path_in_root = format!("/{}", prepared_root.esp_subdir);
+
+    let mut bootctl_args = vec![
+        "install",
+        "--root",
+        root_path,
+        "--esp-path",
+        esp_path_in_root.as_str(),
+        "--boot-path",
+        esp_path_in_root.as_str(),
```
supakeen: This should be the XBOOTLDR path *if* it exists, otherwise it should be omitted?

```
--boot-path=
   Path to the Extended Boot Loader partition, as defined in the UAPI.1 Boot Loader Specification[1]. If not specified, /boot/ is
   checked. It is recommended to mount the Extended Boot Loader partition to /boot/, if possible.
```

supakeen (approved):
One nitpick, one thing I think should be changed.

We used the following redirect during our testing and that made things work, it's largely similar to the new arguments passed here:

Note that we *had* to set the `--entry-token` as it couldn't find one in `/etc/kernel/entry-token` *and* couldn't find `/etc/os-release`. The behavior is as follows:

```
If set to auto (the default), the /etc/kernel/entry-token file will be read if it exists, and the stored value used. Otherwise, if the
local machine ID is initialized it is used. Otherwise, IMAGE_ID= from os-release will be used, if set. Otherwise, ID= from os-release
will be used, if s
```

```
#!/usr/bin/env python3
import subprocess as sp
import sys


def main():
    args = sys.argv[1:]
    for idx, arg in enumerate(args):
        if arg == "--esp-path":
            args[idx+1] = "/boot/efi"
    args.extend(["--root", "/run/osbuild/mounts"])
    args.extend(["--entry-token", "literal:simonos"])

    cmd = ["/usr/bin/bootctl.orig"] + args
    print(" ".join(cmd), file=sys.stderr)
    sp.run(cmd, check=True)


if __name__ == "__main__":
    main()
```

cgwalters:
> Note that we had to set the --entry-token as it couldn't find one in /etc/kernel/entry-token and couldn't find /etc/os-release. The behavior is as follows:

But you're setting `--root` there I believe to the physical root, whereas this PR is setting it to the composefs image root.

supakeen:
> > Note that we had to set the --entry-token as it couldn't find one in /etc/kernel/entry-token and couldn't find /etc/os-release. The behavior is as follows:
> 
> But you're setting `--root` there I believe to the physical root, whereas this PR is setting it to the composefs image root.

That makes sense 🙂.

supakeen:
I can confirm that with the RPM from this PR installed in the container we get past the bootloader installation.

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -1145,6 +1145,109 @@ pub(crate) fn setup_composefs_uki_boot(
...
+
+        // Attach the detached composefs fsmount fd to a real tmpdir path so
+        // that mount(2) and bootctl --root can work with it.
+        let composefs = TempMount::mount_fd(composefs_mnt_fd)
+            .context("Attaching composefs image to temporary directory")?;
+
+        // Per BLS: if only an ESP is present (no XBOOTLDR), mount it to /boot.
+        // If both ESP and XBOOTLDR are present, mount the ESP to /efi instead.
+        // /boot/efi is explicitly discouraged by the spec.
+        let esp_subdir = if composefs
```
Johan-Liebert1: Here I think we're checking for existence of "efi" dir in the EROFS. Shouldn't we be iterating through disk partitions to find/not-find XBOOTLDR?

cgwalters: We were already using the partitions to find the *source*, this is about the mount target, and there's a bigger picture issue, Fedora derivatives are missing `/efi` by default.

supakeen: (bit of a sidenote but it might be a good idea to *move* to using `/boot` + `/efi` in Fedora, I'll discuss with a few people if it's worth a change proposal. Especially now that we have appropriate RPM macros).

Johan-Liebert1 (approved):

cgwalters:
ok yeah a definite pain point here is our rpm artifacts expire, so one can't retry just the flaky jobs after a while.

Also I think what we want here is to change the logic so that `required-checks` only gates on the default checks, not the full extended suite in PRs.

The problem is here if we opt in to more ci with ci/merge label, it then blocks adding to the merge queue until those jobs pass. Whereas we'd get automatic retries in the merge queue.

jeckersb:
> ok yeah a definite pain point here is our rpm artifacts expire, so one can't retry just the flaky jobs after a while.

Yeah this is definitely annoying.  I burned some tokens recently to try and get it to untangle things such that if only one subset of the jobs flakes (say everything passes except fedora-44-x86_64) it would only need to re-build the fedora-44-x86_64 package in order to re-run the fedora-44-x86_64 job.  Because currently it has to rebuild *all* of the packages which then in turn triggers *all* of the jobs to re-run.

But anyway my token burning didn't yield any immediate obvious way to decouple it but maybe throwing more compute at the problem, be it GPU or human brain power, could improve this substantially.

cgwalters:
I previously generated https://github.com/cgwalters/merge-queue-like-bors though I'm a bit wary to deploy it on this repo. Maybe let's start with bcvk or another one?

It *should* fix this problem though.

PR #2196, opened 2026-05-12
install: don't fsfreeze VFAT

supakeen:
When finalize_filesystems is called there are no checks that the filesystem actually supports FIFREEZE/FITHAW.

I ran into this when `bootc install to-filesystem` tried to fsfreeze my ESP; which was mounted at `/boot` as the BLS spec dictates.

Let's at least naively skip fsfreeze on VFAT (probably the most common case for an unsupported filesystem, and ext(2|3|4) is harder to tell apart.

### DIFF
--- a/crates/lib/src/install.rs
+++ b/crates/lib/src/install.rs
@@ -1421,11 +1421,19 @@ pub(crate) fn finalize_filesystem(
         .args(["-o", "remount,ro", path.as_str()])
         .run()?;
     // Finally, freezing (and thawing) the filesystem will flush the journal, which means the next boot is clean.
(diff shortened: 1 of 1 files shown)


supakeen:
Not time sensitive; I can work around it by mounting the ESP to `/boot/efi` instead during the installation.

Note that the `systemd-gpt-auto-generator` on the system that boots *does* mount the ESP to `/boot`.

Johan-Liebert1 (approved):

cgwalters (approved):
This looks right, though I'm not sure why our existing tests don't hit this. It must be something subtle about the to-disk vs to-filesystem flows.

PR #2197, opened 2026-05-13
Test if this fixes CI...

jmarrero:


jmarrero:
Did not want to mess up your work on https://github.com/bootc-dev/bootc/pull/2195 @jeckersb but if this passes by chance, you can grab my commit.

jmarrero:
If this does not work, we could try pinning the dev container before the updated bcvk: `main-aff9676` and dropping my last commit. I have no idea if  the different bvck version from cargo and dev container is a problem but let's see.

jeckersb:
well it passed the basic tests but i'll add the tag to run the full set like on #2195 and we'll see i guess

PR #2198, opened 2026-05-13
ci: Fix kernel downgrade to let scriptlets run with TMPDIR=/var/tmp

cgwalters:
Undo hacky --noscripts - this was working around an rpm-ostree bug (working on a PR for that).

Assisted-by: OpenCode (claude-sonnet-4-6@default)


Johan-Liebert1 (approved):
Reran failed jobs

Issue #1812, opened 2025-11-26
`test-integration-cfs` failed on Centos Stream 9


henrywang:
Run `test-integration-cfs` test failed on Centos Stream 9 only. Same test passed on Centos Stream 10, Fedora 42, 43, and 44.

Failure log:
```shell
$ bcvk libvirt run --name bootc-composefs-bcvk-test --filesystem=ext4 --firmware=uefi-insecure localhost/bootc
Creating libvirt domain 'bootc-composefs-bcvk-test' (install source container image: localhost/bootc)
 INFO Creating default storage pool at "/home/runner/.local/share/libvirt/images"
 INFO Default storage pool created successfully
 INFO Creating base disk: "/home/runner/.local/share/libvirt/images/bootc-base-5174e4b9590ddeb9.qcow2"
Connected (19 seconds elapsed), beginning installation...
Setting up temporary filesystems...
virtiofs mount not found at /run/virtiofs-mnt-hoststorage/, mounting manually...
Starting bootc installation...
Source image: containers-storage:localhost/bootc
Additional args: --filesystem ext4
Getting image source signatures
Copying blob sha256:df2b0c56c68b9f088529ea001d435de9f42170d62cbcc90f6b8e10bca3cd164f
......
Copying config sha256:29adbe22941a6068222d529c5df31b527d60dd27ea64b9c8aeeb0b4f61b0a923
Writing manifest to image destination
Installing image: docker://localhost/bootc:latest
Digest: sha256:e5b61e58e418f0e59c25c5337221ffcd70f453c6a34c6589e37d07d0b9149145
Block setup: direct
       Size: 21474836480
     Serial: <unknown>
      Model: <unknown>
Checking that no-one is using this disk right now ... OK

Disk /dev/vdb: 20 GiB, 21474836480 bytes, 41943040 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes

>>> Script header accepted.
>>> Script header accepted.
>>> Created a new GPT disklabel (GUID: E16F418C-E767-441A-BA61-E68145C76867).
/dev/vdb1: Created a new partition 1 of type 'BIOS boot' and of size 1 MiB.
/dev/vdb2: Created a new partition 2 of type 'EFI System' and of size 1 GiB.
[...]

PR #2201, opened 2026-05-15
composefs: Support transient /etc, transient root, and volatile /var 

cgwalters:
While working on the sealed images, I think having `/etc` transient is definitely something we want to encourage. I hit https://github.com/composefs/composefs-rs/issues/287 right away, and this fixes that.

For other cases, we'll want to also support a transient `/` and to do *that* is a lot more complicated because of SELinux - right now we need a custom early overlay service.

For `/var` I don't think we want to encourage use of overlayfs because it breaks things like podman/docker and any usage of overlayfs in general there. So we now document (and better support) `systemd.volatile=state` for this.

### DIFF
--- a/contrib/packaging/inject-baseconfig
+++ b/contrib/packaging/inject-baseconfig
@@ -1,0 +1,61 @@
+#!/bin/bash
+# Inject base configuration files for CI testing of transient-root/etc/var configurations.
+# Arguments: $1=variant, $2=baseconfigs (comma-separated, may be empty)
+set -xeuo pipefail
+
+VARIANT="${1:-}"
+BASECONFIGS="${2:-}"
+
+# No-op if no baseconfigs specified
+if [ -z "${BASECONFIGS}" ]; then
+  exit 0
+fi
+
+# setup-root-conf.toml is composefs-specific; ostree uses prepare-root.conf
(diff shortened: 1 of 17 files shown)


cgwalters:
I didn't exclude centos-9 here, which then got us into https://github.com/bootc-dev/bootc/issues/1812

contrib/packaging/inject-baseconfig
```diff
@@ -0,0 +1,61 @@
...
+    ;;
+esac
+
+mkdir -p "$(dirname "${TARGET}")"
+
+# Split on commas and process each token
+IFS=',' read -ra TOKENS <<< "${BASECONFIGS}"
+for raw_token in "${TOKENS[@]}"; do
+  # Trim leading/trailing spaces
+  token="${raw_token#"${raw_token%%[![:space:]]*}"}"
```
Johan-Liebert1, line 32: I feel like we should probably use python scripts for stuff like this

Johan-Liebert1 (approved):

contrib/packaging/inject-baseconfig
cgwalters: Yeah, or Rust (could do some xtask-style stuff here).

PR #2204, opened 2026-05-20
project: Use ImageReference from containers_image_proxy

Johan-Liebert1:
Refactor `struct ImageReference` as an alias to
`containers_image_proxy::ImageReference`, also refactor `enum Transport` as Transport from containers_image_proxy

Rename `crate::lib::spec::ImageReference` to `ImageReferenceSig` to avoid confusion with ImageReference from containers_image_proxy

Requires: https://github.com/bootc-dev/containers-image-proxy-rs/pull/138

### DIFF
--- a/crates/ostree-ext/src/cli.rs
+++ b/crates/ostree-ext/src/cli.rs
@@ -43,7 +43,7 @@ pub fn parse_imgref(s: &str) -> Result<OstreeImageReference> {
 
 /// Parse a base [`ImageReference`] from a CLI argument.
 pub fn parse_base_imgref(s: &str) -> Result<ImageReference> {
-    ImageReference::try_from(s)
+    Ok(ImageReference::try_from(s)?)
 }
 
 /// Parse an [`ostree::Repo`] from a CLI argument.
--- a/crates/xtask/src/man.rs
+++ b/crates/xtask/src/man.rs
@@ -10,6 +10,8 @@ use serde::{Deserialize, Serialize};
 use std::{fs, io::Write};
 use xshell::{Shell, cmd};
 
+use crate::out_of_sync_error;
+
 /// Represents a CLI option extracted from the JSON dump
 #[derive(Debug, Serialize, Deserialize)]
 pub struct CliOption {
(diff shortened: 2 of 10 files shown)


kelvinmulonzi:
rerun

crates/ostree-ext/src/cli.rs
```diff
@@ -43,7 +43,7 @@ pub fn parse_imgref(s: &str) -> Result<OstreeImageReference> {
 
 /// Parse a base [`ImageReference`] from a CLI argument.
 pub fn parse_base_imgref(s: &str) -> Result<ImageReference> {
-    ImageReference::try_from(s)
+    Ok(ImageReference::try_from(s)?)
```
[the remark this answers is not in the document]
Johan-Liebert1: These are two different Result types

cgwalters (approved):

crates/xtask/src/man.rs
```diff
@@ -751,7 +751,7 @@ fn check_markdown_subcommands(
     };
     if new_content != content {
         anyhow::bail!(
-            "{} is out of date; run `cargo xtask update-generated` to update it",
+            "{} is out of date; run `cargo xtask update-generated from-code` to update it",
```
cgwalters: We should have a common helper for this *and* it should refer to `just update-generated`

docs/src/host-v1.schema.json
```diff
@@ -287,7 +287,7 @@
           "description": "The host image",
           "anyOf": [
             {
-              "$ref": "#/$defs/ImageReference"
+              "$ref": "#/$defs/ImageReferenceSig"
```
cgwalters: Hmm, I'm not sure we can change the schema like this. Let's keep it as is?

cgwalters (changes requested):

docs/src/host-v1.schema.json
Johan-Liebert1: Okay

crates/xtask/src/man.rs
Johan-Liebert1: Done

cgwalters (approved):

PR #2206, opened 2026-05-21
Composefs GC changes

Johan-Liebert1:
composefs/gc: Run GC for BLS/UKI binaries on finalize

On finalize, after we are done atomically exchanging staged bootloader
entries with the current one, run GC for bootloader binaries that are
no longer referenced

We only GC the bootloader binaries and we do not touch the composefs
repository as that can turn out to be quite expensive. The repo will be
pruned in the next update/switch

### DIFF
--- a/crates/lib/src/bootc_composefs/gc.rs
+++ b/crates/lib/src/bootc_composefs/gc.rs
@@ -213,6 +213,7 @@ pub(crate) async fn composefs_gc(
     storage: &Storage,
     booted_cfs: &BootedComposefs,
     dry_run: bool,
+    prune_repo: bool,
 ) -> Result<GcResult> {
     const COMPOSEFS_GC_JOURNAL_ID: &str = "3b2a1f0e9d8c7b6a5f4e3d2c1b0a9f8e7";
 
@@ -286,6 +287,10 @@ pub(crate) async fn composefs_gc(
             BootType::Bls => delete_kernel_initrd(storage, &get_type1_dir_name(verity), dry_run)?,
             BootType::Uki => delete_uki(storage, verity, dry_run)?,
         }
+    }
+
+    if !prune_repo {
+        return Ok(GcResult::default());
     }
 
     // Identify orphaned deployments: state dirs or bootloader entries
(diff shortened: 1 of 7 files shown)


crates/lib/src/cli.rs
```diff
@@ -722,6 +722,10 @@ pub(crate) enum InternalsOpts {
         /// Implies `--dry-run`.  Intended for use in tests and health-checks.
         #[clap(long)]
         assert_no_op: bool,
+        /// If false, will only GC unreferenced BLS/UKI Binaries without touching the composefs
+        /// repository
+        #[clap(long)]
+        prune_repo: bool,
```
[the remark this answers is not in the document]
Johan-Liebert1: Updated

tmt/tests/booted/test-composefs-gc.nu
```diff
@@ -177,7 +153,7 @@ def fifth_boot [] {
     tmt-reboot
 }
 
-def sixth_boot [i: int] {
+def fifth_boot [] {
```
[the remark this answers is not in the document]
Johan-Liebert1: Oops, missed this. Updated

travier (approved):
Logic looks good. Not tested

crates/lib/src/bootc_composefs/gc.rs
```diff
@@ -213,6 +213,7 @@ pub(crate) async fn composefs_gc(
     storage: &Storage,
     booted_cfs: &BootedComposefs,
     dry_run: bool,
+    prune_repo: bool,
```
cgwalters, line 216: Two boolean arguments is ugly and confusing. Let's have a single `enum GcMode`.

cgwalters (changes requested):

crates/lib/src/bootc_composefs/gc.rs
cgwalters: Or probably `dry_run: bool` with `enum GcDepth` or so.

cgwalters: Eh it's fine we can just do this type of style fixup after too!

cgwalters (approved):

PR #2207, opened 2026-05-21
generator: Fix orphaned /etc/shadow and /etc/gshadow entries before sysusers

### LINKED ISSUE #1179: lint: Verify `/etc/{,g}shadow`
We had a case where ublue developers hit this:

`Mar 07 17:23:41 localhost systemd-sysusers[1022]: /etc/gshadow: Group "plocate" already exists.`

And that becomes pretty fatal as other sysusers don't get added.

Add a bootc lint that verifies that `/etc/gshadow` doesn't have broken entries; we have code that cleans this up in rpm-ostree somewhere in https://github.com/coreos/rpm-ostree/tree/main/rust/src/nameservice or maybe the postprocess.

(Really we should have `bootc container finalize|commit`)

cgwalters:
There's a bit of a trap in the movement from nss-altfiles to systemd-sysusers; if users/groups migrate from the former to the latter, they may leave orphaned entires in the shadow files.

systemd-sysusers then tries to create those users/groups at boot it finds them already in the shadow files and fatally errors.

Add a generator which enables a unit detects this situation and cleans up the shadow entries.

Now in practice: we probably should have made sure that nss-altfiles users don't have shadow entries at all, but that ship has sailed.

Fixes: https://github.com/bootc-dev/bootc/issues/1179

Assisted-by: OpenCode (Claude Sonnet 4.6)

### DIFF
--- a/crates/lib/src/bootc_composefs/finalize.rs
+++ b/crates/lib/src/bootc_composefs/finalize.rs
@@ -118,6 +118,13 @@ pub(crate) async fn composefs_backend_finalize(
     let diff = compute_diff(&pristine_files, &current_files, &new_files)?;
     merge(&current_etc, &current_files, &new_etc, &new_files, &diff)?;
 
+    // Remove /etc/.updated from the new deployment so that ConditionNeedsUpdate=|/etc
+    // services (systemd-sysusers, systemd-tmpfiles) run on the first boot, mirroring
+    // what ostree does in sysroot_finalize_deployment.
+    new_etc
+        .remove_file_optional(".updated")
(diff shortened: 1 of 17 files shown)


renner0e (approved):

cgwalters:
OK, discovered an existing bug here around `ConditionNeedsUpdate` with composefs backend, have a prep commit to fix that now too.

Johan-Liebert1 (approved):

cgwalters:
The flake here was likely https://github.com/bootc-dev/bcvk/issues/153

Issue #2111, opened 2026-04-01
bwrap namespace creation fails during cross-architecture install-to-filesystem


henrywang:
When building a cross-architecture bootc image (aarch64 on x86_64 host) via bootc-image-builder, `bootc install to-filesystem` fails because `bwrap` cannot create a new namespace under QEMU user-mode emulation.

The `bwrap` sandbox is created by bootc (not bootupd) when invoking `bootupctl backend install`. Under QEMU user-mode emulation for cross-architecture builds, the `clone()` syscalls used by `bwrap` to create new namespaces return `EINVAL`.

```shell
org.osbuild.bootc.install-to-filesystem: 6686eb16629b6153c396e2b0e595ee6229ef20e05fbe055fd686d096e57b3bf0 {
  "kernel-args": [
    "rw",
    "console=tty0",
    "console=ttyS0"
  ],
  "target-imgref": "quay.io/bootc-test/hidden:8puv"
}
device/disk (org.osbuild.loopback): loop0 acquired (locked: False)
mount/- (org.osbuild.ext4): mounting /dev/loop0p3 -> /store/tmp/buildroot-tmp-c159qibh/mounts/
mount/boot (org.osbuild.ext4): mounting /dev/loop0p2 -> /store/tmp/buildroot-tmp-c159qibh/mounts/boot
mount/boot-efi (org.osbuild.fat): mounting /dev/loop0p1 -> /store/tmp/buildroot-tmp-c159qibh/mounts/boot/efi
Installing image: docker://quay.io/bootc-test/hidden:8puv
Initializing ostree layout
layers already present: 0; layers needed: 74 (1.1 GB)
Deploying container image...done (2 minutes)
Bootloader: grub
Installing bootloader via bootupd
bwrap: Creating new namespace failed: Invalid argument
error: Installing to filesystem: Installing bootloader: Failed to run command: Command {
    program: "bwrap",
    args: [
        "bwrap",
        "--bind",
        "/run/osbuild/mounts/ostree/deploy/default/deploy/838eff4ba7b6318e4e54925a2426e1aa0863eb2c48f189b7b5dca622fe1da446.0",
        "/",
        "--proc",
        "/proc",
        "--dev-bind",
        "/dev",
        "/dev",
        "--bind",
        "/sys",
        "/sys",
        "--tmpfs",
        "/run",
        "--bind",
        "/run",
        "/run",
        "--bind",
        "/run/osbuild/mounts/boot",
        "/boot",
[...]

PR #2209, opened 2026-05-22
bwrap: Fall back to direct invocation on namespace creation failure

cdellacqua:
Under QEMU user-mode emulation the host kernel can return EINVAL from clone(CLONE_NEWUSER) on bwrap's sandbox setup, and bwrap aborts before spawning the child. This trips bootc install on cross-architecture builds (see issue #2111), where bwrap is used to make bootupctl run from the target image, not the buildroot.

Workaround until the upstream qemu-user/kernel/bwrap interaction is resolved: detect the "Creating new namespace failed" prefix on bwrap's stderr and re-run the target program directly. A `eprintln!` fires on every fallback so the behaviour is visible in logs. Both bootupctl sites in bootloader opt in: the `--filesystem` capability probe and `backend install`.

Assisted-by: Claude Code (Opus 4.7 1M)

### DIFF
--- a/crates/lib/src/bootloader.rs
+++ b/crates/lib/src/bootloader.rs
@@ -94,17 +94,29 @@ pub(crate) fn supports_bootupd(root: &Dir) -> Result<bool> {
 ///
 /// Runs `bootupctl backend install --help` and looks for `--filesystem` in the
 /// output. When `deployment_path` is set the command runs inside a bwrap
-/// container so we probe the binary from the target image.
+/// container so we probe the binary from the target image; if bwrap cannot
+/// create a namespace (see <https://github.com/bootc-dev/bootc/issues/2111>)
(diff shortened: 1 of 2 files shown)


cgwalters:
I get the pain here, but it's not viable in the general case.

What we *can* do here is special case just when `bootc install to-filesystem` is being run from a container (i.e. the source root == target) and then we know direct invocation == container.

However, another fallback path that would likely be quite viable here is to just do a "simple container" where we just set up some mount points directly and `chroot`. In bash it's something like this https://gitlab.com/fedora/bootc/base-images-experimental/-/blob/main/build/dnf-installroot

cdellacqua:
Thanks! Implemented the chroot approach in #2210 and confirmed it fixes the original qemu-user case on real hardware (building a bootc image for an aarch64 Raspberry Pi 4 from an amd64 laptop).

Closing as superseded.

PR #2219, opened 2026-05-29
blockdev: Restore multipath partition number fallback from device path

ckyrouac:
When the sysfs `partition` attribute is missing on multipath devices and lsblk's `partn` field is also unavailable, fall back to extracting trailing digits from the ESP device path (e.g. `/dev/mapper/mpatha2` → `"2"`). This restores the behavior originally added in [bootupd PR #1006](https://github.com/coreos/bootupd/pull/1006) that was lost during the migration to the bootc-internal-blockdev crate.

Assisted-by: Claude Code

### DIFF
--- a/crates/blockdev/src/blockdev.rs
+++ b/crates/blockdev/src/blockdev.rs
@@ -154,7 +154,9 @@ impl Device {
     /// We read `/sys/class/block/<name>/partition` rather than parsing device
     /// names because naming conventions vary across disk types (sd, nvme, dm, etc.).
     /// On multipath devices the sysfs `partition` attribute doesn't exist, so we
-    /// fall back to the `partn` field reported by lsblk.
+    /// fall back to the `partn` field reported by lsblk, then to parsing the
+    /// partition suffix from the ESP device path relative to the parent device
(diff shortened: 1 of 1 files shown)


crates/blockdev/src/blockdev.rs
```diff
@@ -170,6 +172,15 @@ impl Device {
             if let Some(partn) = esp_device.partn {
                 return Ok(partn.to_string());
             }
+            // Last resort: strip the parent device path from the ESP device path,
```
cgwalters, line 175: Hmmm....I'm OK with this but I wonder if there's not some way we can actually find the data by looking at the multipath data?

cgwalters (approved):

crates/blockdev/src/blockdev.rs
Johan-Liebert1: Might be worth looking into `KNAME` output of the `lsblk` command

travier:
Title comment/commit should be updated to reference the full URL for <https://github.com/coreos/bootupd/pull/1006> to not have it point to the bootc repo.

PR #2223, opened 2026-06-02
GrubCC binary integration

### LINKED ISSUE #2212: [ComposeFS Backend] Integration with GrubCC bootloader
GrubCC (Grub Confidential Clusters) is a minified Grub build which would work similar to systemd-boot. We would want to integrate this as another bootloader option for now. Later we should plan on removing all Grub related code and only support systemd-boot and GrubCC

Related
https://redhat.atlassian.net/browse/RHEL-171399

Johan-Liebert1:
Add GrubCC (Grub ConfidentialClusters) as a new bootloader option. This
is a minimal version of grub that's supposed to work exactly like
systemd-boot

Fixes: #2212

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -217,9 +217,11 @@ jobs:
         test_os: ${{ fromJson(needs.compute-ci-level.outputs.integration_os_matrix) }}
         variant: [ostree, composefs]
         filesystem: ["ext4", "xfs"]
-        bootloader: ["grub", "systemd"]
+        # TODO: Remove "grub" once "grub-cc" is stable
+        bootloader: ["grub", "grub-cc", "systemd"]
         boot_type: ["bls", "uki"]
         seal_state: ["sealed", "unsealed"]
+
         exclude:
           # https://github.com/bootc-dev/bootc/issues/1812
           - test_os: centos-9
@@ -244,6 +246,22 @@ jobs:
             boot_type: uki
           - variant: ostree
             bootloader: systemd
+
+          # For now only have grub-cc tests in F44
+          - test_os: fedora-45
+            bootloader: grub-cc
+          - test_os: fedora-43
+            bootloader: grub-cc
+          - test_os: centos-9
+            bootloader: grub-cc
+          - test_os: centos-10
+            bootloader: grub-cc
+          # Not in ostree
+          - variant: ostree
+            bootloader: grub-cc
+          # Not yet "sealed"
+          - bootloader: grub-cc
+            seal_state: sealed
 
     runs-on: ubuntu-24.04
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -59,6 +59,7 @@ COPY --from=target-base /target-rootfs/ /
 ARG SKIP_CONFIGS
 ARG boot_type
 ARG seal_state
+ARG bootloader
 # All network-fetching operations: package installs from distro repos, Copr, Koji.
 # Separated so `just build-fetch --target=fetch` can be retried independently on
 # transient network failures without re-running the configuration phase.
@@ -89,6 +90,20 @@ RUN --mount=type=tmpfs,target=/run --mount=type=tmpfs,target=/tmp \
     if [[ ${#pkgs_to_install[@]} -gt 0 ]]; then
         dnf install -y "${pkgs_to_install[@]}"
     fi
+
+    if [[ "$bootloader" == "grub-cc" ]]; then
+        # We have this until we get grub-cc support in bootupd
+        arch=$(uname -m)
+        curl -L -o /var/grub-cc.rpm "https://kojipkgs.fedoraproject.org/packages/grub2/2.12/59.eln156/x86_64/grub2-efi-x64-cc-2.12-59.eln156.${arch}.rpm"
+        mkdir /var/grub-cc
+        rpm2archive /var/grub-cc.rpm | tar -xvz -C /var/grub-cc
+        file=$(find /var/grub-cc -name '*.efi')
+        mkdir /usr/lib/grub-cc
+        cp $file /usr/lib/grub-cc
+        rm -rvf /var/grub-cc
+        rm -rvf /var/grub-cc.rpm
+    fi
+
 EOF
 
 # Note we don't do any customization here yet
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,8 +53,32 @@ RUN --mount=type=tmpfs,target=/run --mount=type=tmpfs,target=/tmp \
     /run/packaging/enable-compose-repos
 RUN --mount=type=tmpfs,target=/run --mount=type=tmpfs,target=/tmp /usr/libexec/bootc-base-imagectl build-rootfs --manifest=standard /target-rootfs
 
+# Get the latest GrubCC binary
+FROM quay.io/fedora/eln:latest AS grub-cc-download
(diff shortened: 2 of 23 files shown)


Dockerfile
```diff
@@ -89,6 +90,20 @@ RUN --mount=type=tmpfs,target=/run --mount=type=tmpfs,target=/tmp \
...
+        arch=$(uname -m)
+        curl -L -o /var/grub-cc.rpm "https://kojipkgs.fedoraproject.org/packages/grub2/2.12/59.eln156/x86_64/grub2-efi-x64-cc-2.12-59.eln156.${arch}.rpm"
+        mkdir /var/grub-cc
+        rpm2archive /var/grub-cc.rpm | tar -xvz -C /var/grub-cc
+        file=$(find /var/grub-cc -name '*.efi')
+        mkdir /usr/lib/grub-cc
+        cp $file /usr/lib/grub-cc
+        rm -rvf /var/grub-cc
+        rm -rvf /var/grub-cc.rpm
+    fi
```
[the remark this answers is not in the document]
Johan-Liebert1: Currently we only run on x86 afaik. And this is supposed to be temporary

crates/lib/src/spec.rs
```diff
@@ -237,6 +237,8 @@ pub enum Bootloader {
     /// Use Grub as the bootloader
     #[default]
     Grub,
+    /// Use Grub for confidential clusters as the bootloader
+    GrubCC,
```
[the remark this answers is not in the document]
Johan-Liebert1: Missed this one. Updated

crates/lib/src/bootc_composefs/status.rs
```diff
@@ -414,10 +414,14 @@ pub(crate) fn get_bootloader() -> Result<Bootloader> {
...
             if loader.to_lowercase().contains("systemd-boot") {
-                Bootloader::Systemd
-            } else {
-                Bootloader::Grub
+                return Ok(Bootloader::Systemd);
+            }
+
+            if loader.contains("GRUB CC") {
+                return Ok(Bootloader::GrubCC);
             }
```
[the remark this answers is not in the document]
Johan-Liebert1: Better for consistency. Updated

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -1369,13 +1371,48 @@ pub(crate) async fn setup_composefs_boot(
...
             &state.config_opts,
             None,
         )?;
+
+        // FIXME: Remove this hack once we have support in bootupd
+        if matches!(postfetch.detected_bootloader, Bootloader::GrubCC) {
+            root_setup
+                .physical_root
+                .remove_dir_all("boot/grub2")
+                .context("removing grub2")?;
```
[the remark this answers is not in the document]
Johan-Liebert1: this is after bootupd install so this will exist

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -1369,13 +1371,48 @@ pub(crate) async fn setup_composefs_boot(
...
+        Bootloader::Grub | Bootloader::GrubCC
+    ) {
         crate::bootloader::install_via_bootupd(
             &root_setup.device_info,
             &root_setup.physical_root_path,
             &state.config_opts,
             None,
         )?;
+
+        // FIXME: Remove this hack once we have support in bootupd
```
cgwalters, line 1382: Yes, but it'd be clearer if this was in the conditional above right? Wouldn't bootupd bail in this scenario as is now?

crates/lib/src/bootc_composefs/delete.rs
```diff
@@ -153,7 +153,7 @@ fn delete_depl_boot_entries(
             }
         },
 
-        Bootloader::Systemd => {
+        Bootloader::Systemd | Bootloader::GrubCC => {
```
cgwalters: What might be cleaner now is to have e.g.

```
enum BootloaderKind {
  BLSCompatible,
  GRUBClassic
}

impl Bootloader {
  fn kind(&self0) -> BootloaderKind
```

then most of these matches can match on the kind, not the concrete impl

crates/lib/src/bootc_composefs/boot.rs
Johan-Liebert1: Yes, bootupd will fail if it's only grub-cc, but another hack is that we don't remove grub from the Dockerfile currently. We keep it alongside the grub-cc binary. So bootupd installs grub, shim etc and afterwards we replace the grubx64.efi with grub-cc.efi. It's not clean

.github/workflows/ci.yml
```diff
@@ -245,6 +247,22 @@ jobs:
           - variant: ostree
             bootloader: systemd
 
+          # For now only have grub-cc tests in F44
```
cgwalters, line 250: Makes sense for now...but we do want this in c10s in the future right?

Hmmm maybe what we should do is make this an explicit "tier-1" target?

Dockerfile
```diff
@@ -89,6 +90,20 @@ RUN --mount=type=tmpfs,target=/run --mount=type=tmpfs,target=/tmp \
     if [[ ${#pkgs_to_install[@]} -gt 0 ]]; then
         dnf install -y "${pkgs_to_install[@]}"
     fi
+
+    if [[ "$bootloader" == "grub-cc" ]]; then
+        # We have this until we get grub-cc support in bootupd
+        arch=$(uname -m)
+        curl -L -o /var/grub-cc.rpm "https://kojipkgs.fedoraproject.org/packages/grub2/2.12/59.eln156/x86_64/grub2-efi-x64-cc-2.12-59.eln156.${arch}.rpm"
+        mkdir /var/grub-cc
+        rpm2archive /var/grub-cc.rpm | tar -xvz -C /var/grub-cc
```
cgwalters, line 127: I think it's worth a comment for why this is necessary

Dockerfile
```diff
@@ -89,6 +90,20 @@ RUN --mount=type=tmpfs,target=/run --mount=type=tmpfs,target=/tmp \
     if [[ ${#pkgs_to_install[@]} -gt 0 ]]; then
         dnf install -y "${pkgs_to_install[@]}"
     fi
+
+    if [[ "$bootloader" == "grub-cc" ]]; then
+        # We have this until we get grub-cc support in bootupd
+        arch=$(uname -m)
+        curl -L -o /var/grub-cc.rpm "https://kojipkgs.fedoraproject.org/packages/grub2/2.12/59.eln156/x86_64/grub2-efi-x64-cc-2.12-59.eln156.${arch}.rpm"
```
cgwalters: This is likely to be subject to GC in the future and take out our CI.

We could use `koji latest-pkg eln grub`; or a separate stage which uses `FROM quay.io/fedora/eln:latest` + `dnf download`.

Also there's one use of ${arch} and one hardcoded x86_64.

cgwalters (dismissed):
Looks overall sane

.github/workflows/ci.yml
Johan-Liebert1: > but we do want this in c10s in the future right?

yes. On initial testing, the grub-cc binary does work on c10s so we could just hack it up until we get proper packaging

Dockerfile
Johan-Liebert1: Updated to download from a separate stage

Johan-Liebert1:
@cgwalters this needs a re-approval

jeckersb (approved):

PR #2225, opened 2026-06-02
Switch over to weekly releases by default

cgwalters:
Every 3 weeks is just slow enough that leaves people wondering when the next release is, but too fast to really have truly meaningful versions by default.

Let's cut over to weekly releases (like Cockpit and many other projects) by default, and only bump the patch version by default (even if we have some features). We can reserve bumping the minor version for when there are bigger features.

### DIFF
--- a/.github/workflows/scheduled-release.yml
+++ b/.github/workflows/scheduled-release.yml
@@ -2,14 +2,12 @@ name: Create Release PR
 
 on:
   schedule:
-    # Run every 3 weeks on Monday at 8:00 AM UTC
-    # Note: GitHub Actions doesn't support "every 3 weeks" directly,
-    # so we use a workaround by running weekly and checking if it's been 3 weeks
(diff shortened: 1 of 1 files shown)


jeckersb (dismissed):

.github/workflows/scheduled-release.yml
joelcapitao: It's not directly related to this PR but can we also append `from-code` to that command c.f https://github.com/bootc-dev/bootc/pull/2183 ?
I just realized that latest GH cronjob failed https://github.com/bootc-dev/bootc/actions/runs/26747408954/job/78826364015 because of it

cgwalters: Good catch, thanks.

Johan-Liebert1 (approved):

joelcapitao (approved):

PR #2235, opened 2026-06-05
xtask: Use insecure firmware for all non-sealed VM tests

jmarrero:
The stock OVMF Secure Boot key database (shipped with the Ubuntu runner's
edk2/ovmf package) does not include the distro signing keys needed to
verify the shim/grub chain for CentOS Stream or Fedora.  Previously,
only systemd-boot tests explicitly requested --firmware=uefi-insecure
while grub tests inherited bcvk's default of uefi-secure, causing UEFI
to reject the boot with 'Security Violation'.

This worked by coincidence until the CentOS Stream 10 compose re-signed
grub2 (2.12-47) with a different certificate (Signing 802 / CA 8 instead
of Signing 202 / CA 2), which changed the Authenticode signature and
broke whatever fragile verification path was succeeding before.

Fix by using --firmware=uefi-insecure for all non-sealed images.  Sealed
images continue to use uefi-secure with explicitly enrolled test keys.

Signed-off-by: Joseph Marrero Corchado <jmarrero@redhat.com>
Assisted-by: OpenCode (Claude claude-opus-4-6)

### DIFF
--- a/crates/xtask/src/bcvk.rs
+++ b/crates/xtask/src/bcvk.rs
@@ -110,10 +110,14 @@ impl BcvkInstallOpts {
                      Run 'just generate-secureboot-keys' to generate them."
                 );
             }
-        } else if matches!(self.bootloader, Some(Bootloader::Systemd)) {
-            Ok(vec!["--firmware=uefi-insecure".into()])
         } else {
-            Ok(Vec::new())
+            // Use insecure firmware for all non-sealed images.  The stock
+            // OVMF Secure Boot key database does not include the distro
+            // signing keys needed to verify shim/grub, so Secure Boot
+            // verification fails at the firmware level with
+            // "Security Violation".  Sealed images work because they enroll
+            // custom test keys and use a test-signed systemd-boot.
+            Ok(vec!["--firmware=uefi-insecure".into()])
         }
     }
 }


jmarrero:
Need more investigation to submit to grub upstream/rpm maintainers but hopefully this unblocks CI

cgwalters:
I think it will work (for now) to add other compose repos into our build, and then pin grub to the previous version.

jmarrero:
I got an agent trying to re-verify this, testing the issue locally to isolate the grub commit as I am getting mixed results. The latest grub2 EFI bin works correctly on the previous compose and the code changes don't seem to jump at me. Re-trying.

cgwalters:
I added the ci/tier-1 label here so we verify work pre-merge-queue

jmarrero:
It's definitely on the grub path, --bootloader=systemd does not have this issue. Still digging.

cgwalters:
Yes, probably related to https://gitlab.com/redhat/centos-stream/rpms/grub2/-/commit/db5c73de4139ace15b496062cc445f0374518a28

jmarrero:
This is already passing the *test-baseconfigs jobs that were failing before 🤞

Johan-Liebert1 (approved):

jmarrero:
Merge queue does not let us retry a job that fails. This is quite painful. Currently just failing on quay timeouts.

cgwalters:
Thanks for doing this! I dug into this a little bit more, and I think a root here is that https://gitlab.com/redhat/centos-stream/rpms/shim/-/commit/1d00d75dece47e1bae0dc4e74c3d343da3670016 isn't actually built or shipped yet.

IOW it's a shim -> grub thing, not related to the host OS (or more specifically the host OS's virt stack) at all.

I would also like to change the special casing we're doing here to only skip SB on centos-10 by default.

PR #2236, opened 2026-06-05
ostree-ext: Add composefs-repo to ostree-commit synthesis

cgwalters:
Add `container::composefs_import::import_from_composefs_repo`, which walks a composefs `FileSystem` tree (already populated by a composefs-rs OCI pull) and synthesizes an equivalent ostree commit, reflinking file content into the ostree bare repo so the same disk blocks back both stores. This is the inverse of libostree's ostree→EROFS path.

To support exporting such synthesized commits back out, `store.rs` gains `export_commit` (export straight from a commit checksum rather than an image ref) and `for_each_file_checksum` (walk a commit's regular-file content checksums), `export_to_oci` is refactored to take a `LayeredImageState` so both ref- and commit-based exports share it, and `query_image_commit` learns to treat a composefs-synthesized merge commit as its own base. A few existing metadata constants and helpers are promoted to `pub`/`pub(crate)` for use by the new module.

This is library plumbing for bootc's unified storage work; it is unused by any in-tree caller on its own and lands separately to keep that feature's change set reviewable. Includes a standalone integration test (`test_composefs_import_to_ostree`) that builds a small OCI image, imports it through a composefs repo into ostree, and verifies the result.

Assisted-by: OpenCode (claude-opus-4-8)

### DIFF
--- a/crates/ostree-ext/src/container/composefs_import.rs
+++ b/crates/ostree-ext/src/container/composefs_import.rs
@@ -1,0 +1,970 @@
+//! Synthesize an ostree commit directly from a composefs OCI repository.
+//!
+//! This is the inverse of the existing ostree→composefs flow in libostree.
+//! Rather than importing OCI layers through the tar pipeline, we walk the
+//! composefs `FileSystem` tree and write each entry as an ostree object.
+//!
+//! For external regular files we bypass the ostree C library's `write_content`
+//! API and instead write content objects directly to disk.  This lets us
+//! attempt `FICLONE` (reflink) from the composefs object fd, falling back to
+//! `copy_file_range` / read-write.  The ostree SHA256 content checksum is
+//! still computed via the C library's `checksum_file_from_input` (a pure,
+//! repo-independent function) so the serialisation format is never
+//! reimplemented.
+//!
+//! Metadata objects (dirmeta, dirtree, commit) and small inline files continue
+//! to use the ostree C API — they are tiny and benefit from the existing
+//! validation.
+//!
+//! # Why reflink, not hardlink
+//!
+//! A composefs object is content-addressed by its SHA-512 fsverity digest.
+//! Multiple filesystem paths with identical content share a single composefs
+//! object (inode).  In ostree bare mode, however, every content object carries
+//! its own per-inode metadata: uid, gid, mode, and *all* xattrs including
+//! `security.selinux`.  The same composefs object can be referenced by two
+//! ostree objects that need different SELinux labels — for example
+//! `/etc/sudo.conf` (label `etc_t`) and
+//! `/usr/share/doc/sudo/examples/sudo.conf` (label `usr_t`).  Because a
+//! single inode can hold only one `security.selinux` value, hardlinking both
+//! ostree objects to the shared composefs inode is impossible: the second
+//! `setxattr` would overwrite the first, causing `ostree fsck` to report a
+//! checksum mismatch.
+//!
+//! Reflinks (FICLONE) solve this: each ostree object gets its own inode and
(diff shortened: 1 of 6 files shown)


crates/ostree-ext/src/container/composefs_import.rs
```diff
@@ -0,0 +1,970 @@
...
+    }
+    let children = xattrs.iter().map(|(k, v)| {
+        // Append NUL so that ostree's C validator can read the name as a C string.
+        let mut name_bytes = k.as_bytes().to_vec();
+        name_bytes.push(b'\0');
+        // SELinux stores label values as NUL-terminated strings on the filesystem.
+        // The composefs selabel() function strips the NUL, but ostree fsck reads
+        // xattrs back from disk (where the kernel always stores them with NUL) when
+        // verifying checksums.  Ensure our checksum matches by adding the NUL here
+        // for security.selinux values that are missing it.
```
cgwalters, line 146: Gotta say this one was fun

crates/ostree-ext/src/container/composefs_import.rs
```diff
@@ -0,0 +1,970 @@
...
+        {
+            value_owned = [value, &[0u8]].concat();
+            value_owned.as_slice()
+        } else {
+            value
+        };
+        glib::Variant::tuple_from_iter([name_bytes.as_slice().to_variant(), value.to_variant()])
+    });
+    Some(glib::Variant::array_from_iter::<(&[u8], &[u8])>(children))
+}
```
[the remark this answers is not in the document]
cgwalters: OK yeah the training-cutoff confabulations are annoying here, and anyways since this bot is going to be sunset, I did https://github.com/bootc-dev/infra/issues/200

crates/ostree-ext/src/container/composefs_import.rs
```diff
@@ -0,0 +1,970 @@
...
+    image_config: &crate::oci_spec::image::ImageConfiguration,
+    reflinks_required: bool,
+) -> Result<String> {
+    let cancellable = gio::Cancellable::NONE;
+
+    // Reconstruct the merged filesystem tree from the composefs store
+    let mut fs = composefs_oci::image::create_filesystem(composefs_repo, config_digest, None)
+        .context("Reconstructing composefs filesystem tree")?;
+
+    // Apply SELinux labels in-memory before writing any objects.
```
cgwalters, line 752: This should just use the `transform_for_boot`

crates/ostree-ext/src/container/composefs_import.rs
```diff
@@ -0,0 +1,970 @@
+//! Synthesize an ostree commit directly from a composefs OCI repository.
```
cgwalters, line 1: TODO: Let's add an equivalence/diff testing of this versus the current ostree-ext container backend

crates/ostree-ext/src/container/composefs_import.rs
```diff
@@ -0,0 +1,970 @@
...
+    let dirmeta = create_dirmeta(&dir.stat);
+    let meta_csum = orepo
+        .write_metadata(
+            crate::ostree::ObjectType::DirMeta,
+            None,
+            &dirmeta,
+            gio::Cancellable::NONE,
+        )
+        .context("Writing dirmeta")?;
+    mtree.set_metadata_checksum(&meta_csum.to_hex());
```
Johan-Liebert1, line 311: This code block is repeated a few times. Does it make sense to factor it out?

Johan-Liebert1 (dismissed):
lgtm. This is a separate feature so unlikely to break anything. Just one comment

cgwalters:
We can redo this on top of https://github.com/composefs/composefs-rs/pull/347

PR #2240, opened 2026-06-09
Prepwork for #2200

### LINKED ISSUE #2185: Rework UKI generation to enable removing kernel & initrd from image
We added `bootc container ukify` (https://github.com/bootc-dev/bootc/issues/1955 & https://github.com/bootc-dev/bootc/pull/1960) and current implementation takes the kernel and initrd from the rootfs: https://github.com/bootc-dev/bootc/blob/c12eca3e2f28db576b167b5e96f6fe0fe2b2a398/crates/lib/src/ukify.rs#L55-L79

Support in bcvk to extract kernel & initrd from a UKI was added in https://github.com/bootc-dev/bcvk/pull/144.

We have some comments in https://github.com/bootc-dev/bootc/blob/c12eca3e2f28db576b167b5e96f6fe0fe2b2a398/contrib/packaging/finalize-uki#L39 but they are too late as we can not remove files from the image after we've computed its composefs fsverity hash as it would change it.

Thus in order to be able to remove the kernel & initrd from the final image and keep only the UKI, we need to remove them in a stage *before* the sealed-uki stage, but *after* the rechunking stage (to actually get the space back), and then copy them back into the sealed-uki stage where we would pass them to the bootc container ukify command as arguments.

Simplified Containerfile:
```
FROM bootc-base-image as rootfs
RUN Do customisations here

FROM rootfs as rootfs-no-kernel
RUN Remove kernel (and potentially initrd)

# Do the rechunking here
...

# Compute fs-verity hash of rechunked rootfs-no-kernel image
FROM tools as sealed-uki
COPY --from=rootfs kernel & initrd
RUN bootc container ukify --kernel ... --initrd ...

# Copy back UKI
FROM rootfs-no-kernel-rechunked
COPY UKI from sealed-uki
```

I've started investigating how this would look like in: https://github.com/travier/fedora-atomic-desktops-sealed/pull/24.

For the Atomic Desktops use case, this compounds with https://github.com/travier/fedora-atomic-desktops-sealed/issues/23. The boot time is significantly impacted by the size of the initrd which makes up for most of the UKI. Thus I'm looking at building multiple initrds and thus multiple UKIs from a single base image. I have thus split the built into two distinct Containerfiles / image build to not have to pay to cost of rechunking the base image for each build.

A more elegant solution for the initrd size would be to use system extensions for the initrd but that adds another set of feature to add to bootc to manage (and we'll have to figure out how to sign them) so I've not looked at that yet.

Johan-Liebert1:
ukify: Allow passing path to kernel and initramfs

While building a sealed UKI image we'd want to remove the original
kernel + initramfs from the final image and have only the final UKI
present. This was not possible before as `bootc container ukify`
expected kernel + initramfs to be present in `usr/lib/modules` of
container root

We now accept a parameter `kernel_dir` which must be of the format
`/path/$kernel_ver` for `bootc container ukify`

Fixes: #2185

---

test/tmt: Don't run bootloader-none test for composefs

Bootloader set to none is not supported with the composefs backend so we
skip tests with this option for composefs backend

---

image-proxy: Use privileged user when pull from containers storage

We were defaulting to unprivileged user "nobody" when pulling an image,
but pulling from containers-storage was failing as it requires extra
privileges. Default to the current user, usually root, when pulling from
containers-storage

### DIFF
--- a/crates/lib/src/cli.rs
+++ b/crates/lib/src/cli.rs
@@ -442,6 +442,13 @@ pub(crate) enum ContainerOpts {
         #[clap(long)]
         write_dumpfile_to: Option<Utf8PathBuf>,
 
+        /// The directory containing the kernel and initramfs.img
+        /// Must be of the format /parent/$kernel_version
+        ///
+        /// Ex. /boot/6.18.7-100.fc42.x86_64
+        #[clap(long)]
+        kernel_dir: Option<Utf8PathBuf>,
+
         /// Additional arguments to pass to ukify (after `--`).
         #[clap(last = true)]
         args: Vec<OsString>,
@@ -1902,12 +1909,36 @@ async fn run_from_opt(opt: Opt) -> Result<()> {
                 kargs,
                 allow_missing_verity,
                 write_dumpfile_to,
+                kernel_dir,
                 args,
             } => {
+                let kernel = match kernel_dir {
+                    Some(kernel_dir) => {
+                        let kver = kernel_dir
+                            .components()
+                            .last()
+                            .ok_or_else(|| anyhow::anyhow!("Could not determine kernel version"))?;
+
+                        Some(crate::kernel::KernelInternal {
+                            kernel: crate::kernel::Kernel {
+                                unified: false,
+                                version: kver.to_string(),
+                            },
+                            k_type: crate::kernel::KernelType::Vmlinuz {
+                                path: kernel_dir.join("vmlinuz"),
+                                initramfs: kernel_dir.join("initramfs.img"),
+                            },
+                        })
+                    }
+
+                    None => None,
+                };
+
                 crate::ukify::build_ukify(
                     &rootfs,
                     &kargs,
                     &args,
+                    kernel,
                     allow_missing_verity,
                     write_dumpfile_to.as_deref(),
                 )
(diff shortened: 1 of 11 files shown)


cgwalters (approved):

crates/lib/src/cli.rs
```diff
@@ -442,6 +442,13 @@ pub(crate) enum ContainerOpts {
         #[clap(long)]
         write_dumpfile_to: Option<Utf8PathBuf>,
 
+        /// The directory containing the kernel and initramfs.img
```
travier, line 445:
```suggestion
        /// The directory containing the kernel and initramfs.img.
```

crates/lib/src/cli.rs
```diff
@@ -442,6 +442,13 @@ pub(crate) enum ContainerOpts {
...
+        /// Must be of the format /parent/$kernel_version
```
travier, line 446: Maybe?

```suggestion
        /// Must be of the format `/parent/$kernel_version`.
```

tmt/tests/booted/test-install-no-boot-dir.nu
```diff
@@ -19,7 +24,7 @@ def main [] {
     mount -o loop disk.img /var/mnt
 
     setenforce 0
-
+
```
travier, line 27:
```suggestion

```

PR #2244, opened 2026-06-10
Update shlex to 2.0.1

jeckersb:
The 2.0.0 release had a bug which broke the docs build, which in turn
causes `cargo doc` to fail for bootc as well.

We explicitly use `--no-deps` in our validate job when building our
docs so we don't catch things like this in other crates, but it's
annoying when I want to generate all the docs locally.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -291,7 +291,7 @@ dependencies = [
  "rustix",
  "serde",
  "serde_json",
- "shlex 2.0.0",
+ "shlex 2.0.1",
  "similar-asserts",
  "static_assertions",
  "tempfile",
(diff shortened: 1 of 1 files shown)


jeckersb:
Actually nevermind this is already included in #2226 I didn't realize that one is still stuck trying to get through the merge queue :sob:

jeckersb:
Ok nevermind, despite #2226 listing 

`[shlex](https://crates.io/crates/shlex) 	workspace.dependencies 	patch 	2.0.0 → 2.0.1`

in the description, it didn't actually update shlex...

cgwalters (approved):

PR #2245, opened 2026-06-11
Do not recursively clean tempdir

Johan-Liebert1:
Ran into this issue by chance in bootupd where the following drop ordering

```rust
drop(tempdir)
drop(mount) // unmounts thing mounted at tempdir
```

was causing all the contents of the mounted device to be deleted because the tempdir was being deleted. We might run into the same issue with our Tempdir impl if we fail to unount the ESP and tempdir is dropped deleting everything in the ESP.

To prevent that, explicitly set `disable_cleanup` to false on the Tempdir and only enable cleanup after we've successfully unmounted whatever was mounted

### DIFF
--- a/crates/mount/src/tempmount.rs
+++ b/crates/mount/src/tempmount.rs
@@ -49,11 +49,43 @@ impl Drop for MountGuard {
     }
 }
 
+/// Holds a tempdir with custom Drop impl
+#[derive(Debug)]
+pub struct MountpointTempdir(tempfile::TempDir);
+
+impl std::ops::Deref for MountpointTempdir {
+    type Target = tempfile::TempDir;
+    fn deref(&self) -> &tempfile::TempDir {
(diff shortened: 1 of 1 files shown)


cgwalters (commented):
Oh, that is horrifying.

How about this instead: We have a `MountpointTempdir` whose `Drop` *only* does `rmdir` (not recursively).

Of course, in many cases I think the real fix here is for us to always use the new mount API to just mount a fd, not a tempdir.

Johan-Liebert1:
> Of course, in many cases I think the real fix here is for us to always use the new mount API to just mount a fd, not a tempdir

we still need a mount point with that right? We could also always mount the ESP at `/boot` since that's what the spec says?

> We have a MountpointTempdir whose Drop only does rmdir (not recursively)

Done

cgwalters (approved):

PR #2248, opened 2026-06-12
Add support for multiple composefs= digests

cgwalters:
This adapts bootc to build on top of the work in https://github.com/composefs/composefs-rs/pull/297

A toplevel goal here is supporting *both* the v1 and v2 EROFS formats, which means we'll work with RHEL9 era systems.

Right now `bootc container ukify` still generates `composefs=` i.e. v2, but I'd like to change that to do both - it's a pretty cheap thing (the main cost is generating the fsverity digests).

### DIFF
--- a/contrib/packaging/seal-uki
+++ b/contrib/packaging/seal-uki
@@ -4,6 +4,7 @@ set -xeuo pipefail
 
 missing_verity=()
 dumpfile_args=()
+erofs_version=v1
 
 while [ ! -z "${1:-}" ]; do
     case "$1" in
@@ -41,6 +42,12 @@ while [ ! -z "${1:-}" ]; do
 
         "--write-dumpfile-to")
             dumpfile_args=(--write-dumpfile-to "$2")
+            shift
+            shift
+        ;;
+
+        "--erofs-version")
+            erofs_version="$2"
             shift
             shift
         ;;
@@ -92,4 +99,4 @@ containerukifyargs=(--rootfs "${target}")
 
 # Build the UKI using bootc container ukify
 # This computes the composefs digest, reads kargs from kargs.d, and invokes ukify
-bootc container ukify "${containerukifyargs[@]}" "${kernel_params[@]}" "${missing_verity[@]}" "${dumpfile_args[@]}" -- "${ukifyargs[@]}"
+bootc container ukify "${containerukifyargs[@]}" "${kernel_params[@]}" "${missing_verity[@]}" "${dumpfile_args[@]}" --erofs-version="${erofs_version}" -- "${ukifyargs[@]}"
--- a/crates/lib/src/bootc_composefs/boot.rs
+++ b/crates/lib/src/bootc_composefs/boot.rs
@@ -76,6 +76,7 @@ use cap_std_ext::{
     dirext::CapStdExtDirExt,
 };
 use clap::ValueEnum;
+use composefs::erofs::format::FormatVersion;
 use composefs::fs::read_file;
 use composefs::fsverity::{FsVerityHashValue, Sha512HashValue};
 use composefs::tree::{FileSystem, RegularFile};
@@ -90,16 +91,19 @@ use composefs_ctl::composefs;
 use composefs_ctl::composefs_boot;
 use composefs_ctl::composefs_oci;
 use fn_error_context::context;
-use linux_kernel_cmdline::utf8::{Cmdline, Parameter};
+use linux_kernel_cmdline::utf8::{Cmdline, Parameter, ParameterKey};
 use ostree_ext::composefs::dumpfile;
 use rustix::{mount::MountFlags, path::Arg};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
 use crate::bootc_composefs::state::{get_booted_bls, write_composefs_state};
-use crate::bootc_composefs::status::ComposefsCmdline;
+use crate::bootc_composefs::status::build_composefs_karg;
 use crate::bootc_kargs::compute_new_kargs;
-use crate::composefs_consts::{TYPE1_BOOT_DIR_PREFIX, TYPE1_ENT_PATH, TYPE1_ENT_PATH_STAGED};
+use crate::composefs_consts::{
+    COMPOSEFS_CMDLINE, COMPOSEFS_DIGEST_CMDLINE, TYPE1_BOOT_DIR_PREFIX, TYPE1_ENT_PATH,
+    TYPE1_ENT_PATH_STAGED,
+};
 use crate::parsers::bls_config::{BLSConfig, BLSConfigType, EFIKey};
 use crate::spec::BootloaderKind;
 use crate::task::Task;
@@ -657,6 +661,15 @@ struct BLSEntryPath {
     config_path: Utf8PathBuf,
 }
 
+/// Replace either karg spelling to ensure only the selected EROFS format remains.
+fn replace_composefs_karg(cmdline: &mut Cmdline, new_karg: &str) -> Result<()> {
+    cmdline.remove(&ParameterKey::from(COMPOSEFS_CMDLINE));
+    cmdline.remove(&ParameterKey::from(COMPOSEFS_DIGEST_CMDLINE));
+    let parameter = Parameter::parse(new_karg).context("Parsing composefs kernel parameter")?;
+    cmdline.add_or_modify(&parameter);
+    Ok(())
+}
+
 /// Sets up and writes BLS entries and binaries (VMLinuz + Initrd) to disk
 ///
 /// # Returns
@@ -666,6 +679,7 @@ pub(crate) fn setup_composefs_bls_boot(
     setup_type: BootSetupType,
     repo: &crate::store::ComposefsRepository,
     id: &Sha512HashValue,
+    format_version: FormatVersion,
     entry: &ComposefsBootEntry<Sha512HashValue>,
     mounted_erofs: &Dir,
 ) -> Result<String> {
@@ -684,9 +698,12 @@ pub(crate) fn setup_composefs_bls_boot(
                 }
             }
 
-            let composefs_cmdline =
-                ComposefsCmdline::build(&id_hex, state.composefs_options.allow_missing_verity);
-            cmdline_options.extend(&Cmdline::from(&composefs_cmdline.to_string()));
+            let composefs_cmdline = build_composefs_karg(
+                id.clone(),
+                format_version,
+                state.composefs_options.allow_missing_verity,
+            );
+            cmdline_options.extend(&Cmdline::from(&composefs_cmdline));
 
             // If there's a separate /boot partition, add a systemd.mount-extra
             // karg so systemd mounts it after reboot. This avoids writing to
@@ -732,14 +749,14 @@ pub(crate) fn setup_composefs_bls_boot(
                 _ => anyhow::bail!("Found NonEFI config"),
             };
 
-            // Copy all cmdline args, replacing only `composefs=`
-            let cfs_cmdline =
-                ComposefsCmdline::build(&id_hex, booted_cfs.cmdline.allow_missing_fsverity)
-                    .to_string();
-
-            let param = Parameter::parse(&cfs_cmdline)
-                .context("Failed to create 'composefs=' parameter")?;
-            cmdline.add_or_modify(&param);
+            replace_composefs_karg(
+                &mut cmdline,
+                &build_composefs_karg(
+                    id.clone(),
+                    format_version,
+                    booted_cfs.cmdline.allow_missing_fsverity,
+                ),
+            )?;
 
             // Locate ESP partition device by walking up to the root disk(s)
             let root_dev = bootc_blockdev::list_dev_by_dir(&storage.physical_root)?;
@@ -963,6 +980,7 @@ struct UKIInfo {
     version: Option<String>,
     os_id: Option<String>,
     boot_digest: String,
+    composefs_digest: Sha512HashValue,
 }
 
 /// Determines the directory (under `mounted_efi`) that a PE binary should be written to.
@@ -1008,6 +1026,7 @@ fn write_pe_to_esp(
     file_path: &Utf8Path,
     pe_type: PEType,
     uki_id: &Sha512HashValue,
+    boot_ids: &[Sha512HashValue],
     missing_fsverity_allowed: bool,
     mounted_efi: impl AsRef<Path>,
 ) -> Result<Option<UKIInfo>> {
@@ -1034,7 +1053,7 @@ fn write_pe_to_esp(
         let composefs_info = ComposefsBootCmdline::<Sha512HashValue>::from_cmdline(&cmdline)
             .context("Parsing composefs=")?
             .ok_or_else(|| anyhow::anyhow!("No composefs image in UKI cmdline"))?;
-        let composefs_cmdline = composefs_info.digest();
+        let composefs_digest = composefs_info.digest().clone();
         let missing_verity_allowed_cmdline = composefs_info.is_insecure();
 
         // If the UKI cmdline does not match what the user has passed as cmdline option
@@ -1053,16 +1072,17 @@ fn write_pe_to_esp(
             _ => { /* no-op */ }
         }
 
-        let file_name = file_path.file_name();
-
-        if *composefs_cmdline != *uki_id {
+        if !boot_ids.contains(&composefs_digest) {
             return Err(UKIDigestMismatch {
-                actual: composefs_cmdline.to_hex(),
+                actual: composefs_digest.to_hex(),
                 expected: uki_id.to_hex(),
-                uki_name: file_name.map(|x| x.to_string()),
+                uki_name: file_path.file_name().map(|name| name.to_string()),
             }
             .into());
         }
+        composefs_info
+            .validate_digest(boot_ids)
+            .context("Validating UKI composefs digest")?;
 
         uki_reader.seek(SeekFrom::Start(0))?;
         let osrel = uki::get_text_section_buffered(&mut uki_reader, ".osrel")?;
@@ -1079,6 +1099,7 @@ fn write_pe_to_esp(
             version: parsed_osrel.get_version(),
             os_id: parsed_osrel.get_value(&["ID"]),
             boot_digest,
+            composefs_digest,
         });
     }
 
@@ -1088,8 +1109,12 @@ fn write_pe_to_esp(
     let pe_dir = Dir::open_ambient_dir(&final_pe_path, ambient_authority())
         .with_context(|| format!("Opening {final_pe_path:?}"))?;
 
+    let pe_name_owned;
     let pe_name = match pe_type {
-        PEType::Uki => &get_uki_name(&uki_id.to_hex()),
+        PEType::Uki => {
+            pe_name_owned = get_uki_name(&boot_label.as_ref().unwrap().composefs_digest.to_hex());
+            &pe_name_owned
+        }
         PEType::UkiAddon | PEType::GlobalUkiAddon => file_path
             .components()
             .last()
@@ -1274,8 +1299,9 @@ pub(crate) fn setup_composefs_uki_boot(
     setup_type: BootSetupType,
     repo: &crate::store::ComposefsRepository,
     id: &Sha512HashValue,
+    boot_ids: &[Sha512HashValue],
     entries: Vec<ComposefsBootEntry<Sha512HashValue>>,
-) -> Result<String> {
+) -> Result<(String, Sha512HashValue)> {
     let (root_path, esp_device, bootloader, missing_fsverity_allowed, uki_addons) = match setup_type
     {
         BootSetupType::Setup((root_setup, state, postfetch)) => {
@@ -1361,6 +1387,7 @@ pub(crate) fn setup_composefs_uki_boot(
                     utf8_file_path,
                     entry.pe_type,
                     &id,
+                    boot_ids,
                     missing_fsverity_allowed,
                     esp_mount.dir.path(),
                 )?;
@@ -1376,18 +1403,27 @@ pub(crate) fn setup_composefs_uki_boot(
         uki_info.ok_or_else(|| anyhow::anyhow!("Failed to get version and boot label from UKI"))?;
 
     let boot_digest = uki_info.boot_digest.clone();
+    let deploy_id = uki_info.composefs_digest.clone();
 
     match bootloader.kind()? {
-        BootloaderKind::GRUBClassic => {
-            write_grub_uki_menuentry(root_path, &setup_type, uki_info.boot_label, id, &esp_device)?
-        }
+        BootloaderKind::GRUBClassic => write_grub_uki_menuentry(
+            root_path,
+            &setup_type,
+            uki_info.boot_label,
+            &deploy_id,
+            &esp_device,
+        )?,
 
-        BootloaderKind::BLSCompatible => {
-            write_systemd_uki_config(&esp_mount.fd, &setup_type, uki_info, id, &bootloader)?
-        }
+        BootloaderKind::BLSCompatible => write_systemd_uki_config(
+            &esp_mount.fd,
+            &setup_type,
+            uki_info,
+            &deploy_id,
+            &bootloader,
+        )?,
     };
 
-    Ok(boot_digest)
+    Ok((boot_digest, deploy_id))
 }
 
 /// A composefs image attached to a temporary directory with the ESP and a
@@ -1611,6 +1647,12 @@ pub(crate) async fn setup_composefs_boot(
     )
     .context("Generating bootable EROFS image")?;
 
+    let oci_img =
+        composefs_oci::oci_image::OciImage::open(&*repo, &pull_result.manifest_digest, None)
+            .context("Opening OCI image to read boot image refs")?;
+    let boot_id_v1 = oci_img.boot_image_ref_v1().cloned();
+    let boot_id_v2 = oci_img.boot_image_ref_v2().cloned();
+
     // Reconstruct the OCI filesystem to discover boot entries (kernel, initramfs, etc.).
     let fs = composefs_oci::image::create_filesystem(
         &*repo,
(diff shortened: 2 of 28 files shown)


cgwalters:
OK I think the chain here should be:

- Merge https://github.com/composefs/composefs-rs/pull/314
- Rebase on this on top

jmarrero (dismissed):
lgtm

just needs a rebase I think.

The only nit is that 
```
            config.erofs_formats = composefs_ctl::composefs::erofs::format::FormatConfig {
                default: composefs_ctl::composefs::erofs::format::FormatVersion::V1,
                extra: [composefs_ctl::composefs::erofs::format::FormatVersion::V2].into(),
            };
```

Is duplicated 3 times here. So if we add a new version in the future we need to care about 3 places. Maybe it should be a helper function?

contrib/packaging/seal-uki
```diff
@@ -15,6 +15,9 @@ allow_missing_verity=$1
 shift
 seal_state=$1
 shift
+# EROFS format version to pass to bootc container ukify (optional, default: v2)
+erofs_version=${1:-v2}
```
Johan-Liebert1: Kind of conflicts with https://github.com/composefs/composefs-rs/pull/330. We'd probably want to have the same defaults everywhere

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -783,6 +785,12 @@ struct UKIInfo {
     version: Option<String>,
     os_id: Option<String>,
     boot_digest: String,
+    /// The composefs image digest parsed from (and validated against) the UKI's
+    /// own cmdline.  This is the authoritative deployment key for UKI boots:
```
Johan-Liebert1: This is for every boot right, not just UKIs?

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -811,8 +820,11 @@ fn write_pe_to_esp(
     if matches!(pe_type, PEType::Uki) {
         let cmdline = uki::get_cmdline_buffered(&mut uki_reader).context("Getting UKI cmdline")?;
 
-        let (composefs_cmdline, missing_verity_allowed_cmdline) =
-            get_cmdline_composefs::<Sha512HashValue>(&cmdline).context("Parsing composefs=")?;
+        let composefs_info = BootComposefsCmdline::<Sha512HashValue>::from_cmdline(&cmdline)
+            .context("Parsing composefs=")?
+            .ok_or_else(|| anyhow::anyhow!("No composefs= or composefs.digest.v1= karg found in UKI cmdline"))?;
+        let composefs_cmdline = composefs_info.digest().clone();
```
Johan-Liebert1: This name is a bit confusing. afaiu this is only the digest and not the entire cmdline?

tmt/tests/Dockerfile.upgrade
```diff
@@ -45,7 +46,7 @@ if [ "${filesystem}" = "xfs" ]; then
 fi
 
 if test "${boot_type}" = "uki"; then
-  /run/packaging/seal-uki /run/target /out /run/secrets "${allow_missing_verity}" "${seal_state}"
+  /run/packaging/seal-uki /run/target /out /run/secrets "${allow_missing_verity}" "${seal_state}" "${erofs_version}"
```
Johan-Liebert1: We also need this in `tmt/tests/booted/test-install-to-filesystem-var-mount.sh`

crates/lib/src/bootc_composefs/update.rs
```diff
@@ -264,56 +300,83 @@ pub(crate) async fn do_upgrade(
...
+        .into_iter()
+        .flatten()
+        .collect();
+    let boot_ids: Vec<&Sha512HashValue> = boot_ids_owned.iter().collect();
+
+    // The deployment key must equal the digest in the next-boot composefs karg
+    // (see setup_composefs_boot for the full rationale).  Provisional value for
+    // BLS (where bootc writes the karg from this same id); overridden for UKI by
+    // the digest the UKI cmdline actually carries.
+    let provisional_deploy_id = boot_id_v2.clone().unwrap_or_else(|| id.clone());
```
Johan-Liebert1: `id` here is confusing especially with both `boot_id_v1/v2` defined. I believe it's the erofs digest corresponding to the erofs version that the repo is currently using?

crates/lib/src/bootc_composefs/update.rs
```diff
@@ -264,56 +300,83 @@ pub(crate) async fn do_upgrade(
...
+            &provisional_deploy_id,
+            &boot_ids,
             entries,
         )?,
     };
 
+    // Authoritative collision check against the final deploy key.  For UKI this
+    // may differ from the provisional checked above (the UKI may carry a
+    // non-default digest), so this is the load-bearing guarantee.
+    ensure_no_deploy_collision(host, &deploy_id)?;
```
Johan-Liebert1, line 378: why do we need to do this again?

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -1164,17 +1187,30 @@ pub(crate) fn setup_composefs_uki_boot(
 
     let boot_digest = uki_info.boot_digest.clone();
 
+    // The deploy key for a UKI boot is the composefs digest baked into the UKI
+    // cmdline (already validated against `boot_ids` in `write_pe_to_esp`).
+    // setup-root opens `state/deploy/<this>` using that same karg, so we must
+    // key the deployment off exactly this value -- whether the UKI was sealed
+    // with the V2 (default) or V1 EROFS digest.
+    let deploy_id = uki_info.composefs_cmdline.clone();
```
Johan-Liebert1: we shouldn't need to clone this?

Johan-Liebert1 (changes requested):
Lots of `.clone()` that I believe shouldn't be needed, I might be wrong though. Supporting both V1 and V2 are great, but it is a bit messy, not sure if there's a better way to do this. Also, some inconsistencies here and there (esp in comments) regarding whether V1 is the default or V2

cgwalters:
In the general case we may need to add support for "older bootc version" which includes not just the v2 digest but the xattr filtering logic too? See https://github.com/composefs/composefs-rs/pull/337

cgwalters:
OK, this one wants  https://github.com/bootc-dev/bootc/pull/2290 to land first which fixes our composefs mounts on c9s.

cgwalters:
> OK, this one wants https://github.com/bootc-dev/bootc/pull/2290 to land first which fixes our composefs mounts on c9s.

OK! Finally that landed. I rebased this, but it's still somewhat raw, especially the last commit.

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -1112,6 +1197,216 @@ fn write_pe_to_esp(
...
 }
 
+/// Reject an insecure UKI before any persistent ESP state is created when the
+/// repository is configured to require fs-verity.  The repository policy is
+/// authoritative here: an image's `composefs=?` marker is only usable when
+/// the repository itself was opened with the explicit missing-verity option.
+fn validate_uki_fsverity_policy(
+    repo_requires_fsverity: bool,
+    missing_fsverity_allowed: bool,
+    uki_allows_missing_fsverity: bool,
```
cgwalters: Too many `bool` here...I think we may need an `enum`

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -1112,6 +1197,216 @@ fn write_pe_to_esp(
...
+            "Use --allow-missing-fsverity only when missing fs-verity is explicitly supported for this install."
+        };
+        anyhow::bail!(
+            "The UKI requests insecure composefs operation, but this repository requires fs-verity. {option_hint}"
+        );
+    }
+    Ok(())
+}
+
+/// Validate every primary UKI before any bootloader or ESP operation.  The
```
cgwalters, line 1257: Hmmm I'm not sure, I think we need to be robust to corrupted state and allow recovery. Bailing early hurts that.

This seems more like a `bootc internals fsck` style thing.

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -1112,6 +1197,216 @@ fn write_pe_to_esp(
...
+        };
+        let cmdline = uki::get_cmdline_buffered(&mut uki_reader).context("Getting UKI cmdline")?;
+        let composefs_info = ComposefsBootCmdline::<Sha512HashValue>::from_cmdline(&cmdline)
+            .context("Parsing composefs=")?
+            .ok_or_else(|| anyhow::anyhow!("No composefs image in UKI cmdline"))?;
+        let uki_name = entry
+            .file_path
+            .file_name()
+            .ok_or_else(|| anyhow!("Could not get UKI file name"))?
+            .to_string_lossy()
```
cgwalters: No. We should require UTF-8.

crates/lib/src/ukify.rs
```diff
@@ -19,8 +20,112 @@ use composefs_ctl::composefs;
 
 use crate::bootc_composefs::digest::compute_composefs_digest;
 use crate::bootc_composefs::status::build_composefs_karg;
+use crate::cli::ErofsVersionArg;
 use crate::kernel::KernelInternal;
 
+const COMPOSEFS_DIGEST_V1_FEATURE: &str = "/usr/lib/bootc/initramfs-features/composefs-digest-v1";
+const COMPOSEFS_DIGEST_V1_FEATURE_CONTENT: &[u8] = b"composefs-digest-v1 state-v1\n";
+
+/// Query `lsinitrd` without unpacking or executing any initramfs contents.
```
cgwalters: No, this is awful. I don't want `ukify` to be parsing the initramfs.

Among other things I don't want to hard depend on `dracut`.

I think we should just default to injecting *both* EROFS kargs right?

Johan-Liebert1: If we just want the cmdline, we have a function in composefs-rs for that. `get_uki_cmdline_buffered`

crates/initramfs/src/lib.rs
Johan-Liebert1, line 1: Minor thing here, but this is now definitely out of sync with https://github.com/composefs/composefs-rs/blob/main/crates/composefs-setup-root/src/main.rs

Not sure if we'd want to keep them in sync

crates/initramfs/src/lib.rs
```diff
@@ -353,15 +354,89 @@ pub fn mount_composefs_image(
...
+    .context("Creating filesystem mount")?;
 
     set_mount_readonly(&rootfs)?;
 
     Ok(rootfs)
 }
 
+fn parse_composefs_candidates(cmdline: &str) -> Result<Vec<ComposefsCmdline<Sha512HashValue>>> {
+    let mut candidates = Vec::new();
+    for token in split_cmdline(cmdline) {
```
Johan-Liebert1: We should really be using the kernel-cmdline crate here

crates/initramfs/src/lib.rs
```diff
@@ -353,15 +354,89 @@ pub fn mount_composefs_image(
...
+        .map(|candidate| candidate.digest().to_hex())
+        .collect::<Vec<_>>()
+        .join(", ");
+    anyhow::bail!("no composefs image found (tried: {tried})")
+}
+
+fn mount_composefs_candidate(
+    sysroot: &OwnedFd,
+    candidate: &ComposefsCmdline<Sha512HashValue>,
+    allow_missing_fsverity: bool,
```
Johan-Liebert1, line 444: I think it's worth documenting that `allow_missing_verity` is coming from the repo and not the cmdline

crates/initramfs/src/lib.rs
```diff
@@ -353,15 +354,89 @@ pub fn mount_composefs_image(
...
 
     set_mount_readonly(&rootfs)?;
 
     Ok(rootfs)
 }
 
+fn parse_composefs_candidates(cmdline: &str) -> Result<Vec<ComposefsCmdline<Sha512HashValue>>> {
+    let mut candidates = Vec::new();
+    for token in split_cmdline(cmdline) {
+        if token.starts_with(&format!("{KARG_COMPOSEFS_DIGEST}="))
```
Johan-Liebert1: I think we should only allow one of each, as in at max `composefs.digest=abc123 composefs=a1b2c3`. I don't think we should allow multiple of `composefs.digest=` or `composefs=` in the kernel cmdline

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -657,6 +715,15 @@ struct BLSEntryPath {
     config_path: Utf8PathBuf,
 }
 
+/// Replace either karg spelling to ensure only the selected EROFS format remains.
+fn replace_composefs_karg(cmdline: &mut Cmdline, new_karg: &str) -> Result<()> {
+    cmdline.remove(&ParameterKey::from(COMPOSEFS_CMDLINE));
+    cmdline.remove(&ParameterKey::from(COMPOSEFS_DIGEST_CMDLINE));
```
Johan-Liebert1: In `initrarmfs/src/lib.rs` these are imported from composefs-rs under the names `KARG_V2` and `KARG_COMPOSEFS_DIGEST` respectively. We should just use one import location, or at least have the same names for constants

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -1112,6 +1197,216 @@ fn write_pe_to_esp(
...
+        }
+        composefs_info
+            .validate_digest(boot_ids)
+            .context("Validating UKI composefs digest")?;
+    }
+    Ok(())
+}
+
+/// Scans `entries` for the primary UKI (`PEType::Uki`, not an addon) and
+/// extracts the `composefs=` digest embedded in its kernel cmdline.
```
Johan-Liebert1, line 1356: or the `composefs.digest=` digest

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -1112,6 +1197,216 @@ fn write_pe_to_esp(
...
+                std::fs::File::from(repo.open_object(id)?)
+            }
+            RegularFile::Inline(..) | RegularFile::Sparse(..) => {
+                anyhow::bail!("UKI file is not a regular external object")
+            }
+        };
+        let cmdline = uki::get_cmdline_buffered(&mut uki_reader).context("Getting UKI cmdline")?;
+        let composefs_info = ComposefsBootCmdline::<Sha512HashValue>::from_cmdline(&cmdline)
+            .context("Parsing composefs=")?
+            .ok_or_else(|| anyhow::anyhow!("No composefs image in UKI cmdline"))?;
```
Johan-Liebert1:
```suggestion
            .ok_or_else(|| anyhow::anyhow!("No composefs digest in UKI cmdline"))?;
```

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -1616,12 +1924,48 @@ pub(crate) async fn setup_composefs_boot(
...
         &pull_result.config_digest,
         Some(&pull_result.config_verity),
-        &Default::default(),
+        &composefs_oci::OciTransformOptions::default(),
     )
     .context("Creating composefs filesystem for boot entry discovery")?;
     let entries =
         get_boot_resources(&fs, &*repo).context("Extracting boot entries from OCI image")?;
 
+    // If the UKI was built by tooling using a different xattr filtering
```
Johan-Liebert1: We already do this when pulling the repo. Why again?

Johan-Liebert1: Also, these functions have nothing to do with "boot" itself. They should be in a separate file, like `boot_utils.rs` or something

crates/lib/src/bootc_composefs/status.rs
```diff
@@ -88,24 +91,30 @@ impl ComposefsCmdline {
         }
     }
 
-    pub(crate) fn build(digest: &str, allow_missing_fsverity: bool) -> Self {
-        ComposefsCmdline {
-            allow_missing_fsverity,
-            digest: digest.into(),
+    /// Search for either supported composefs kernel command line parameter.
+    pub(crate) fn find_in_cmdline(cmdline: &Cmdline) -> Option<Self> {
+        let parsed = BootComposefsCmdline::<Sha512HashValue>::from_cmdline(cmdline).ok()??;
```
Johan-Liebert1: Not a fan of `??` without any `context`

tmt/tests/booted/readonly/046-test-erofs-version.nu
```diff
@@ -0,0 +1,72 @@
+use std assert
+use tap.nu
+
+tap begin "verify composefs UKI EROFS version boots correctly"
+
+let is_composefs = (tap is_composefs)
+
+if not $is_composefs {
```
Johan-Liebert1, line 11: can use `skip_if_ostree`

tmt/tests/booted/test-install-to-filesystem-var-mount.sh
Johan-Liebert1, line 1: This test specifically is failing with

```
content: error: Installing to disk: Setting up composefs boot: The UKI requests insecure composefs operation, but this repository requires fs-verity. Use --allow-missing-fsverity only when missing fs-verity is explicitly supported for this install.
content: Connection to localhost closed.
```

Johan-Liebert1: I guess in the install path `--allow-missing-verity` is not being respected?

crates/lib/src/bootc_composefs/status.rs
cgwalters: Yeah it's questionable so to speak

crates/lib/src/deploy.rs
```diff
@@ -449,6 +450,21 @@ fn check_disk_space_inner(
     Ok(())
 }
 
+/// Reserve room for the compressed OCI staging copy, the private inspection
+/// repository, and the final composefs import before downloading any layers.
+pub(crate) fn check_disk_space_for_composefs_staging(
```
Johan-Liebert1: We already have `check_disk_space_composefs`. Does this do anything different?

crates/lib/src/install.rs
```diff
@@ -2022,45 +2057,108 @@ async fn install_to_filesystem_impl(
...
+            let img_manifest_config = match source_manifest {
+                Some(manifest) => manifest,
+                None => get_container_manifest_and_config(&fetch_ref).await?,
+            };
             crate::store::ensure_composefs_dir(&rootfs.physical_root)?;
-            // Use init_path since the repo may not exist yet during install
+            // Use init_path since the repo may not exist yet during install.
+            // Generate both V1 and V2 EROFS images (see initialize_composefs_repository);
+            // this config must match the one used there since it re-inits the same repo.
+            let allow_missing_fsverity = allow_missing_verity;
```
Johan-Liebert1: why the reassign?

crates/lib/src/install.rs
```diff
@@ -2022,45 +2057,108 @@ async fn install_to_filesystem_impl(
...
+                crate::bootc_composefs::repo::stage_external_source(
+                    &source,
+                    rootfs.physical_root_path.as_std_path(),
+                    state.source.digest.as_deref(),
+                )
+                .await?,
+            )
+        };
+        let staged_policy = if let Some(staged) = staged_source.as_ref() {
+            crate::bootc_composefs::repo::inspect_staged_uki_policy(
```
Johan-Liebert1: This creates a new repo anyway, in a tempdir sure, but this feels incredibly wasteful. We do a similar operation in `prepare_install`, which does require the source imgref to be the same as the image being installed and I guess this solves that particular issue, but I don't know if we *NEED* to do this. Especially, if the install isn't even a UKI install

crates/lib/src/install.rs
```diff
@@ -2022,45 +2057,108 @@ async fn install_to_filesystem_impl(
...
+            )?;
+            Some(manifest)
+        };
+        let staged_source = if state.source.in_host_mountns {
+            None
+        } else {
+            let source = state.source.imageref.clone();
+            let source: ostree_ext::container::ImageReference = source;
+            Some(
+                crate::bootc_composefs::repo::stage_external_source(
```
Johan-Liebert1: Why not just initialize the repository at this point? I don't see much benefit in being this thorough with UKI installs to the point where some operations feel almost wasteful

cgwalters: Yes. This is redone!

Johan-Liebert1 (dismissed):
Tested a local install and everything looks good. We now have four EROFS images per deployment (bootable-[v1,v2], non-bootable-[v1,v2]). It'd be great to have refs for them so they can be linked to a particular deployment. We do have `bootc cfs .. linked-erofs-images` which kinda does do this, so not a blocker

Johan-Liebert1:
The f44 UKI failures seem related to https://bugzilla.redhat.com/show_bug.cgi?id=2507393

I see the following in audit logs

```
type=AVC
msg=audit(1789986405.413:65):
avc: denied { create }
pid=750
comm="systemd-tpm2-se"
name=".#nvpcr-anchor.d612e3beb8f447cca6274026257a9463.cred64701887318784f4"
scontext=system_u:system_r:init_t:s0
tcontext=system_u:object_r:dosfs_t:s0
tclass=file
permissive=0
```

and the service fails with

```
[root@fedora ~]# systemctl status systemd-tpm2-setup.service
× systemd-tpm2-setup.service - TPM SRK Setup
     Loaded: loaded (/usr/lib/systemd/system/systemd-tpm2-setup.service; static)
    Drop-In: /usr/lib/systemd/system/service.d
             └─10-timeout-abort.conf
     Active: failed (Result: exit-code) since Mon 2026-09-21 10:26:45 UTC; 56s ago
 Invocation: 1ec53e897d3049278a3b2edde716b6d0
       Docs: man:systemd-tpm2-setup.service(8)
    Process: 750 ExecStart=/usr/lib/systemd/systemd-tpm2-setup --graceful (code=exited, status=1/FAILURE)
   Main PID: 750 (code=exited, status=1/FAILURE)
   Mem peak: 3M
        CPU: 29ms

Sep 21 10:26:45 fedora systemd-tpm2-setup[750]: SRK already stored in the TPM.
Sep 21 10:26:45 fedora systemd-tpm2-setup[750]: SRK fingerprint is b127294a29a4cdb1d41ce5f472ce078d9230cbc79d9867d1e84c780bb6e75c45.
Sep 21 10:26:45 fedora systemd-tpm2-setup[750]: SRK public key saved to '/var/lib/systemd/tpm2-srk-public-key.pem' in PEM format.
Sep 21 10:26:45 fedora systemd-tpm2-setup[750]: SRK public key saved to '/var/lib/systemd/tpm2-srk-public-key.tpm2b_public' in TPM2B_PUBLIC format.
Sep 21 10:26:45 fedora systemd-tpm2-setup[750]: Successfully written anchor secret to '/var/lib/systemd/nvpcr/nvpcr-anchor.cred'.
Sep 21 10:26:45 fedora systemd-tpm2-setup[750]: Failed to write anchor secret file to '/boot/loader/credentials/nvpcr-anchor.d612e3beb8f447cca6274026257a9463.cred': Permission denied
Sep 21 10:26:45 fedora systemd-tpm2-setup[750]: 2 NvPCRs already initialized.
Sep 21 10:26:45 fedora systemd[1]: systemd-tpm2-setup.service: Main process exited, code=exited, status=1/FAILURE
Sep 21 10:26:45 fedora systemd[1]: systemd-tpm2-setup.service: Failed with result 'exit-code'.
Sep 21 10:26:45 fedora systemd[1]: Failed to start systemd-tpm2-setup.service - TPM SRK Setup.
```

Johan-Liebert1 (dismissed):

cgwalters:
Thinking this one should be a reason to bump to 1.17.x!

cgwalters:
> We now have four EROFS images per deployment (bootable-[v1,v2], non-bootable-[v1,v2]). It'd be great to have refs for them so they can be linked to a particular deployment. 

All variants of this are stored in the splitstream and discoverable.

crates/lib/src/bootc_composefs/repo.rs
```diff
@@ -93,17 +95,100 @@ pub(crate) fn composefs_repository_config(allow_missing_fsverity: bool) -> Repos
...
+        );
+    }
+    Ok(())
+}
+
+/// Replace only the metadata of a repository created by this install.  The
+/// replacement is generated by composefs-rs itself, so format and feature
+/// settings cannot drift from the normal initialization path.
+pub(crate) fn finalize_fresh_repository_policy(
+    rootfs: &std::path::Path,
```
cgwalters: Ideally we use fd-relative APIs in general

crates/lib/src/bootc_composefs/repo.rs
```diff
@@ -93,17 +95,100 @@ pub(crate) fn composefs_repository_config(allow_missing_fsverity: bool) -> Repos
...
+    anyhow::ensure!(
+        target_metadata == replacement_metadata,
+        "composefs metadata generated for policy replacement does not match the repository"
+    );
+
+    let replacement_meta = temp.path().join("meta.json");
+    std::fs::rename(&replacement_meta, &target_meta)
+        .context("Atomically replacing composefs metadata")?;
+    std::fs::File::open(&target_meta)?.sync_all()?;
+    std::fs::File::open(&repo_path)?.sync_all()?;
```
cgwalters: And this should use https://docs.rs/cap-std-ext/latest/cap_std_ext/dirext/trait.CapStdExtDirExt.html#tymethod.atomic_replace_with we can sync in the closure before rename

crates/initramfs/src/lib.rs
```diff
@@ -379,13 +379,36 @@ pub fn mount_composefs_image(
...
 fn parse_composefs_candidates(cmdline: &str) -> Result<Vec<ComposefsCmdline<Sha512HashValue>>> {
     let mut candidates = Vec::new();
-    for token in split_cmdline(cmdline) {
-        if token.starts_with(&format!("{KARG_COMPOSEFS_DIGEST}="))
-            || token.starts_with(&format!("{KARG_V2}="))
-        {
-            if let Some(candidate) = ComposefsCmdline::<Sha512HashValue>::from_cmdline(token)? {
-                candidates.push(candidate);
-            }
+    let mut seen_digest = false;
```
cgwalters: Would be cleaner as `mut Option<&>`

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -1037,6 +1032,136 @@ struct UKIInfo {
     composefs_digest: Sha512HashValue,
 }
 
+/// Parse every composefs argument emitted by supported bootc UKIs.  Do not use
+/// `ComposefsCmdline::from_cmdline()` for this: it deliberately returns one
```
cgwalters: Hmm. Maybe we should just extend that API to say which has priority?

crates/lib/src/bootc_composefs/repo.rs
```diff
@@ -112,7 +197,9 @@ pub(crate) async fn initialize_composefs_repository(
     root_setup: &RootSetup,
     allow_missing_fsverity: bool,
     use_unified: bool,
-) -> Result<PullResult<Sha512HashValue>> {
+    fetch_imgref: Option<&containers_image_proxy::ImageReference>,
+    manifest: &ostree_ext::oci_spec::image::ImageManifest,
```
cgwalters: Hmm this is now really "initialize and pull image" which I think we should keep to two cleanly separate steps still.

docs/src/man/bootc-container-ukify.8.md
```diff
@@ -31,6 +31,16 @@ Any additional arguments after `--` are passed through to ukify unchanged.
 
     Make fs-verity validation optional in case the filesystem doesn't support it
 
+**--erofs-version**=*EROFS_VERSION*
```
cgwalters, line 40: Hmm actually I'm changing my mind, I think we should actually just generate *both* digests by default in `bootc container ukify`, and then `--erofs-version=v1` drops the v2 compat. I don't see any good use case for `--erofs-version=v2` but for symmetry (reproducing an older UKI?) I guess we can support it too.

crates/lib/src/bootc_composefs/repo.rs
Johan-Liebert1: Yes, also we're doing a bunch of stuff here related to the composefs repo itself, which I think should belong in `composefs-rs` as helpers on the Repository object

Johan-Liebert1:
> The old client already parses composefs.digest= first and stages V1 directly

how would the older client parse V1 cmdline arg?

cgwalters:
> how would the older client parse V1 cmdline arg?

It just parses `composefs=` which can historically now be v2 or v1 - where did you see that text?

Johan-Liebert1:
In the copilot review

> The documented 1.16 bridge path doesn't happen for ≥1.16.4. The old client already parses composefs.digest= first and stages V1 directly; the V2 fallback is never used and the V2 image isn't created. test-49-composefs-1-16-bridge pins 1.16.0, which predates the V1 default, so it can't catch the 1.16.4+ behavior.

> It just parses composefs= which can historically now be v2 or v1

okay, kinda makes sense then

Johan-Liebert1 (approved):
Looks good to me now. @cgwalters I've disabled auto merge in case you want to change something here

PR #2249, opened 2026-06-12
switch: fix --apply not rebooting when already staged

jlebon:
When `bootc switch --apply` is called targeting an image that is already
staged, the spec comparison short-circuits with 'Image specification is
unchanged' and returns early without rebooting. This is the same class
of bug that was fixed for `bootc upgrade --apply` in commit 0067a47c.

Fix both the ostree and composefs switch paths.

Full AI use (reproduced bug, patched bootc, verified bugfix). I verified
the verification and the patch.

### DIFF
--- a/tmt/tests/booted/test-image-upgrade-reboot.nu
+++ b/tmt/tests/booted/test-image-upgrade-reboot.nu
@@ -8,7 +8,8 @@
 # This test does:
 # bootc image copy-to-storage
 # podman build <from that image>
-# bootc switch <into that image> --apply
+# bootc switch <into that image> (stage only)
+# bootc switch --apply <same image> (spec unchanged, must still reboot)
 # Verify we boot into the new image
 #
 # For composefs builds, it additionally verifies that composefs is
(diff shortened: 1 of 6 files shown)


tmt/tests/booted/test-image-upgrade-reboot.nu
```diff
@@ -65,10 +66,13 @@ def initial_build [] {
         $st.status.booted.composefs.verity | save /var/original-verity
     }
 
-    # Now, switch into the new image
-    print $"Applying ($imgsrc)"
+    # Now, switch into the new image. First stage it, then run switch --apply for
+    # the same target. This exercises the case where --apply must still reboot for
+    # staged deployments.
+    print $"Staging ($imgsrc)"
```
jlebon, line 72: We're technically losing coverage here on doing a one-shot `switch --apply` which both had to stage and reboot together.

It didn't feel worth adding another reboot to test this since it's trivial (while the early return reboot into staged here is a more subtle gotcha), but happy to do so if folks prefer.

Johan-Liebert1 (dismissed):

jlebon:
Anything else on this one?

cgwalters (dismissed):

jlebon:
OK there seems to be an inconsistency between the composefs and ostree backends. In ostree, the `host.spec.image` is considered to be the staged image first with fallback to the booted image. Whereas in the composefs backend, it always uses the booted image. I assume that's an oversight?

jlebon:
> In ostree, the `host.spec.image` is considered to be the staged image first with fallback to the booted image. Whereas in the composefs backend, it always uses the booted image.

Pushed a commit to fix the composefs backend now.

composefs passes locally now but would appreciate if someone could add the `ci/merge` label.

cgwalters (approved):

cgwalters:
> OK there seems to be an inconsistency between the composefs and ostree backends. In ostree, the host.spec.image is considered to be the staged image first with fallback to the booted image. Whereas in the composefs backend, it always uses the booted image. I assume that's an oversight?

Yeah, thanks for looking at this and finding it!

PR #2253, opened 2026-06-14
sd-boot: Only use --random-seed flag on systemd >= 257

yeetypete:
https://github.com/bootc-dev/bootc/pull/2134 introduced passing the `--random-seed` flag to `bootctl install`. However, this flag is only available on systemd >= 257.

Since it would be great if the latest bootc version would work for older systemd versions (e.g. 255 which comes in ubuntu 24.04), this PR adds some logic to query the bootctl version and gate the addition of this flag only on systemd versions >= 257. Some tests are also added for parsing `bootctl --version` output.

### DIFF
--- a/crates/lib/src/bootloader.rs
+++ b/crates/lib/src/bootloader.rs
@@ -33,6 +33,10 @@ const SYSTEMD_KEY_DIR: &str = "loader/keys";
 /// bootc does not use the entry-token at all.
 const KERNEL_INSTALL_CONF_ROOT: &str = "/tmp";
 
+/// First systemd release whose `bootctl install` accepts `--random-seed`.
+/// See: <https://www.freedesktop.org/software/systemd/man/latest/bootctl.html>
+const BOOTCTL_RANDOM_SEED_MIN_VERSION: u32 = 257;
+
 /// Mount the first ESP found among backing devices at /boot/efi.
 ///
 /// This is used by the install-alongside path to clean stale bootloader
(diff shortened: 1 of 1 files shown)


crates/lib/src/bootloader.rs
```diff
@@ -312,6 +321,28 @@ pub(crate) fn install_systemd_boot(
     Ok(())
 }
 
+/// First systemd release whose `bootctl install` accepts `--random-seed`.
+/// See: <https://www.freedesktop.org/software/systemd/man/latest/bootctl.html>
+const BOOTCTL_RANDOM_SEED_MIN_VERSION: u32 = 257;
+
+#[context("Querying bootctl version")]
+fn bootctl_systemd_version() -> Result<u32> {
```
Johan-Liebert1, line 329: This checks the systemd version in the source image, which might be different than the target image. At this point, we'd have already generated the EROFS. So I think it'd be better to chroot into the mounted EROFS and then get the version

crates/lib/src/bootloader.rs
```diff
@@ -312,6 +321,28 @@ pub(crate) fn install_systemd_boot(
     Ok(())
 }
 
+/// First systemd release whose `bootctl install` accepts `--random-seed`.
+/// See: <https://www.freedesktop.org/software/systemd/man/latest/bootctl.html>
+const BOOTCTL_RANDOM_SEED_MIN_VERSION: u32 = 257;
```
Johan-Liebert1: Minor nit. Even though this is fine, prefer defining variables before they're used

Johan-Liebert1 (changes requested):

crates/lib/src/bootloader.rs
yeetypete: Moved this up to the top with the other consts

crates/lib/src/bootloader.rs
yeetypete: @Johan-Liebert1 thanks for the review. Currently `bootctl install` runs using the the version in the source image. Should we change this to use the target image version as well? Otherwise we end up gating the flag for the wrong systemd version (we gate for the target when the source version is used).

In my case these were the same because I was running `bootc install to-disk` inside the container I was deploying but I guess there are probably some implications I'm not aware of with this change.

Johan-Liebert1: Oh yeah, in the ostree case we chroot into the new deployment and invoke bootupd from there. I think that should be a separate effort, unrealted to this PR

Johan-Liebert1 (approved):
LGTM. We definitely want https://github.com/bootc-dev/bootc/pull/1816 for composefs-backend as well

PR #2257, opened 2026-06-17
docs: Revamp image requirements

### LINKED ISSUE #2237: Base image layout requirements for installations using `--composefs-backend`
The [documentation for the composefs backend](https://bootc.dev/bootc/experimental-composefs.html) doesn't explain how to structure an image that will be compatible with bootc to install an image using said backend (instead of ostree). There should be some description of what needs to land where (ex: kernel images) in the image to ensure bootc can install it, and it boots.

[The lone example](https://bootc.dev/bootc/experimental-composefs.html#build-pattern-compute-digest-and-generate-uki-in-one-stage) in the composefs docs does not appear to describe a base image build. Though, it does appear the image layout is likely not entirely the same as ostree. A [script from bootcrew](https://github.com/bootcrew/mono/blob/main/shared/bootc-rootfs.sh) might be the best suggestion yet at what is required to build a compatible image.

Documentation should be available to enable early base image build trials with this new backend.

cgwalters:
Closes: https://github.com/bootc-dev/bootc/issues/2237

### DIFF
--- a/docs/src/bootc-images.md
+++ b/docs/src/bootc-images.md
@@ -7,32 +7,50 @@ activity.
 However, there are a number of basic requirements and integration
 points, some of which have distribution-specific variants.
 
-Further at the current time, the bootc project makes a lot
-of use of ostree, and this can appear in the base image
-requirements.
+## Generic requirements (composefs or ostree backends)
 
-## ostree-in-container
+### `/sysroot`
 
-With [bootc 1.1.3](https://github.com/bootc-dev/bootc/releases/tag/v1.1.3)
-or later, it is no longer required to have a `/ostree` directory
-present in the base image.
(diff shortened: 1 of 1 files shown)


jmpolom:
See [comments](https://github.com/bootc-dev/bootc/issues/2237#issuecomment-4736874391) about additional content needed for this to be complete.

cgwalters:
> See https://github.com/bootc-dev/bootc/issues/2237#issuecomment-4736874391 about additional content needed for this to be complete.

I don't want this PR to snowball into "all the issues" - I think what we have here is a clear and isolated "container image filesystem and layout" improvement. We'll absolutely track and fix the other (related) issues, thanks!

ckyrouac (approved):

PR #2264, opened 2026-06-22
ostree-ext: Adapt to new containers policy config load order

### LINKED ISSUE #2258: Adapt to new containers policy config load order
The config file load order for the policy has been updated in https://github.com/podman-container-tools/container-libs/pull/711 & https://github.com/podman-container-tools/container-libs/pull/795.

From https://github.com/podman-container-tools/container-libs/blob/main/image/docs/containers-policy.json.5.md:

> By default, the policy is read from $XDG_CONFIG_HOME/containers/policy.json (or from $HOME/.config/containers/policy.json if $XDG_CONFIG_HOME is unset), if it exists; otherwise from /etc/containers/policy.json; otherwise from /usr/share/containers/policy.json. Applications performing verification may allow using a different policy instead.
>
> If CONTAINERS_POLICY_JSON is set, it specifies the policy file to use, unless overridden by application-specific configuration.

This has just landed in Rawhide (https://bodhi.fedoraproject.org/updates/FEDORA-2026-2419096432) and the CoreOS tests failed there but the update was pushed.

We should thus update the bootc code following that change.

---

From https://github.com/coreos/fedora-coreos-config/pull/4220 / https://github.com/coreos/fedora-coreos-tracker/issues/2171

cgwalters:
Closes: https://github.com/bootc-dev/bootc/issues/2258

The real fix here will be to migrate to `--require-signed` from skopeo in the future.

Generated-by: AI (with only a few tweaks)

### DIFF
--- a/crates/ostree-ext/src/container/skopeo.rs
+++ b/crates/ostree-ext/src/container/skopeo.rs
@@ -2,6 +2,7 @@
 
 use super::ImageReference;
 use anyhow::{Context, Result};
+use cap_std_ext::cap_std;
 use cap_std_ext::cmdext::{CapStdExtCommandExt, CmdFds};
 use containers_image_proxy::oci_spec::image as oci_image;
 use fn_error_context::context;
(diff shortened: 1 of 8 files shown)


travier (approved):
LGTM

travier:
CC @Johan-Liebert1

crates/ostree-ext/src/container/skopeo.rs
```diff
@@ -17,9 +17,83 @@ use tokio::process::Command;
...
+    )
+}
+
+/// Inner helper that is fully parameterised so it can be unit-tested without
+/// touching the real environment or filesystem.
+fn resolve_policy_path(
+    env_override: Option<&Path>,
+    xdg_config_home: Option<&Path>,
+    home: Option<&Path>,
+    exists: impl Fn(&Path) -> bool,
```
Johan-Liebert1: This does not need to be a function

crates/ostree-ext/src/container/skopeo.rs
```diff
@@ -17,9 +17,83 @@ use tokio::process::Command;
...
+        if exists(&p) {
+            tracing::debug!("Using user policy path: {}", p.display());
+            return Ok(p);
+        }
+    }
+
+    // 3–4. System paths.
+    for &candidate in SYSTEM_POLICY_PATHS {
+        let p = Path::new(candidate);
+        if exists(p) {
```
Johan-Liebert1: We can just call p.exists() instead of having a closure here

Johan-Liebert1 (changes requested):

crates/ostree-ext/src/container/skopeo.rs
cgwalters: It was needed because the unit test fixture replaced it.

However, I reworked this to create a tempdir instead, which is a lot cleaner.

Johan-Liebert1 (approved):

PR #2266, opened 2026-06-24
[TEST] test https://github.com/composefs/composefs-rs/pull/321

giuseppe:


giuseppe:
@cgwalters is this test what you had in mind?

Dockerfile
```diff
@@ -180,6 +180,19 @@ ENV SOURCE_DATE_EPOCH=${SOURCE_DATE_EPOCH}
...
+# This replaces the C libcomposefs.so in the target container.
+FROM buildroot as composefs-capi
+RUN --mount=type=tmpfs,target=/run --mount=type=tmpfs,target=/tmp \
+    --mount=type=cache,target=/composefs-rs/target --mount=type=cache,target=/var/roothome <<EORUN
+set -xeuo pipefail
+git clone --depth=1 -b cdynlib https://github.com/giuseppe/composefs-rs.git /composefs-rs/src
+cd /composefs-rs/src
+cargo build --release -p composefs-capi
+mkdir -p /out
+cp target/release/libcomposefs_capi.so /out/
```
cgwalters: Would be better to have e.g. `make install DESTDIR=` support. Maybe it should be `make install-capi`.

cgwalters (commented):
Yes, doing it this way runs more tests, but there is also support for revdep testing in composefs-rs that we could wire things up to directly as well

cgwalters:
For PRs like this, add the `ci/merge` label to trigger all of the tests

giuseppe:
@cgwalters the failures do not seem related

cgwalters:
Yeah the quay.io flakes on the build phase stopped tests from even running, retried those.

Issue #2268, opened 2026-06-26
GrubCC does not support 'uki' key in BLSConfig


Johan-Liebert1:
Boot fails with

```
BdsDxe: loading Boot0004 "Fedora" from HD(2,GPT,BF4F367A-79D1-4FFB-9C90-BCD9BC70F532,0x1000,0x200000)/\EFI\fedora\shimx64.efi
BdsDxe: starting Boot0004 "Fedora" from HD(2,GPT,BF4F367A-79D1-4FFB-9C90-BCD9BC70F532,0x1000,0x200000)/\EFI\fedGRUB version 2.12

   Minimal BASH-like line editing is supported. For the first word, TAB lists possible command completions. Anywhere else TAB lists possible device or file completions. To enable less(1)-like paging, "set pager=1".                    


grub> 
```

GrubCC version `1:2.12-60.eln156`

PR #2269, opened 2026-06-26
Support `uki` key in BLSConfig

### LINKED ISSUE #2263: composefs: Use `uki` key for Type 1 configs pointing to a UKI
We should now use the ["new"](https://github.com/uapi-group/specifications/commit/3f2bd8236d7f9ce6dedf8bda9cadffd0d363cb08) `uki` key from the BLS spec in Type 1 configs when referencing a UKI. See: https://uapi-group.org/specifications/specs/boot_loader_specification/#type-1-boot-loader-entry-keys

To not break updates, we should still accept configs using the `efi` key but newly generated configs should use the `uki` key only.

Both systemd-boot and the grub-cc build should support that now.

---

From https://discussion.fedoraproject.org/t/f45-change-proposal-grub-efi-for-confidential-computing-self-contained/193574/7

Johan-Liebert1:
Systemd >= v258 supports 'uki' key in bls configs. Update BLSConfig
struct to reflect the same. We still support the 'efi' key for
backwards compatibility

During writing of the BLSConfig files we check for systemd version and
use the appropriate key

GrubCC does not support the 'uki' key right now
See: https://github.com/bootc-dev/bootc/issues/2268

Closes: #2263

### DIFF
--- a/crates/lib/src/bootloader.rs
+++ b/crates/lib/src/bootloader.rs
@@ -326,7 +326,7 @@ pub(crate) fn install_systemd_boot(
 }
 
 #[context("Querying bootctl version")]
-fn bootctl_systemd_version() -> Result<u32> {
+pub(crate) fn bootctl_systemd_version() -> Result<u32> {
     let out = Command::new("bootctl").arg("--version").run_get_string()?;
     parse_systemd_version(&out)
 }
--- a/crates/lib/src/bootloader.rs
+++ b/crates/lib/src/bootloader.rs
@@ -1,5 +1,6 @@
 use std::fs::create_dir_all;
 use std::process::Command;
+use std::sync::OnceLock;
 
 use anyhow::{Context, Result, anyhow, bail};
 use bootc_utils::{ChrootCmd, CommandRunExt};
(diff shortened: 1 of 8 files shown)


crates/lib/src/bootloader.rs
```diff
@@ -326,7 +326,7 @@ pub(crate) fn install_systemd_boot(
 }
 
 #[context("Querying bootctl version")]
-fn bootctl_systemd_version() -> Result<u32> {
+pub(crate) fn bootctl_systemd_version() -> Result<u32> {
```
cgwalters, line 330: This isn't new, but since we're using it more...

One issue here is that this won't do the right thing in a "sysroot" style install. We may need to do another explicit chroot style operation in the install logic.

Second, since we query this a few times, we should cache the result in a OnceLock or so

cgwalters (dismissed):

crates/lib/src/bootloader.rs
Johan-Liebert1: > One issue here is that this won't do the right thing in a "sysroot" style install. We may need to do another explicit chroot style operation in the install logic.

tried doing that, but bootupd fails trying to look up block device for composefs root mount. I have a fix in https://github.com/coreos/bootupd/pull/1113 which I can separate out in another PR


> Second, since we query this a few times, we should cache the result in a OnceLock or so

Yup. Sounds good

Johan-Liebert1: Updated caching version result in OnceLock. The bootupd install stuff will need some time as we  need a bootupd release, have it in the image etc

cgwalters (approved):

PR #2273, opened 2026-06-30
xtask/tmt: Abort run and dump diagnostics on VM launch failure

cgwalters:
Disk space is one of our flakes. We should not attempt to continue if bcvk fails to launch a VM. If that happens, let's gather disk usage instead.

### DIFF
--- a/crates/xtask/src/tmt.rs
+++ b/crates/xtask/src/tmt.rs
@@ -130,6 +130,48 @@ fn distro_supports_bind_storage_ro(distro: &str) -> bool {
     !distro.starts_with(DISTRO_CENTOS_9)
 }
 
+/// Collect and print diagnostics useful for understanding host disk-space
+/// exhaustion. This is invoked when launching a VM fails, which we treat as an
+/// infrastructure failure (as opposed to a test failure). The most common cause
+/// of such failures in CI is the host filesystem running out of space, so we
+/// dump what is consuming it: container images, libvirt VMs/disks, the tmt log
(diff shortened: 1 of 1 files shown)


crates/xtask/src/tmt.rs
```diff
@@ -130,6 +130,48 @@ fn distro_supports_bind_storage_ro(distro: &str) -> bool {
...
+    // Broad view of the runner's home directory, where caches, the tmt
+    // workdir, and libvirt storage frequently accumulate. Use `du --max-depth=1`
+    // on $HOME directly rather than a `$HOME/*` glob: the latter silently skips
+    // hidden directories like ~/.cache and ~/.local, which is exactly where the
+    // worst offenders tend to live. Pipe through `sort -h` for readability;
+    // since xshell does not provide pipes, run it through a shell.
+    if let Ok(home) = std::env::var("HOME") {
+        println!("\n--- du -h --max-depth=1 {home} (sorted) ---");
+        let script = "du -h --max-depth=1 \"$1\" 2>/dev/null | sort -h";
+        let _ = cmd!(sh, "sh -c {script} sh {home}").ignore_status().run();
```
Johan-Liebert1, line 169:
```suggestion
        let _ = cmd!(sh, "'sh -c {script}' sh {home}").ignore_status().run();
```

Johan-Liebert1: nvm, seems like xshell handles this well

Johan-Liebert1 (approved):

PR #2274, opened 2026-06-30
ci: Don't let heavy jobs block merge-queue entry in PRs

cgwalters:
The goal of the `ci/merge` label is to test things on PRs, but in case of flakes, we want to allow adding to the merge queue and let retries happen there.

This updated pattern matches https://github.com/cgwalters/merge-queue-like-bors/commit/025748f129c0cb0b20f842aada0ab3a4e061b09b

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -558,14 +558,36 @@ jobs:
           name: container-export-test-${{ env.ARCH }}
           path: target/anaconda-test/*.log
 
-  # Sentinel job for required checks - configure this job name in repository settings.
-  # Accepts 'skipped' as success so that merge_group-only jobs don't block PRs.
-  required-checks:
-    if: always()
+  # Runs only in merge_group / workflow_dispatch and gates the full heavy suite.
+  # Skipped on PRs, so required-checks (below) sees it as 'skipped' == success there.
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -574,3 +576,23 @@ jobs:
...
+  required-checks:
+    if: always()
+    needs: [compute-ci-level, validate, cargo-deny, docs, required-checks-merge]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check required jobs
+        env:
+          NEEDS: ${{ toJson(needs) }}
+        run: |
+          FAILED=$(echo "$NEEDS" | jq -r 'to_entries[] | select(.value.result | IN("success","skipped") | not) | .key')
```
Johan-Liebert1, line 594: maybe we should check if only `required-checks-merge` was skipped and not any other ones?

cgwalters: It's kind of intentional because I think we should support skipping other jobs too. We don't do much of that yet but we could.

Johan-Liebert1 (approved):

Johan-Liebert1:
`Plan: /tmt/plans/integration/plan-32-multi-device-esp` failing. I have a fix here https://github.com/bootc-dev/bootc/pull/2200

PR #2275, opened 2026-07-02
store: Support read-only /sysroot on a live ISO

cgwalters:
We want to support at least read-only operation when mounted from a physically read-only medium (like a squashfs/EROFS for /sysroot).

Check the backing block device status, and propagate its read-only state up.

Generated-by: AI

### DIFF
--- a/crates/lib/src/store/mod.rs
+++ b/crates/lib/src/store/mod.rs
@@ -108,7 +108,7 @@ use ostree_ext::container_utils::ostree_booted;
 use ostree_ext::prelude::FileExt;
 use ostree_ext::sysroot::SysrootLock;
 use ostree_ext::{gio, ostree};
-use rustix::fs::Mode;
+use rustix::{fd::AsFd, fs::Mode, fs::StatVfsMountFlags};
 
 use composefs::fsverity::Sha512HashValue;
 use composefs::repository::RepositoryConfig;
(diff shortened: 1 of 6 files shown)


crates/lib/src/store/mod.rs
```diff
@@ -507,6 +558,16 @@ impl Storage {
         })
     }
 
+    /// Ensure the storage is writable, erroring with a clear message if the
+    /// underlying `/sysroot` is on a read-only block device (e.g. a live ISO).
+    pub(crate) fn require_writable(&self) -> Result<()> {
+        anyhow::ensure!(
+            !self.is_ro,
+            "Cannot perform this operation: /sysroot is on a read-only block device (e.g. a live ISO)"
```
Johan-Liebert1, line 566: Nit: This one just feels like the LLM taking things too literally

```suggestion
            "Cannot perform this operation: /sysroot is on a read-only block device"
```

crates/lib/src/store/mod.rs
```diff
@@ -302,15 +302,51 @@ pub(crate) enum BootedStorageKind<'a> {
...
+    }
+
+    // The backing block device is physically read-only (e.g. a live ISO); there
+    // is no point attempting a remount, and write operations are not possible.
+    if matches!(bootc_blockdev::is_dir_backing_device_ro(d), Ok(Some(true))) {
+        return Ok(true);
+    }
+
+    // The mount is read-only but the device is writable: this is the normal
+    // ostree case where /sysroot is mounted read-only by default. Remount it
```
Johan-Liebert1, line 346: Here and a few places else: we shouldn't only be mentioning ostree as the composefs backend also does the same

crates/lib/src/store/mod.rs
```diff
@@ -302,15 +302,51 @@ pub(crate) enum BootedStorageKind<'a> {
...
+///
+/// Note that ostree mounts `/sysroot` read-only by default even on writable
+/// systems and remounts it writable on demand, so being read-only is not by
+/// itself meaningful: we only conclude the system is read-only if a remount
+/// attempt *fails* and the backing block device confirms it.
+fn get_physical_root_and_run() -> Result<(Dir, Dir, bool)> {
+    let d = Dir::open_ambient_dir("/sysroot", cap_std::ambient_authority())
+        .context("Opening /sysroot")?;
+    let is_ro = sysroot_is_read_only(&d)?;
+    let physical_root = d.open_dir(".").context("Opening /sysroot")?;
```
Johan-Liebert1, line 317: Do we need to reopen sysroot here?

Johan-Liebert1 (changes requested):

cgwalters:
OK this one should just need a re-stamp!

Johan-Liebert1 (approved):

PR #2276, opened 2026-07-03
initramfs: tolerate pre-existing /run/systemd/volatile-root symlink

ricardosalveti:
ostree-prepare-root is growing the same gpt-auto workaround for composefs roots (creating the /run/systemd/volatile-root symlink so systemd's blockdev_get_root() can resolve the backing device, see https://github.com/systemd/systemd/issues/35017), proposed in https://github.com/ostreedev/ostree/pull/3608. It runs before bootc-root-setup.service, so once that lands the symlink already exists by the time gpt_workaround() runs, and the unconditional symlink() call makes the service fail.

Both creators point the symlink at the /dev/block/MAJ:MIN alias of the same physical root device, so simply ignore EEXIST, matching how ensure_dir() handles already-existing directories.

### DIFF
--- a/crates/initramfs/src/lib.rs
+++ b/crates/initramfs/src/lib.rs
@@ -419,8 +419,13 @@ pub fn gpt_workaround() -> Result<()> {
         major(rootdev.st_rdev),
         minor(rootdev.st_rdev)
     );
-    symlink(target, "/run/systemd/volatile-root")?;
-    Ok(())
+    // Handle the case where the symlink was already created before bootc
+    // runs, e.g. by ostree-prepare-root
+    // (https://github.com/ostreedev/ostree/pull/3608).
+    match symlink(target, "/run/systemd/volatile-root") {
+        Ok(()) | Err(Errno::EXIST) => Ok(()),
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
Thanks!

(That said, not related to your PR but I think what we really want to do is actually add the overlayfs API to query backing layers)

crates/initramfs/src/lib.rs
```diff
@@ -419,8 +419,13 @@ pub fn gpt_workaround() -> Result<()> {
         major(rootdev.st_rdev),
         minor(rootdev.st_rdev)
     );
-    symlink(target, "/run/systemd/volatile-root")?;
-    Ok(())
+    // Handle the case where the symlink was already created before bootc
```
Johan-Liebert1, line 422: We'd also want this in https://github.com/composefs/composefs-rs/blob/main/crates/composefs-setup-root/src/main.rs#L249

ricardosalveti: Sure, created https://github.com/composefs/composefs-rs/pull/354.

ricardosalveti:
Doesn't look like the CI failures are related to this change.

PR #2277, opened 2026-07-03
Composefs changes, disable failing ostree test 

### LINKED ISSUE #90: Bootc using composefs reports the image id instead of the digest
The `bootc status --json` reports the image id instead of the image digest as it is the case when using ostree:
```json
  "status": {
    "booted": {
      "cachedUpdate": null,
      "composefs": {
[..]
      "image": {
        "architecture": "amd64",
        "image": {
          "image": "ghcr.io/bootc-dev/bink/node:v1.35-fedora-44",
          "transport": "registry"
        },
        "imageDigest": "sha256:cd5d55566e9ca3a49ea588e9d44c54f4b48e1387646c273d1f56e8ad33b8ecac",
        "timestamp": "2026-06-29T12:00:31.727356920Z",
        "version": null
      },
[..]
```
```bash
podman inspect ghcr.io/bootc-dev/bink/node:v1.35-fedora-44
[
     {
          "Id": "cd5d55566e9ca3a49ea588e9d44c54f4b48e1387646c273d1f56e8ad33b8ecac",
          "Digest": "sha256:646a1b971d9bb3f2d3f2d43c63bcb721bcdd803a8db36d1e298dc4b828ba90b2",
          "RepoTags": [
               "ghcr.io/bootc-dev/bink/node:v1.35-fedora-44"
          ],
          "RepoDigests": [
               "ghcr.io/bootc-dev/bink/node@sha256:646a1b971d9bb3f2d3f2d43c63bcb721bcdd803a8db36d1e298dc4b828ba90b2"
          ],
[..]
```
### How to reproduce
Start a bink cluster with the composefs image and verify the bootc status:
```console
bink cluster start --node-image ghcr.io/bootc-dev/bink/node:v1.35-fedora-44-disk-composefs --cluster-name composefs
bink node ssh node1 --cluster-name composefs
[root@node1 ~]# sudo bootc status --json
```

### LINKED ISSUE #2278: composefs finalize: /etc merge fails with "a path led outside of the filesystem" when a removed path is a symlink in the new deployment's /etc
## Description

`bootc composefs-finalize-staged` aborts with:

```
error: Merging: a path led outside of the filesystem
```

whenever the /etc diff's **removed** set contains a path that still exists in the *new* deployment's `/etc` as a symlink whose target resolves outside `/etc` (e.g. a `.wants` enablement symlink pointing at `/usr/lib/systemd/system/...`). Because `bootc-finalize-staged.service` runs this at shutdown, the staged deployment silently never applies — the system reboots into the old image every time.

## Root cause

In `crates/etc-merge/src/lib.rs`, the removed-files loop (currently ~line 796 on `main`):

```rust
for removed in &diff.removed {
    let stat = new_etc_fd.metadata_optional(&removed)?;
    ...
```

`cap_std` `metadata()`/`metadata_optional()` **follows symlinks**. When the removed path exists in the new etc as a symlink pointing outside the `Dir`'s root (absolute target, or relative target escaping `/etc`), cap-std rejects the resolution with "a path led outside of the filesystem" and finalize fails. `symlink_metadata` (lstat semantics) is needed here — the code immediately afterwards already handles `stat.is_symlink()` via `remove_file`, so it clearly intends not to follow.

## Reproduction (bootc 1.16.2, composefs backend, Debian trixie-based image)

1. Image ships a unit enabled via `/etc/systemd/system/multi-user.target.wants/foo.service -> /usr/lib/systemd/system/foo.service`, and the unit runs `ExecStartPost=systemctl disable %n` (self-disabling first-boot service).
2. `bootc install to-disk --composefs-backend`, boot. First boot removes the `.wants` symlink from live `/etc`.
3. `bootc switch <new-image>` (any image that still ships the symlink), reboot.
4. `bootc-finalize-staged.service` fails at shutdown with the error above; old deployment boots again. `bootc config-diff` shows the removed symlink path.

Restoring the removed symlink in the live `/etc` makes finalize succeed, confirming the diagnosis.

## Environment

- bootc 1.16.2 (also inspected `main` — `metadata_optional` still present in the removed loop)
- composefs backend (`--composefs-backend`), systemd-boot, BLS
- Debian trixie-based bootable container (frostyard/snosi)

Reviewed by Brian Ketelsen, Generated by: Claude/fable

Johan-Liebert1:
We were writing out the ImageId in output to `bootc status` instead of writing the manifest digest

Fixes: bootc-dev/bink#90

---

etc-merge: Use `symlink_metadata_optional`

cap_std's `metadata_optional` follows symlinks and errors out if the
symlink target is outside the current root

Use `symlink_metadata_optional` when checking for removed files in new
etc so that symlinks that resolve to a path outside of new_etc don't
get followed

Fixes: #2278

---

tests: Skip test-loader-entries-source for ostree v2026.2

This test ends up failing for systems with ostree v2026.2. Disable it
until we have a fix

### DIFF
--- a/crates/lib/src/bootc_composefs/status.rs
+++ b/crates/lib/src/bootc_composefs/status.rs
@@ -7,6 +7,7 @@ use composefs_ctl::composefs::fsverity::Sha512HashValue;
 use composefs_ctl::composefs_oci;
 use composefs_oci::OciImage;
 use fn_error_context::context;
+use openssl::sha::Sha256;
 use serde::{Deserialize, Serialize};
 
 use crate::{
@@ -480,7 +481,10 @@ pub(crate) fn get_bootloader() -> Result<Bootloader> {
 /// Falls back to reading legacy `.imginfo` files for backwards compatibility
 /// with deployments created before the manifest digest was stored in `.origin`.
 #[context("Reading image info for deployment {deployment_id}")]
-pub(crate) fn get_imginfo(storage: &Storage, deployment_id: &str) -> Result<ImgConfigManifest> {
+pub(crate) fn get_imginfo(
+    storage: &Storage,
+    deployment_id: &str,
+) -> Result<(ImgConfigManifest, String)> {
     let ini = read_origin(&storage.physical_root, deployment_id)?
         .ok_or_else(|| anyhow::anyhow!("No origin file for deployment {deployment_id}"))?;
 
@@ -501,7 +505,7 @@ pub(crate) fn get_imginfo(storage: &Storage, deployment_id: &str) -> Result<ImgC
             .cloned()
             .ok_or_else(|| anyhow::anyhow!("OCI image has no config (artifact?)"))?;
 
-        return Ok(ImgConfigManifest { config, manifest });
+        return Ok((ImgConfigManifest { config, manifest }, manifest_digest_str));
     }
 
     // Fallback: read legacy .imginfo file for deployments created before
@@ -528,7 +532,11 @@ pub(crate) fn get_imginfo(storage: &Storage, deployment_id: &str) -> Result<ImgC
     let img_conf = serde_json::from_str::<ImgConfigManifest>(&buffer)
         .context("Failed to parse .imginfo file as JSON")?;
 
-    Ok(img_conf)
+    // Compute the manifest digest
+    let mut hasher = Sha256::new();
+    hasher.update(&serde_json::to_vec(&img_conf.manifest).context("Serializing image manifest")?);
+    let manifest_digest_str = hex::encode(hasher.finish());
+    Ok((img_conf, manifest_digest_str))
 }
 
 #[context("Getting composefs deployment metadata")]
@@ -543,9 +551,8 @@ fn boot_entry_from_composefs_deployment(
             let ostree_img_ref = OstreeImageReference::from_str(&img_name_from_config)?;
             let img_ref = crate::spec::ImageReference::from(ostree_img_ref);
 
-            let img_conf = get_imginfo(storage, &verity)?;
+            let (img_conf, image_digest) = get_imginfo(storage, &verity)?;
 
-            let image_digest = img_conf.manifest.config().digest().to_string();
             let architecture = img_conf.config.architecture().to_string();
             let version = img_conf
                 .manifest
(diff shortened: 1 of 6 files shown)


crates/lib/src/bootc_composefs/status.rs
```diff
@@ -528,7 +532,11 @@ pub(crate) fn get_imginfo(storage: &Storage, deployment_id: &str) -> Result<ImgC
     let img_conf = serde_json::from_str::<ImgConfigManifest>(&buffer)
         .context("Failed to parse .imginfo file as JSON")?;
 
-    Ok(img_conf)
+    // Compute the manifest digest
+    let mut hasher = Sha256::new();
+    hasher.update(&serde_json::to_vec(&img_conf.manifest).context("Serializing image manifest")?);
```
jlebon, line 537: This is re-serializing the manifest object. But for our purposes here, we want to be certain we end up with the same manifest digest as the remote. Are we guaranteed a match here? Ideally we'd just hash the raw manifest string we originally fetched from the remote at pull time, which IIRC is how this works on the ostree backend.

Johan-Liebert1: This is more "legacy" code for backwards compatibility for older systems where we used to store the config + manifest in an .imginfo file. To answer your question, yes, it should, in theory, produce the same digest since we pretty much used to store the manifest and config unchanged. Now though, we do what ostree does and store the manifest digest string inside the .origin file

cgwalters (dismissed):

crates/lib/src/bootc_composefs/status.rs
cgwalters: > To answer your question, yes, it should, in theory, produce the same digest since we pretty much used to store the manifest and config unchanged. 

I think the point Jonathan is making here is that we should hash the raw data and not re-serialize.

But @Johan-Liebert1 is also right in this is just the legacy path, which I think we can schedule to remove at some point and isn't really relevant anymore.

cgwalters (dismissed):

cgwalters:
> This test ends up failing for systems with ostree v2026.2. Disable it until we have a fix

https://github.com/ostreedev/ostree/issues/3609 will track this.

cgwalters (approved):

PR #2283, opened 2026-07-07
tests: Skip test-loader-entries-source for ostree v2026.2

Johan-Liebert1:
This test ends up failing for systems with ostree v2026.2. Disable it until we have a fix


Johan-Liebert1:
Rolled into https://github.com/bootc-dev/bootc/pull/2277

PR #2290, opened 2026-07-07
deploy: Pull bound images before staging via composefs mount

### LINKED ISSUE #2013: Logically-bound images aren't reliably present
### Discussed in https://github.com/bootc-dev/bootc/discussions/2011

<div type='discussions-op-text'>

<sup>Originally posted by **csssuf** February 18, 2026</sup>
As I've been working with logically-bound images, I've encountered a couple of challenges around reliability. These can more or less be summed up by saying that errors pulling LBIs are not really handled, which makes it tough to rely on `bootc switch`/`upgrade` when LBIs are involved.

The two main challenges are:
- If an LBI cannot be pulled, there is no clean way to retry it.
- If any LBIs fail to be pulled, the new deployment is still staged for next boot.

These two issues combine to make updates with any LBI changes nearly impossible to apply atomically - we expect LBIs to be available immediately on boot, but we cannot retry them if the initial fetch fails. Right now, we're working around this by using physically-embedded images, but this is painful for other reasons (we lose the layering from the embedded images, so updates are huge, and loading images into c/storage is fallible at boot-time).

I'm not sure what the best option here is - off the top of my head, it seems safest to unstage a deployment (if this is possible) if we can't fetch its LBIs; that way, we're back in the state we started in and it's up to the caller to retry the `bootc` operation that failed.</div>

cgwalters:
We were still staging a new bootloader entry even when we failed to pull a LBI. Switch to only doing an image pull and not a deployment, then pulling referenced LBIs from that.

Closes: https://github.com/bootc-dev/bootc/issues/2013

### DIFF
--- a/crates/utils/src/command.rs
+++ b/crates/utils/src/command.rs
@@ -9,6 +9,16 @@ use std::{
 
 use anyhow::{Context, Result};
 
+/// Create a seekable, filesystem-independent file for command output.
+fn command_output_file() -> Result<std::fs::File> {
+    // bootc's command helpers run from a systemd generator. Generators on
+    // systemd 252 and older may see a read-only /tmp (253+ provides a private
+    // writable /tmp), so output capture must not rely on filesystem temp files.
+    rustix::fs::memfd_create("bootc-command-output", rustix::fs::MemfdFlags::CLOEXEC)
+        .map(std::fs::File::from)
+        .context("create memfd for command output")
+}
+
 /// Helpers intended for [`std::process::Command`].
 pub trait CommandRunExt {
     /// Log (at debug level) the full child commandline.
@@ -139,7 +149,7 @@ impl CommandRunExt for Command {
 
     /// Synchronously execute the child, and return an error if the child exited unsuccessfully.
     fn run_capture_stderr(&mut self) -> Result<()> {
-        let stderr = tempfile::tempfile()?;
+        let stderr = command_output_file()?;
         self.stderr(stderr.try_clone()?);
         tracing::trace!("exec: {self:?}");
         self.status()?.check_status_with_stderr(stderr)
@@ -168,7 +178,7 @@ impl CommandRunExt for Command {
     }
 
     fn run_get_output(&mut self) -> Result<Box<dyn std::io::BufRead>> {
-        let mut stdout = tempfile::tempfile()?;
+        let mut stdout = command_output_file()?;
         self.stdout(stdout.try_clone()?);
         self.run_capture_stderr()?;
         stdout.seek(std::io::SeekFrom::Start(0)).context("seek")?;
@@ -220,7 +230,7 @@ pub trait AsyncCommandRunExt {
 
 impl AsyncCommandRunExt for tokio::process::Command {
     async fn run(&mut self) -> Result<()> {
-        let stderr = tempfile::tempfile()?;
+        let stderr = command_output_file()?;
         self.stderr(stderr.try_clone()?);
         self.status().await?.check_status_with_stderr(stderr)
     }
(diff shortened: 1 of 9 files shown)


gursewak1997:
Needs `cargo fmt`

gursewak1997 (dismissed):

jmarrero (dismissed):
lgtm

cgwalters:
> : error: Switching: Switching (ostree): Staging: Mounting ostree commit b579414cd3d4d034ac8c9377579de4033025dfefc637bfdab95aa5ec8b9ddb5b: composefs_fsmount: Block device required (os error 15)

Looks legit

cgwalters:
OK, this one should be good to go now

jmarrero (dismissed):

cgwalters:
Hmm, hitting https://github.com/bootc-dev/bootc/issues/2340 on this on centos-9.

cgwalters:
OK this rolls in the fix for that

cgwalters:
Looks like there are some potential latent issues with general tests on c9s

cgwalters:
I don't yet have an idea of what's going on with the shadow generator here, looking

crates/utils/src/command.rs
```diff
@@ -9,6 +9,16 @@ use std::{
 
 use anyhow::{Context, Result};
 
+/// Create a seekable, filesystem-independent file for command output.
+fn command_output_file() -> Result<std::fs::File> {
+    // bootc's command helpers run from a systemd generator. Generators on
+    // systemd 252 and older may see a read-only /tmp (253+ provides a private
+    // writable /tmp), so output capture must not rely on filesystem temp files.
```
cgwalters, line 16: This took SO MUCH time to figure out

jeckersb: Oof that's a rough one!

jeckersb (approved):

Johan-Liebert1 (dismissed):

cgwalters:
OK I think https://github.com/bootc-dev/actions/pull/53 will fix this. That was really painful.

cgwalters:
OK, trying to hack up a newer qemu on top of 24.04 is probably viable, but comes with its own chain of downsides. I am leaning towards heading straight to `ubuntu-26.04`; a downside is it's still technically in preview, and then ran smack into a problem where our build uses bwrap inside podman, and in 26.04 there's an apparmor profile for bwrap that breaks what we're doing. (IMO this path-based matching is a serious downside of apparmor; SELinux gets this right because `podman run` switches security contexts, which already limits things itself)

jeckersb (dismissed):

tmt/plans/integration.fmf
```diff
@@ -94,6 +94,10 @@ execute:
     how: fmf
     test:
       - /tmt/tests/tests/test-24-image-upgrade-reboot
+  adjust:
+  - when: distro == centos-9 and boot_type == uki and seal_state == sealed
```
Johan-Liebert1, line 98: Didn't know tmt could do this

Johan-Liebert1 (approved):

cgwalters:
:tada: 

This one turned out to be quite a saga, the core problem was mostly RHEL9 specific issues:

- A qemu bug https://gitlab.com/virtio-fs/virtiofsd/-/work_items/213#note_2695259613 that was only fixed in recent OS versions, meaning we had to jump our base runner env to ubuntu-26.04 - while this bug could cause problems on RHEL10+ it primarily bit RHEL9 due to guest configuration
- Then upgrading to ubuntu-26.04 of course had its own fallout such as https://github.com/bootc-dev/actions/pull/55
- Also on RHEL9 systemd doesn't support the credential injection we use in bcvk, which means we can't (easily) do the trick with binding the host storage, which broke sealed upgrade tests

PR #2291, opened 2026-07-07
feat: --require-digest option for bootc upgrade/switch

### LINKED ISSUE #2279: [Feature request] Option to require specific digest for bootc upgrade
### Motivation

There are various scenarios where one might want to perform some additional verification steps before staging an update; for example, verifying SLSA provenance for the image is one common scenario. Currently, this runs into a TOCTOU race condition: if you do provenance verification (for example) on the latest version of an image, and then run `bootc upgrade`, there's no guarantee that a new version of the image wasn't pushed in between these two steps, potentially resulting in the update downloading and staging an image whose provenance hasn't been locally verified.

### Feature request

An option to `bootc upgrade` that allows an update to proceed only if it has a given digest would solve this issue. Using `--require-digest` as a placeholder name for this proposed option, `bootc upgrade --require-digest sha256:...` would behave identically to `bootc upgrade`, except that it would check that the image to be pulled has the given digest.

If the digests match, the update would proceed as normal; if the digests don't match, the command would fail with a specific documented error code. This would allow the caller to detect the error, redo their extra verification steps, and try the update again.

### Possible alternatives

If there's currently a way to avoid this TOCTOU issue in bootc, I'm not aware of it (and would be interested to hear about it).

Another possible solution would be to add some sort of "pre-update hook" mechanism where bootc could be made to run a provided script before an update, passing some metadata such as the update digest as arguments to the script. But that seems significantly more complicated for minimal advantage over the option I proposed here.

HastD:
Add a new option, `--require-digest`, to `bootc upgrade` and `bootc switch`. When provided, bootc checks that the image to be downloaded/staged has the given digest; if it does not, the operation will fail with an error message.

This is useful for avoiding TOCTOU issues when performing some external checks (such as provenance verification) on an image before updating to that image.

Resolves #2279.

### DIFF
--- a/crates/lib/src/bootc_composefs/switch.rs
+++ b/crates/lib/src/bootc_composefs/switch.rs
@@ -48,6 +48,7 @@ pub(crate) async fn switch_composefs(
         bootc.operation = "switch",
         bootc.target_image = target_imgref.to_string(),
         bootc.apply_mode = opts.apply,
+        bootc.require_digest = opts.require_digest.as_ref().map(ToString::to_string),
         "Starting composefs switch operation",
     );
 
@@ -77,6 +78,7 @@ pub(crate) async fn switch_composefs(
         apply: opts.apply,
         download_only: false,
         use_unified,
+        require_digest: opts.require_digest,
     };
 
     if let Some(cfg_verity) = image {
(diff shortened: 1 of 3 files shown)


HastD:
Marking as draft for now since we're discussing at #2279 whether this is a good way to solve the problem.

HastD:
Closing per discussion at #2279 that concluded that this isn't necessary.

PR #2293, opened 2026-07-07
ci: A few improvements

cgwalters:
Mainly reducing disk usage to hopefully reduce flakes. A *very* surprising discovery is that some GHA runners have 70G disks, others have 140G, and there's apparently no way to steer them. This explains a good bit of the disk usage flakes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -307,7 +307,7 @@ jobs:
           test ${{ matrix.test_os }} = "${used_vid}"
 
       - name: Unit and container integration tests
-        run: just test-container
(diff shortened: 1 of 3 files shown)


Johan-Liebert1 (approved):

cgwalters:
OK I'm forcing this one over the queue

HastD:
If you're still having issues with disk space, I have an action that frees disk space relatively quickly: https://github.com/HastD/free-disk-space
It can be configured to only do anything if a given amount isn't already available, so for example with `skip-if-available: "64G"` it'll do nothing on runners that happen to have a 140 GB disk, while freeing about 40 GB of space in around 90 seconds on runners with a 70 GB disk.

Issue #2136, opened 2026-04-10
bootc expects `org.opencontainers.image.version` to be set as a label, not an annotation


fpoirotte:
Hello,

I stumbled across a minor issue with the way bootc looks for an image's version (when using `bootc status`).
It expects the version to be set as a label either with the key `org.opencontainers.image.version` or with the key `version` (see [this function](https://github.com/bootc-dev/bootc/blob/b1fe5d697855189f64ba87cff5cfb058f91792f6/crates/ostree-ext/src/container/mod.rs#L605)).

However, the OpenContainers image specification only defines `org.opencontainers.image.version` as a key inside annotations (see [here](https://specs.opencontainers.org/image-spec/annotations/#pre-defined-annotation-keys)).

To match the OpenContainers spec, would it be possible to change `version_for_config` to look for `org.opencontainers.image.version` in the annotations, then look for `version` in the labels, or would this be considered too much of a breaking change?

Best regards,
François

PR #2294, opened 2026-07-08
Misc. Improvements Draft PR

antheas:
Here are some improvements I came up with while working with bootc lately. They are by no means ready to merge (and vibed). Hopefully they are an inspiration.

Here is a small summary.

 1) The current RPM for bootc compiles bootc three times because of the docs and tests instead of re-using the first build. Commit  is junk and I vibed it, but I need to be able to build the spec for my workflow. It would be nice to fix so it compiles once.
 2) `bootc internals ostree-ext container encapsulate --contentmeta` still adds the empty package layer. It probably shouldn't. The caller can add an extra layer if they desire. I am unaware of someone using the previous manifest option with rpm-ostree in either case.
 3) `bootc internals ostree-ext container encapsulate` and ABI is still single threaded, and because of the current architecture it's very hard to fix (by hand). But it is around 3 times faster with 4 workers (~3min to 40s in testing)
 4) Bootc does not preserve the .version key during merge commits so ostree does not write the version on the bootloader. Perhaps the source here should be the OCI label, but for my usecase the ostree tag works fine.
 5) `bootc internals ostree-ext container unencapsulate` always does a checkout, even when a commit has no derived layers. This adds 10-20 seconds on deployment times universally and spins fans, even though it is not needed. It is also important for my usecase because it means you cannot do unencapsulate on a `bare-split-attrs` repo currently, because it does not support checkouts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,28 +32,42 @@ prefix ?= /usr
 CARGO_FEATURES_DEFAULT ?= $(shell . /usr/lib/os-release; if echo "$$ID_LIKE" |grep -qF rhel; then echo rhsm; fi)
 # You can set this to override all cargo features, including the defaults
 CARGO_FEATURES ?= $(CARGO_FEATURES_DEFAULT)
+BOOTC_PROFILE ?= $(if $(wildcard target/rpm/bootc),rpm,release)
+BOOTC_TARGET_DIR = target/$(BOOTC_PROFILE)
 
 # Build all binaries
 .PHONY: bin
 bin: manpages
-	cargo build --release --features "$(CARGO_FEATURES)" --bins
+	if [ ! -x "$(BOOTC_TARGET_DIR)/bootc" ]; then \
+		cargo build --release --features "$(CARGO_FEATURES)" --bins; \
+	fi
 
-# Note this cargo build is run without features (such as rhsm)
 .PHONY: manpages
 manpages:
-	cargo run --release --package xtask -- manpages
+	if [ -x "$(BOOTC_TARGET_DIR)/bootc" ]; then \
+		rm -rf target/man; \
+		if [ -x "$(BOOTC_TARGET_DIR)/xtask" ]; then \
+			BOOTC_DOCGEN_BIN="$(BOOTC_TARGET_DIR)/bootc" "$(BOOTC_TARGET_DIR)/xtask" manpages; \
+		else \
+			BOOTC_DOCGEN_BIN="$(BOOTC_TARGET_DIR)/bootc" cargo run --release --package xtask -- manpages; \
+		fi; \
+	else \
+		mkdir -p target/man; \
+		printf '.TH BOOTC 8\n.SH NAME\nbootc \\- bootable container system\n' > target/man/bootc.8; \
+		printf '.TH BOOTC-SETUP-ROOT-CONF 5\n.SH NAME\nbootc-setup-root.conf \\- bootc setup-root configuration\n' > target/man/bootc-setup-root-conf.5; \
+	fi
 
 .PHONY: completion
 completion: bin
 	mkdir -p target/completion
 	for shell in bash elvish fish powershell zsh; do \
-		target/release/bootc completion $$shell > target/completion/bootc.$$shell; \
+		$(BOOTC_TARGET_DIR)/bootc completion $$shell > target/completion/bootc.$$shell; \
 	done
 
 STORAGE_RELATIVE_PATH ?= $(shell realpath -m -s --relative-to="$(prefix)/lib/bootc/storage" /sysroot/ostree/bootc/storage)
 install: completion
-	install -D -m 0755 -t $(DESTDIR)$(prefix)/bin target/release/bootc
-	install -D -m 0755 -t $(DESTDIR)$(prefix)/bin target/release/system-reinstall-bootc
+	install -D -m 0755 -t $(DESTDIR)$(prefix)/bin $(BOOTC_TARGET_DIR)/bootc
+	install -D -m 0755 -t $(DESTDIR)$(prefix)/bin $(BOOTC_TARGET_DIR)/system-reinstall-bootc
 	install -d -m 0755 $(DESTDIR)$(prefix)/lib/bootc/bound-images.d
 	install -d -m 0755 $(DESTDIR)$(prefix)/lib/bootc/kargs.d
 	ln -s "$(STORAGE_RELATIVE_PATH)" "$(DESTDIR)$(prefix)/lib/bootc/storage"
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -32,28 +32,42 @@ prefix ?= /usr
...
 # You can set this to override all cargo features, including the defaults
 CARGO_FEATURES ?= $(CARGO_FEATURES_DEFAULT)
+BOOTC_PROFILE ?= $(if $(wildcard target/rpm/bootc),rpm,release)
+BOOTC_TARGET_DIR = target/$(BOOTC_PROFILE)
 
 # Build all binaries
 .PHONY: bin
 bin: manpages
-	cargo build --release --features "$(CARGO_FEATURES)" --bins
+	if [ ! -x "$(BOOTC_TARGET_DIR)/bootc" ]; then \
```
cgwalters, line 41: This completely breaks the buildsystem, no? We'd just silently be reusing old binaries...

Makefile
```diff
@@ -32,28 +32,42 @@ prefix ?= /usr
...
+		cargo build --release --features "$(CARGO_FEATURES)" --bins; \
+	fi
 
-# Note this cargo build is run without features (such as rhsm)
 .PHONY: manpages
 manpages:
-	cargo run --release --package xtask -- manpages
+	if [ -x "$(BOOTC_TARGET_DIR)/bootc" ]; then \
```
cgwalters, line 47: But yes, this is really really messy how we have our documentation depend on running the binary.

Maybe what is easier is to try to move instead to where we just *validate* the generated docs as a separate phase, but don't default to linking the two together.

It's effectively what we did for the tmt stuff

crates/ostree-ext/src/container/encapsulate.rs
```diff
@@ -82,32 +108,292 @@ fn commit_meta_to_labels<'a>(
...
+    let worker_opts = LayerExportOpts::from(opts);
+    let mut handles = Vec::with_capacity(worker_count);
+    for _ in 0..worker_count {
+        let queue = Arc::clone(&queue);
+        let result_tx = result_tx.clone();
+        let repo = repo.clone();
+        let commit = commit.to_string();
+        let ocidir = ociw.dir().try_clone().context("Cloning OCI directory")?;
+        let opts = worker_opts.clone();
+        handles.push(thread::spawn(move || {
```
cgwalters, line 339: I think it'd be cleaner to use https://doc.rust-lang.org/stable/std/thread/fn.scope.html - or use rayon. (Or use tokio)

crates/ostree-ext/src/chunking.rs
```diff
@@ -855,8 +851,6 @@ fn basic_packing<'a>(
         r.push(max_freq_components);
     }
 
-    // Allocate an empty bin for new packages
```
cgwalters, line 858: I agree we never used this in practice, and we can probably drop it now. OTOH, this code is basically obsoleted by https://github.com/coreos/chunkah anyways?

crates/ostree-ext/src/container/store.rs
```diff
@@ -1578,6 +1582,9 @@ impl ImageImporter {
             META_CONFIG,
             import.config.to_canon_json_string()?.to_variant(),
         );
+        if let Some(version) = image_version {
```
cgwalters, line 1585: This may relate to https://github.com/bootc-dev/bootc/issues/2136

cgwalters (changes requested):
Thanks for submitting this, I think it'd make sense to split this into distinct PRs?

antheas:
Hi, yes, this PR is not designed to merge

It is more of an issue with some draft solutions attached, moreso for inspiration.

Which ones do you think should spinoff as draft PRs and which should land in an issue?

Makefile
antheas: yes, that patch is not good and not designed for merge. But it did its job for me.

It is very hard to iterate with bootc while having it compile three times to deploy

crates/ostree-ext/src/chunking.rs
antheas: that is a nice tool. I will keep it in mind for non-ostree images

ostree-ext provides a fast deploy path for ostree backends so it will have to stay around until the composefs backend is stable with similar performance to the fast ostree deploy path, then add 2-3 more years for migration or a way to specify to bootc interim migration paths so e.g., version 2027.3 switches to composefs and points ostree instances to 2027.2 that has the machinery to migrate from ostree to composefs while being built for ostree. Ah.

crates/ostree-ext/src/container/store.rs
antheas: Related but not quite. It's true that for images built with docker / buildah they need to be tagged twice so docker compatible registries that only read the leaf manifest can see the tag and tools like podman / skopeo that read the inner config can pick up those

This fixes the bootloader version exposed by ostree. Without this entries appear like PRETTY_NAME. This PR supersedes the one for PRETTY_BOOTLOADER

Ideally on the spinoff, bootc could be taught to set the version tag on the merge commit based on org.opencontainers.image.version while falling back to ostree

crates/ostree-ext/src/container/encapsulate.rs
antheas: Yeah, this was a fast fixup that _works_. I wouldn't be happy merging it, the deploy speedup is real though

antheas:
By the way, I had another look at chunkah and https://github.com/dmnks/rpm/commit/1825517392c1a04b1f3bc6a89d308fdda2927edd

Great work on landing --parkdb! I cannot wait to try it out.

I went through my image and fixed some layer invalidations, specifically the initramfs plus some libdnf weirdness.

For chunkah, you need to make sure /usr/lib/sysimage/libdnf5/ is unpackaged, because it is owned by libdnf5 and it will corrupt its layer when packages change. /etc/os-release should also be unpackaged (owned by fedora-release). bootupd is also a bit naughty and timestamps EFI.json (/usr/lib/bootupd/updates/EFI.json needs to be unpackaged) Then, dropping these three files is sound:

```
    /usr/lib/sysimage/libdnf5/transaction_history.sqlite
    /usr/lib/sysimage/libdnf5/transaction_history.sqlite-shm
    /usr/lib/sysimage/libdnf5/transaction_history.sqlite-wal
```

They are ghost files with timestamps, although I am unsure of the implication of removing them.

For the initramfs, it includes os-release so some care needs to be taken for the image version to be added after the initramfs is built. There is also a sysusers dracut bug, where when it adds dbus it adds a modification day since epoch to /etc/shadow (we should probably open an issue for that). This causes the initramfs to change every day. But with those fixed, assuming no packages that touch the initramfs such as kernel and firmware are the same, the initramfs is the same! So, the layer stays the same and soft reboot works!

With those improvements, if I rebuild the image from scratch, only the commit layer and unpackaged layer change, with a total of ~80mb, and bootc from this PR deploys the image in ~10 seconds. It is so fast that I cannot measure it.

You need to keep that in mind before deprecating the rpm-ostree rechunker (which I do not use) or `bootc internals ostree-ext` (which I do use), as chunkah does not have the fast deploy path

EDIT: Deployment takes 13 seconds, diff is 59mb

PR #2296, opened 2026-07-08
governance: Add member, reviewer and committer roles

jmarrero:
Add three new contributor ladder tiers to the governance document: member, reviewer and committer. All existing maintainers are listed as committers in MAINTAINERS.md.

This was drafted using Opus 4.6 and with https://kubernetes.io/docs/contribute/participate/roles-and-responsibilities/ & https://github.com/containerd/.project/blob/main/GOVERNANCE.md as examples. Iterated a bit to make sure Opus did not rewrite everything and just added the relevant sections.

*DRAFT while we discuss*

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -6,6 +6,9 @@ This governance explains how the project is run.
 - [Values](#values)
 - [Maintainers](#maintainers)
 - [Becoming a Maintainer](#becoming-a-maintainer)
+- [Members](#members)
+- [Reviewers](#reviewers)
+- [Committers](#committers)
 - [Meetings](#meetings)
 - [CNCF Resources](#cncf-resources)
 - [Code of Conduct Enforcement](#code-of-conduct)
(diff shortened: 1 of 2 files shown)


cgwalters:
But this file got copied into https://github.com/bootc-dev/community/blob/main/GOVERNANCE.md - see my concern in https://github.com/bootc-dev/community/pull/1#discussion_r3426181932

I think we want this to apply to all repos in the org, right? So we should do this change in the community repo, and then fix all the other repos to point to it

SNThrailkill:
Reading everything over I think I agree based on current repos in the org. We can probably do a blanket policy for the org that applies to all repos. The mechanism for doing it I'm not sure on.

bupd:
> I think we want this to apply to all repos in the org, right? So we should do this change in the community repo, and then fix all the other repos to point to it

This is the way to go have relevant files like MAINTAINERS.md & GOVERNANCE.md etc. but these files under the repo should have only one line which points to the community repo which contains the MAINTAINERS.md

jmarrero:
ok moved to: https://github.com/bootc-dev/community/pull/43

PR #2298, opened 2026-07-08
docs/composefs: Update with latest

cgwalters:
In particular, this commits us to supporting in-place upgrades from the current state!

We will leave things like the EROFS v1 switch as a followon, ensuring that we continue to support that as an upgrade path.

### DIFF
--- a/docs/src/experimental-composefs.md
+++ b/docs/src/experimental-composefs.md
@@ -5,14 +5,34 @@ do provide feedback on them.
 
 ## Overview
 
-The composefs backend is an experimental alternative storage backend that uses [composefs-rs](https://github.com/containers/composefs-rs) instead of ostree for storing and managing bootc system deployments.
+The composefs backend is an experimental alternative storage backend that uses [composefs-rs](https://github.com/composefs/composefs-rs) instead of ostree for storing and managing bootc system deployments.
 
-**Status**: Experimental. The composefs backend is under active development and not yet suitable for production use. The feature is always compiled in as of bootc v1.10.1.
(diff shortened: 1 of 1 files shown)


jeckersb (approved):
🎉

docs/src/experimental-composefs.md
```diff
@@ -61,59 +81,86 @@ valid use case is to temporarily disable it in order to test a change locally
...
 
-The key to building sealed images is using a multi-stage Dockerfile where a separate stage mounts the target rootfs, computes its composefs digest, and generates the signed UKI in one step:
+Building a sealed image involves three stages: build the rootfs, split the kernel and initramfs out of it, and generate the signed UKI from the split rootfs in a tools stage:
 
 ```dockerfile
 # Build your rootfs with all packages and configuration
 FROM <base-image> as rootfs
 RUN apt|dnf|zypper install ... && bootc container lint --fatal-warnings
 
+# Split the kernel and initramfs out of the rootfs. This moves
```
travier, line 100: We might want to include a rechunking step here to effectively drop the kernel & initrd from the base image.

travier (approved):
Looking good! One suggestion.

docs/src/experimental-composefs.md
cgwalters: This isn't intended to be a true reference example, we already have too many of those. I'll be sure the rechunking lands in other (downstream) docs in general!

jeckersb:
Hm `required-checks` appears to be wedged here, going to close and re-open to try and force it to sort things out...

Issue #542, opened 2024-05-16
Consider designing bootc as a library to be used by installer applications


jmpolom:
Right now bootc only seems envisioned as a simple CLI application to install and lifecycle an ostree based system using ostree containers as the install/update payload. Given the inevitability that a use case will necessitate a higher level "installer", bootc could have an API available to use in a higher level rust application that could function itself as the installer. This would enable a distro to create a proper installer application while fully integrating the capabilities of bootc (IE: more evolved than a shell script installer).

jeckersb:
The vast majority of the code already lives as a separate library crate under https://github.com/containers/bootc/tree/main/lib.  The current bootc cli is really just a thin wrapper around the library.  It looks like we don't currently publish that on crates.io for general use, but we could start doing so.  At the very least if someone wanted to use it as a dependency today they could do so via a cargo git dependency.

jmpolom:
Yeah I think my point was to perhaps consider this as more of a prominent feature. Obviously it could be included somehow today but the focus seems on it being a standalone tool at the CLI level.

jmpolom:
> Having higher level installers wrap/integrate with `bootc install` was very much a big design goal from the start. I think we have the essentials of that now with `bootc install to-filesystem` being there.
> 
> This intersects with both #440 and #522
> 
> What specific issues for an installer though do you forsee being improved with an API versus a fork/exec model? The main ones are usually things like progress reporting, but we could also do that via either e.g. capnpnp or varlink over a fd.

Better integration potential and not bringing in a secondary IPC scheme to support (thus requiring whatever interacts with bootc to also support said scheme).

I could agree that just using bootc as an external process checks all the boxes but unfortunately the CLI output right now is not very useful/helpful for integrating with an installer that uses `install to-filesystem`. See #536.

If the idea is to keep relations with bootc to being something that's just exec'd via the shell or some wrapping application there need to be serious improvements to the ability to configure it to to produce useful, parseable output (such as outputting a path to the ostree deployment directory -- there are probably others here). `ostree container image deploy` has a flag to get the deployed commit hash. Accoutrements like this are needed for full viability of this model. You could probably cover the vast majority of needs with a few options to tailor the outputs.

For the cases where the bootc CLI interface doesn't cleanly provide what's needed though, an API affords the installer creator maximum ability to get whatever is needed pretty much just subject to the limitations of the language and libraries they're willing to bring in.

[...]

Issue #522, opened 2024-05-13
go SDK/CGO support


hexfusion:
As the popularity of bootc grows a central location for go types/examples would be useful for projects. Speaking with @cgwalters he is willing to consider something similar to rpmostree-client-go[1] for bootc but there is some intersection with [2].

Curious if instead of creating go client as a os exec wrapper, exposing some of the Rust functionality via CGO? Please include me in future conversations on the topic if possible :pray: 

[1] https://github.com/coreos/rpmostree-client-go/
[2] https://github.com/containers/bootc/issues/518
[3] https://github.com/bytecodealliance/wasmtime/blob/b869b66bce8a57c9e83db4e3bcd6809d88e94a5f/crates/c-api/src/types/extern.rs#L39
[4]  https://github.com/bytecodealliance/wasmtime-go/blob/cabc733f95623fb897b2c35266124192ffca94b4/externtype.go#L9C12-L9C29

cgwalters:
This issue also intersects with https://github.com/containers/bootc/issues/2

cgwalters:
I've added this to the 1.2 milestone (next month) as it's pretty important for https://github.com/coreos/rpm-ostree/issues/4994

As of currently...I am slightly leaning towards varlink.

cgwalters:
I looked a bit more at users of gRPC over Unix domain socket (what we definitely want here) and came upon https://github.com/hyperium/h2/pull/487 which is also pretty telling (in the opposite direction, most gRPC users are over TCP).

@sashagrunert any opinions on this?

saschagrunert:
https://github.com/hyperium/h2/pull/487 is still unresolved and I personally think that's an issue. If using TCP is an alternative then gRPC seems more mature than Cap’n Proto. We still made good results using Cap’n Proto in conmon-rs, though. :)

cgwalters:
We definitely don't want TCP here, it opens up huge questions around authorization, firewalling, lifecycle binding etc. even when operating just on loopback.

I like Cap'n Proto - they promote the speed thing but I think the fact that it's actually capability based (ref "security" on this page https://capnproto.org/rpc.html ) - I don't think that's true of gRPC and it's definitely not true of varlink. But the flip side is capnp is very complex (on the wire format and implementation).

I dunno...

antheas:
In https://github.com/containers/bootc/pull/921 I introduce progress reporting via jsonl.

[...]

PR #2299, opened 2026-07-08
status: Add --sysroot flag to query non-booted sysroots

cgwalters:
Add `bootc status --sysroot <path>` to query the deployment state of an sysroot at an arbitrary path. This is designed for install tooling (Anaconda, osbuild, custom installers) that needs to discover deployment metadata immediately after installation without rebooting.

The returned Host uses the same schema as `bootc status --json`. Since the target sysroot is not the running system, `booted` and `staged` are null -- all deployments appear in `otherDeployments`, ordered by priority (first entry is what will boot next).

xref https://github.com/bootc-dev/bootc/issues/542

### DIFF
--- a/crates/lib/src/bootc_composefs/status.rs
+++ b/crates/lib/src/bootc_composefs/status.rs
@@ -12,7 +12,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::{
     bootc_composefs::{
-        boot::BootType,
+        boot::{BootType, EFI_LINUX},
         selinux::are_selinux_policies_compatible,
         state::{get_composefs_usr_overlay_status, read_origin},
         utils::{compute_store_boot_digest_for_uki, get_uki_cmdline},
@@ -34,6 +34,7 @@ use crate::{
 use std::str::FromStr;
 
 use bootc_utils::try_deserialize_timestamp;
+use camino::Utf8Path;
 use cap_std_ext::{cap_std::fs::Dir, dirext::CapStdExtDirExt};
 use ostree_container::OstreeImageReference;
 use ostree_ext::container::{self as ostree_container};
@@ -252,8 +253,8 @@ fn get_sorted_grub_uki_boot_entries_helper<'a>(
 pub(crate) fn get_sorted_type1_boot_entries(
     boot_dir: &Dir,
     ascending: bool,
+    bootloader: Bootloader,
 ) -> Result<Vec<BLSConfig>> {
-    let bootloader = get_bootloader()?;
     get_sorted_type1_boot_entries_helper(boot_dir, ascending, false, bootloader)
 }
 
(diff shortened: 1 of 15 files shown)


crates/lib/src/bootc_composefs/status.rs
```diff
@@ -614,6 +627,77 @@ pub(crate) async fn get_composefs_status(
...
+    let (boot_dir, bootloader) = if sysroot_dir.exists("boot/grub2") {
+        (
+            sysroot_dir.open_dir("boot").context("Opening boot dir")?,
+            Bootloader::Grub,
+        )
+    } else if let Some(esp_dir) = sysroot_dir
+        .open_dir_optional("boot/efi")?
+        .filter(|d| d.exists(TYPE1_ENT_PATH) || d.exists(EFI_LINUX))
+    {
+        (esp_dir, Bootloader::Systemd)
```
Johan-Liebert1, line 660: This could also be GrubCC but we have no way of knowing I guess

crates/lib/src/bootc_composefs/status.rs
```diff
@@ -614,6 +627,77 @@ pub(crate) async fn get_composefs_status(
...
+    } else {
+        anyhow::bail!(
+            "Could not find a supported bootloader layout under {sysroot_path}/boot \
+             (expected boot/grub2 or boot/efi/{TYPE1_ENT_PATH})"
+        );
+    };
+
+    let storage = Storage::new_composefs_sysroot(sysroot_path, sysroot_dir, boot_dir)?;
+    let bootloader_entries = list_bootloader_entries(&storage, bootloader)?;
```
Johan-Liebert1, line 669: I'm not sure if we should do this. We shouldn't care about the bootloaders or bootloader entries when querying an arbitrary sysroot. We can definitely base our assumptions by looking at the `sysroot/composefs/images` or `sysroot/state` directories. This would cut down a lot of changes in this PR

Johan-Liebert1 (commented):
Looks good overall. Just not sure about whether we need the bootloader stuff. We could definitely allow querying even if an ESP isn't mounted at `sysroot/boot`

crates/lib/src/bootc_composefs/status.rs
cgwalters: The bootloader entries are the source of truth for what gets booted, which is a key concept here in the general case. That said, our current API is much more designed around what a booted system looks like.

That said, there's already e.g. `bootctl --list` and arguably what we want I think is an extension of that that additionally returns ostree/composefs metadata instead of trying to shim things into our existing status API?

Or perhaps we try to implement that new API as part of https://github.com/bootc-dev/bootc/issues/522

Johan-Liebert1: #522 seems like a separate effort to me, what we want here is a very small subset of that imo

> The bootloader entries are the source of truth for what gets booted, which is a key concept here in the general case

true, but the whole point here is that this might not be a booted system

I don't think this approach is bad, just that from a user's perspective I'd assume just having a sysroot would be enough for the query without having to mount esp or boot partitions in it

cgwalters:
OK while these are different things, I think we need to land https://github.com/bootc-dev/bootc/issues/2300 first.

PR #2301, opened 2026-07-09
Etc merge changes

Johan-Liebert1:
etc-merge: Handle unmergable paths properly

We would have unmergable paths if a modified file was changed to a
directory in the new etc or vice-versa. One of the major issues was that
we were erroring out during "merge" which is not ideal. A new vector now
keeps track of all unmergable paths and the reason they're not mergable.

Ref: composefs/composefs-rs/issues/335

---

cfs/stage: Compute etc diff before staging

Before we write the staged deployment file at `/run/composefs` and
before we create new boot entries, we now compute the etc diff in
advance to check whethere we will face any merge conflicts while
finalizing the deployment. If we find conflicts we exit with and error
and detailing where conflicts were found

---

tmt: Add tests for etc merge conflict

This was predominantly AI generated with some minor tweaks from me.

Makes sure that bootc upgrade/switch fails if merge conflicts are found
during three way etc merge.

### DIFF
--- a/crates/etc-merge/src/lib.rs
+++ b/crates/etc-merge/src/lib.rs
@@ -80,8 +80,8 @@ fn stat_eq_ignore_mtime(this: &Stat, other: &Stat) -> bool {
 
 #[derive(Debug)]
 pub struct UnmergablePaths {
-    path: PathBuf,
-    reason: String,
+    pub path: PathBuf,
+    pub reason: String,
 }
 
 /// Represents the differences between two directory trees.
@@ -95,7 +95,7 @@ pub struct Diff {
     /// Paths that exist in the pristine /etc but not in the current one
     removed: Vec<PathBuf>,
     /// Paths that are unmergable
-    unmergable_paths: Vec<UnmergablePaths>,
+    pub unmergable_paths: Vec<UnmergablePaths>,
 }
 
 fn collect_all_files(
@@ -178,6 +178,7 @@ fn get_deletions(
     Ok(())
 }
 
+#[context("Checking mergability: {current_path:?}")]
 fn check_if_mergable(
     new: &Directory<CustomMetadata>,
     current_inode: &Inode<CustomMetadata>,
@@ -224,7 +225,7 @@ fn check_if_mergable(
                         ),
                     });
                 }
-                Err(ImageError::NotFound(..)) => {},
+                Err(ImageError::NotFound(..)) | Err(ImageError::NotADirectory(..)) => {}
                 Err(e) => Err(e)?,
             }
         }
@@ -497,6 +498,19 @@ pub fn compute_diff(
     Ok(diff)
 }
 
+pub fn print_unmergable_paths(diff: &Diff, writer: &mut impl Write) {
+    use owo_colors::OwoColorize;
+
+    for unmergable in &diff.unmergable_paths {
+        let _ = writeln!(
+            writer,
+            "{} {}",
+            ModificationType::Unmergable.magenta(),
+            unmergable.reason
+        );
+    }
+}
+
 /// Prints a colorized summary of differences to standard output.
 pub fn print_diff(diff: &Diff, writer: &mut impl Write) {
     use owo_colors::OwoColorize;
@@ -513,14 +527,7 @@ pub fn print_diff(diff: &Diff, writer: &mut impl Write) {
         let _ = writeln!(writer, "{} {removed:?}", ModificationType::Removed.red());
     }
 
-    for unmergable in &diff.unmergable_paths {
-        let _ = writeln!(
-            writer,
-            "{} {}",
-            ModificationType::Unmergable.magenta(),
-            unmergable.reason
-        );
-    }
+    print_unmergable_paths(diff, writer);
 }
 
 #[context("Collecting xattrs")]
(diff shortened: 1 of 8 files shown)


crates/etc-merge/src/lib.rs
```diff
@@ -190,6 +239,7 @@ fn get_modifications(
     current: &Directory<CustomMetadata>,
     pristine_leaves: &[Leaf<CustomMetadata>],
     current_leaves: &[Leaf<CustomMetadata>],
+    new_leaves: &[Leaf<CustomMetadata>],
```
jeckersb: This is unused, the only "usage" is to pass it via recursion below.  AI spotted this but I'm surprised the compiler doesn't catch it.

Edit: actually clippy does catch it!

```
warning: parameter is only used in recursion
   --> crates/etc-merge/src/lib.rs:242:5
    |
242 |     new_leaves: &[Leaf<CustomMetadata>],
    |     ^^^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_new_leaves`
    |
note: parameter used here
   --> crates/etc-merge/src/lib.rs:269:29
    |
269 | ...                   new_leaves,
    |                       ^^^^^^^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.97.0/index.html#only_used_in_recursion
    = note: `#[warn(clippy::only_used_in_recursion)]` on by default
```

crates/etc-merge/src/lib.rs
```diff
@@ -170,6 +178,47 @@ fn get_deletions(
     Ok(())
 }
 
+fn check_if_mergable(
```
jeckersb, line 182: This should probably return `Result` since below it calls `get_directory` which can fail, and currently it's just swallowing any unexpected `Err` but it should probably re-raise instead?

crates/etc-merge/src/lib.rs
```diff
@@ -216,26 +266,45 @@ fn get_modifications(
...
+                            Err(ImageError::NotFound(..)) | Err(ImageError::NotADirectory(..)) => {
+                                // This directory was deleted in new_etc
+                                // If it was modified in the current etc, we want this back
+                                //
+                                // Was a directory in current etc, but is now
+                                // a file/symlink in the new etc
+                                if diff.added.len() != total_added {
+                                    diff.added.insert(total_added, current_path.clone());
+                                } else if diff.modified.len() != total_modified {
+                                    diff.modified.insert(total_added, current_path.clone());
```
jeckersb: Should `total_added` here be `total_modified` instead?  That's how it was in the old version and I think it should still be the case here

crates/etc-merge/src/lib.rs
```diff
@@ -170,6 +178,47 @@ fn get_deletions(
     Ok(())
 }
 
+fn check_if_mergable(
+    new: &Directory<CustomMetadata>,
+    current_inode: &Inode<CustomMetadata>,
+    current_path: &PathBuf,
```
jeckersb: Should just be `&Path` per clippy

```
warning: writing `&PathBuf` instead of `&Path` involves a new object where a slice will do
   --> crates/etc-merge/src/lib.rs:184:19
    |
184 |     current_path: &PathBuf,
    |                   ^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.97.0/index.html#ptr_arg
    = note: `#[warn(clippy::ptr_arg)]` on by default
help: change this to
    |
184 ~     current_path: &Path,
185 |     diff: &mut Diff,
...
194 |                 diff.unmergable_paths.push(UnmergablePaths {
195 ~                     path: current_path.to_path_buf(),
196 |                     reason: format!(
...
210 |                 diff.unmergable_paths.push(UnmergablePaths {
211 ~                     path: current_path.to_path_buf(),
    |
```

crates/lib/src/bootc_composefs/finalize.rs
```diff
@@ -37,16 +42,22 @@ pub(crate) async fn get_etc_diff(storage: &Storage, booted_cfs: &BootedComposefs
...
         &pristine_files,
         &current_files,
-        &FileSystem::new(Stat::uninitialized()),
+        &new_etc_files
+            .as_ref()
+            .unwrap_or(&FileSystem::new(Stat::uninitialized())),
     )?;
 
-    print_diff(&diff, &mut std::io::stdout());
+    if print {
```
jeckersb: This logic should probably just go in the caller?  As far as I can tell, this function is only called in two places, and only the one in cli.rs passes `print: true`

crates/etc-merge/src/lib.rs
Johan-Liebert1: This must've been a cherry pick artifact. Removed

crates/etc-merge/src/lib.rs
Johan-Liebert1: Yes, updated

crates/etc-merge/src/lib.rs
Johan-Liebert1: Fixed

crates/etc-merge/src/lib.rs
Johan-Liebert1: This is always called after we get a new directory so it should be safe every time, but yeah better to be explicit

crates/lib/src/bootc_composefs/finalize.rs
Johan-Liebert1: Yup. Updated

jeckersb (approved):

PR #2307, opened 2026-07-10
ostree-ext: Remove privilege dropping for container pulls

gursewak1997:
The setpriv-based isolation was fragile (systemd-only), never ported to composefs-rs, and will be obsolete with the podman pull migration.

Ref: https://github.com/bootc-dev/bootc/pull/2303#issuecomment-4935319952


cgwalters (approved):

crates/ostree-ext/src/container/mod.rs
```diff
@@ -428,23 +428,17 @@ impl ManifestDiff<'_> {
...
-    merge_default_container_proxy_opts_with_isolation(config, user)
+    merge_default_container_proxy_opts_with_isolation(config, None)
 }
 
 /// Apply default configuration for container image pulls, with optional support
 /// for isolation as an unprivileged user.
 pub fn merge_default_container_proxy_opts_with_isolation(
     config: &mut containers_image_proxy::ImageProxyConfig,
-    isolation_user: Option<&str>,
+    _isolation_user: Option<&str>,
```
cgwalters, line 441: Probably should have just dropped this parameter, can do after

PR #2314, opened 2026-07-14
install: Use systemd-repart for partitioning

### LINKED ISSUE #2132: `install to-disk` should create bigger ESP (likely 2G) for composefs+UKI installations by default
Currently, `bootc install to-disk` will create a 1GB ESP, which is likely too small for most use case with a UKI.

As data points, Atomic Desktops UKIs are ~250MB, so 3 UKIs (during updates) will be 75% full and with bootloaders, firmware updates and potentially other firmware files in there, it would be already full and we would have no room for the future. 2GB should give us some room.

For the Atomic Desktops, this will likely be configured in Anaconda but it should probably be configurable in bootc as it heavily depends on what you put in the initrd.

Johan-Liebert1:
If `systemd-repart` binary is present and we find one of the directories associated with systemd-repart configurations, then use systemd-repart for partitioning, else fallback to sfdisk

Also, update the default ESP size to 2G for composefs installs. This only applies to sfdisk path.

Closes: #2132


crates/lib/src/install/baseline.rs
```diff
@@ -190,6 +192,193 @@ pub(crate) fn udev_settle() -> Result<()> {
...
+
+    if !has_config {
+        return Ok(None);
+    }
+
+    let output = Command::new("systemd-repart")
+        .arg("--dry-run=no")
+        .arg("--empty=allow")
+        .arg("--no-pager")
+        .arg("--json=pretty")
```
cgwalters: I would say it's better to use its varlink API now; I was already proposing pulling in zlink for https://github.com/bootc-dev/bcvk/pull/222

crates/lib/src/install/baseline.rs
```diff
@@ -190,6 +192,193 @@ pub(crate) fn udev_settle() -> Result<()> {
...
+        match part.partition_type.as_str() {
+            "esp" => esp_partno = Some(partno),
+            "xbootldr" => boot_partno = Some(partno),
+            t if t.starts_with("root-") => rootpn = Some(partno),
+            _ => {}
+        }
+    }
+
+    let rootpn =
+        rootpn.ok_or_else(|| anyhow::anyhow!("systemd-repart output missing root partition"))?;
```
cgwalters, line 349: I think we should support images that have repart.d definitions that are expected to run on firstboot in the target environment, without defining the default root either.

I would flip this around a bit - we default to running repart.d definitions for the root and esp/xbootldr, but *probably* not the other ones? And we let the other ones naturally run on firstboot.

cgwalters (commented):
Thanks for starting this! ACK on the principle.

Needs docs and tests at least.

crates/lib/src/install/baseline.rs
Johan-Liebert1: > I think we should support images that have repart.d definitions that are expected to run on firstboot in the target environment, without defining the default root either.

so this will entail us creating a root partition if it's not defined?

Agree with the second point. We can only include root,esp and xbootldr

crates/lib/src/install/baseline.rs
Johan-Liebert1: varlink support for systemd-repart seems to have been added in v259. Should we have two separate paths?

Johan-Liebert1: I don't think it's worth it having two separate paths (with varlink and without), unless there's some obvious advantage with varlink

crates/lib/src/install/baseline.rs
Johan-Liebert1: Re-reading the comment, I think I get it now. The main question then would be how would be figure out the sizes of the root partition? Doing a simple `--dry-run` with systemd-repart is probably the best way

Johan-Liebert1:
Currently we check fs-verity compatibility for composefs installations using the filesystem param passed via cli or in our config files. With systemd-repart, we will be allowing any filesystem to be used for root partition (not just the ones defined in our Filesystem enum).

This poses an issue with checking fs-verity compatibility of filesystems that are not defined in our enum. This is probably going to be a design choice, should we: 

Read repart.d definitions, add more fs options in our enum to still be able to check for fs-verity compatibility
or
Simply ignore the check

cc @cgwalters @jeckersb

crates/lib/src/install/baseline.rs
cgwalters: > The main question then would be how would be figure out the sizes of the root partition? 

We honor the repart defaults i.e. something like 
```
[Partition]
Type=root
```

should be all that's needed to have that fill the remaining space.

For cloud scenarios a common thing is that the raw block device is expanded dynamically, and I think there repart will run again and should handle expansion of the root.

cgwalters:
> Currently we check fs-verity compatibility for composefs installations using the filesystem param passed via cli or in our config files. With systemd-repart, we will be allowing any filesystem to be used for root partition (not just the ones defined in our Filesystem enum).

I think ideally we do fsverity support detection early on dynamically instead of having a filesystem allowlist.

But I'd vote for now we just let the install fail if fsverity is required and we fail to enable it instead of trying to match filesystems via repart.d

Johan-Liebert1:
Weird that we're only getting disk space error on f43

```
content: + podman run --rm --privileged -v /dev:/dev -v /run/udev:/run/udev -v /var/disk.img:/disk.img --pid=host --security-opt label=type:unconfined_t --env BOOTC_BOOTLOADER_DEBUG=1 localhost/bootc-repart-noroot bootc install to-disk --filesystem ext4 --root-size 5G --disable-selinux --via-loopback --source-imgref containers-storage:localhost/bootc-repart-noroot /disk.img
content: Automatically enabling --generic-image when installing via loopback
content: notice: Target has SELinux enabled, overriding to disable
content: Installing image: docker://localhost/bootc-repart-noroot
content: Block setup: direct
content:        Size: 10737418240
content:      Serial: <unknown>
content:       Model: <unknown>
content:  Partitions: Discoverable
content: Initializing ostree layout
content: layers already present: 0; layers needed: 13 (2.1 GB)
content: error: Installing to disk: Creating ostree deployment: Pulling: Importing: Writing relabeled layer 30721fef5053e6f825058143840fc073e309af149cb21783931acf9844affc6a to mtree: Writing content object: min-free-space-percent '3%' would be exceeded, at least 118.7 MB requested
```

Increased root partition size

Johan-Liebert1:
Finally, tests pass. This should be good for review now @cgwalters @jeckersb

tmt/tests/booted/test-install-repart.nu
```diff
@@ -33,6 +33,12 @@ def run_install_to_disk [
     target_image: string
     extra_bootc_args: list<string>
 ] {
+    # Remove usr overlay state for ostree
+    # We still need this even though we now have usr-overlay check in ostree branch
+    # as well, because ostree checks for a file inside /run/ostree/deployment-state/...
+    # for usr-overlay status
+    rm -rvf /run/ostree/deployment-state
+
```
cgwalters, line 41: Hmm but we're kind of piling on hacks here... (not your fault to be clear)

docs/src/bootc-install.md
```diff
@@ -424,6 +424,114 @@ If you're building tooling that uses `bootc install to-filesystem`, you should:
...
+and xbootldr, for example `/home`, swap, or `/var`.  Because all definitions
+run at install time, systemd-repart allocates space for all of them during
+installation.
+
+### DPS auto-mount and symlinked directories
+
+`systemd-gpt-auto-generator` maps DPS partition types to fixed mount points
+(e.g. `Type=home` mounts at `/home`).  If the mount point is a symlink
+as is common in ostree-based systems where `/home -> /var/home`, the
+auto-generated mount unit will fail.
```
cgwalters, line 489: I think we should argue to fix this in systemd btw

docs/src/bootc-install.md
```diff
@@ -424,6 +424,114 @@ If you're building tooling that uses `bootc install to-filesystem`, you should:
...
+3. `Format=` in the repart.d root partition definition
+
+If none of these provide a filesystem type, the installation will fail.
+
+### Firstboot definitions
+
+Images may include repart.d definitions for partitions beyond root, ESP,
+and xbootldr, for example `/home`, swap, or `/var`.  Because all definitions
+run at install time, systemd-repart allocates space for all of them during
+installation.
```
cgwalters, line 482: Hmmm hmm...I worry about this breaking use cases people have where they actually relied on the repart definitions running on firstboot.

Think about use cases like "generic cloud image with repart definition for instance-local drive" - or really any case where the block devices aren't available at the time we happen to run `bootc install` (basically bootc-image-builder style flows typically).

Actually we already have this knob: `--generic-image`. I think if that is set, we should *only* honor the root repart definition (but do use the other ones for calculating space constraints?)

cgwalters (commented):
Not a full review yet

tmt/tests/booted/test-install-repart.nu
Johan-Liebert1: yeah, just the check in bootc didn't work

docs/src/bootc-install.md
Johan-Liebert1: > but do use the other ones for calculating space constraints?

Right, so we end up with `root_size` option regardless in the repart case as we can just calculate that. Sounds good.

cgwalters:
This one is a good one to perhaps bring up in a team call?

docs/src/bootc-install.md
cgwalters, line 46: We had a side chat on this on the topic of documentation - would make sense to convert this to a man page so "does this version of bootc support repart" could be knowable from the docs downstream

PR #2315, opened 2026-07-15
UKI print dump diff

### LINKED ISSUE #2313: ukify: Generate dumpfile by default
We could generate a dumpfile by default and put it in `/boot/dumpfile`
If there's a verity mismatch between the UKI cmdline and the final EROFS, we could print the diff

Would help immensely for debugging issues like https://github.com/bootc-dev/bootc/discussions/1984

Johan-Liebert1:
cfs/uki: Print dumpfile diff on UKI digest mismatch

If we have a UKI digest mismatch error, we now print the dumpfile diff,
if a dumpfile is present at `/boot/$kver.dump`

Closes: #2313

### DIFF
--- a/crates/lib/src/bootc_composefs/boot.rs
+++ b/crates/lib/src/bootc_composefs/boot.rs
@@ -62,6 +62,7 @@
 //! 2. **Secondary**: Currently booted deployment (rollback option)
 
 use std::cell::Cell;
+use std::ffi::OsStr;
 use std::fs::create_dir_all;
 use std::io::{Read, Seek, SeekFrom, Write};
 use std::path::Path;
@@ -90,6 +91,7 @@ use composefs_ctl::composefs_boot;
 use composefs_ctl::composefs_oci;
 use fn_error_context::context;
 use linux_kernel_cmdline::utf8::{Cmdline, Parameter};
+use ostree_ext::composefs::dumpfile;
 use rustix::{mount::MountFlags, path::Arg};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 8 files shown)


crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -1460,17 +1474,82 @@ pub(crate) async fn setup_composefs_boot(
...
+                        // We expect the dumpfile to be named the same as the UKI
+                        // Ex. UKI      - 6.19.14-108.fc42.x86_64.efi
+                        //     Dumpfile - 6.19.14-108.fc42.x86_64.dump
+                        let dumpfile_name = mismatch
+                            .uki_name
+                            .as_ref()
+                            .and_then(|x| x.strip_suffix(EFI_EXT).map(|x| format!("{x}.dump")));
+
+                        let Some(dumpfile_name) = &dumpfile_name else {
+                            tracing::warn!("Dumpfile not found for diff");
```
cgwalters: This is a normal situation, not a warning case.

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -1460,17 +1474,82 @@ pub(crate) async fn setup_composefs_boot(
...
+
+                        // TODO: This can use the `dump_files` API once we have
+                        // https://github.com/composefs/composefs-rs/pull/359
+                        let mut original = tempdir.create("original")?;
+                        original.write_all(&embedded)?;
+
+                        let mut tmpfile = tempdir.create("current")?;
+                        dumpfile::write_dumpfile(&mut tmpfile, &fs).context("Writing dumpfile")?;
+
+                        let mut cmd = std::process::Command::new("diff");
```
cgwalters, line 1611: I think we can use https://crates.io/crates/similar

Johan-Liebert1: This doesn't seem to accept files as params, only strings, ranges etc. Reading the entire diff into memory is okay-ish I guess as it's usually around < 20M

crates/lib/src/bootc_composefs/boot.rs
Johan-Liebert1: Right. I'll change it to info

Johan-Liebert1: or debug maybe

cgwalters (approved):

cgwalters (approved):

PR #2322, opened 2026-07-16
cfs: Restore download progress for update and switch

### LINKED ISSUE #2316: [Composefs Backend] As of version 1.16, download progress for chunks is no longer shown by "bootc update" and "bootc switch"
When using newer versions of bootc on the composefs backend, nothing is printed to the console after invoking "bootc update" or "bootc switch" unless the update could not be performed. When an update exists and is being fetched the cursor simply waits in place for a while and then the prompt returns, indicating the action succeeded. Upgrades and rebases still _work_ completely fine, but the lack of a progress indicator seems like an obvious bug to me.

This may possibly be related to the implementation of unified storage under composefs?

cgwalters:
Closes: #2316

composefs switched to having a progress API instead of writing progress directly to the tty itself; we need to wire it up now to our progress infrastructure.

Assisted-by: AI

### DIFF
--- a/crates/lib/src/bootc_composefs/mod.rs
+++ b/crates/lib/src/bootc_composefs/mod.rs
@@ -5,6 +5,7 @@ pub(crate) mod digest;
 pub(crate) mod export;
 pub(crate) mod finalize;
 pub(crate) mod gc;
+pub(crate) mod progress;
 pub(crate) mod repo;
 pub(crate) mod rollback;
 pub(crate) mod selinux;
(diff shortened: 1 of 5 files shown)


jmarrero (approved):
lgtm

renner0e:
I tested this PR on a fedora kinoite based image and the progress bar only returned when not using unified storage

cgwalters:
> I tested this PR on a fedora kinoite based image and the progress bar only returned when not using unified storage

I don't believe we had a real progress bar at all for unified storage (podman pull) before - that one wasn't a regression, but something to fix separately.

renner0e:
> > I tested this PR on a fedora kinoite based image and the progress bar only returned when not using unified storage
> 
> I don't believe we had a real progress bar at all for unified storage (podman pull) before - that one wasn't a regression, but something to fix separately.

Alright, thanks for the clarification.

cgwalters:
(It requires podman 6.0 with https://github.com/podman-container-tools/podman/pull/28224 - and see related comments )

cgwalters:
Man, the flakes :cry:

PR #2302, opened 2026-07-09
bootloader: Chroot into composefs image root for bootupd installs

cgwalters:

PR #2323, opened 2026-07-17
[composefs-backend]: `install to-existing-root` fixes

Johan-Liebert1:
system-reinstall: Propagate boot mount in container

systemd gpt-auto-generator has automount for /boot which we need to
propagate inside the container for the service to mount the ESP at
/boot on access

Also, before remounting `/boot` as rw, check if it's readonly so we
don't fail the re-mount operation

---

composefs: Fix `install to-existing-root`

Fix a few issues with composefs path for `install to-existing-root`.

1. Composefs repository initialization
Since we mount `/sysroot:ro` at `/target`, composefs repository
initialization would fail with `Read only filesystem`. Fix it by
remounting `/target/sysroot` read-write

2. Handle bootloader
Get the current bootloader by reading `LoaderInfo` and try to install
the same one

---

cfs/to-existing-root: Bootloader support

To figure out which bootloader to install, get the host's bootloader
(read from LoaderInfo from efivars) and store it in state. Before
finalizing the bootloader, check if the detected bootloader is
compatible with the current image (as the bootloader on host and the one in
the new image might differ)

### DIFF
--- a/contrib/packaging/switch-to-sdboot
+++ b/contrib/packaging/switch-to-sdboot
@@ -27,7 +27,6 @@ done
 if [ "${#pkgs_to_remove[@]}" -gt 0 ]; then
   rpm -e "${pkgs_to_remove[@]}"
 fi
-rm -vrf /usr/lib/bootupd/updates
 
 # First install the unsigned systemd-boot RPM to get the package in place
 rpm -Uvh --replacepkgs "${src}"/*.rpm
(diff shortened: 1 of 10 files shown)


Johan-Liebert1:
Tests failing because I forgot we remove bootupd stuff when we install systemd-boot. I'll fixup the tests

Johan-Liebert1:
This turned out to be a bit more complicated than anticipated. Turning this into draft for now

Johan-Liebert1:
The issue is bootupd mounts ESP at `/target/sysroot/boot` which ends up with it adding the incorrect UUID (the ESP's) in `bootuuid.cfg`

Johan-Liebert1:
Opened https://github.com/coreos/bootupd/pull/1126

cgwalters:
Yes, I think I worked around this in https://github.com/bootc-dev/bootc/pull/2302 hackily too

Johan-Liebert1:
This needs new bootupd release to include coreos/bootupd#1126

https://src.fedoraproject.org/rpms/rust-bootupd/pull-request/27#

PR #2329, opened 2026-07-20
store: Fix upgrade with existing V2 composefs repo

jmarrero:
After composefs-rs switched the default EROFS format from V2 to V1 (composefs/composefs-rs#330), get_ensure_composefs() fails because init_path() refuses to re-initialize an existing V2 repo with V1.

Fix by trying open_path() first, which respects whatever format is on disk. Fall back to init_path() only for fresh directories, and open_upgrade() for pre-meta.json repos. Add a regression test.

Assisted-by: opencode (Claude claude-opus-4-6)

### DIFF
--- a/crates/lib/src/store/mod.rs
+++ b/crates/lib/src/store/mod.rs
@@ -111,7 +111,7 @@ use ostree_ext::{gio, ostree};
 use rustix::fs::Mode;
 
 use composefs::fsverity::Sha512HashValue;
-use composefs::repository::RepositoryConfig;
+use composefs::repository::{RepositoryConfig, RepositoryOpenError};
 use composefs_ctl::composefs;
 
 use crate::bootc_composefs::backwards_compat::bcompat_boot::prepend_custom_prefix;
@@ -608,6 +608,11 @@ impl Storage {
     ///
     /// This lazily opens the composefs repository, creating the directory if needed
     /// and bootstrapping verity settings from the ostree configuration.
+    ///
+    /// If the repository already exists on disk, it is opened as-is, preserving
+    /// whatever EROFS format version it was created with (e.g. V2 from an older
+    /// composefs-rs).  A fresh repository is only initialized when no `meta.json`
+    /// is found, using the current default format version from composefs-rs.
     pub(crate) fn get_ensure_composefs(&self) -> Result<Arc<ComposefsRepository>> {
         if let Some(composefs) = self.composefs.get() {
             return Ok(Arc::clone(composefs));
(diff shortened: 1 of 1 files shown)


crates/lib/src/store/mod.rs
```diff
@@ -710,4 +748,71 @@ mod tests {
...
+    /// After composefs-rs PR #330, the default switched to V1. Calling `init_path`
+    /// with V1 config on an existing V2 repository triggers a "repository already
+    /// initialized with different configuration" error.
+    ///
+    /// This test verifies that:
+    /// 1. `init_path` with V1 correctly fails on an existing V2 repo (the bug).
+    /// 2. `open_path` succeeds on the V2 repo regardless of the current default.
+    /// 3. A fresh directory without meta.json falls back to `init_path` correctly.
+    #[test]
+    fn test_open_existing_v2_repo_after_v1_default_change() -> Result<()> {
```
cgwalters: This is a pretty synthetic test...it's not running any of the relevant code in bootc, so it's almost just a unit test for composefs; which has a bit of value, but I'm pretty sure we already cover it there.

I think the integration testing gap here is an upgrade from an older system with unified storage enabled for ostree right?

crates/lib/src/store/mod.rs
```diff
@@ -620,14 +625,47 @@ impl Storage {
...
-        } else {
-            config.set_insecure()
+
+        // First, try to open an existing repository.  This respects whatever
+        // EROFS format version (V1 or V2) was persisted in meta.json, avoiding
+        // the "already initialized with different configuration" error that
+        // occurs when the composefs-rs default format version changes between
+        // bootc builds (e.g. V2 → V1 after composefs-rs PR #330).
+        let composefs_dir = self.physical_root.open_dir(COMPOSEFS)?;
+        let composefs = match ComposefsRepository::open_path(&composefs_dir, ".") {
```
cgwalters, line 635: I think we have another variant of this in the initramfs? We should do an audit of other places where we open the composefs repo.

tmt/tests/booted/test-unified-upgrade.nu
```diff
@@ -0,0 +1,85 @@
+# number: 47
+# tmt:
+#   summary: Test ostree upgrade with unified storage (composefs repo reopen)
```
cgwalters: I think we already have something like this - this looks like a generic unified storage test.

The problem is we need an *old* bootc with unified storage set. We already have upgrade tests, I think this would be another variant of those.

IOW: Does this test fail without the fix? I don't think it would

crates/lib/src/store/mod.rs
jmarrero: Got the agent to go thru the code, nothing else found. "Only get_ensure_composefs() is affected"

cgwalters: Agree on this

tmt/tests/booted/test-unified-upgrade.nu
cgwalters: I think for now I'd say we should merge without an added test case; the code looks right.

I did https://github.com/composefs/composefs-rs/pull/363 which I think we can switch to instead in the future, and it has sufficient unit/integration tests.

cgwalters (dismissed):

cgwalters:
Needs cargo fmt

cgwalters (approved):

crates/lib/src/store/mod.rs
```diff
@@ -620,14 +625,45 @@ impl Storage {
...
-            config
-        } else {
-            config.set_insecure()
+
+        // First, try to open an existing repository.  This respects whatever
+        // EROFS format version (V1 or V2) was persisted in meta.json, avoiding
+        // the "already initialized with different configuration" error that
+        // occurs when the composefs-rs default format version changes between
+        // bootc builds (e.g. V2 → V1 after composefs-rs PR #330).
+        let composefs_dir = self.physical_root.open_dir(COMPOSEFS)?;
```
Johan-Liebert1, line 634: We have function `initialize_composefs_repository` which does something similar, like initializing the repo. It is supposed to be called during install time, so it shouldn't really run into this particular error, but there are code duplication and I think it's better to consolidate these two functions

cgwalters: I think when installing to an existing root, it could. We definitely want to align install with upgrade as much as possible.

That said, this can be a followup.

PR #2342, opened 2026-07-27
Switch download only

### LINKED ISSUE #2137: Support `--download-only` for `bootc switch`
For parity with upgrade and because there's legitimate demand for it.

https://github.com/bootc-dev/bootc/issues/1320#issuecomment-3640647056

> @jlebon is there any chance to also allow this flag for `bootc-switch`?
Working on a project where we are currently using `bootc-switch` rather than `bootc-upgrade` as we need to be able switch to exact  versions, rather than relying on the latest available.
At the moment we download the images outside of bootc and then use  `bootc switch --transport oci`. Having the --download-only flag would make this more streamlined.

Johan-Liebert1:
switch: Add `download_only` and `from_download` options

Similar to what we have for the update operation

---

cfs/switch: Handle download only

Handle `download_only` and `from_downloaded` to only download and stage
the image but not finalize, and finalize from an already staged only
downloded deployment respectively

---

ostree/switch: Handle download only

Similar to ostree update, handle download only for bootc switch

---

cli/switch: Make `target` optional

If we have `--from-downloaded` then make `target` optional

---

Closes: #2137

### DIFF
--- a/crates/lib/src/bootc_composefs/switch.rs
+++ b/crates/lib/src/bootc_composefs/switch.rs
@@ -4,7 +4,10 @@ use fn_error_context::context;
 use crate::{
     bootc_composefs::{
         status::get_composefs_status,
-        update::{DoUpgradeOpts, UpdateAction, do_upgrade, is_image_pulled, validate_update},
+        update::{
+            DoUpgradeOpts, UpdateAction, apply_upgrade_from_downloaded, do_upgrade,
+            is_image_pulled, validate_update,
+        },
     },
     cli::{SwitchOpts, imgref_for_switch},
     progress_jsonl::ProgressWriter,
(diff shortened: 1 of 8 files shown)


crates/lib/src/bootc_composefs/switch.rs
```diff
@@ -17,13 +20,28 @@ pub(crate) async fn switch_composefs(
...
+        soft_reboot: opts.soft_reboot,
+        apply: opts.apply,
+        download_only: opts.download_opts.download_only,
+        use_unified: false,
+        quiet: opts.quiet,
+        prog,
+    };
+
+    if opts.download_opts.from_downloaded {
+        return apply_upgrade_from_downloaded(storage, booted_cfs, &host, &do_upgrade_opts).await;
```
jeckersb, line 40: Do we care that staged deployments don't have any provenance whether they came from `upgrade` or `switch`?  As-is one could do `bootc upgrade --download-only && bootc switch --from-downloaded` to apply an upgrade.  There's no way to distinguish between an "upgrade-staged" vs "switch-staged" deployment.

I worry that a typical user that doesn't understand or care about all of the implementation details is not going to realize that upgrades and switches can be used interchangeably in this way and might be surprised if they think that upgrades and switches use distinct staged deployments.

Then again I guess the ostree backend already does it this way too, so it's at least consistent.

Maybe something we address in a followup.

jeckersb (dismissed):

crates/lib/src/bootc_composefs/switch.rs
Johan-Liebert1: Yeah, I had that thought as well and was thinking about something like `bootc staged --apply`, but decided against it.

> Maybe something we address in a followup

we can also take this topic async if you'd like

jeckersb:
https://github.com/bootc-dev/bootc/pull/2361 should fix the `/var/roothome/.ssh` thing, let's hopefully land that first and then rebase

tmt/tests/booted/test-download-only-switch.nu
```diff
@@ -0,0 +1,158 @@
...
+    let imgsrc = imgsrc
+    # This test only works in local mode
+    assert ($imgsrc | str ends-with "-local") "This test requires local mode"
+
+    bootc image copy-to-storage
+
+    # Create test file v1 on host
+    "v1" | save testing-bootc-switch-apply
+
+    # A simple derived container (v1) that adds a file
```
cgwalters: At this point we must have like 20 copies of this. Not new, but...

jeckersb (approved):

PR #2343, opened 2026-07-27
policy: Handle absolute symlinks in system policy path resolution

### LINKED ISSUE #2339: `skopeo.rs`'s container policy resolving does not follow symlinks
At [Zirconium](https://zirconium.gay/) we utilize a pattern from system of using `/usr/share/factory` to manage `/etc` and other system directories, one of the uses of that pattern is symlinking `/etc/containers/policy.json` to a path in `/usr` so users cannot break their policy verification at update time as easily, and can reset them if necessary. It seems that since commit `a2bc817222e7f33147662d0802da0f62246609c7` the handling for the container policy does not resolve symlinks, and instead just opens the symlink file itself, that behavior breaks updates because `bootc upgrade` doesn't recognize the file, see:

```
root@taxifolia:/home/tulip# bootc --version
bootc 1.16.4
root@taxifolia:/home/tulip# ln -sf /usr/share/factory/etc/containers/policy.json /etc/containers/policy.json
root@taxifolia:/home/tulip# bootc upgrade
error: Upgrading: Preparing import: Fetching manifest: Resolving containers policy path: No containers policy.json found; checked $CONTAINERS_POLICY_JSON, user config dir, and system paths
root@taxifolia:/home/tulip# rm -f /etc/containers/policy.json
root@taxifolia:/home/tulip# cp -f /usr/share/factory/etc/containers/policy.json /etc/containers/policy.json
root@taxifolia:/home/tulip# bootc upgrade
No changes in ostree-image-signed:docker://ghcr.io/tulilirockz/taxifolia:latest => sha256:bebc994639bd3778d7ec89f9ed9123cc00a6a54e9798c1681b90d43ecd5cb297
No update available.
```

See: 
https://github.com/bootc-dev/bootc/blame/a2bc817222e7f33147662d0802da0f62246609c7/crates/ostree-ext/src/container/skopeo.rs#L85-L96

Related issue in Zirconium:
https://github.com/zirconium-dev/zirconium/issues/354

Johan-Liebert1:
cap-std refuses to follow absolute symlinks since they would escape the sandbox, returning PermissionDenied. When /etc/containers/policy.json is a symlink to e.g. /usr/share/containers/policy.json, this caused policy resolution to silently skip the file.

Switch to `cap_std_ext::RootDir` that handles the above semantics by itself

Test written by Claude

Fixes: #2339

### DIFF
--- a/crates/ostree-ext/src/container/skopeo.rs
+++ b/crates/ostree-ext/src/container/skopeo.rs
@@ -2,6 +2,7 @@
 
 use super::ImageReference;
 use anyhow::{Context, Result};
+use cap_std_ext::RootDir;
 use cap_std_ext::cap_std;
 use cap_std_ext::cmdext::{CapStdExtCommandExt, CmdFds};
 use containers_image_proxy::oci_spec::image as oci_image;
@@ -46,12 +47,14 @@ const USER_POLICY_SUFFIX: &str = "containers/policy.json";
 /// Absolute paths (from env vars) have their leading `/` stripped so they
 /// resolve under `root`. Passing `root` opened on `/` gives normal behaviour;
 /// tests can pass a cap-std `Dir` backed by a temporary directory.
+///
+/// We use `RootDir` to handle absolute symlinks
 fn resolve_policy_path(
-    root: &cap_std::fs::Dir,
+    root: &RootDir,
     env_override: Option<&Path>,
     xdg_config_home: Option<&Path>,
     home: Option<&Path>,
-) -> Result<cap_std::fs::File> {
+) -> Result<std::fs::File> {
     // Helper: strip a leading `/` so the path is relative to root.
     fn strip_abs(p: &Path) -> &Path {
         p.strip_prefix("/").unwrap_or(p)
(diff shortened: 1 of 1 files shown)


crates/ostree-ext/src/container/skopeo.rs
cgwalters: Let's use https://docs.rs/cap-std-ext/5.1.2/cap_std_ext/struct.RootDir.html instead which handles absolute links.

cgwalters (changes requested):

crates/ostree-ext/src/container/skopeo.rs
Johan-Liebert1: That's nicer. Updated

crates/ostree-ext/src/container/skopeo.rs
```diff
@@ -311,13 +326,65 @@ mod tests {
...
-        assert!(resolve_policy_path(&td3, None, None, None).is_err());
+        let root3 = to_root_dir(&td3);
+        assert!(resolve_policy_path(&root3, None, None, None).is_err());
+
+        Ok(())
+    }
+
+    /// Create an absolute symlink inside a cap-std Dir. We need to go via
+    /// procfs because cap-std (by design) refuses to create symlinks with
+    /// absolute targets.
```
cgwalters, line 344: https://docs.rs/cap-std/latest/cap_std/fs/struct.Dir.html#method.symlink_contents exists

cgwalters (dismissed):

crates/ostree-ext/src/container/skopeo.rs
Johan-Liebert1: This is from `cap-std` which would error out on absolute symlinks

cgwalters (approved):

PR #2346, opened 2026-07-27
install/aleph: add `.aleph-version.json` symlink

jbtrystram:
This adds a compatibility symlink along the created `.bootc-aleph.json` named `.aleph-version.json`.

This compat symlink allow for an easier migration from osbuild's aleph stage [1] to the aleph provided by `bootc install`. This also create a precedent for a "standard" aleph well-known filename that isn't tied to the name of the tool creating it.

Note that coreOS and bootupd relies on this aleph file to exist.

[1] https://github.com/osbuild/osbuild/blob/main/stages/org.osbuild.ostree.aleph

See https://github.com/coreos/fedora-coreos-tracker/issues/2187

### DIFF
--- a/crates/lib/src/install/aleph.rs
+++ b/crates/lib/src/install/aleph.rs
@@ -8,9 +8,15 @@ use ostree_ext::{container as ostree_container, oci_spec};
 use serde::Serialize;
 
 use super::SELinuxFinalState;
+use crate::utils::symlink_idempotent;
 
 /// Path to initially deployed version information
 pub(crate) const BOOTC_ALEPH_PATH: &str = ".bootc-aleph.json";
+/// Compatibility symlink target.
+/// Legacy from coreOS and osbuild where the
+/// aleph data is accessible at this expected path.
+/// See <https://github.com/coreos/fedora-coreos-tracker/issues/2187>
+const ALEPH_SYMLINK: &str = ".aleph-version.json";
 
 /// The "aleph" version information is injected into /root/.bootc-aleph.json
 /// and contains the image ID that was initially used to install.  This can
@@ -80,11 +86,16 @@ impl InstallAleph {
         Ok(r)
     }
 
-    /// Serialize to a file in the target root.
+    /// Serialize to a file in the target root, and create a compatibility symlink.
     pub(crate) fn write_to(&self, root: &Dir) -> Result<()> {
         root.atomic_replace_with(BOOTC_ALEPH_PATH, |f| {
             anyhow::Ok(self.to_canon_json_writer(f)?)
         })
-        .context("Writing aleph version")
+        .context("Writing aleph version")?;
+        // Create a symlink so the aleph data is also accessible at the
+        // legacy path.
+        symlink_idempotent(root, BOOTC_ALEPH_PATH, ALEPH_SYMLINK)
+            .context("Creating aleph symlink")?;
+        Ok(())
     }
 }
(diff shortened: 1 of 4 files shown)


crates/lib/src/install/aleph.rs
```diff
@@ -80,11 +82,15 @@ impl InstallAleph {
...
-    /// Serialize to a file in the target root.
+    /// Serialize to a file in the target root, and create a compatibility symlink.
     pub(crate) fn write_to(&self, root: &Dir) -> Result<()> {
         root.atomic_replace_with(BOOTC_ALEPH_PATH, |f| {
             anyhow::Ok(self.to_canon_json_writer(f)?)
         })
-        .context("Writing aleph version")
+        .context("Writing aleph version")?;
+        // Create a symlink so the aleph data is also accessible via `.aleph-version.json`
+        root.symlink(BOOTC_ALEPH_PATH, BOOTC_ALEPH_SYMLINK)
```
jeckersb: I think this would fail if the symlink already exists for whatever reason.  Probably if the link already exists and it has the expected target this should return success?

cgwalters: In short "should be idempotent"

crates/lib/src/install/aleph.rs
```diff
@@ -11,6 +11,8 @@ use super::SELinuxFinalState;
 
 /// Path to initially deployed version information
 pub(crate) const BOOTC_ALEPH_PATH: &str = ".bootc-aleph.json";
+/// Compatibility symlink so the aleph data is also accessible as `.aleph-version.json`
+const BOOTC_ALEPH_SYMLINK: &str = ".aleph-version.json";
```
cgwalters: But it's not (IMO) a `BOOTC_ALEPH_` anymore right? We've just made up a new file name that's used as a kind of hidden API between osbuild and coreos and bootc right?

While it's accessible via git log anyways, I think the comment here should link to the FCOS tracker issue is I believe that's the canonical thing defining its existence/rationale - it's not directly part of this project.

cgwalters:
I guess I'll just keep repeating that I feel like the aleph stuff was a mistake, but we've also spent way too much time talking about it relative to its importance; 

cc https://github.com/coreos/fedora-coreos-tracker/issues/2187#issuecomment-4927815995

So anyways LGTM after issues above are addressed

crates/lib/src/install/aleph.rs
jbtrystram: updated

crates/lib/src/install/aleph.rs
```diff
@@ -80,11 +85,33 @@ impl InstallAleph {
...
+    /// Serialize to a file in the target root, and create a compatibility symlink.
     pub(crate) fn write_to(&self, root: &Dir) -> Result<()> {
         root.atomic_replace_with(BOOTC_ALEPH_PATH, |f| {
             anyhow::Ok(self.to_canon_json_writer(f)?)
         })
-        .context("Writing aleph version")
+        .context("Writing aleph version")?;
+        // Create a symlink so the aleph data is also accessible
+        // at the legacy path.
+        match root.symlink(BOOTC_ALEPH_PATH, ALEPH_SYMLINK) {
```
cgwalters: I would have factored out a little `symlink_idempotent` helper, something to do for the future

cgwalters (dismissed):

crates/lib/src/install/aleph.rs
jbtrystram: I just did that while addressing the missing DCO and other linter complains :)

cgwalters:
Sorry I meant that's something we can do as a separate PR, didn't have to block this one.

PR #2353, opened 2026-07-30
composefs+UKI: Try to generate image with matching digest 

cgwalters:
We had two changes in composefs-rs:

- xattr filtering
- v1 vs v2 EROFS

When we switched to v1 for the EROFS generation by default, it
broke UKI upgrades when the system was deployed with an older
bootc.

Unfortunately, there's no easy way to fix this (short of
reverting the change to v1 by default).

But at least with this approach, one can more easily temporarily
use a newer bootc on the client, and it will be able to adapt
to e.g. having an *older* bootc (perhaps using v2) in the target
environment.

Assisted-by: AI

### DIFF
--- a/crates/lib/src/bootc_composefs/boot.rs
+++ b/crates/lib/src/bootc_composefs/boot.rs
@@ -1112,6 +1112,122 @@ fn write_pe_to_esp(
     Ok(boot_label)
 }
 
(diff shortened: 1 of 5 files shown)


cgwalters:
Merging with https://github.com/bootc-dev/bootc/pull/2248

PR #2361, opened 2026-08-03
tmpfiles: Canonicalize declared tmpfiles.d paths against the rootfs

cgwalters:
It turned out that a systemd version skew caused systemd-tmpfiles to rerun, which created `/root/.ssh` - and that triggered our tmpfiles lint.

We need to handle the symlink aliases more strongly - do that by walking the generated paths and comparing device/inode.

Assisted-by: AI


jeckersb (approved):

cgwalters:
OK this one got ejected by just one flake, in the bib test. I'm following the plan to drop out the merge queue requiement for now.

PR #2362, opened 2026-08-03
ci: Lower default karma for Bodhi updates

cgwalters:
Per previous meeting, we have an ongoing conflict between the 1 week schedules for both things.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -149,3 +149,7 @@ jobs:
     dist_git_branches:
       # Fedora rawhide updates are created automatically
       - fedora-branched
+    # Our 1 week releases conflict with Fedora's default 1 week auto-shipping;
+    # Requiring 3 human identities to gate doesn't make sense when we have CI.
+    bodhi_extra_params:
+      stable_karma: 1


jeckersb (approved):

cgwalters:
@jeckersb FYI nothing will get through the queue until https://github.com/bootc-dev/bootc/pull/2361 lands at least

jeckersb:
> @jeckersb FYI nothing will get through the queue until #2361 lands at least

Yeah I realized that and jumped it to the front of the queue.  First time I've attempted to do that so we'll see how it works out!

cgwalters:
Yes, though note that will only work because I recently edited the queue policy per https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#managing-a-merge-queue to set "Only merge non-failing pull requests" to false. (Which itself should probably help actually! In fact one could say I probably messed up originally in thinking the merge queue would reduce flakes, but it wouldn't unless that's set to false)

Issue #2177, opened 2026-04-28
test flakes tracker


cgwalters:
# CI Flakes Tracker

## 1. COPR repo unavailable

```
Error: Failed to download metadata for repo
'copr:copr.fedorainfracloud.org:group_CoreOS:continuous':
Cannot download repomd.xml: All mirrors were tried
```
The COPR continuous repo is transiently unavailable during the container build,
causing `dnf` to fail. Happens specifically on centos-10 jobs that pull from
the CoreOS COPR. Retry would likely succeed.

## 2. quay.io 502

**Error:**
```
error: Installing to filesystem: Creating ostree deployment: Pulling:
Importing: Unencapsulating base: failed to invoke method GetBlob:
fetching blob: received unexpected HTTP status: 502 Bad Gateway
```
The in-VM install test hit a transient 502 from the container registry while
pulling an image. Pure network flake.

## 3. GHA runner killed mid-TMT

**Symptom:** Job ran unit tests successfully, TMT step started but produced no
conclusion after 2h15m. No artifact uploaded. Run marked failed with no logged
error. Runner appears to have been killed externally (OOM or host issue).

## 4. `test_container_write_derive` assertion failure

**Frequency:** Once (fedora-44, ostree, xfs, grub, bls, unsealed — Unit and container integration tests)  
**Error:**
```
assertion `left == right` failed
  left: "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"
 right: "newderivedfile v0"
at crates/ostree-ext/tests/it/main.rs:1713
```
File written into a derived container was read back as all null bytes. The same
test passed in `fedora-43, ostree` in the same queue batch. Likely a race
condition or tmpfs/storage timing issue on the fedora-44 runner; not seen
before or since.

jeckersb:
https://github.com/bootc-dev/bootc/actions/runs/27350791739/job/80817797747 we got:

```
Error: No space left on device : '/home/runner/actions-runner/cached/2.335.1/_diag/pages/c1de0eb0-4981-4a5f-8a5d-c197f4a11e80_760c16b0-69cb-4987-9094-c1d6fd2a21fd_1.log'
```

in the `test-integration (centos-10, ostree, xfs, grub, bls, unsealed)` job

cgwalters:
> Error: No space left on device 

[...]

PR #2363, opened 2026-08-04
ci: Dynamically re-gate heavy tests on PRs when no merge queue is active

cgwalters:
Our CI jobs were wired to the merge queue semantics, I broke this when dropping the merge queue requirement.

xref #2177

Change the logic to detect a merge queue dynamically, then if we decide to flip back on the queue things should work the same as before. But here we need to run the tests.

Also while we're here make use of the (new-ish) `ubuntu-slim` for jobs that are just computing state as that's cheaper and should be quicker to schedule.

Assisted-by: AI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,22 +43,58 @@ jobs:
   # Determine the OS matrix for CI jobs based on context:
   #   - merge_group / workflow_dispatch / ci/merge label: all OSes
   #   - ci/tier-1 label: centos-10 only (fast feedback on the primary target)
-  #   - plain PR: no heavy jobs
+  #   - PR with no merge queue configured: all OSes (the PR is the only gate,
+  #     so it must run the full suite; see the merge-queue check below)
+  #   - plain PR (merge queue active): no heavy jobs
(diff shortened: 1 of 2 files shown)


cgwalters:
OK let's get this one in and then cut another release to get the fix out for https://github.com/bootc-dev/bootc/issues/2364

jeckersb (dismissed):

cgwalters:
Ah and CI was failing here because I hadn't rebased, argh. Done now...

cgwalters:
OK this needs one more stamp

jeckersb (dismissed):

gursewak1997 (dismissed):

Johan-Liebert1:
I've disabled the UKI dumpfile test for now as it has other issues, especially in the upgrade path. upgrade/switch operations work, but the test itself needs some fixing. Also, I forgot to include the dumpfile print in the upgrade flow. I'll create another PR fixing that

Johan-Liebert1 (approved):

cgwalters:
Argh, two more flakes. It's annoying one can't restart the tests until all have finished.

PR #2365, opened 2026-08-04
composefs: Drop pull-time repo handle before running GC

### LINKED ISSUE #2364: [Composefs-native backend] bootc update/switch hangs after completing the staging process
While as far as I can tell the actual staging process completes fine, the command itself will hang when executed in an interactive terminal or script and never (?) exit.

Progress bars now display mostly correctly (although they often show 0:00 ETA immediately, not sure if that is just my net being fast or an actual issue); layers that already exist simply say "Already exists" instead of at least giving the layer hash but it is mostly functional. And of course, the composefs issue brought up in https://github.com/bootc-dev/bootc/issues/2324 is now fixed due to being patched in composefs-rs and the version of that being updated.

However, this new regression is still quite serious. For systems that have an update on a timer or on boot (which is common practice), this may prevent parts of the system updating or even inhibit booting. ~~I will update if anything goes bad when I boot into the new deployment that was downloaded but exploring it using `bootc internals cfs mount` seems not to show any problems.~~

Update: actually booting into the deployment works completely fine.

Steps to reproduce:
- Have bootc 1.16.7 on the composefs backend ~~(unified storage may be required?)~~
- Run `bootc update` or `bootc switch`
- Observe that the command does not exit even when left for a very long time
- However, the deployment that was downloaded is staged

cgwalters:
We were opening the repo twice, which caused a deadlock.

Fixes: https://github.com/bootc-dev/bootc/issues/2364

Assisted-by: AI

### DIFF
--- a/crates/lib/src/bootc_composefs/gc.rs
+++ b/crates/lib/src/bootc_composefs/gc.rs
@@ -488,6 +488,10 @@ pub(crate) async fn composefs_gc(
     // images for deployments that predate the manifest→image link;
     // once all deployments have been pulled with the new code, these
     // become redundant.
+    // `Repository::gc()` takes an exclusive flock on the repository file as its
+    // first action. Callers must ensure no other `Repository` handle on that
+    // same underlying file is still alive when this runs, or it will deadlock
+    // (see update.rs's do_upgrade() for an example of getting this wrong).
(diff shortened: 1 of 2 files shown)


Johan-Liebert1 (approved):
Oh, that was a hidden one, didn't realize

PR #2366, opened 2026-08-04
selinux: Ship policy module for install_t domain transitions

gursewak1997:
## Problem

When systemd directly runs a binary labeled `install_exec_t`, the SELinux policy transitions the process to `install_t` (which grants `mac_admin`). But when users wrap bootc in their own systemd services, the wrapper runs as `initrc_t` or `unconfined_service_t`, and there are no transition rules from those domains — so bootc never gets `install_t`. This silently breaks cross-major-version updates that need to write arbitrary SELinux labels.

This has been a pain point since the ostree days. Unlike rpm-ostree (a daemon where systemd handled the transition), bootc encourages users to wrap it in custom services where the entrypoint binary won't be `install_exec_t`.

Repro (from the issue):
```bash
# Direct exec — works:
systemd-run -u testservice -qPG -- /usr/bin/testservice  # → install_t

# Wrapper — broken:
systemd-run -u testservice -qPG -- /bin/sh -c '/usr/bin/testservice && echo done'  # → initrc_t
```

## Solution

Ship a small SELinux policy module that adds the missing `type_transition` rules from `initrc_t` and `unconfined_service_t` to `install_t` via `install_exec_t`. The module is built during the RPM build and loaded via `semodule` in `%post`.

Ref: RHEL-117256

Assisted-by: AI

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,6 +50,14 @@ completion: bin
 		target/release/bootc completion $$shell > target/completion/bootc.$$shell; \
 	done
 
+SELINUX_PP = contrib/selinux/bootc.pp
+.PHONY: selinux
+selinux: $(SELINUX_PP)
+$(SELINUX_PP): contrib/selinux/bootc.te
+	checkmodule -M -m -o contrib/selinux/bootc.mod $<
+	semodule_package -o $@ -m contrib/selinux/bootc.mod
+	rm -f contrib/selinux/bootc.mod
+
 STORAGE_RELATIVE_PATH ?= $(shell realpath -m -s --relative-to="$(prefix)/lib/bootc/storage" /sysroot/ostree/bootc/storage)
 install: completion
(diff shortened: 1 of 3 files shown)


cgwalters:
Thanks for working on this!

I think we need an integration test for this. Also it'd be good to "crosscheck" that the test fails without this patch.

I'm OK shipping this here but IMO it actually has nothing to do with bootc at all - this is purely a (Fedora?) SELinux policy issue and as I said in the Jira I think it could affect e.g. dnf too.

I'd like to at least spend some time/tokens on what a patch for [the base selinux-policy](https://github.com/fedora-selinux/selinux-policy) would look like.

gursewak1997:
> I'd like to at least spend some time/tokens on what a patch for [the base selinux-policy](https://github.com/fedora-selinux/selinux-policy) would look like.

I opened https://github.com/fedora-selinux/selinux-policy/pull/3327

PR #2368, opened 2026-08-05
lsm: treat `unlabeled_t` SELinux context as unlabeled for relabeling

joelcapitao:
Modify `has_security_selinux()` to return `Unlabeled` when a file has the `unlabeled_t` SELinux type. This ensures files created in permissive mode (where the kernel does not apply type transitions) get properly relabeled during `bootc install to-filesystem`.

Previously, files with `unlabeled_t` were considered "already labeled" and skipped during relabeling passes, causing SELinux policy mismatches on the final disk image.

Note that `/sysroot/boot` cannot be relabeled when it is masked by a mounted filesystem (e.g., when osbuild pre-mounts the boot partition). This is acceptable because `/sysroot/boot` is primarily used as a potential mount point for the boot partition, created by `ostree admin init-fs`. In the final booted system, `/boot` is mounted separately and its contents are properly labeled.

### DIFF
--- a/crates/lib/src/lsm.rs
+++ b/crates/lib/src/lsm.rs
@@ -1,4 +1,5 @@
 use std::borrow::Cow;
+use std::ffi::CString;
 use std::io::Write;
 use std::os::fd::AsRawFd;
 use std::os::unix::process::CommandExt;
@@ -23,7 +24,16 @@ use rustix::fd::AsFd;
 const SELINUXFS: &str = "/sys/fs/selinux";
 /// The SELinux xattr
 const SELINUX_XATTR: &[u8] = b"security.selinux\0";
+/// The kernel initial SID used for objects without a label.
+const SELINUX_INITIAL_SID_UNLABELED: &str = "unlabeled";
 const SELF_CURRENT: &str = "/proc/self/attr/current";
+
+fn unlabeled_type() -> Result<CString> {
+    let context =
+        selinux::SecurityContext::of_initial_kernel_context(SELINUX_INITIAL_SID_UNLABELED, true)
+            .context("Querying the SELinux unlabeled initial context")?;
+    context_type(context.as_bytes())
+}
 
 #[context("Querying selinux availability")]
 pub(crate) fn selinux_enabled() -> Result<bool> {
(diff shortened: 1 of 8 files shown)


crates/lib/src/lsm.rs
```diff
@@ -262,7 +262,17 @@ pub(crate) fn has_security_selinux(root: &Dir, path: &Utf8Path) -> Result<SELinu
     let mut buf = [0u8; 2048];
     let fdpath = format!("/proc/self/fd/{}/{path}", root.as_raw_fd());
     match rustix::fs::lgetxattr(fdpath, "security.selinux", &mut buf) {
-        Ok(_) => Ok(SELinuxLabelState::Labeled),
+        Ok(len) => {
+            // Check if the label is unlabeled_t - treat it as unlabeled.
+            // This can happen when files are created with SELinux in permissive mode,
+            // where the kernel doesn't apply type transitions and assigns unlabeled_t.
```
cgwalters, line 268: I don't think it's true that permissive mode doesn't do type transitions

crates/lib/src/lsm.rs
```diff
@@ -262,7 +262,17 @@ pub(crate) fn has_security_selinux(root: &Dir, path: &Utf8Path) -> Result<SELinu
...
+            let label = std::str::from_utf8(&buf[..len]).unwrap_or("");
+            if label.contains(":unlabeled_t:") {
```
cgwalters, line 270: I'm not happy about parsing security contexts like this, there's official APIs for this

I also don't want to hardcode unlabeled_t ideally.

Hmmmmmm...I think what's giong on here is the kernel returns this type when the xattr is missing - maybe ideally we can figure out a way to query that state. I bet the stock selinux userspace tooling has code for this.

crates/lib/src/lsm.rs
```diff
@@ -262,7 +262,17 @@ pub(crate) fn has_security_selinux(root: &Dir, path: &Utf8Path) -> Result<SELinu
     let mut buf = [0u8; 2048];
     let fdpath = format!("/proc/self/fd/{}/{path}", root.as_raw_fd());
     match rustix::fs::lgetxattr(fdpath, "security.selinux", &mut buf) {
-        Ok(_) => Ok(SELinuxLabelState::Labeled),
+        Ok(len) => {
+            // Check if the label is unlabeled_t - treat it as unlabeled.
+            // This can happen when files are created with SELinux in permissive mode,
+            // where the kernel doesn't apply type transitions and assigns unlabeled_t.
+            let label = std::str::from_utf8(&buf[..len]).unwrap_or("");
```
cgwalters, line 269: Silently discarding non-UTF8 security contexts feels odd, we can just scan bytes

joelcapitao:
Thank you Colin for the review. Unfortunately, I haven't had the time to amend it with your suggestions (RHCOS on Konflux is high prio) and will be on PTO starting tomorrow for 2 weeks. I'll work on that once back.

cgwalters:
To emphasize I think it is not correct to parse SELinux security contexts in userspace at all, and especially not to hardcode specific type names. The idea is access decisions are centralized, and there are APIs for interacting with policy.

But digging in a bit...what we're doing here in bootc is somewhat unusual in that we're trying to special case "not labeled at all". It's quite tricky because on a system without SELinux enabled (say an Ubuntu host targeting a bootc SELinux-enabled host) when we call `getxattr("security.selinux")` we get `ENODATA` which this code handles.

What we get if SELinux is enabled on the host though is this `unlabeled_t` (conceptually on a SELinux system, *all* objects have security contexts and `unlabeled_t` is a stand-in for unknown).

I think though we *can* get from the policy is this unlabeled type. I am having an agent dig at this.

cgwalters:
OK I pushed some changes to this. @joelcapitao want to look? @jmarrero can you approve?

jmarrero (approved):
lgtm

supakeen:
I think the initial guard here checks for the hosts (?) `sys/fs/selinux/enforce` to exist but then wants to read from `sys/fs/selinux/initial_contexts`. The existence of the first is no guarantee that the second is available and this probably is the cause of issues (https://github.com/coreos/fedora-coreos-tracker/issues/2219).

This /might/ be related to the cached selinux_mnt in libselinux for the process; it would be set to NULL if selinuxfs is read-only (https://github.com/SELinuxProject/selinux/blob/main/libselinux/src/init.c#L30-L51). There's supposedly a re-exec in there but I don't think it happens at all with an `unconfined_t` because `test_install_t` runs `chcon` which is a direct `setattr`?

cgwalters:
Thanks, let's move discussion of this to https://github.com/bootc-dev/bootc/issues/2444

PR #2370, opened 2026-08-07
Uki print dump diff fix

Johan-Liebert1:
uki: Fix dumpfile diffing

The dumpfile diff was not being performed in the update/switch
operations which was causing the dumpfile test to fail. Refactor out the
diffing code to also run in the update/switch operations.

Another major issue was `/boot` being masked in the EROFS which caused
us to not find the dumpfile when reading from the EROFS. Update to read
from the filesystem, create tmpfiles and run diff on the tmpfiles

---

tmt: Fix and re-enable uki-dumpfile test

We were not computing the bootable digest which caused assertions to
fail

### DIFF
--- a/crates/lib/src/bootc_composefs/boot.rs
+++ b/crates/lib/src/bootc_composefs/boot.rs
@@ -64,7 +64,8 @@
 use std::cell::Cell;
 use std::fs::create_dir_all;
 use std::io::{Read, Seek, SeekFrom, Write};
-use std::path::{Path, PathBuf};
+use std::os::fd::AsFd;
+use std::path::Path;
 use std::sync::Arc;
 
 use anyhow::{Context, Result, anyhow, bail};
@@ -77,7 +78,7 @@ use cap_std_ext::{
 use clap::ValueEnum;
 use composefs::fs::read_file;
 use composefs::fsverity::{FsVerityHashValue, Sha512HashValue};
-use composefs::tree::RegularFile;
+use composefs::tree::{FileSystem, RegularFile};
 use composefs_boot::bootloader::{
     BootEntry as ComposefsBootEntry, EFI_ADDON_DIR_EXT, EFI_ADDON_FILE_EXT, EFI_EXT, PEType,
     UsrLibModulesVmlinuz, get_boot_resources,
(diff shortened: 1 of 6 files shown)


Johan-Liebert1:
:/ grub-cc tests failing due to some upstream bug where LoaderInfo efivar returns Grub as the bootloader

Johan-Liebert1:
It was probably this commit https://src.fedoraproject.org/rpms/grub2/c/fedae6ebf3e06c6a02dd8aa90e4277292575eb86?branch=rawhide

hack/provision-fetch.sh
```diff
@@ -6,9 +6,7 @@
 # This script is idempotent: re-running it after a partial failure is safe.
 set -xeu
 
-cloudinit=0
 case ${1:-} in
-  cloudinit) cloudinit=1 ;;
```
gursewak1997, line 11: `hack/Containerfile.packit:32` still calls `./provision-derived.sh` cloudinit, hich forwards "$@" to this script. It will now hit the * branch and exit 1, breaking the Packit container build. 
Either keep cloudinit) as a no-op here, or also update provision-derived.sh and Containerfile.packit.

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -144,6 +148,93 @@ pub(crate) struct UKIDigestMismatch {
...
+        return;
+    };
+
+    let Ok(tempdir) = tempfile::tempdir() else {
+        return;
+    };
+    let path = tempdir.path();
+    let Ok(tempdir_cap) = Dir::open_ambient_dir(path, ambient_authority()) else {
+        return;
+    };
```
gursewak1997, line 178: AI caught: the `let Ok(x) = ... else { return; }` pattern silently swallows errors. This function has ~5 early returns that discard the error with no logging (tempfile::tempdir(), Dir::open_ambient_dir, both tempdir_cap.create() calls, write_all). 

The `read_dumpfile_from_fs` and `write_dumpfile` failures already use `tracing::debug!` which is the right pattern; let's do the same for the rest

gursewak1997 (commented):
Looks good mostly; just a few nits

hack/provision-fetch.sh
Johan-Liebert1: Set it up as a no-op

gursewak1997 (dismissed):

Johan-Liebert1:
https://github.com/bootc-dev/bootc/actions/runs/31562640440/job/94023684221?pr=2370 is failing due to the deadlock in https://github.com/bootc-dev/bootc/pull/2315 and ig we don't have a new enough bootc in the base images yet

gursewak1997 (approved):

PR #2380, opened 2026-08-13
GC UKI and UKI Addons from Composefs Repository

### LINKED ISSUE #373: erofs: Save `/boot's` structure when creating an image using `--bootable`
There are times when we would need to figure out what was originally in `/boot` before we mask it with `--bootable`. One notable example is listing UKI Addons that are/were present in the image. We can of course reconstruct the entire filesystem using the splitstream but that feels extremely wasteful.

I'm proposing a small EROFS (or maybe even just a dumpfile) of `/boot` stored alongside the EROFS

Johan-Liebert1:
cfs/gc: GC UKI and UKI Addons

Make use of the non-bootable EROFS to GC everything in the boot
directory of the deployment being GC'd. This is only done for UKI boots
as for normal Type1 booted systems, there should be nothing in /boot

---

tmt: Add tests for .boot images and UKI asset cleanup

Test whether UKI assets are being cleaned up from the objects directory
during GC

---

cfs/filesystem: Pass manifest verity to create_filesystem
There are two reasons to do this

1. We skip the expensive operation of reading every layer and computing
   its hash to preserve integrity.

2. We end up also reading whiteouts in `/boot` which end up erroring out
   since those objects don't exist anymore

### DIFF
--- a/crates/lib/src/bootc_composefs/gc.rs
+++ b/crates/lib/src/bootc_composefs/gc.rs
@@ -494,8 +494,13 @@ pub(crate) async fn composefs_gc(
     // These won't be GC'd by the above `repo.gc` as the EROFS have
     // `/boot` masked
     for verity in all_orphans {
-        let verity_to_sha = Sha512HashValue::from_hex(verity)
-            .with_context(|| anyhow::anyhow!("Bad fsverity {verity}"))?;
+        let verity_to_sha = Sha512HashValue::from_hex(verity);
+
+        let Ok(verity_to_sha) = verity_to_sha else {
+            // We do not want to block on GC if we find corrupted state
+            tracing::warn!("Bad FSVerity '{verity}' when performing GC on orphan, skipping");
+            continue;
+        };
 
         let linked_images = linked_erofs_images(&booted_cfs.repo, &verity_to_sha)
             .with_context(|| anyhow::anyhow!("Getting linked images for {verity}"))?;
--- a/crates/lib/src/bootc_composefs/gc.rs
+++ b/crates/lib/src/bootc_composefs/gc.rs
@@ -12,6 +12,10 @@ use composefs_boot::bootloader::EFI_EXT;
 use composefs_ctl::composefs;
 use composefs_ctl::composefs_boot;
 use composefs_ctl::composefs_oci;
+use ostree_ext::composefs::fsverity::Sha512HashValue;
+use ostree_ext::composefs_oci::linked_erofs_images;
+use rustix::fs::AtFlags;
+use rustix::fs::{statat, unlinkat};
 
 use crate::{
     bootc_composefs::{
@@ -347,7 +351,7 @@ pub(crate) async fn composefs_gc(
 
     // Collect the set of manifest digests referenced by live deployments,
     // and track EROFS image verities as fallback additional_roots for
-    // deployments that predate the manifest→image link.
+    // deployments that predate the manifest -> image link.
     let mut live_manifest_digests: Vec<composefs_oci::OciDigest> = Vec::new();
     let mut additional_roots = Vec::new();
     // Container image names for containers-storage pruning.
@@ -367,7 +371,7 @@ pub(crate) async fn composefs_gc(
         }
 
         // Keep the EROFS image as an additional root until all deployments
-        // have manifest→image refs. Once a deployment is pulled with the
+        // have manifest -> image refs. Once a deployment is pulled with the
         // new code, its EROFS image is reachable from the manifest and
         // this entry becomes redundant (but harmless).
         additional_roots.push(verity.clone());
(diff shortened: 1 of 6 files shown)


cgwalters:
Why can't we just read the non-bootable tree? That's already supported in composefs-rs directly - the same filesystem tree one would get to run as a container image. Yes, we won't have e.g. SELinux labeling, but that's not a problem.

cgwalters (changes requested):
See above

Johan-Liebert1:
> Why can't we just read the non-bootable tree?

I didn't even realise we saved this :|
We can totally just use this image

Johan-Liebert1:
Okay, going through the code a bit more, we're doing things extremely inefficiently. We're recreating the filesystem multiple times, one for non-bootable fs and once for bootable. I had some optimizations here https://github.com/bootc-dev/bootc/pull/2380/changes/edb2ef01d7cce71f2b5219d15f8cb0d05742ab58, but the current stuff requires some more refactoring

cgwalters:
> Okay, going through the code a bit more, we're doing things extremely inefficiently. We're recreating the filesystem multiple times, one for non-bootable fs and once for bootable.

What's inefficient about that? It's just O(metadata) so I'd be surprised if it was taking a noticeable amount of CPU etc.

Johan-Liebert1:
`image::create_filesystem` does untar and parse for all layers. This is mostly on the bootc side since we're not returning the created fs from composefs APIs the same filesystem is being created twice for one install/upgrade, one for the bootable erofs and one for the non-bootable. Also, currently we're not even passing the verity digest of the image to `create_filesystem` which hashes all of the layers for verification

cgwalters:
> image::create_filesystem does untar and parse for all layers.

Hmm, but we only need to process the tar *metadata* not the data, that seems like a plain bug to fix in composefs-rs - like we want a tar reading path that only gives headers from split stream.

> Also, currently we're not even passing the verity digest of the image to create_filesystem which hashes all of the layers for verification

I don't understand this one.

But at this point can you spawn an agent on this and let's move this to composefs-rs?

All that needs to be done on the bootc side I think is to read the non-bootable image to access `/boot`, and we fix any perf issues in composefs-rs.

Johan-Liebert1:
> I don't understand this one.

Here https://github.com/bootc-dev/bootc/blob/main/crates/lib/src/bootc_composefs/boot.rs#L1486, we pass config verity as `None`, which in turn invokes https://github.com/composefs/composefs-rs/blob/main/crates/composefs-oci/src/image.rs#L124

> But at this point can you spawn an agent on this and let's move this to composefs-rs?

This would be a combined effort really. I think the easiest solution is to simply return the created fs from composefs APIs and have other functions take in an `Option<Filesystem<..>>` as a param. But yeah, that can be a followup

Johan-Liebert1:
Also, regarding just this PR, I don't believe there's any way we link the bootable and non-bootable EROFS-es? I see we have them as GC links in the config splitstream, but that requires reading the splitstream metadata. Should we create a ref in bootc for easier accessibility, similar to a ref I'm creating here https://github.com/bootc-dev/bootc/pull/2380/changes/6660f94d371e8e6e717fb983e66278a9d8dd63b8?

cgwalters:
Hmm...yes we may be missing APIs for this, will look

Johan-Liebert1:
@cgwalters I think this should be good now

crates/lib/src/bootc_composefs/gc.rs
```diff
@@ -484,6 +488,83 @@ pub(crate) async fn composefs_gc(
         }
     }
 
+    let (mut objects_bytes, mut objects_removed) = (0, 0);
+
+    // Now GC the UKI/UKI Addons from `.boot` EROFS if we have them
```
cgwalters, line 493: We don't have any integration tests for this though...a bit of a risk

crates/lib/src/bootc_composefs/gc.rs
```diff
@@ -484,6 +488,83 @@ pub(crate) async fn composefs_gc(
...
+            };
+
+            tracing::debug!(
+                "{}: objects/{path:?}",
+                if gc_opts.dry_run {
+                    "would remove"
+                } else {
+                    "removing"
+                },
+            );
```
cgwalters, line 550: This construction is a bit awkward I think it'd be simpler to use structured logging and pass `dry_run=gc_opts.dry_run` or so

crates/lib/src/bootc_composefs/gc.rs
```diff
@@ -484,6 +488,83 @@ pub(crate) async fn composefs_gc(
...
+            }
+
+            // Get file size before removing
+            if let Ok(stat) = statat(&objects_dir, path.as_ref(), AtFlags::empty()) {
+                objects_bytes += stat.st_size as u64;
+            }
+
+            objects_removed += 1;
+
+            unlinkat(&objects_dir, path.as_ref(), AtFlags::empty())
```
cgwalters, line 563: Here and elsewhere we need to be able to recover from corrupted state - in this particular case should gracefully handle ENOENT

cgwalters (approved):
See above

crates/lib/src/bootc_composefs/gc.rs
Johan-Liebert1: We do here https://github.com/bootc-dev/bootc/pull/2380/changes#diff-113194a272e290654a78e8fe283666117ac5bd7391e8813d3b269937f0fd15e1

crates/lib/src/bootc_composefs/gc.rs
Johan-Liebert1: This is something that I'm unsure about. I think the GC op should be aware of what exists and what does not and we should definitely error out if something we absolutely expected does not exist. I think it's worth discussing in the weekly meet

Related #2442

PR #2381, opened 2026-08-14
docs: Document tmpfs-backed storage for usr-overlay

jmarrero:
The transient overlay created by `bootc usr-overlay` is backed by a
tmpfs filesystem, meaning installed packages consume system RAM rather
than persistent disk.  The kernel default sizes this tmpfs at 50% of
physical RAM, and there is currently no pre-flight space check before
the overlay fills up.

This is not obvious from the existing documentation and can lead to
confusing ENOSPC errors on memory-constrained systems such as CI
runners or small VMs, where users may assume the overlay uses disk
storage.

Assisted-by: AI
I am knowledgeable in this problem domain and reviewed the change carefully.

### DIFF
--- a/docs/src/man/bootc-usr-overlay.8.md
+++ b/docs/src/man/bootc-usr-overlay.8.md
@@ -19,6 +19,23 @@ A common pattern is wanting to use tracing/debugging tools, such as
 such as `apt` or `dnf` can apply changes into this transient overlay
 that will be discarded on reboot.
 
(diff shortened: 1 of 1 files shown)


Johan-Liebert1 (approved):

cgwalters (approved):
Hmm you know I think we ended up changing this in ostree at some point, or tried to for similar reasons.

PR #2382, opened 2026-08-14
tmpfiles: Tolerate EACCES when resolving declared paths

### LINKED ISSUE #2379: `bootc container lint` fails with EACCES on `/sys/fs/selinux/checkreqprot` under rootless podman — regression in 1.16.7 (var-tmpfiles)
**Environment**
- `bootc` 1.16.7 (regression; 1.16.6 works)
- `podman` 5.8.4, rootless, `crun` runtime
- Host: Fedora 43, unprivileged user (no `--privileged`)
- Originally found on RHEL `rhel-bootc` images; fully reproduced below with public CentOS Stream bootc images

**Description**

`bootc container lint` fails with a fatal error instead of a warning when running inside an unprivileged/rootless container, because the `var-tmpfiles` lint now tries to access `/sys/fs/selinux/checkreqprot` and gets `EACCES` in that environment.

```
error: Linting: Unexpected runtime error running lint var-tmpfiles: I/O error on /sys/fs/selinux/checkreqprot: Permission denied (os error 13)
```

`selinux-policy` ships a tmpfiles.d rule for that path:
```
/usr/lib/tmpfiles.d/selinux-policy.conf:4:w /sys/fs/selinux/checkreqprot - - - - 0
```

**Reproducer**

I used public images and bootc builds, without special privileges. Current CentOS bootc images don't ship with bootc 1.16.7, so I installed it directly from Stream Koji. Specifically, I used `centos-bootc:c10s` tag, with SHA [2b7e3b1abf8d](https://quay.io/repository/centos-bootc/centos-bootc/manifest/sha256:2b7e3b1abf8db094d1efb083721dc0f72e6feeef2355fc16ea010d0266b2bb95)

Using bootc 1.16.6 (as shipped in the image) — passes with an unrelated warning
```
-> podman run --rm quay.io/centos-bootc/centos-bootc:c10s sh -c "rpm -q bootc && bootc container lint"
bootc-1.16.6-1.el10.x86_64
Lint warning: var-tmpfiles: Found non-directory/non-symlink files in /var:
  var/roothome/buildinfo/content-sets.json

Checks passed: 12
Checks skipped: 1
```

Upgrade to bootc 1.16.7 (public CentOS Stream koji build) — fails
```
-> podman run --rm quay.io/centos-bootc/centos-bootc:c10s sh -c "rpm -U --nodeps https://kojihub.stream.centos.org/kojifiles/vol/koji02/packages/bootc/1.16.7/1.el10/x86_64/bootc-1.16.7-1.el10.x86_64.rpm && rpm -q bootc && bootc container lint"
bootc-1.16.7-1.el10.x86_64
error: Linting: Unexpected runtime error running lint var-tmpfiles: I/O error on /sys/fs/selinux/checkreqprot: Permission denied (os error 13)
```

**Additional information**

`selinux-policy`/`libselinux` are unchanged between the two runs (only `bootc` itself was upgraded in place via the RPM above), so this is not a selinux-policy change — it's purely a bootc code change between 1.16.6 and 1.16.7.

Bisecting the release, the commit `cfec5cc0` ("tmpfiles: Canonicalize declared tmpfiles.d paths against the rootfs") seems to have relation to this change. It changed `var-tmpfiles` to walk and canonicalize (stat/inode-compare) every path declared in tmpfiles.d — including entries under `/sys` that were previously never touched by this lint. When that stat/open hits `EACCES` (as it does for `/sys/fs/selinux/checkreqprot` under an unprivileged container), the new code path treats it as an unexpected/fatal error rather than tolerating it.

This looks similar to #1481 (var-tmpfiles crashing on an unhandled I/O errno — there it was `ENOSYS` under QEMU emulation, here it's `EACCES` under rootless podman): the lint needs to tolerate "can't stat this path in this sandboxed environment" for paths outside the actual container rootfs content, rather than propagating any I/O error as fatal.

I also tried to reproduce it with `quay.io/fedora/fedora-bootc:latest`, but the lint passes there. Fedora's `selinux-policy` (`selinux-policy-44.5-1.fc44`) does not ship a tmpfiles.d entry for `checkreqprot` at all — that entry appears to be specific to RHEL/CentOS Stream's `selinux-policy`. This is consistent with the root cause: the bug is triggered by any tmpfiles.d entry pointing at a restricted/virtual path under `/sys`, not by something Fedora-specific in bootc's code.

**Impact**

This breaks any CI/build pipeline that runs `bootc container lint` as part of `podman build` without extra privileges — a very common setup — turning a previously-working build into a hard failure on upgrade to 1.16.7.

**Workaround**

`bootc container lint --skip var-tmpfiles`

gursewak1997:
## Summary

The path canonicalization added in cfec5cc0 opens the parent directory of every declared tmpfiles.d path to get its `(dev, ino)` identity. Entries like selinux-policy's `w /sys/fs/selinux/checkreqprot` point at virtual filesystem paths that return EACCES in rootless containers. Treat these the same as a missing parent — return `None` — since the `/var` walker cannot encounter inaccessible paths either.

Fixes: https://github.com/containers/bootc/issues/2379

## Test plan

- Unit test `test_resolve_parent_identity_permission_denied` creates a mode-0o000 directory and verifies EACCES returns `Ok(None)` instead of a fatal error
- Verified patched binary passes `bootc container lint` in a rootless `quay.io/centos-bootc/centos-bootc:c10s` container (which ships the triggering `selinux-policy.conf` entry)

Assisted-by: AI

### DIFF
--- a/crates/tmpfiles/src/path_resolution.rs
+++ b/crates/tmpfiles/src/path_resolution.rs
@@ -110,10 +110,12 @@ impl PathResolver {
             return Ok(Some(PathIdentity::new(self.root_dev, self.root_ino, leaf)));
         }
 
-        let Some(parent_file) = self.root_dir.open_optional(parent).map_err(to_err)? else {
-            // Some component of the parent doesn't exist: see the docs above
-            // on why `None` is correct here.
-            return Ok(None);
+        let parent_file = match self.root_dir.open_optional(parent) {
+            Ok(Some(f)) => f,
+            Ok(None) => return Ok(None),
+            // Tolerate inaccessible paths (EACCES in rootless containers).
+            Err(e) if e.kind() == std::io::ErrorKind::PermissionDenied => return Ok(None),
+            Err(e) => return Err(to_err(e)),
         };
         let parent_meta = parent_file.metadata().map_err(to_err)?;
         Ok(Some(PathIdentity::new(
@@ -126,6 +128,8 @@ impl PathResolver {
 
 #[cfg(test)]
 mod tests {
+    use cap_std::fs::PermissionsExt as _;
+
     use super::*;
 
     fn newroot() -> Result<cap_std_ext::cap_tempfile::TempDir> {
@@ -264,6 +268,23 @@ mod tests {
             .unwrap_err();
         assert!(matches!(err, Error::PathIo { .. }), "{err:?}");
 
+        Ok(())
+    }
+
+    #[test]
+    fn test_resolve_parent_identity_permission_denied() -> anyhow::Result<()> {
+        if rustix::process::getuid().is_root() {
+            return Ok(());
+        }
+        let rootfs = &newroot()?;
+        rootfs.create_dir("noaccess")?;
+        rootfs.set_permissions("noaccess", cap_std::fs::Permissions::from_mode(0o000))?;
+        let resolver = PathResolver::new(rootfs)?;
+        let resolved = resolver.resolve_parent_identity(Path::new("/noaccess/child/leaf"))?;
+        assert_eq!(resolved, None, "EACCES parent should resolve to None");
+
+        // Restore permissions so the temp dir can be cleaned up.
+        rootfs.set_permissions("noaccess", cap_std::fs::Permissions::from_mode(0o755))?;
         Ok(())
     }
 


crates/tmpfiles/src/path_resolution.rs
```diff
@@ -110,10 +110,20 @@ impl PathResolver {
...
 
-        let Some(parent_file) = self.root_dir.open_optional(parent).map_err(to_err)? else {
-            // Some component of the parent doesn't exist: see the docs above
-            // on why `None` is correct here.
-            return Ok(None);
+        let parent_file = match self.root_dir.open_optional(parent) {
+            Ok(Some(f)) => f,
+            Ok(None) => return Ok(None),
+            // Tolerate inaccessible paths (EACCES in rootless containers,
+            // ENOSYS under QEMU emulation).
```
cgwalters: ENOSYS is a totally different thing, I don't think we should conflate with EPERM.

In the qemu/ENOSYS case the functionality won't work *at all* for *any file*. I think it'd be cleaner for that case to do a "does it work" probe and have a clear warning if it doesn't.

It's just a different bug.

cgwalters (changes requested):

cgwalters (approved):

PR #2384, opened 2026-08-17
install: Add --preserve-var and --merge-etc to to-existing-root

ckyrouac:
This is AI generated with lots of guidance.

---

Add two flags to `bootc install to-existing-root` that are most useful
when migrating a package-mode system to a bootc image built externally:

  --preserve-var   Copy the running system's /var data into the new
                    deployment, and write a package-mode GRUB rollback
                    boot entry.

  --merge-etc      Apply the running system's /etc customisations onto
                    the new deployment via a 3-way merge.

Typical usage:

  podman run --rm --privileged --pid=host --user=root:root \
      -v /dev:/dev --security-opt label=type:unconfined_t \
      -v /:/target -v /var/lib/containers:/var/lib/containers \
      <fleet-image> \
      bootc install to-existing-root \
          --preserve-var \
          --merge-etc \
          --acknowledge-destructive

### DIFF
--- a/crates/etc-merge/src/lib.rs
+++ b/crates/etc-merge/src/lib.rs
@@ -788,7 +788,11 @@ fn merge_leaf(
     };
 
     if matches!(new_inode, Some(Inode::Directory(..))) {
-        anyhow::bail!("Modified config file {file:?} newly defaults to directory. Cannot merge")
+        tracing::warn!(
+            "Modified config file {file:?} newly defaults to a directory in the new image; \
+             keeping the image's directory and skipping host customization"
+        );
+        return Ok(());
     };
 
     // If a new file with the same path exists, we delete it
@@ -800,9 +804,23 @@ fn merge_leaf(
         // Using rustix's symlinkat here as we might have absolute symlinks which clash with ambient_authority
         symlinkat(&**target, new_etc_fd, file).context(format!("Creating symlink {file:?}"))?;
     } else {
-        current_etc_fd
+        let copy_result = current_etc_fd
             .copy(&file, new_etc_fd, &file)
-            .with_context(|| format!("Copying file {file:?}"))?;
+            .with_context(|| format!("Copying file {file:?}"));
+        // If a path component in the current /etc leads through an absolute symlink that
+        // escapes the cap-std root (e.g. /etc/alternatives symlinks), cap-std raises
+        // "a path led outside of the filesystem".  Warn and skip rather than aborting the
+        // whole merge; the image's existing content wins for that path.
+        if let Err(ref e) = copy_result {
+            if e.to_string().contains("a path led outside of the filesystem") {
+                tracing::warn!(
+                    "Skipping {file:?}: path escapes /etc sandbox (absolute symlink component); \
+                     keeping image's version"
+                );
+                return Ok(());
+            }
+        }
+        copy_result?;
     };
 
     rustix::fs::chownat(
@@ -916,8 +934,8 @@ pub fn merge(
     .context("Merging modified files")?;
 
     for removed in &diff.removed {
-        // Use symlink_metadata_optional so that symlinks that resolve to a path
-        // outside the new_etc_fd don't get followed
+        // Use symlink_metadata (lstat) so we don't follow absolute symlinks out
+        // of the cap-std sandbox (e.g. /etc/ssl/cert.pem → /etc/pki/…).
         let stat = new_etc_fd.symlink_metadata_optional(&removed)?;
 
         let Some(stat) = stat else {
@@ -1291,11 +1309,10 @@ mod tests {
 
         let merge_res = merge(&c, &current_etc_files, &n, &new_etc_files.unwrap(), &diff);
 
-        assert!(merge_res.is_err());
-        assert_eq!(
-            merge_res.unwrap_err().root_cause().to_string(),
-            "Modified config file \"file-to-dir\" newly defaults to directory. Cannot merge"
-        );
+        // The image's directory wins over the host's modified file; merge succeeds with a warning.
+        assert!(merge_res.is_ok(), "Expected merge to succeed: {:?}", merge_res);
+        // The directory should still exist in new_etc (image's directory wins)
+        assert!(n.metadata("file-to-dir").unwrap().is_dir());
 
         Ok(())
     }
--- a/crates/lib/src/install/migrate.rs
+++ b/crates/lib/src/install/migrate.rs
@@ -1,0 +1,630 @@
+//! # Package-mode to image-mode migration helpers
+//!
+//! This module implements the post-install state-preservation steps that make
+//! `bootc install to-existing-root` useful for converting a live package-mode
+//! (RPM/DEB) system to a bootc image-mode deployment without losing data.
+//!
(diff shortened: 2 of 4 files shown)


crates/etc-merge/src/lib.rs
```diff
@@ -800,9 +804,23 @@ fn merge_leaf(
...
+        let copy_result = current_etc_fd
             .copy(&file, new_etc_fd, &file)
-            .with_context(|| format!("Copying file {file:?}"))?;
+            .with_context(|| format!("Copying file {file:?}"));
+        // If a path component in the current /etc leads through an absolute symlink that
+        // escapes the cap-std root (e.g. /etc/alternatives symlinks), cap-std raises
+        // "a path led outside of the filesystem".  Warn and skip rather than aborting the
+        // whole merge; the image's existing content wins for that path.
+        if let Err(ref e) = copy_result {
+            if e.to_string().contains("a path led outside of the filesystem") {
```
cgwalters, line 815: Let's avoid string matching errors, we can use downcasts

But perhaps better we should be using https://docs.rs/cap-std-ext/5.1.2/cap_std_ext/struct.RootDir.html ? And yes we'd need a `copy` api there.

Perhaps simplest to open the parent dir rooted, then do a copy that way

crates/lib/src/install/migrate.rs
```diff
@@ -0,0 +1,630 @@
...
+//! needs to be transferred to make the migrated system functional.
+//!
+//! ## Rollback BLS entry (`--preserve-var` also saves the kernel)
+//!
+//! When `--preserve-var` is passed, the running kernel and initramfs are saved
+//! to `<root_path>/var/lib/pkgmode-rollback/` **before** the install wipes
+//! `/boot`.  After the install a third BLS entry (`pkgmode-rollback.conf`) is
+//! written into the active `loader.N/entries/` directory so GRUB presents a
+//! "Previous OS" option.  The entry is invisible to ostree (which only reads
+//! `ostree-*.conf` files) and to bootupd (which does not touch `loader/entries/`).
```
cgwalters, line 61: This is kind of hacky, I think it'd be cleaner to just not delete the entries (and kernel/initramfs) at all which should be a separate option.

crates/lib/src/install/migrate.rs
```diff
@@ -0,0 +1,630 @@
...
+/// `root_path` is the host root as seen from inside the install container.
+/// `kver` is the running kernel version string returned by `save_pkgmode_kernel`.
+#[context("Running post-install /var preservation and rollback entry")]
+pub(crate) fn preserve_var_and_write_rollback(root_path: &Path, kver: &str) -> Result<()> {
+    // Find the newly-created deployment directory.
+    let deploy_dir = find_deploy_dir(root_path)
+        .context("Locating new ostree deployment directory")?;
+    println!("  Deployment directory: {}", deploy_dir.display());
+
+    // The deployment's var/ is two levels up from the deploy dir:
```
cgwalters, line 152: There's APIs for this

crates/lib/src/install/migrate.rs
```diff
@@ -0,0 +1,630 @@
...
+/// already present at `<root_path>/sysroot/ostree`, `root_path` is adjusted
+/// to `<root_path>/sysroot` before we are called, so the deploy tree is then
+/// at `<root_path>/ostree/deploy`.  On a plain package-mode root (Fedora Cloud,
+/// RHEL, etc.) the ostree sysroot is installed directly under the root, so the
+/// path is `<root_path>/ostree/deploy` as well.
+///
+/// We enumerate `<root_path>/ostree/deploy/` and return the most recently
+/// modified entry under each stateroot's `deploy/` subdirectory.
+#[context("Locating new ostree deployment directory")]
+fn find_deploy_dir(root_path: &Path) -> Result<PathBuf> {
```
cgwalters, line 213: There's APIs for htis too

crates/lib/src/install/migrate.rs
```diff
@@ -0,0 +1,630 @@
...
+
+/// Run the full post-install migration pipeline:
+///
+/// 1. Preserve `/var` (reflink copy or bind-mount unit injection)
+/// 2. Write the package-mode rollback BLS entry
+///
+/// `root_path` is the host root as seen from inside the install container.
+/// `kver` is the running kernel version string returned by `save_pkgmode_kernel`.
+#[context("Running post-install /var preservation and rollback entry")]
+pub(crate) fn preserve_var_and_write_rollback(root_path: &Path, kver: &str) -> Result<()> {
```
cgwalters, line 146: In general let's try to use cap-std

crates/lib/src/install/migrate.rs
```diff
@@ -0,0 +1,630 @@
...
+    } else {
+        println!("  Filesystem does not support reflinks — falling back to full copy (Strategy D)");
+        preserve_var_copy(src_var, new_var)
+    }
+}
+
+/// Returns true if the filesystem hosting `new_var` supports reflinks.
+///
+/// Probes by attempting a zero-byte reflink from `src_var` into `new_var`.
+fn reflinks_supported(src_var: &Path, new_var: &Path) -> bool {
```
cgwalters, line 272: Our storage code already probes this I believe

crates/lib/src/install/migrate.rs
```diff
@@ -0,0 +1,630 @@
...
+        // Skip plain top-level ephemeral names.
+        if SKIP_TOP.iter().any(|s| *s == name_str.as_ref()) {
+            println!("    Skipping {} (ephemeral)", src_var.join(name_str.as_ref()).display());
+            continue;
+        }
+
+        let src_entry = src_var.join(name_str.as_ref());
+        let dst_entry = new_var.join(name_str.as_ref());
+
+        // For `log/`, skip only the `journal` subdirectory (large; regenerated by journald).
```
cgwalters, line 317: Why is this special cased?

crates/lib/src/install/migrate.rs
```diff
@@ -0,0 +1,630 @@
...
+        if name_str == "log" {
+            copy_dir_skip_subdir(&src_entry, &dst_entry, "journal", true)?;
+            continue;
+        }
+
+        // For `lib/`, skip `containers` (podman/docker storage — overlay mounts cannot
```
cgwalters, line 323: Ditto, I don't think we should have special cases buried in this code. We *should* make it configurable of course in general, via e.g. `--preserve-var-skip=log/,lib/containers` or something

crates/etc-merge/src/lib.rs
```diff
@@ -788,7 +788,11 @@ fn merge_leaf(
     };
 
     if matches!(new_inode, Some(Inode::Directory(..))) {
-        anyhow::bail!("Modified config file {file:?} newly defaults to directory. Cannot merge")
+        tracing::warn!(
+            "Modified config file {file:?} newly defaults to a directory in the new image; \
+             keeping the image's directory and skipping host customization"
+        );
```
Johan-Liebert1: I don't know about this. Since the user had made changes to something in etc, I don't think we should just warn and move on.

cgwalters:
In a live chat about this the problem that `/var/lib/rpm` may exist came up - thanks @marrusl .

We should support a config file for this functionality (or reuse existing bootc install config) that supports exclusions e.g.

PR #2385, opened 2026-08-17
CI networking cleanup

ckyrouac:
Some AI generated changes to CI to remove network access during test execution. Opening as a draft to test and iterate.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -154,7 +154,7 @@ jobs:
       - name: Fetch external dependencies (with retry)
         run: just build-fetch
 
-      - name: Integration tests
+      - name: Build images and test binaries
         run: |
           set -xeu
           # Build images as regular user, then copy to root's podman storage
@@ -168,13 +168,19 @@ jobs:
 
           # Build test binaries before any sudo commands to avoid cargo permission issues
           cargo build --release -p tests-integration
+          cargo build --release -p system-reinstall-bootc
 
           sudo podman build -t localhost/bootc-fsverity -f ci/Containerfile.install-fsverity
 
           df -h /
           sudo install -m 0755 target/release/tests-integration /usr/bin/bootc-integration-tests
+          sudo install -m 0755 target/release/system-reinstall-bootc /usr/bin/system-reinstall-bootc
           sudo rm target -rf
           df -h /
+
+      - name: Integration tests
+        run: |
+          set -xeu
           # The ostree-container tests
           sudo podman run --privileged --pid=host -v /:/run/host -v $(pwd):/src:ro -v /var/tmp:/var/tmp \
             --tmpfs /var/lib/containers \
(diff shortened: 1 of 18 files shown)


jeckersb:
xref https://github.com/bootc-dev/bcvk/pull/311 since these are somewhat overlapping, although offhand I'm not familiar with what `--restrict-network` is doing here

hack/lbi/podman.image
```diff
@@ -1,2 +1,2 @@
 [Image]
-Image=registry.access.redhat.com/ubi9/podman:latest
+Image=docker.io/library/alpine:latest
```
cgwalters, line 2: But this isn't podman anymore...

tmt/tests/booted/test-logically-bound-switch.nu
```diff
@@ -32,6 +33,19 @@ def initial_setup [] {
     bootc image copy-to-storage
     podman images
     podman image inspect localhost/bootc | from json
+
+    # Pre-seed the bound images used by this test into bootc's own storage.
```
cgwalters, line 37: The thing that worries me a bit about this is that the additional image store stuff is already somewhat special magic, and how AIS interacts with bootc/composefs storage and *especially* bootc's own use of AIS for LBIs...is...complicated.

I mean, isn't this change effectively subverting bootc's own image pulls?

In other words, I think we are still going to need tests that interact with registries directly.

The more I think about this the more I feel we should be using https://github.com/cgwalters/cstor-dist or so instead.

ckyrouac: thanks for looking at this! I haven't given this much attention. I just wanted to push it up as a starting point. I agree using cstor-dist makes sense. I was working on that awhile ago on another branch. I'll try to merge that code into this PR.

.github/workflows/ci.yml
```diff
@@ -499,6 +501,9 @@ jobs:
           name: packages-${{ matrix.test_os }}
           path: target/packages/
 
+      - name: Fetch external dependencies (with retry)
```
cgwalters, line 504: I think we can factor this out as a prep PR?

.github/workflows/ci.yml
```diff
@@ -168,13 +168,19 @@ jobs:
 
           # Build test binaries before any sudo commands to avoid cargo permission issues
           cargo build --release -p tests-integration
+          cargo build --release -p system-reinstall-bootc
```
cgwalters, line 171: This part isn't totally clear to me why we're doing this but probably makes sense?

I wonder if this should be part of a `just` verb too

hack/packages.txt
```diff
@@ -13,3 +13,5 @@ dosfstools
 e2fsprogs
 # Required by  bib-build test
 qemu-img
+# Required by 030-test-composefs.nu (pre-fetches busybox OCI image during build)
+skopeo
```
cgwalters, line 17: Can easily move this to a prep PR too

cgwalters:
Per discussion we're going to close this but we'll take at least some parts of it to another PR

PR #2386, opened 2026-08-18
podstorage: Add early check for missing podman binary

### LINKED ISSUE #2284: bootc 1.16.0+ fails with an unclear error message when missing `podman`
Originally I thought I was hitting a bug in bootc but it turns out I was just missing a `podman` binary in my image. I'm keeping the original issue content for archival and searchability.

---

Tested versions: 1.16.0-1.16.3
Install command: `install to-disk --composefs-backend --via-loopback /data/bootable.img --filesystem "${filesystem}" --wipe --bootloader systemd`

Running `bootc switch` or `bootc upgrade` on the booted system throws `error: Upgrading composefs: Creating imgstore: Initializing images: No such file or directory (os error 2)`.

Other things I found

- `/sysroot/ostree/bootc` exists and is a symlink to `../composefs/bootc`
- a `storage.tmp` directory is left behind under `/sysroot/ostree/bootc`
- enabling `--experimental-unified-storage` during `install-to-disk` triggers the same error

If needed I can create and link a full repository where this can be easily reproduced, my source repo is private.

---

I tried poking around with Copilot to see if we can track down why this happens, here's what we found:

From source inspection, the symlink itself does **not** appear to be the problem. On composefs installs, bootc expects `ostree/bootc -> ../composefs/bootc`, and in this case `/sysroot/ostree/bootc` already exists as that symlink.

The failure appears to happen later, during **first-time initialization of bootc-owned containers-storage (`imgstore`)**.

Relevant code path:
- `bootc switch` / `bootc upgrade` on composefs may call `image_exists_in_unified_storage()`
- that calls `Storage::get_ensure_imgstore()`
- which calls `CStorage::create()`
- `CStorage::create()` creates `ostree/bootc/storage.tmp` and then runs `podman images` to bootstrap the storage
- the error string matches that exact step: `Creating imgstore: Initializing images: No such file or directory`

The strongest clue is that:
- `storage.tmp` is left behind after the failure
- enabling `--experimental-unified-storage` during install triggers the **same error during install**, which goes through the same `CStorage::create()` path

So the working hypothesis is:
1. the bug is in **imgstore bootstrap**, not symlink creation
2. specifically, first-time initialization of bootc-owned containers-storage under composefs is failing inside `CStorage::create()`
3. there may also be a secondary issue where unified-storage auto-detection eagerly initializes imgstore on non-unified composefs installs, instead of checking for its existence first

Relevant areas:
- `crates/lib/src/podstorage.rs` (`CStorage::create`, `new_podman_cmd_in`)
- `crates/lib/src/store/mod.rs` (`get_ensure_imgstore`)
- `crates/lib/src/deploy.rs` (`image_exists_in_unified_storage`)
- `crates/lib/src/bootc_composefs/repo.rs` (unified install path)

This is still a hypothesis from code reading + repro behavior, but it seems more likely than the original missing-symlink theory.

HarshwardhanPatil07:
When podman is not installed, "No such file or directory (os error 2)" is shown, which users misinterpret as a missing file or directory on disk. 

Before fix:
error: Upgrading: No such file or directory (os error 2)

After fix:
error: Upgrading: Creating imgstorage: nonexistent_podman executable not found in PATH; it is required for container image storage operations

Steps to reproduce:
1. cargo build
2. BOOTC_variant=ostree just build
3. bcvk libvirt run --name bootc-repro --replace --ssh-wait localhost/bootc
4. bcvk libvirt ssh bootc-repro -- 'bash -c "BOOTC_EXP_EXTERNAL_CONTAINER_TOOL=nonexistent_podman bootc upgrade

Closes: https://github.com/bootc-dev/bootc/issues/2284

### DIFF
--- a/crates/lib/src/podstorage.rs
+++ b/crates/lib/src/podstorage.rs
@@ -290,6 +290,14 @@ impl CStorage {
         sepolicy: Option<&ostree::SePolicy>,
     ) -> Result<Self> {
         Self::init_globals()?;
+
+        let bin = bootc_utils::podman_bin();
+        if !crate::utils::have_executable(bin)? {
+            anyhow::bail!(
+                "{bin} executable not found in PATH; it is required for container image storage operations"
+            );
+        }
+
         let subpath = &Self::subpath();
 
         // SAFETY: We know there's a parent


Johan-Liebert1 (dismissed):

cgwalters:
Perhaps `bootc container lint` should also check our runtime dependencies exist?

HarshwardhanPatil07:
Yes @cgwalters, Thank you! i will add it

crates/lib/src/lints.rs
```diff
@@ -895,6 +895,34 @@ fn prune_known_run_paths(paths: &mut BTreeSet<Utf8PathBuf>) {
...
+#[distributed_slice(LINTS)]
+static LINT_RUNTIME_DEPS: Lint = Lint::new_warning(
+    "runtime-deps",
+    "Check that required runtime dependencies are present in the image.",
+    check_runtime_deps,
+);
+fn check_runtime_deps(root: &Dir, _config: &LintExecutionConfig) -> LintResult {
+    let mut bins: Vec<&str> = vec![bootc_utils::podman_bin(), bootc_utils::skopeo_bin()];
+    bins.dedup();
+    const BIN_DIRS: &[&str] = &["usr/bin", "usr/sbin", "bin", "sbin"];
```
cgwalters: We have other code in this repo to search for executables, I don't think we need to reimplement $PATH lookups.

Hmm, another way to do this would be to have something like `storage::check_dependencies()` and centralize things there?

That would also ensure we work with that experimental override variable...

crates/lib/src/lints.rs
```diff
@@ -895,6 +895,34 @@ fn prune_known_run_paths(paths: &mut BTreeSet<Utf8PathBuf>) {
...
 }
 
+#[distributed_slice(LINTS)]
+static LINT_RUNTIME_DEPS: Lint = Lint::new_warning(
+    "runtime-deps",
+    "Check that required runtime dependencies are present in the image.",
+    check_runtime_deps,
+);
+fn check_runtime_deps(root: &Dir, _config: &LintExecutionConfig) -> LintResult {
+    let mut bins: Vec<&str> = vec![bootc_utils::podman_bin(), bootc_utils::skopeo_bin()];
```
cgwalters: A while ago I tried to have this list be included in Cargo.toml

cgwalters (changes requested):

cgwalters:
But this all said sorry I would have probably said we could just merge the first commit right now, and do the lint changes as a separate followup PR. The first commit is clearly safe and correct.

HarshwardhanPatil07:
@cgwalters Thanks! I will remove the second commit and continue in followup pr. Thank you!

Johan-Liebert1 (approved):

Johan-Liebert1:
Closing and reopening as GH packages have expired

cgwalters (approved):

cgwalters:
Ah my bad I had a stale "requested changes"

PR #2391, opened 2026-08-18
ci: Skip heavy test jobs for docs-only PRs

gursewak1997:
Skip heavy test jobs for docs-only PRs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,23 @@ jobs:
       upgrade_os_matrix: ${{ steps.matrix.outputs.upgrade_os_matrix }}
       run_heavy: ${{ steps.matrix.outputs.run_heavy }}
       merge_queue_enabled: ${{ steps.merge-queue.outputs.enabled }}
+      docs_only: ${{ steps.docs-check.outputs.docs_only }}
     steps:
+      # On pull requests, check whether only documentation/governance files
+      # (.md) changed.  When that is the case the heavy test jobs are skipped,
+      # saving runner time for process-only PRs (e.g. releases.md, roadmap.md).
+      - name: Check for docs-only changes
+        id: docs-check
+        if: github.event_name == 'pull_request'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          if gh pr diff "${{ github.event.pull_request.number }}" --name-only | grep -qvE '\.md$'; then
+            echo "docs_only=false" >> "$GITHUB_OUTPUT"
+          else
+            echo "docs_only=true" >> "$GITHUB_OUTPUT"
+          fi
+
       # Detect at runtime whether `main` is currently protected by a GitHub
       # merge queue (see https://github.com/bootc-dev/bootc/issues/2177).
       # The tiered CI below only makes sense with a queue: it lets heavy jobs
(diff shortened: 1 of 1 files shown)


Johan-Liebert1 (dismissed):
LGTM

.github/workflows/ci.yml
```diff
@@ -57,7 +57,29 @@ jobs:
...
+      # (.md) changed.  When that is the case every other CI job is skipped,
+      # saving runner time and reducing noise for process-only PRs
+      # (e.g. releases.md, roadmap.md, CONTRIBUTING.md).
+      - name: Check for docs-only changes
+        id: docs-check
+        if: github.event_name == 'pull_request'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          FILES=$(gh api "repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/files" \
```
cgwalters: Hmm, perhaps cleaner to interact in terms of `git diff`? Aren't there existing actions/tools for this?

.github/workflows/ci.yml
```diff
@@ -92,6 +114,16 @@ jobs:
           LABELS='${{ toJson(github.event.pull_request.labels.*.name) }}'
           EVENT='${{ github.event_name }}'
           MERGE_QUEUE_ENABLED='${{ steps.merge-queue.outputs.enabled }}'
+          DOCS_ONLY='${{ steps.docs-check.outputs.docs_only }}'
+
+          # Docs-only PRs: skip all heavy and light build/test jobs.
+          if [[ "$DOCS_ONLY" == "true" ]]; then
```
cgwalters: I think we should consolidate this with the "Plain PR" logic below

cgwalters (approved):

cgwalters:
So I got suspicious when merging this because currently our CI was down...and this job changed CI itself, not just docs, so CI should have been failing.

Basically  `compute-ci-level` doesn't run `actions/checkout`, so when `docs-check` calls
`gh pr diff "$PR" --name-only`, `gh` can't find a local git repo to infer
`owner/repo` from and fails:

```
failed to run git: fatal: not a git repository (or any parent up to mount point /)
```

That error goes to stderr, so stdout is empty. Piped into
`grep -qvE '\.md$'`, empty input means "found no non-md line", so `grep`
exits 1 and the `if` falls into the `else` branch — `docs_only` gets set to
`true` unconditionally, regardless of what actually changed.

I have to emphasize just how absolutely terrible of a language shell script is, and this is a *perfect* example of that.

We need to add a checkout step and also correctly test exit status, which is best by separating fetch from grep.

cgwalters:
Since we're redoing this let's also please try harder to clean up this code, I think we should only be testing the docs conditional once for example.

Also what may work better is to factor out this stuff into e.g. `ci/analyze` or so as an independent script (not embedded in the yaml) so we can more easily run it outside of GHA to test etc.

cgwalters:
Bigger picture one thing we could try to do is factor out most of this logic into https://github.com/bootc-dev/actions perhaps and try to apply it consistently across all of the repos that way.

(And when we do that we should write more of it in typescript probably, not bash)

gursewak1997:
This should have been caught earlier but starting the factoring out: https://github.com/bootc-dev/actions/pull/51

cgwalters:
Sorry I'm giving slightly conflicting messages but it may make the most sense to do a quick PR that tries to fix this in the existing shell script and land that ASAP, but have a parallel effort to do the generic actions and js logic?

gursewak1997:
That was my take but then I went on an ambitious tangent and thought we could land something in actions soon (today) and get something up and running. Let me fix this and make sure we are not skipping CI:/

cgwalters:
No worries, and thanks for working on this!

BTW one thing I've been doing because this stuff is hard to test is basically having agents force push to https://github.com/bootc-dev/ci-sandbox and testing things there.

(Though we should perhaps make separate $username-ci-sandbox repos to avoid conflict?)

PR #2392, opened 2026-08-19
test-install: Fix chunkah hang by removing host systemd/dbus leak, without breaking journal logging

Johan-Liebert1:
`priv-integration.sh` was intermittently hanging at the chunkah build step (a nested `podman run` invoked via `systemd-run -dP --wait`).

The root cause isn't fuse-overlayfs performance -- it's that the outer `sudo podman run --privileged --pid=host ...` invocation bind-mounts the *host's* real `/run/systemd` and `/run/dbus` into the container. Every `systemd-run` call inside therefore talks to the host's actual systemd manager, which forks the wrapped process into the *host's* mount/IPC namespaces instead of the container's. When that escaped process is itself a `podman run` (the chunkah build), it contends forever on host-level podman locks (`/dev/shm/libpod_lock`) against the real host-level `sudo podman run --privileged` process that's still running the CI job -- a deadlock.

This was confirmed empirically by comparing `/proc/<pid>/ns/mnt`/`ns/ipc` of the hung process against host PID 1, and reproduced reliably in isolation.

**Update:** an earlier version of this PR simply dropped both `/run/systemd` and `/run/dbus` and deleted the `systemd-run` wrappers. That fixes the hang, but a full end-to-end run turned up a second, unrelated issue it introduced: a few lines above the chunkah step, the script asserts on a "layers already present" message via `nsenter -m -t 1 journalctl _COMM=bootc`. ostree-ext's structured logging silently skips writing to the journal at all unless `libsystemd::daemon::booted()` (a check for `/run/systemd/system`) is true, so without `/run/systemd` that check has nothing to find and fails -- not a hang, just a different, real regression.

The fix now keeps `/run/systemd` mounted (only `/run/dbus` is dropped -- nothing here ever needed it) so that check keeps working, and has `priv-integration.sh` itself `umount /run/systemd` right after that check, before any of the systemd-run-free podman/build work later in the script. Cgroup delegation for nested `podman run` turns out to be re-evaluated dynamically on each invocation based on whatever systemd sockets are currently visible, not fixed once at outer-container-creation time, so this mid-script unmount is enough: with no live systemd socket at that point, podman's own cgroup-manager detection falls back to `cgroupfs` automatically and the nested chunkah container stays properly inside the outer container's own namespaces. None of the four `systemd-run -dP --wait` call sites need systemd supervision -- they're synchronous builds/copies that were only ever wrapped to get a cgroup for the nested container, which podman now arranges on its own.

Also keeps the `--log-level=debug`/`RUST_LOG=chunkah=debug` logging from the original version of this PR as a separate commit, since it's independently useful regardless of this fix.

Assisted-by: AI
I reviewed this investigation and the resulting patch; the root cause (and this later correction to it) was confirmed via repeated CI reproduction in a sandbox repo, including a full unmodified end-to-end run of `priv-integration.sh` with the exact patch applied, before landing here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -207,7 +207,7 @@ jobs:
           # The ostree-container tests
           sudo podman run --privileged --pid=host -v /:/run/host -v $(pwd):/src:ro -v /var/tmp:/var/tmp \
             --tmpfs /var/lib/containers \
-            -v /run/dbus:/run/dbus -v /run/systemd:/run/systemd localhost/bootc /src/crates/ostree-ext/ci/priv-integration.sh
+            -v /run/systemd:/run/systemd localhost/bootc /src/crates/ostree-ext/ci/priv-integration.sh
           # Nondestructive but privileged tests
           sudo bootc-integration-tests host-privileged localhost/bootc-install
           # Install tests
--- a/crates/ostree-ext/ci/priv-integration.sh
+++ b/crates/ostree-ext/ci/priv-integration.sh
@@ -91,6 +91,16 @@ if ! grep 'layers already present: ' logs.txt; then
     exit 1
 fi
 
+# The check above requires /run/systemd (bind-mounted from the host) to be
+# present: ostree_ext's structured logging silently skips writing to the
+# journal otherwise. But leaving it mounted for the rest of the script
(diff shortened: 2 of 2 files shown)


Johan-Liebert1 (commented):
lgtm, can't approve my own PR though

cgwalters (approved):

PR #2393, opened 2026-08-19
Update composefs-rs to v0.9.0 tag

cgwalters:
Switch from pinning a specific commit rev to pinning the v0.9.0 git tag, which is easier to track against upstream releases.

This version splits UKI addons into two kinds: PEType::UkiAddon, scoped to a single UKI (found alongside it under a `<uki>.efi.extra.d` directory), and the new PEType::GlobalUkiAddon, which applies to every UKI per systemd-stub's addon spec (found under /boot/loader/addons).

Extract the ESP output directory computation for a PE binary into its own pe_output_dir() helper so it's independently testable, and teach it to route global addons to the shared "loader/addons" directory on the ESP instead of namespacing them under a deployment's UKI directory like per-UKI addons. Global addons are still gated behind --uki-addon like their per-UKI counterparts, since installing arbitrary addons found in the image without explicit opt-in would be surprising.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -738,8 +738,8 @@ checksum = "55b672471b4e9f9e95499ea597ff64941a309b2cdbffcc46f2cc5e2d971fd335"
 
 [[package]]
 name = "composefs"
-version = "0.7.0"
-source = "git+https://github.com/composefs/composefs-rs?rev=0fbc853325ff92dbb842e715611bed7ddf9fe169#0fbc853325ff92dbb842e715611bed7ddf9fe169"
+version = "0.9.0"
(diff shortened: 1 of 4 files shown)


jeckersb (dismissed):

Johan-Liebert1:
Closing and repoening to retrigger CI

Johan-Liebert1 (dismissed):

jeckersb:
Needs rebased, `main` removed the `PathBuf` import so this is a sneaky merge non-conflict failing CI.

Johan-Liebert1:
/copilot rebase this PR on top of main

Johan-Liebert1:
Welp, that didn't work. I could swear we had a rebase command. Rebased and pushed

Johan-Liebert1 (approved):

PR #2394, opened 2026-08-19
ci: Fix docs-only detection in compute-ci-level

gursewak1997:
The previous attempt (merged in #2391) failed silently because compute-ci-level had no checkout step, so gh pr diff could not infer the repository and produced empty output. The pipe into grep swallowed the error, setting docs_only=true for all PRs.

Fix by adding a sparse checkout for PR events, separating the gh pr diff fetch from the grep so failures abort via set -euo pipefail, and guarding empty output explicitly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -162,7 +162,10 @@ jobs:
   # this could be a proper matrix)
   install-tests:
     name: "Test install"
-    if: needs.compute-ci-level.outputs.run_heavy == 'true'
+    # Disabled: this job is hanging in the chunkah/podman step even after
+    # switching to native overlay (see "test-install: Fix chunkah hang").
+    # Re-enable once the underlying hang is root-caused.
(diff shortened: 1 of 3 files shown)


cgwalters (dismissed):

Johan-Liebert1:
Hmm.. Test Install still failing

cgwalters:
My offhand guess around the install test is that some package got updated. I have an agent looking at this.

For now, I'd vote to just make it non-gating.

Johan-Liebert1:
Yeah, sgtm. Every `podman` command is just hanging indefinitely

cgwalters:
OK, I'll take this

ckyrouac (approved):

cgwalters:
We're back to green :heavy_check_mark: CI :tada: :tada:

jeckersb:
> We're back to green ✔️ CI 🎉 🎉

yes

PR #2395, opened 2026-08-20
lib/container_export: Always skip /tmp and /var/tmp during export

cgwalters:
CentOS 10 UKI CI jobs fail `bootc container export --format=tar` with:

```
error: Exporting container: Exporting to tar: No label found in policy '...' for /var/tmp/rhc)
```

`rhc`'s post-install scriptlet drops runtime state under `/var/tmp` during
image build, and the SELinux targeted policy has no file-context entry for
it. This reproduces reliably on the composefs+uki matrix legs, likely
because that build path takes long enough for the scriptlet's async write
to land before the image layer is committed (the file can be present or
absent on other legs depending on timing). Same `rhc` package version was
observed in both passing and failing CI runs, so this isn't a version
regression in `rhc` itself.

Rather than tolerating arbitrary unlabeled paths anywhere in the tree
(which risks silently exporting genuinely mislabeled files), this extends
the existing `SKIP_PATHS` list to always exclude `/tmp` and `/var/tmp`.
These are meant to hold only ephemeral, runtime-created content -
`ostree-ext::commit`'s `FORCE_CLEAN_PATHS` already treats the same two
paths (plus `/run` and `/var/cache`) this way for regular ostree commits,
so dropping them from tar export is consistent with existing behavior.

Adds a unit test exercising `export_filesystem_walk()` directly (with
SELinux labeling disabled) against a synthetic root, verifying `/tmp` and
`/var/tmp` content is dropped while everything else is kept.

Assisted-by: AI
I am knowledgeable in this problem domain and reviewed it carefully.

### DIFF
--- a/crates/lib/src/container_export.rs
+++ b/crates/lib/src/container_export.rs
@@ -126,8 +126,13 @@ fn tar_header_dir_root() -> tar::Header {
 }
 
 /// Paths that should be skipped during export.
-/// These are bootc/ostree-specific paths that shouldn't be in the exported tarball.
-const SKIP_PATHS: &[&str] = &["sysroot/ostree"];
+/// - `sysroot/ostree` is bootc/ostree-specific and shouldn't be in the exported tarball.
+/// - `tmp` and `var/tmp` are meant to hold only ephemeral, runtime-created content (the
+///   same paths `ostree-ext::commit` always cleans before committing). They can end up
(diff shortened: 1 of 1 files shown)


cgwalters:
Folded into #2394 instead, since our CI is currently blocked by both issues and it's simpler to land the fixes together.

cgwalters:
xref https://github.com/RedHatInsights/rhc/pull/467#issuecomment-5358465451

PR #2406, opened 2026-08-24
composefs: Fix inverted rollback queued detection

### LINKED ISSUE #2405: composefs: rollback detection inverted on the systemd-boot UKI backend
On a composefs-native sealed host (systemd-boot, UKI, Type1 BLS entries with `uki` keys), `bootc status` reports `rollbackQueued: true` immediately after a successful staged upgrade boot, and `false` when a rollback genuinely is queued. The `BLSConfigType::EFI` arm of the first-entry check in `composefs_deployment_status_from` is missing the negation that the `NonEFI` and GRUB UKI arms both have, so it reports a queued rollback exactly when the default entry does reference the booted deployment:

https://github.com/bootc-dev/bootc/blob/534c76ea/crates/lib/src/bootc_composefs/status.rs#L988-L995

Observed on a physical host (bootc 1.16.9-1.fc44, Fedora 44, Secure Boot, UKI):

- Staged an upgrade, verified `bootOrder: default` / `rollbackQueued: false`, rebooted. `bootc-finalize-staged.service` ran at shutdown (journal-attested) and the new deployment booted (`LoaderEntrySelected` names the new entry; booted digest correct).
- `bootc status` then reported `bootOrder: rollback` / `rollbackQueued: true`, while `bootctl list --esp-path` marked the booted deployment's entry `(default)` - it carries the primary sort-key, and sd-boot sorts sort-keys ascending.
- Trusting the status, we ran `bootc rollback` to clear the queue. It printed "Reverting queued rollback state" and swapped the entries - making the old deployment sd-boot's real default while `bootc status` now read `default` / `false`. The documented remedy armed a silent regression.
- A second `bootc rollback` (printing "Next boot: rollback deployment") restored the correct order - the messaging inverts together with the status.

Everything derived from the flag inverts with it: `bootc rollback` takes its direction and messaging from the same spec, and the guard that refuses to delete the rollback deployment while it is queued to boot (`host.status.rollback_queued` in `delete.rs`) fires in the wrong cases - a genuinely queued rollback deployment reads as safe to delete.

A/B on the same on-disk entry set (sd-boot default = booted deployment, confirmed via `bootctl`): stock 1.16.9 reports `{"bootOrder":"rollback","rollbackQueued":true}`; with the fix it reports `{"bootOrder":"default","rollbackQueued":false}`.

Possibly related but distinct: #2387 (soft-reboot promotion, same files).

andrewdunndev:
The `BLSConfigType::EFI` arm of the rollback detection in `composefs_deployment_status_from` is missing the negation the `NonEFI` and GRUB UKI arms have, so on a systemd-boot UKI host `bootc status` reports `rollbackQueued: true` after every successful staged upgrade boot and `false` when a rollback really is queued. `bootc rollback` messaging/direction and the rollback-deployment delete guard invert with it. Details and on-host receipts in #2405.

The check is extracted into `rollback_queued_from_first_entry()` so the BLS arms share one negation, with regression tests for both entry types in both directions - fixture entries are parsed and sorted through `get_sorted_type1_boot_entries_helper`, asserting the entry carrying `primary_sort_key` is the default and that the verdict flips with the booted digest.

Testing: full `bootc-lib` suite green; the new tests fail on the unfixed code (re-inverting either arm's negation is caught); on a physical Fedora 44 sealed UKI host whose sd-boot default entry is the booted deployment (confirmed with `bootctl list`), stock 1.16.9 reports `rollbackQueued: true` and the patched binary reports `false` on the same on-disk state.

Fixes #2405

### DIFF
--- a/crates/lib/src/bootc_composefs/status.rs
+++ b/crates/lib/src/bootc_composefs/status.rs
@@ -960,63 +960,61 @@ async fn composefs_deployment_status_from(
     let booted_cfs = host.require_composefs_booted()?;
 
     let mut grub_menu_string = String::new();
-    let (is_rollback_queued, sorted_bls_config, grub_menu_entries) = match booted_cfs
-        .bootloader
-        .kind()?
-    {
-        BootloaderKind::GRUBClassic => match boot_type {
-            BootType::Bls => {
-                let bls_configs = get_sorted_type1_boot_entries(boot_dir, false)?;
+    let (is_rollback_queued, sorted_bls_config, grub_menu_entries) =
+        match booted_cfs.bootloader.kind()? {
+            BootloaderKind::GRUBClassic => match boot_type {
+                BootType::Bls => {
+                    let bls_configs = get_sorted_type1_boot_entries(boot_dir, true)?;
+                    let bls_config = bls_configs
+                        .first()
+                        .ok_or_else(|| anyhow::anyhow!("First boot entry not found"))?;
+
+                    match &bls_config.cfg_type {
+                        BLSConfigType::NonEFI { .. } => {
+                            let is_rollback_queued = rollback_queued_from_first_entry(
+                                bls_config,
+                                booted_composefs_digest.as_ref(),
+                            )?;
+
+                            (is_rollback_queued, Some(bls_configs), None)
+                        }
+
+                        BLSConfigType::EFI { .. } => {
+                            anyhow::bail!("Found 'efi' field in Type1 boot entry")
+                        }
+
+                        BLSConfigType::Unknown => anyhow::bail!("Unknown BLS Config Type"),
+                    }
+                }
+
+                BootType::Uki => {
+                    let menuentries =
+                        get_sorted_grub_uki_boot_entries(boot_dir, &mut grub_menu_string)?;
+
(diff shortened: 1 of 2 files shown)


Johan-Liebert1 (changes requested):
This gets really annoying due to Grub's sorting being different than that of systemd-boot. GrubCC is its own separate case with it almost working like systemd-boot but sorting entries like Grub.

I believe the change is good, but could you please add this test to `tmt/tests/booted/test-rollback.nu` so we can have integration tests run for all bootloaders?

andrewdunndev:
Added in the tmt commit: test-rollback.nu asserts rollbackQueued at each transition (false on a fresh boot, true once queued, the double rollback queues then unqueues, false again after booting the target), so every bootloader config the plans run exercises it.

Writing it surfaced the same inversion in the GRUBClassic/Bls arm: ascending=false put the non-default entry first, and the new asserts would have tripped on the grub lanes. The second commit sorts ascending, routes that arm through the shared helper, adds a Grub-parameterized unit test pinning that the primary entry sorts first under the filename-release comparator, and corrects the earlier fixtures to production's sort-key/filename pairing.

Johan-Liebert1 (approved):
lgtm

Issue #2408, opened 2026-08-24
composefs backend: zstd:chunked images fail to pull with "unexpected EOF reading tar entry"


andrewdunndev:
## Environment

- bootc 1.16.9-1.fc44, composefs backend, sealed UKI + systemd-boot
- Image: OCI, layers `application/vnd.oci.image.layer.v1.tar+zstd` with a zstd:chunked TOC annotation, pushed by `buildah push --compression-format zstd:chunked`

## Reproduction

1. Publish a bootc image with `buildah push --compression-format zstd:chunked`.
2. Run `bootc upgrade` on a composefs-backend host whose origin references it.

Fails deterministically in seconds:

```
error: Upgrading composefs: ... Unable to pull container image docker://...:
Failed to pull config Descriptor { media_type: ImageConfig, digest: sha256:073a8cb2...,
size: 23125, ... }: unexpected EOF reading tar entry
```

The registry side was verified intact before filing: `skopeo inspect` passes, the named config blob fetches byte-exact (size and sha256 match the descriptor), and podman and skopeo consume the same image without complaint.

## Cause (from source, at the composefs-rs rev v1.16.9 pins - 0fbc8533, unchanged on current main f22ecda)

- `crates/composefs-oci/src/layer.rs` `decompress_async()` routes `MediaType::ImageLayerZstd` through a plain `async_compression` `ZstdDecoder`. A zstd:chunked blob is a multi-frame stream (including skippable frames); the decoder stops at the first frame boundary, truncating the decompressed tar, and `crates/composefs-oci/src/tar.rs` bails with `unexpected EOF reading tar entry`. Chunked support for this path is composefs/composefs-rs#137, a draft dormant since June 2025.
- The error is misattributed: `crates/composefs-oci/src/skopeo.rs` `pull()` wraps `ensure_config_with_layers()` - the config fetch and every layer import - in a single `with_context("Failed to pull config ...")`, so a layer-side failure is reported against the config descriptor. That sent me down an hour of registry forensics on a blob that was byte-perfect. #1703 looks like another failure class surfacing under the same wrapper.

This is not #509: that issue tracks zstd:chunked in the containers-storage/podman subsystem; this is the composefs backend's own pull path (composefs-rs), which podman never exercises.

## Why this is worse than a missing feature

[...]

PR #2409, opened 2026-08-24
docs/composefs: Document zstd:chunked pull limitation

andrewdunndev:
The composefs backend cannot currently pull zstd:chunked images (#2408): the decoder stops at the first zstd frame, so the pull fails with "unexpected EOF reading tar entry". Note the limitation, the fix in flight, and the publisher-side workaround under Known Issues.

Generated-by: AI
I measured the limitation on my own host and reviewed the line.


cgwalters (approved):

cgwalters:
(The question is if we'll remember to remove this when we apply the fix...)

andrewdunndev:
The docs entry links #2408, which can't close until a composefs-rs update brings the fix in, so that's the standing reminder; I'll send the removal PR alongside that and close it out.

PR #2410, opened 2026-08-25
podstorage: Skip `podman image exists` for non-storage transports

yeetypete:
`podman image exists` only resolves names in local storage, so a reference with any other transport (`oci:`, `docker://`, ...) makes it fail with an error on stderr. Check the transport up front using the `ImageReference` parser and return `false` directly, which is what the caller already inferred from the non-zero exit status. We can then skip running `podman image exists` alltogether and avoid the stderr messages.

Example "failure" cases where we get some unwanted stderr from podman:

Case 1: switching to a locally built image via --transport oci, with the OCI layout present on disk. The image is exported to an OCI directory with `podman push`, then `bootc switch` is pointed at that directory.

```text
podman push --quiet --compression-format zstd docker.io/yeetypete/bootc-ubuntu:26.04 oci:image.oci:26.04-49b8e14
sudo bootc switch --transport oci  /var/home/psiegel/bootc-ubuntu/image.oci:26.04-49b8e14
[sudo: authenticate] Password:
Error: unsupported transport "oci" for looking up local images
Error: unsupported transport "oci" for looking up local images
Importing 129 layers from OCI layout
```

Case 2: switching to a registry image while the booted deployment still refers to an oci: layout, and that layout directory no longer exists on disk. Note that the switch itself is not aborted.

```text
Error: parsing reference "oci:/var/home/psiegel/bootc-ubuntu/image.oci:26.04-49b8e14": invalid reference format
```

Assisted-by: AI

### DIFF
--- a/crates/lib/src/podstorage.rs
+++ b/crates/lib/src/podstorage.rs
@@ -26,6 +26,7 @@ use cap_std_ext::cmdext::{CapStdExtCommandExt, CmdFds};
 use cap_std_ext::dirext::CapStdExtDirExt;
 use cap_std_ext::{cap_std, cap_tempfile};
 use fn_error_context::context;
+use ostree_ext::container::{ImageReference, Transport};
 use ostree_ext::ostree::{self};
 use std::os::fd::OwnedFd;
 use tokio::process::Command as AsyncCommand;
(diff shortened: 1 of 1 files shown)


Johan-Liebert1 (approved):
lgtm

PR #2411, opened 2026-08-26
Unshare fix for boot.automount

### LINKED ISSUE #2402: auto-updates fail with `opendir(boot): Operation not permitted` when /boot is a systemd automount that has idled out
⚠️ DISCLAIMER. Report is generated by Claude Code while it was troubleshooting one of my bench systems.
I've read the report and indeed seems to be true.
If you don't want to read an LLM report, please stop here. Feel free to close the report.

---

## Summary

`bootc-fetch-apply-updates.service` does not declare `RequiresMountsFor=/boot`. On systems where `/boot` is a `systemd-gpt-auto-generator` automount with `TimeoutIdleSec=120`, `/boot` unmounts ~2 minutes after boot and the update timer then fires against an unmounted `/boot`, failing with:

```
error: Initializing storage: opendir(boot): Operation not permitted
```

The same failure hits `bootc status` and `bootc upgrade` run manually.

ostree hardened its own units against this exact hazard in ostreedev/ostree#2544; bootc's updater unit never got the equivalent.

## Environment

- bootc `1.16.7-1.fc44`
- ostree `2026.3-1.fc44`
- systemd `259.8-1.fc44`
- Base image `quay.io/fedora/fedora-bootc:44`
- BIOS/GRUB boot: `sda1` BIOS boot, `sda2` vfat ESP (unused), `sda3` xfs root
- `/` is a composefs overlay; no `/etc/fstab`

## Symptom

```
Aug 21 18:40:55 systemd[1]: Starting bootc-fetch-apply-updates.service - Apply bootc updates...
Aug 21 18:40:56 bootc[2194]: error: Initializing storage: opendir(boot): Operation not permitted
Aug 21 18:40:56 systemd[1]: bootc-fetch-apply-updates.service: Failed with result 'exit-code'.
```

Repeats on every timer firing. `opendir(...)` is libostree's `glnx_throw_errno_prefix` style, i.e. `EPERM` from `openat(sysroot_fd, "boot")`.

## Reproducer

Deterministic on an affected system:

```console
$ sudo umount /boot          # simulate the automount idling out
$ sudo bootc status
error: Status: opendir(boot): Operation not permitted
```

Note bootc's access does **not** trigger the automount — `/boot` remains unmounted afterwards. (I suspect this is because bootc re-execs under `unshare -m` and autofs won't service the request from that private mount namespace, but I haven't confirmed that.)

Touch `/boot` by any other means and bootc works again:

```console
$ ls /boot >/dev/null && sudo bootc status   # succeeds
```

## Root cause

Two generators both claim `/boot`:

- `ostree-system-generator` -> `/run/systemd/generator/boot.mount` (`What=/sysroot/boot`, `Options=bind`) — the real mount
- `systemd-gpt-auto-generator` -> `/run/systemd/generator.late/boot.automount` (`Where=/boot`, `TimeoutIdleSec=120`, described as "EFI System Partition Automount") — targeting the otherwise-unused ESP

When the autofs idle timer fires it tears down ostree's `/boot` bind mount. The ostree units survive this because they declare the dependency:

```console
$ systemctl show boot.mount -p RequiredBy
RequiredBy=ostree-finalize-staged.service ostree-boot-complete.service \
           ostree-finalize-staged-hold.service rpm-ostree-fix-shadow-mode.service local-fs.target

$ systemctl show bootc-fetch-apply-updates.service -p RequiresMountsFor
RequiresMountsFor=
```

## Why this is newly visible

This system ran fine for two weeks on ostree `2026.2` and broke on upgrade to `2026.3`, with bootc, systemd and the kernel at identical versions across both deployments. The behaviour change is ostreedev/ostree@62109dca ("prepare-root: create `/run/systemd/volatile-root` for composefs"), whose commit message states:

> With composefs enabled both are overlayfs mounts with an anonymous `st_dev` … As a result systemd-gpt-auto-generator silently skips all partition discovery: the ESP is never automounted on /boot

Before 2026.3, gpt-auto could not resolve the composefs root's block device and skipped partition discovery entirely, so `boot.automount` was never generated. Journal evidence — `Set up automount boot.automount` appears only in deployments running 2026.3:

```
boot -3: ostree 2026.2  automount=0
boot -2: ostree 2026.3  automount=1   <- /boot unmounted 2m41s after boot
boot -1: ostree 2026.2  automount=0
boot  0: ostree 2026.3  automount=1
```

So ostree 2026.3 makes gpt-auto work as intended, which in turn exposes the missing dependency in bootc. I'd expect this to affect any composefs-based bootc system with a discoverable ESP as 2026.3 rolls out.

## Proposed fix

Mirror what ostree does — add to `bootc-fetch-apply-updates.service`:

```ini
[Unit]
RequiresMountsFor=/boot
```

Verified as effective here via a drop-in: systemd then pulls in `boot.mount` and orders the service after it.

Two secondary suggestions:

1. `opendir(boot): Operation not permitted` is opaque for what is really "/boot is not mounted". A dedicated check with a clearer message would save a lot of digging.
2. Any other bootc command touching `/boot` (`bootc status`, `bootc upgrade`, `bootc switch`) has the same exposure when run from an idle shell, and can't be fixed by unit dependencies. Triggering the automount explicitly, or reporting a clear error, would help.

## References

- ostreedev/ostree#2544 — ostree's fix for the same hazard on its own units
- ostreedev/ostree@62109dca — the change that makes gpt-auto discover partitions on composefs roots
- systemd/systemd#13099 — gpt-auto generates `boot.automount` despite an existing `boot.mount`
- systemd/systemd#35017 — gpt-auto with overlay roots

Johan-Liebert1:
Pass --propagation=slave so that if systemd gpt-auto-generator has
automount for /boot it gets propagated inside the new mount ns
for us to be able to access the /boot mount even if it's expired

Without this all bootc commands fail with

```
error: Initializing storage: opendir(boot): Operation not permitted
```

Closes: https://github.com/bootc-dev/bootc/issues/2402

### DIFF
--- a/tmt/tests/booted/readonly/054-test-boot-automount.nu
+++ b/tmt/tests/booted/readonly/054-test-boot-automount.nu
@@ -1,0 +1,40 @@
+use tap.nu
+
+# Test if we can still access /boot after it has expired
+
+let automounted = (
+    do { ^findmnt --mountpoint /boot --types autofs --noheadings } | complete
+).exit_code == 0
+
+if not $automounted {
+    print "/boot is not automounted. Exiting"
+    exit 0
+}
+
+$env.SYSTEMD_PAGER = ''
+
+systemctl cat -l boot.automount
+
+bootc status
+
+# Expire boot.autmount
+mkdir /etc/systemd/system/boot.automount.d
+
+echo "
+[Automount]
+TimeoutIdleSec=1
+" | save --force /etc/systemd/system/boot.automount.d/override.conf
+
+systemctl daemon-reload
+systemctl restart boot.automount
+
+print "After overriding timeout"
+systemctl cat -l boot.automount
+
+# Wait for automount to expire
+sleep 5sec
+
+# Make sure bootc status works
+bootc status
+
+tap ok
--- a/crates/lib/src/cli.rs
+++ b/crates/lib/src/cli.rs
@@ -1033,7 +1033,14 @@ pub(crate) fn ensure_self_unshared_mount_namespace() -> Result<()> {
             anyhow::bail!("Failed to unshare mount namespace");
         }
     }
-    bootc_utils::reexec::reexec_with_guardenv(recurse_env, &["unshare", "-m", "--"])
+
+    // Pass --propagation=slave so that if systemd gpt-auto-generator has
+    // automount for /boot it gets propagated inside the new mount ns
+    // for us to be able to access the /boot mount even if it's expired
+    bootc_utils::reexec::reexec_with_guardenv(
+        recurse_env,
+        &["unshare", "-m", "--propagation=slave", "--"],
+    )
 }
 
 /// Load global storage state, expecting that we're booted into a bootc system.


cgwalters:
I don't understand how this would fix the bug; isn't the root problem that we have dueling/conflicting generators for `/boot`?

Is the default propagation here private? And so by making it slave...if something mounts `/boot` in the outer mountns it mounts it for us? We almost certainly want `slave` semantics anyways, so this seems like a good idea.

But what would trigger the mount in the default mountns?

I don't understand how this fixes the bug.

Johan-Liebert1:
> Is the default propagation here private? 

yes, the default is private.

I believe the issue is with private mount namespace, the mount request is not actually sent to systemd and thus the automount is not performed. Following journalctl logs


```bash
# Tried to access /boot on PID1 mounts 

Aug 26 12:25:09 fedora systemd[1]: boot.automount: Got automount request for /boot, triggered by 1513 (bash)                                                                                   
Aug 26 12:25:09 fedora systemd[1]: Mounting boot.mount - /boot...                                                                                                                              
Aug 26 12:25:09 fedora systemd[1]: Mounted boot.mount - /boot.                                                                                                                                 
                                                                                                                                                                                               
                                                                                                                                                                                               
# boot.automount Unmounts /boot as I have a timer of 2 seconds
                                                                                                                                
Aug 26 12:25:19 fedora systemd[1]: Unmounting boot.mount - /boot...                                                                                                                            
Aug 26 12:25:19 fedora systemd[1]: boot.mount: Deactivated successfully.                                                                                                                       
Aug 26 12:25:19 fedora systemd[1]: Unmounted boot.mount - /boot.  

1. Ran `unshare -m /bin/bash; cd /boot` 

No logs appear as automount request was not sent to systemd
```

Johan-Liebert1:
> And so by making it slave...if something mounts /boot in the outer mountns it mounts it for us?

Currently (with private ns), no. Since our mount ns is private, we don't see outer mounts done after unsharing and they don't see ours

With slave mount ns, we do get mounts from outer mount ns, though they don't see ours iirc. For mounts to propagate both ways, we need `propagation=shared` I think

tmt/tests/booted/readonly/054-test-boot-automount.nu
```diff
@@ -0,0 +1,40 @@
...
+" | save --force /etc/systemd/system/boot.automount.d/override.conf
+
+systemctl daemon-reload
+systemctl restart boot.automount
+
+print "After overriding timeout"
+systemctl cat -l boot.automount
+
+# Wait for automount to expire
+sleep 5sec
```
cgwalters, line 35: Hardcoded sleeps invite race conditions, most of these should be edge triggered or at least polling loops with a timeout.

In this case I'm pretty sure we can monitor for changes in the unit via the dbus api.

cgwalters (approved):
Hmm...okay after digging in a bit more, I think using `slave` propagation is almost always the right thing to do for use cases like ours; as the docs say otherwise things like e.g. plugging in a USB stick wouldn't be visible. But of course since we're not a daemon, in practice it mostly doesn't matter.

So I think this change is right on the face of it, but I don't think it's a *complete* fix since I think we still have dueling `/boot` going on. But, that's actually more of an ostree side bug.

tmt/tests/booted/readonly/054-test-boot-automount.nu
Johan-Liebert1: > So I think this change is right on the face of it, but I don't think it's a complete fix since I think we still have dueling /boot going on. But, that's actually more of an ostree side bug.

Yeah, I think you're right. I've re-opened the issue

PR #2415, opened 2026-08-26
docs: Add initramfs day two options and future

jmarrero:
Admins might need to add machine-specific initramfs content after install
(a udev rule to unlock local storage is the classic case) and reach for
editing /boot, which bootc overwrites on update.  Document the supported
path: build a machine-local derived image that regenerates the
initramfs, so the image stays the source of truth.

Also record why rpm-ostree's client-side initramfs is incompatible with
bootc upgrade, and the UKI-addon direction.

Assisted-by: AI

### DIFF
--- a/docs/src/initramfs.md
+++ b/docs/src/initramfs.md
@@ -60,7 +60,69 @@ $ sudo systemctl reboot
 See [Booting local builds](booting-local-builds.md) for more about the
 `containers-storage` transport. Rebuild and switch to this derived image
 whenever either the base image or the machine-specific initramfs configuration
-changes. A systemd service and timer can automate those steps if required.
+changes. A systemd service and timer can automate the rebuild if required.
+
+### Automate the local build
+
+For example, store the build context in `/var/lib/machine-bootc` and create a
+system service:
+
+```systemd
+# /etc/systemd/system/machine-bootc-build.service
+[Unit]
+Description=Build the machine-local bootc image
+After=network-online.target
+Wants=network-online.target
+
+[Service]
+Type=oneshot
+WorkingDirectory=/var/lib/machine-bootc
+ExecStart=/usr/bin/podman build --security-opt=label=disable --pull=newer --tag localhost/machine-bootc:latest .
+```
+
+Schedule the build with a timer:
+
+```systemd
+# /etc/systemd/system/machine-bootc-build.timer
+[Unit]
+Description=Build the machine-local bootc image daily
+
+[Timer]
+OnCalendar=*-*-* 06:00:00
+Persistent=true
+
+[Install]
+WantedBy=timers.target
+```
+
+Enable the timer:
+
+```console
+$ sudo systemctl daemon-reload
+$ sudo systemctl enable --now machine-bootc-build.timer
+```
+
+This example automates only the build. It intentionally leaves staging the new
+image and rebooting as explicit operations. A deployment can be automated too,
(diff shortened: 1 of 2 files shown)


docs/src/initramfs.md
```diff
@@ -0,0 +1,139 @@
+# Managing the initramfs after installation
+
+The initramfs is part of the container image and is updated together with the
+image. On systems using the OSTree backend with a split kernel and initramfs,
+the canonical path is `/usr/lib/modules/$kver/initramfs.img`. Editing the copy
+in `/boot` is not supported because `bootc` replaces it from the container
+image during an update.
+
+Some systems need machine-specific initramfs content after installation, such
+as a udev rule required to unlock local storage. Until `bootc` has a dedicated
```
cgwalters, line 10: Well the details of what one wants to do in the initramfs matter a lot here.

Like, does the udev rule itself actually need to be machine specific?

A general pattern that also works is to have "machine specific" configuration embedded in *generic* images that dispatches on machine specific identifiers hardware MAC addresses are a classic option, but not the only one.

docs/src/initramfs.md
```diff
@@ -0,0 +1,139 @@
...
+
+The `Containerfile` can derive from the image shown by `bootc status` and
+regenerate its initramfs:
+
+```Dockerfile
+FROM quay.io/example/example-bootc:latest
+
+COPY 98-thunderbolt.rules /etc/udev/rules.d/98-thunderbolt.rules
+
+RUN set -eu; \
```
cgwalters, line 36: This is operating system specific, I think we should just link to https://docs.fedoraproject.org/en-US/bootc/initramfs/

docs/src/initramfs.md
```diff
@@ -0,0 +1,139 @@
...
+base image and the content being added.
+
+Build the image locally, then stage it through the Podman container storage:
+
+```console
+$ sudo podman build --security-opt=label=disable --pull=newer \
+    --tag localhost/machine-bootc:latest .
+$ sudo bootc switch --transport containers-storage localhost/machine-bootc:latest
+$ sudo systemctl reboot
+```
```
cgwalters, line 58: Do we need to duplicate these instructions vs just referencing the booting-local-builds.md?

docs/src/initramfs.md
```diff
@@ -0,0 +1,139 @@
...
+
+See [Booting local builds](booting-local-builds.md) for more about the
+`containers-storage` transport. Rebuild and switch to this derived image
+whenever either the base image or the machine-specific initramfs configuration
+changes. A systemd service and timer can automate the rebuild if required.
+
+### Automate the local build
```
cgwalters, line 65: This is a fully generic instruction that I think makes more sense in the booting-local-builds.md

BTW, the real problem with this is correct change detection; if you don't have reproducible builds then you're going to be pointlessly rebooting your machine some days...

jmarrero: moved and added a note about change detection.

docs/src/initramfs.md
jmarrero: done

jmarrero:
I also had Claude Code with Fable create Containers and test these with bcvk.

cgwalters (approved):

cgwalters:
> I also had Claude Code with Fable create Containers and test these with bcvk.

Would be good to put that stuff somewhere, maybe as part of revitalizing https://gitlab.com/fedora/bootc/tests/bootc-workflow-test

PR #2416, opened 2026-08-27
tmt: Introduce `skip_for_ostree`

Johan-Liebert1:
So we don't spawn VMs for tests like "composefs-gc" just to do nothing and exit


cgwalters (approved):
Looks sane but I think this would all be cleaner if we had "tags" on the test that had to match, e.g. could also filter tests by OS version etc

cgwalters (commented):
Also can't we drop some existing "exit 0" in the tests?

PR #2421, opened 2026-08-31
lint: Check required runtime dependencies

### LINKED ISSUE #2420: container lint: Check required runtime dependencies
## Problem

A bootc image can pass `bootc container lint` while missing binaries required by bootc at runtime.

Related review suggestion:
https://github.com/bootc-dev/bootc/pull/2386#issuecomment-5332427490

HarshwardhanPatil07:
Container images can pass linting while omitting binaries that bootc needs after deployment, leaving failures to be discovered during switch or upgrade. Source the existing workspace dependency metadata and report missing binaries while preserving the external container-tool override.

Missing dependencies:
```
Lint warning: runtime-deps: Missing required runtime dependencies: chcon, ostree, podman, setpriv, skopeo, systemctl, zstd
Checks passed: 0
Checks skipped: 14
Warnings: 1
```
After adding all dependencies:
```
Checks passed: 1
Checks skipped: 14
```

Closes: https://github.com/bootc-dev/bootc/issues/2420

Assisted-by: AI

### DIFF
--- a/crates/lib/build.rs
+++ b/crates/lib/build.rs
@@ -1,0 +1,18 @@
+fn main() {
+    let metadata = cargo_metadata::MetadataCommand::new()
+        .current_dir(env!("CARGO_MANIFEST_DIR"))
+        .no_deps()
+        .exec()
+        .expect("running cargo metadata");
+    let workspace_manifest = metadata.workspace_root.join("Cargo.toml");
+    println!("cargo::rerun-if-changed={workspace_manifest}");
+
+    let bins = metadata.workspace_metadata["binary-dependencies"]["bins"]
+        .as_array()
+        .expect("workspace.metadata.binary-dependencies.bins must be an array");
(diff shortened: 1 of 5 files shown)


crates/lib/build.rs
```diff
@@ -0,0 +1,20 @@
+use std::path::Path;
+
+fn main() {
+    let workspace_manifest = Path::new(env!("CARGO_MANIFEST_DIR")).join("../../Cargo.toml");
```
cgwalters: There is a cargo-metadata crate we use elsewhere.

I'm not an expert in build.rs but I'd be a little surprised if there wasn't a better way to access this data

cgwalters (dismissed):

crates/lib/build.rs
HarshwardhanPatil07: Thank you! added

cgwalters (approved):

PR #2429, opened 2026-09-01
lsm: Relabel unlabeled_t contexts

jmarrero:
The final physical-root labeling pass currently skips any path with a security.selinux xattr, including paths explicitly carrying the unusable unlabeled_t type. Treat that type like a missing label so the target policy replaces it while preserving other existing OSTree labels.


cgwalters:
This is a duplicate of https://github.com/bootc-dev/bootc/pull/2368

PR #2435, opened 2026-09-04
packit: Use aliases for Fedora

jeckersb:
Packit provides aliases[1] for various targets in Fedora in order to
avoid hard-coding specific versions.  Use these instead of hard-coding
various lists of version numbers.  We should be able to simply consume
the aliases and things will adjust automatically as the Fedora
lifecycle progresses.

[1] https://packit.dev/docs/configuration#aliases

Signed-off-by: John Eckersberg <dev@eckersberg.com>

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -30,7 +30,7 @@ actions:
 
 jobs:
   # copr_build on PRs: only runs when ci/tier-1 or ci/merge label is present.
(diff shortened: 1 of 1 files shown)


ckyrouac (approved):

cgwalters (approved):
Nice!

PR #2436, opened 2026-09-04
docs: Document that systemd-confext and sysext are unsupported

mohan-shash:
## Summary
- Add a new `systemd-confext and systemd-sysext` section under `/etc` in `filesystem.md`
- State explicitly that these mechanisms are not *currently* supported on bootc-managed systems
- Explain the overlayfs stacking interaction with composefs and transient `/etc`
- Point readers to supported per-host configuration alternatives and the particles design doc

This addresses recurring questions where the presence of `systemd-confext` / `systemd-sysext` in the OS image is mistaken for bootc support.

Assisted-by: AI - Cursor

### DIFF
--- a/docs/src/filesystem.md
+++ b/docs/src/filesystem.md
@@ -111,6 +111,36 @@ view of `/etc`. This should generally be considered an internal implementation d
 of bootc/ostree. Do *not* explicitly put files into this location, it can create
 undefined behavior. There is a check for this in `bootc container lint`.
 
+### systemd-confext and systemd-sysext
+
+[systemd-confext](https://www.freedesktop.org/software/systemd/man/systemd-confext.html)
+and [systemd-sysext](https://www.freedesktop.org/software/systemd/man/systemd-sysext.html)
+are not currently supported on bootc-managed systems.
+
+These tools may be present in the operating system image (for example, because
+they ship in the `systemd` package), but bootc has not validated them on deployed
+systems. Using them on a bootc host is not recommended.
+
+A key reason is that bootc systems commonly use composefs for the root filesystem,
+and may also enable [transient root](#enabling-transient-root) and/or
+[transient `/etc`](#enabling-transient-etc). Each of these features uses overlayfs
+stacking. The Linux kernel limits filesystem stacking depth
+(`FILESYSTEM_MAX_STACK_DEPTH = 2`), so combining composefs with transient `/etc`
+already consumes the available depth. Adding confext or sysext overlays on top
+of that is likely to fail.
+
+For these reasons, transient `/etc` and systemd-confext should be treated as
+mutually exclusive today.
+
+For per-host configuration, use the patterns described in
+[Building images: Configuration](building/guidance.md#configuration) instead:
+image-embedded configuration (prefer `/usr` where possible), persistent `/etc`
+with day-2 configuration management tools, or machine-local kernel arguments
+via `rpm-ostree kargs` or `/usr/lib/bootc/kargs.d`.
+
+For more on the design rationale, see
+[Relationship with systemd "particles"](relationship-particles.md).
+
 ## `/var`
 
 Content in `/var` persists by default; it is however supported to make it or subdirectories


docs/src/filesystem.md
```diff
@@ -111,6 +111,36 @@ view of `/etc`. This should generally be considered an internal implementation d
...
+A key reason is that bootc systems commonly use composefs for the root filesystem,
+and may also enable [transient root](#enabling-transient-root) and/or
+[transient `/etc`](#enabling-transient-etc). Each of these features uses overlayfs
+stacking. The Linux kernel limits filesystem stacking depth
+(`FILESYSTEM_MAX_STACK_DEPTH = 2`), so combining composefs with transient `/etc`
+already consumes the available depth. Adding confext or sysext overlays on top
+of that is likely to fail.
+
+For these reasons, transient `/etc` and systemd-confext should be treated as
+mutually exclusive today.
```
HarshwardhanPatil07: This reason made me more sense if we could have 

"**_transient `/etc` and systemd-confext should be treated as mutually exclusive today._**" line above the line 

> A key reason is that bootc systems commonly use 

as a beginner who read this first time, this is what came in my mind


```suggestion
For these reasons, transient `/etc` and systemd-confext should be treated as
mutually exclusive today.

A key reason is that bootc systems commonly use composefs for the root filesystem,
and may also enable [transient root](#enabling-transient-root) and/or
[transient `/etc`](#enabling-transient-etc). Each of these features uses overlayfs
stacking. The Linux kernel limits filesystem stacking depth
(`FILESYSTEM_MAX_STACK_DEPTH = 2`), so combining composefs with transient `/etc`
already consumes the available depth. Adding confext or sysext overlays on top
of that is likely to fail.

```

HarshwardhanPatil07 (approved):

HarshwardhanPatil07:
Thank you for writing this!

docs/src/filesystem.md
```diff
@@ -111,6 +111,36 @@ view of `/etc`. This should generally be considered an internal implementation d
...
 
+### systemd-confext and systemd-sysext
+
+[systemd-confext](https://www.freedesktop.org/software/systemd/man/systemd-confext.html)
+and [systemd-sysext](https://www.freedesktop.org/software/systemd/man/systemd-sysext.html)
+are not currently supported on bootc-managed systems.
+
+These tools may be present in the operating system image (for example, because
+they ship in the `systemd` package), but bootc has not validated them on deployed
+systems. Using them on a bootc host is not recommended.
```
cgwalters: This paragraph has high overlap with previous

docs/src/filesystem.md
```diff
@@ -111,6 +111,36 @@ view of `/etc`. This should generally be considered an internal implementation d
 of bootc/ostree. Do *not* explicitly put files into this location, it can create
 undefined behavior. There is a check for this in `bootc container lint`.
 
+### systemd-confext and systemd-sysext
+
+[systemd-confext](https://www.freedesktop.org/software/systemd/man/systemd-confext.html)
+and [systemd-sysext](https://www.freedesktop.org/software/systemd/man/systemd-sysext.html)
+are not currently supported on bootc-managed systems.
```
cgwalters, line 118: This is a very complex topic as there are some bootc-specific concerns as noted here around bootc's own use of overlayfs, but then there are also OS-specific integration concerns (SELinux etc.) that are distinct to particular OS variants.

I think we should weaken this and basically have two things:

1) Conflicts with non-default use of transient overlays for `/` and `/etc`
2) May be OS specific concerns, consult the OS/distro docs (and we should definitely tackle this more elaborately in e.g. the Fedora atomic/bootc docs)

docs/src/filesystem.md
```diff
@@ -111,6 +111,36 @@ view of `/etc`. This should generally be considered an internal implementation d
...
+
+These tools may be present in the operating system image (for example, because
+they ship in the `systemd` package), but bootc has not validated them on deployed
+systems. Using them on a bootc host is not recommended.
+
+A key reason is that bootc systems commonly use composefs for the root filesystem,
+and may also enable [transient root](#enabling-transient-root) and/or
+[transient `/etc`](#enabling-transient-etc). Each of these features uses overlayfs
+stacking. The Linux kernel limits filesystem stacking depth
```
cgwalters, line 132: Hmm, I think this is fixable by using `etc = none` in current composefs backend, which doesn't have a corresponding ostree knob (but should).

Then confext could take over `/etc`.

cgwalters (changes requested):

docs/src/filesystem.md
mohan-shash: should we file an issue for this?

docs/src/filesystem.md
```diff
@@ -111,6 +111,48 @@ view of `/etc`. This should generally be considered an internal implementation d
 of bootc/ostree. Do *not* explicitly put files into this location, it can create
 undefined behavior. There is a check for this in `bootc container lint`.
 
+### systemd-confext and systemd-sysext
+
+[systemd-confext](https://www.freedesktop.org/software/systemd/man/systemd-confext.html)
+and [systemd-sysext](https://www.freedesktop.org/software/systemd/man/systemd-sysext.html)
```
cgwalters, line 117: I think sysext is different and in theory it should work

cgwalters (approved):
I think we'll be following up on this

PR #2437, opened 2026-09-05
Remove condition on existance of /run/ostree-booted for bootc systemd units/timers

godvino:
This file is created by ostree backend and won't be present for systems using composefs backend

### DIFF
--- a/systemd/bootc-fetch-apply-updates.timer
+++ b/systemd/bootc-fetch-apply-updates.timer
@@ -1,7 +1,8 @@
 [Unit]
 Description=Apply bootc updates
 Documentation=man:bootc(8)
-ConditionPathExists=/run/ostree-booted
+ConditionPathExists=|/run/ostree-booted
+ConditionKernelCommandLine=|composefs
 
 [Timer]
 OnBootSec=1h
--- a/systemd/bootc-fetch-apply-updates.service
+++ b/systemd/bootc-fetch-apply-updates.service
@@ -1,7 +1,8 @@
 [Unit]
 Description=Apply bootc updates
 Documentation=man:bootc(8)
-ConditionPathExists=/run/ostree-booted
+ConditionPathExists=|/run/ostree-booted
+ConditionKernelCommandLine=|composefs
 
 [Service]
 Type=oneshot


Johan-Liebert1 (changes requested):
We shouldn't outright remove this. This should be OR'd with something like `ConditionKernelCommandLine=composefs`

godvino:
> We shouldn't outright remove this. This should be OR'd with something like `ConditionKernelCommandLine=composefs`

Done.

Johan-Liebert1 (approved):
https://github.com/bootc-dev/bootc/pull/2248 adds another kernel cmdline for composefs so we should not forget to add that in here as well once it's merged

Johan-Liebert1:
Also, @godvino could you sign your commit with `git commit -s --amend --no-edit`. We require signed commits

godvino:
> Also, @godvino could you sign your commit with `git commit -s --amend --no-edit`. We require signed commits

Done.

Johan-Liebert1:
Hmm... DCO's still failing.  That command should've added a 

```git
Signed-off-by: <name> <email> 
```
Do you have name and email properly set up for git?

godvino:
> Hmm... DCO's still failing. That command should've added a
> 
> ```
> Signed-off-by: <name> <email> 
> ```
> 
> Do you have name and email properly set up for git?

Should be fixed now.

systemd/bootc-fetch-apply-updates.timer
```diff
@@ -1,7 +1,8 @@
 [Unit]
 Description=Apply bootc updates
 Documentation=man:bootc(8)
-ConditionPathExists=/run/ostree-booted
+ConditionPathExists=|/run/ostree-booted
+ConditionKernelCommandLine=|composefs
```
cgwalters, line 5: So this is a severe oversight on our part, and part of this is that we were just incrementally getting the tests passing for composefs and didn't have a conformance suite for this.

However I now have a serious worry that fixing this could seriously break people who were using the composefs backend and basically relying on us *not* doing this (i.e. they have their own update systems).

Also of note, this change will also break after https://github.com/bootc-dev/bootc/pull/2248 lands in the general case.

If we have to I would argue that we add `/run/composefs-booted`, I think that's been discussed elsewhere.

(Though strictly speaking a better check is for the source of `/`'s mount containing `composefs` perhaps, that's what we do in bootc's generator itself)

cgwalters (changes requested):
Per above, at least this needs some more consideration

systemd/bootc-fetch-apply-updates.timer
cgwalters: Also, I now kind of regret shipping the builtin timer. It wasn't...wrong exactly but it's too simplistic and basically everyone needs to replace it, and if they need to do that then we shouldn't have had it on at all...

Johan-Liebert1: Maybe we can disable this timer by default and people who want it can enable it?

cgwalters: It'd be a very breaking change to disable the timer for the ostree backend too.

I am not seeing an easy path out of this situation without VERY LOUD warnings on either side.

OK, one technical thing we could do: look at the systemd journal and see whether *any previous boots* had something else do an update but...so heuristic and messy, I am not sure it'd be worth it.

PR #2440, opened 2026-09-08
cfs/boot: Handle separate /boot mount

### LINKED ISSUE #2399: Native composefs with a separate boot filesystem does not make it available at `/sysroot/boot`
**Source version analyzed:** bootc v1.16.7
**Tested operating system:** Fedora 44 (aarch64, UEFI, GRUB)

## Summary

A native-composefs installation with a separate boot filesystem mounts that
filesystem read-only at `/boot`, but bootc's composefs runtime code expects the
same filesystem at `/sysroot/boot`. Nothing mounts it there. As a result, bootc
reads the underlying physical root's stale installation-time `/boot` directory
instead of the real boot filesystem.

This causes `bootc status` and consumers that call it, including DNF, to fail
while looking for BLS entries. The same path is used by composefs upgrade,
rollback, finalization, deletion, and garbage-collection code.

## Installation scenario

The system was installed using `bootc install to-filesystem` with the native
composefs backend, GRUB, a Btrfs root filesystem, a separate ext4 boot
filesystem, and a VFAT EFI System Partition. The relevant arguments included:

```text
--composefs-backend
--bootloader=grub
--root-mount-spec=/dev/disk/by-label/root
--boot-mount-spec=UUID=<boot-filesystem-uuid>
```

After boot, the mount topology was:

```text
$ findmnt /sysroot
TARGET   SOURCE                  FSTYPE OPTIONS
/sysroot /dev/mapper/root[/root] btrfs  ro,noatime,seclabel,compress=zstd:3,...

$ findmnt /boot
TARGET SOURCE           FSTYPE OPTIONS
/boot  /dev/nvme0n1p2  ext4   ro,relatime,seclabel

$ findmnt /sysroot/boot
# no result

$ findmnt /boot/efi
# no result
```

`/sysroot/boot` existed as an ordinary directory containing remnants from the
installation process, but it was not the separate boot filesystem.

## Failure

`bootc status` failed while attempting to read BLS entries from the wrong
directory:

```text
error: Status: Prepending custom prefix to EFI and BLS entries: Getting sorted Type1 boot entries: No such file or directory (os error 2)
```

This also prevented `dnf update` from proceeding because its bootc status check
returned the same error. SELinux audit messages printed around the command were
unrelated; this was a missing mount/path problem, not a labeling denial.

## Source analysis

The install and runtime paths disagree:

- In `crates/lib/src/install.rs`, a non-empty `--boot-mount-spec` becomes a
  `MountSpec` targeting `/boot` and is forced read-only.
- In `crates/lib/src/bootc_composefs/boot.rs`,
  `setup_composefs_bls_boot()` converts that spec into a kernel argument of the
  form `systemd.mount-extra=<source>:/boot:<fstype>:<options>`.
- In `crates/lib/src/store/mod.rs`, `BootedStorage::new()` handles a
  composefs/GRUB system by opening `boot` relative to the physical root. On a
  running system that physical root is `/sysroot`, so the resulting boot
  directory is `/sysroot/boot`.
- `Storage::bls_boot_binaries_dir()` explicitly documents `/sysroot/boot` as
  the GRUB Type 1 boot-binary location. The status and composefs lifecycle code
  then reads `loader/entries` relative to that directory.

Remounting `/sysroot` read-write in bootc's private mount namespace does not
solve this. A filesystem mounted at `/boot` is not thereby also mounted at
`/sysroot/boot`.

The ESP is discovered and temporarily mounted by some bootc operations, but
the installed system also lacked the conventional persistent `/boot/efi`
mount. That leaves bootloader tooling outside those narrowly scoped bootc
operations without the expected ESP path.

## Demonstrated workaround

Installing the following native systemd mount topology fixes the failure:

1. Mount the boot filesystem by UUID read-write at `/sysroot/boot`.
2. Bind-mount `/sysroot/boot` read-only at `/boot`.
3. Mount the ESP by UUID read-write at `/boot/efi`.

In unit terms, the essential relationships are:

```ini
# sysroot-boot.mount
[Mount]
What=/dev/disk/by-uuid/<boot-filesystem-uuid>
Where=/sysroot/boot
Type=ext4
Options=rw

# boot.mount
[Mount]
What=/sysroot/boot
Where=/boot
Type=none
Options=bind,ro

# boot-efi.mount
[Mount]
What=/dev/disk/by-uuid/<esp-filesystem-uuid>
Where=/boot/efi
Type=vfat
Options=rw,fmask=0177,dmask=0077
```

With these units active, `/sysroot/boot` is the writable authoritative boot
filesystem expected by bootc, `/boot` remains protected from ordinary package
manager writes, and the ESP is available at its conventional path.

## Expected behavior

`bootc install to-filesystem --composefs-backend` should leave a separate boot
filesystem in a runtime mount arrangement consistent with bootc's own
composefs storage API. At minimum, a successful installation using
`--boot-mount-spec` should allow `bootc status`, upgrade, rollback, staged
finalization, and garbage collection to access the actual BLS entries.

Possible fixes include:

- Have native-composefs installation generate the persistent systemd mount
  topology above, including appropriate ordering and UUID references.
- Or change the composefs runtime implementation to operate consistently on
  the existing `/boot` mount, remounting or cloning it privately when writes
  are required, and audit every caller that currently assumes
  `/sysroot/boot`.

A regression test should install native composefs with a separate boot
filesystem, boot it, verify both the public read-only and private physical boot
paths, and exercise `bootc status` plus a staged upgrade.

---

**AI assistance:** OpenAI Codex (GPT-5) helped analyze the observed mount state,
review the bootc v1.16.7 source, and draft this report.

Johan-Liebert1:
If /boot is mounted as XBOOTLDR partition then grub configs are stored there as we were unconditionally searching for grub configs in `/sysroot/boot`, which would fail every time. Fix this by checking if `/boot` is a mountpoint and then checking if we have grub configs in there.

This is only an issue with Grub as GrubCC and SystemdBoot both do not store anything inside of `/sysroot/boot` and will (should) always have the ESP mounted at /boot

Closes: #2399

### DIFF
--- a/crates/lib/src/store/mod.rs
+++ b/crates/lib/src/store/mod.rs
@@ -95,6 +95,7 @@ use std::ops::Deref;
 use std::sync::Arc;
 
 use anyhow::{Context, Result};
+use bootc_mount::run_findmnt;
 use bootc_mount::tempmount::TempMount;
 use camino::Utf8PathBuf;
 use cap_std_ext::cap_std;
@@ -104,6 +105,7 @@ use cap_std_ext::cap_std::fs::{
 use cap_std_ext::dirext::CapStdExtDirExt;
 use fn_error_context::context;
 
+use ocidir::cap_std::ambient_authority;
 use ostree_ext::container_utils::ostree_booted;
 use ostree_ext::prelude::FileExt;
 use ostree_ext::sysroot::SysrootLock;
@@ -361,6 +363,80 @@ fn sysroot_is_read_only(d: &Dir) -> Result<bool> {
     Ok(false)
 }
 
+#[context("Finding boot for Grub")]
+fn get_boot_dir_for_grub(physical_root: &Dir) -> Result<Dir> {
+    // We have this so systemd's boot.automount shouldn't expire until
+    // this function finishes execution as we have a handle to /boot
+    let boot =
+        Dir::open_ambient_dir("/boot", ambient_authority()).context("Failed to open /boot")?;
+
+    let is_boot_mntpnt = boot
+        .is_mountpoint(".")
+        .context("Checking if /boot is a mountpoint")?;
+
+    // /boot is not a mount point for bootloader Grub, so we have to
+    // have stuff in /sysroot/boot
+    if !matches!(is_boot_mntpnt, Some(true)) {
+        return physical_root
+            .open_dir("boot")
+            .context("Opening boot in physical root");
+    }
+
+    // /boot is a mountpoint
+    // Figure out if it's the ESP or XBOOTLDR
+    let mnt_res = run_findmnt(&[], None, Some("/boot")).context("Finding /boot mount info")?;
+
+    let mut boot_fs = None;
+
+    for mount in mnt_res.filesystems {
+        if mount.source.starts_with("systemd") {
+            // systemd automount, useless for getting any info
+            continue;
+        }
+
+        if let Some(already_found) = boot_fs {
+            // Really shouldn't happen, but for sanity
+            anyhow::bail!(
+                "Found multiple mounts on /boot. Found {}, already had {already_found}",
+                mount.fstype
+            );
+        };
+
+        boot_fs = Some(mount.fstype);
+    }
+
+    let boot_fs = boot_fs.ok_or_else(|| anyhow::anyhow!("Failed to get filesystem for /boot"))?;
+
+    // NOTE: It would be ideal here to check for DPS UUID but we can't be sure that the
+    // device that /boot is mounted as will have DPS compatible UUID
+    //
+    // The best effort we can have is to check the fstype
+    match boot_fs.as_ref() {
+        // /boot is ESP so we have grub configs in /sysroot/boot
+        "vfat" => {
+            return physical_root
+                .open_dir("boot")
+                .context("Opening boot in physical root");
+        }
+
+        // XBOOTLDR, so grub configs should hopefully be here
+        // but check just in case
+        "ext4" | "xfs" | "btrfs" => {
+            if boot.is_dir("grub2") {
+                return Ok(boot);
+            }
+
+            return physical_root
+                .open_dir("boot")
+                .context("Opening boot in physical root");
+        }
+
+        fstype => {
+            anyhow::bail!("Unknown fstype {fstype} for /boot")
+        }
+    };
+}
+
 impl BootedStorage {
     /// Create a new booted storage accessor for the given environment.
     ///
@@ -390,9 +466,8 @@ impl BootedStorage {
                 };
 
                 let boot_dir = match get_bootloader()?.kind()? {
-                    BootloaderKind::GRUBClassic => {
-                        physical_root.open_dir("boot").context("Opening boot")?
-                    }
+                    // We can have a separate /boot and not /sysroot/boot
+                    BootloaderKind::GRUBClassic => get_boot_dir_for_grub(&physical_root)?,
                     // NOTE: Handle XBOOTLDR partitions here if and when we use it
                     BootloaderKind::BLSCompatible => {
                         esp_mount.fd.try_clone().context("Cloning fd")?


cgwalters:
> will (should) always have the ESP mounted at /boot

Actually won't it be at `/efi` if that dir exists? We need to handle that

Johan-Liebert1:
Oh, you're right, that's the case with both ESP and XBOOTLDR. So this needs to be updated for the other two bootloaders as well. I'll add that

Johan-Liebert1:
So this isn't actually an issue for us since we manually search for the ESP and temp mount it to get a handle. Trying the following setup, everything works as intended

```shell
[root@fedora boot]# findmnt /boot
TARGET SOURCE    FSTYPE OPTIONS
/boot  systemd-1 autofs rw,relatime,fd=89,pgrp=1,timeout=120,minproto=5,maxproto=5,direct,pipe_ino=9419
/boot  /dev/vda2 ext4   ro,relatime,seclabel

[root@fedora boot]# findmnt /efi
TARGET SOURCE    FSTYPE OPTIONS
/efi   systemd-1 autofs rw,relatime,fd=94,pgrp=1,timeout=120,minproto=5,maxproto=5,direct,pipe_ino=9447
/efi   /dev/vda1 vfat   rw,nosuid,nodev,noexec,relatime,nosymfollow,fmask=0177,dmask=0077,codepage=437,iocharset=ascii,shortname=mixed,errors=remount-ro
```

```shell
[root@fedora boot]# fdisk -l
Disk /dev/vda: 10 GiB, 10737418240 bytes, 20971520 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: gpt
Disk identifier: 78C1B509-AC00-4417-ADB9-8B2C1E481990

Device       Start      End  Sectors Size Type
/dev/vda1     2048  2099199  2097152   1G EFI System
/dev/vda2  2099200  4196351  2097152   1G Linux extended boot
/dev/vda3  4196352 20969471 16773120   8G Linux root (x86-64)
```

```shell
[root@fedora boot]# bootc status
[  240.313722] EXT4-fs (vda3): re-mounted df0fd0e2-f9f1-4e2d-b90f-515d284dde04.
● Booted image: localhost/bootc-bls
        Digest: sha256:026acb94e25c5b9a4c2f45a790377f8699434df90be3d033b6f1652cfaa524e3 (amd64)
        Verity: 641f49373ed6d35f6b274d8e8216ba39e63aa4378d150f404a09ba54f200f46241b1a72639b966513912da3715c3258f9259c345168bfe1909d5ce98f2cbc718
       Version: 44.20260908.0 (2026-09-09T03:29:24Z)
```

Johan-Liebert1:
Also, currently, we just put everything in the ESP (for SystemdBoot and GrubCC) so I don't think this would be an issue for us. @cgwalters hope I'm not missing some use case here?

crates/lib/src/store/mod.rs
```diff
@@ -361,6 +363,80 @@ fn sysroot_is_read_only(d: &Dir) -> Result<bool> {
...
+        // /boot is ESP so we have grub configs in /sysroot/boot
+        "vfat" => {
+            return physical_root
+                .open_dir("boot")
+                .context("Opening boot in physical root");
+        }
+
+        // XBOOTLDR, so grub configs should hopefully be here
+        // but check just in case
+        "ext4" | "xfs" | "btrfs" => {
```
cgwalters: I see no reason to hardcode this

crates/lib/src/store/mod.rs
```diff
@@ -361,6 +363,80 @@ fn sysroot_is_read_only(d: &Dir) -> Result<bool> {
...
+            );
+        };
+
+        boot_fs = Some(mount.fstype);
+    }
+
+    let boot_fs = boot_fs.ok_or_else(|| anyhow::anyhow!("Failed to get filesystem for /boot"))?;
+
+    // NOTE: It would be ideal here to check for DPS UUID but we can't be sure that the
+    // device that /boot is mounted as will have DPS compatible UUID
```
cgwalters: But we *could* check it

cgwalters (changes requested):

crates/lib/src/store/mod.rs
```diff
@@ -361,6 +363,80 @@ fn sysroot_is_read_only(d: &Dir) -> Result<bool> {
...
+            return physical_root
+                .open_dir("boot")
+                .context("Opening boot in physical root");
+        }
+
+        // XBOOTLDR, so grub configs should hopefully be here
+        // but check just in case
+        "ext4" | "xfs" | "btrfs" => {
+            if boot.is_dir("grub2") {
+                return Ok(boot);
```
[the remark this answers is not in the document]
cgwalters: Hmm right libostree has some code to handle this I think we do need to handle readonly XBOOTLDR (or classic mount)

crates/lib/src/store/mod.rs
Johan-Liebert1: Should this be an unconditional match?

crates/lib/src/store/mod.rs
Johan-Liebert1: Updated

crates/lib/src/store/mod.rs
Johan-Liebert1: Made the check unconditional

crates/lib/src/store/mod.rs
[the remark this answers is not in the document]
Johan-Liebert1: updated

cgwalters (approved):

PR #2443, opened 2026-09-08
registry pull retries

HarshwardhanPatil07:
## What

Add bounded retries for registry-backed OSTree image pulls when the containers-image proxy reports a `GetBlob` request-initiation failure.

This also routes the normal, non-unified OSTree installation path through the shared retrying pull function.

Related: https://github.com/bootc-dev/bootc/issues/2177

## Why

CI has intermittently failed during commands such as `bootc install to-filesystem` when registries like Quay return transient errors, including `502 Bad Gateway` and network timeouts.

The existing CI retry wrapper covers preparatory Podman operations, but it does not cover image pulls performed internally by bootc. Consequently, one temporary registry failure can fail the entire integration job.

Retries are appropriate here because registry availability is outside bootc's control and repeating an image pull is safe.

The proxy error does not expose the underlying HTTP status. Therefore, bootc cannot distinguish a transient `502` from every permanent `GetBlob` initiation failure. The retry count bounds the worst-case additional delay to 60 seconds.

### DIFF
--- a/crates/lib/src/deploy.rs
+++ b/crates/lib/src/deploy.rs
@@ -48,6 +48,7 @@ use std::collections::HashSet;
 use std::io::{BufRead, Write};
 use std::os::fd::AsFd;
 use std::process::Command;
+use std::time::Duration;
 
 use anyhow::{Context, Result, anyhow};
 use bootc_utils::skopeo_bin;
@@ -75,6 +76,13 @@ use crate::utils::async_task_with_spinner;
 
 // TODO use https://github.com/ostreedev/ostree-rs-ext/pull/493/commits/afc1837ff383681b947de30c0cefc70080a4f87a
 const BASE_IMAGE_PREFIX: &str = "ostree/container/baseimage/bootc";
+
+// Match the default attempt count and delay used by the Justfile's build-fetch
+// retry helper. A failed attempt has to rebuild the importer, so retries are
+// intentionally made at the whole-pull boundary instead of independently for
+// every layer.
+const PULL_MAX_ATTEMPTS: u32 = 3;
+const PULL_RETRY_DELAY: Duration = Duration::from_secs(30);
 
 /// Create an ImageProxyConfig with bootc's user agent prefix set.
 ///
@@ -769,8 +777,59 @@ pub(crate) async fn pull_from_prepared(
     Ok(Box::new((*import).into()))
 }
 
-/// Wrapper for pulling a container image, wiring up status output.
-pub(crate) async fn pull(
+fn is_retryable_pull_error(transport: &str, error: &anyhow::Error) -> bool {
+    if transport != "registry" {
+        return false;
+    }
+
+    // The legacy GetBlob proxy method does not preserve a typed distinction
+    // between transient registry failures and errors such as a missing blob.
+    // Retry the opaque registry error at this top-level boundary, with the
+    // attempt limit above preventing an unbounded delay.
+    error.chain().any(|source| {
+        matches!(
+            source.downcast_ref::<ostree_ext::containers_image_proxy::Error>(),
+            Some(
+                ostree_ext::containers_image_proxy::Error::RequestInitiationFailure {
+                    method,
+                    ..
+                }
+            ) if method.as_ref() == "GetBlob"
+        )
+    })
+}
+
+async fn retry_pull_operation<F, Fut, T>(
+    transport: &str,
+    mut operation: F,
+    retry_delay: Duration,
+) -> Result<T>
+where
+    F: FnMut() -> Fut,
+    Fut: std::future::Future<Output = Result<T>>,
+{
+    for attempt in 1..=PULL_MAX_ATTEMPTS {
+        match operation().await {
+            Ok(value) => return Ok(value),
+            Err(error)
+                if attempt < PULL_MAX_ATTEMPTS && is_retryable_pull_error(transport, &error) =>
+            {
+                tracing::warn!(
+                    attempt,
+                    max_attempts = PULL_MAX_ATTEMPTS,
+                    retry_delay_seconds = retry_delay.as_secs(),
+                    error = %error,
+                    "Container image pull failed; retrying"
+                );
+                tokio::time::sleep(retry_delay).await;
+            }
+            Err(error) => return Err(error),
+        }
+    }
+    unreachable!("the pull attempt range is non-empty")
+}
+
+async fn pull_once(
     repo: &ostree::Repo,
     imgref: &ImageReference,
     target_imgref: Option<&OstreeImageReference>,
@@ -808,6 +867,32 @@ pub(crate) async fn pull(
             Ok(pull_from_prepared(imgref, quiet, prog, *prepared_image_meta).await?)
         }
     }
+}
+
+/// Wrapper for pulling a container image, wiring up status output.
+pub(crate) async fn pull(
+    repo: &ostree::Repo,
+    imgref: &ImageReference,
+    target_imgref: Option<&OstreeImageReference>,
+    quiet: bool,
+    prog: ProgressWriter,
+    booted_deployment: Option<&ostree::Deployment>,
+) -> Result<Box<ImageState>> {
+    retry_pull_operation(
+        &imgref.transport,
+        || {
+            pull_once(
+                repo,
+                imgref,
+                target_imgref,
+                quiet,
+                prog.clone(),
+                booted_deployment,
+            )
+        },
+        PULL_RETRY_DELAY,
+    )
+    .await
 }
 
 pub(crate) async fn wipe_ostree(sysroot: Sysroot) -> Result<()> {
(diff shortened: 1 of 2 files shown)


crates/lib/src/install.rs
```diff
@@ -1073,26 +1073,34 @@ async fn install_container(
...
+        match prepared {
+            PreparedPullResult::AlreadyPresent(existing) => existing,
+            PreparedPullResult::Ready(image_meta) => {
+                crate::deploy::check_disk_space_ostree(repo, &image_meta, &spec_imgref)?;
+                pull_from_prepared(&spec_imgref, false, ProgressWriter::default(), *image_meta)
+                    .await?
+            }
         }
+    } else {
+        pull(
```
Johan-Liebert1, line 1095: Won't this skip the disk space check?

crates/lib/src/deploy.rs
```diff
@@ -769,8 +777,59 @@ pub(crate) async fn pull_from_prepared(
...
+                    ..
+                }
+            ) if method.as_ref() == "GetBlob"
+        )
+    })
+}
+
+async fn retry_pull_operation<F, Fut, T>(
+    transport: &str,
+    mut operation: F,
```
Johan-Liebert1, line 804: Do we need this argument? It's just `pull_once` that's going to be called right?

crates/lib/src/deploy.rs
```diff
@@ -769,8 +777,59 @@ pub(crate) async fn pull_from_prepared(
...
+fn is_retryable_pull_error(transport: &str, error: &anyhow::Error) -> bool {
+    if transport != "registry" {
+        return false;
+    }
+
+    // The legacy GetBlob proxy method does not preserve a typed distinction
+    // between transient registry failures and errors such as a missing blob.
+    // Retry the opaque registry error at this top-level boundary, with the
+    // attempt limit above preventing an unbounded delay.
+    error.chain().any(|source| {
```
Johan-Liebert1, line 789: Not too sure about this. What happens if the image itself doesn't exist? Will we still keep retrying?

crates/lib/src/deploy.rs
```diff
@@ -769,8 +777,59 @@ pub(crate) async fn pull_from_prepared(
...
+                    retry_delay_seconds = retry_delay.as_secs(),
+                    error = %error,
+                    "Container image pull failed; retrying"
+                );
+                tokio::time::sleep(retry_delay).await;
+            }
+            Err(error) => return Err(error),
+        }
+    }
+    unreachable!("the pull attempt range is non-empty")
```
Johan-Liebert1, line 829: Let's have this be an error instead

crates/lib/src/deploy.rs
```diff
@@ -769,8 +777,59 @@ pub(crate) async fn pull_from_prepared(
...
+                }
+            ) if method.as_ref() == "GetBlob"
+        )
+    })
+}
+
+async fn retry_pull_operation<F, Fut, T>(
+    transport: &str,
+    mut operation: F,
+    retry_delay: Duration,
```
Johan-Liebert1, line 805: `PULL_MAX_ATTEMPTS` is used as is, why does this need to be an argument?

crates/lib/src/deploy.rs
cgwalters: Yes I think this should only use https://github.com/bootc-dev/containers-image-proxy-rs/blob/154f4868c2fbf0ea60bad91651389260e5f1e948/src/imageproxy.rs#L91

crates/lib/src/deploy.rs
```diff
@@ -76,6 +77,13 @@ use crate::utils::async_task_with_spinner;
 // TODO use https://github.com/ostreedev/ostree-rs-ext/pull/493/commits/afc1837ff383681b947de30c0cefc70080a4f87a
 const BASE_IMAGE_PREFIX: &str = "ostree/container/baseimage/bootc";
 
+// Match the default attempt count and delay used by the Justfile's build-fetch
+// retry helper. A failed attempt has to rebuild the importer, so retries are
+// intentionally made at the whole-pull boundary instead of independently for
+// every layer.
+const PULL_MAX_ATTEMPTS: u32 = 3;
+const PULL_RETRY_DELAY: Duration = Duration::from_secs(30);
```
cgwalters, line 85: Conceptually I think we should be matching what e.g. podman does by default. What our build system happens to do is a different unrelated thing.

On that topic see https://github.com/podman-container-tools/container-libs/pull/951

In the short term, we can just copy the same defaults and add a link to that issue as a TODO to allow having bootc be configurable in the same way.

crates/lib/src/deploy.rs
```diff
@@ -769,8 +777,59 @@ pub(crate) async fn pull_from_prepared(
     Ok(Box::new((*import).into()))
 }
 
-/// Wrapper for pulling a container image, wiring up status output.
-pub(crate) async fn pull(
+fn is_retryable_pull_error(transport: &str, error: &anyhow::Error) -> bool {
+    if transport != "registry" {
```
cgwalters, line 781: I think it'd be cleaner to skip the retry loop at a higher level

crates/lib/src/install.rs
cgwalters: It's also losing the separation between "prepare" and "pull" for the non-unified path.

prepare is "fetch manifest to use for change detection", "pull" = "download whole image".

There's a bigger picture issue here in that we're not doing retries for the first, but we need to in the general case as we can hit flakes there too (DNS, TCP etc).

cgwalters (changes requested):

crates/lib/src/install.rs
HarshwardhanPatil07: Thank you! restored it to original now the check runs after preparation and before pull

crates/lib/src/deploy.rs
HarshwardhanPatil07: It now retries only `GetBlobError::Retryable`

crates/lib/src/deploy.rs
HarshwardhanPatil07: Yes, Thank you!

HarshwardhanPatil07:
PTAL @Johan-Liebert1 @cgwalters

crates/lib/src/deploy.rs
```diff
@@ -769,8 +778,53 @@ pub(crate) async fn pull_from_prepared(
...
 }
 
-/// Wrapper for pulling a container image, wiring up status output.
-pub(crate) async fn pull(
+fn is_retryable_pull_error(error: &anyhow::Error) -> bool {
+    // TODO: Also classify transient prepare/OpenImage failures once
+    // containers-image-proxy exposes a typed error for them. Its current
+    // RequestInitiationFailure cannot safely distinguish DNS/TCP failures from
+    // permanent errors such as a missing image or a signature-policy rejection.
+    error.chain().any(|source| {
```
cgwalters, line 821: Yeah worth filing an issue and linking to it

crates/lib/src/deploy.rs
```diff
@@ -769,8 +778,53 @@ pub(crate) async fn pull_from_prepared(
...
+    Fut: std::future::Future<Output = Result<T>>,
+{
+    let mut retries = 0;
+    loop {
+        match operation().await {
+            Ok(value) => return Ok(value),
+            Err(error) if retries < PULL_MAX_RETRIES && is_retryable_pull_error(&error) => {
+                let retry_delay = pull_retry_delay(retries);
+                let attempt = retries + 1;
+                tracing::warn!(
```
cgwalters: We don't currently have a ton of usage of `warn!`. I have a bit of a worry this is going to corrupt e.g. terminal progress rendering.

There's a kind of ongoing clash there between our usage of indicatif and tracing...we may need to do something tricky like hook into tracing and detect when indicatif is active and use that instead?

One thing we should perhaps do here is pass down the pull progress struct, and use that as our abstraction instead (yes this would be a much bigger task).

---

A general problem here is that we have no testing of our TUI progress rendering, because it's a bit tricky to do.

Did you test this out interactively? One good way to simulate pull failures is to e.g. inject iptables rules to reject packets to quay.io etc.

crates/lib/src/install.rs
```diff
@@ -1022,6 +1024,21 @@ async fn initialize_ostree_root(state: &State, root_setup: &RootSetup) -> Result
     Ok((storage, has_ostree))
 }
 
+async fn pull_ostree_install_once(
+    repo: &ostree::Repo,
+    imgref: &ImageReference,
+    target_imgref: &ostree_container::OstreeImageReference,
+) -> Result<Box<crate::deploy::ImageState>> {
+    let prepared = prepare_for_pull(repo, imgref, Some(target_imgref), None).await?;
+    match prepared {
```
Johan-Liebert1, line 1044: This match is duplicated down below. Rust will tell us if we add another enum variant, but still, feels refactorable

crates/lib/src/deploy.rs
HarshwardhanPatil07: @Johan-Liebert1 helped me debugging this but we saw that progress bar was not rendering in first place with main branch

Johan-Liebert1: Wait, no, I found the issue in my script. I didn't have `-it` in the podman command so the progress bar never showed up

Johan-Liebert1: And I remember I had removed `-it` because the progress bar didn't render properly on my vim emulated terminal

HarshwardhanPatil07: Okay, I tested this interactively by rejecting both IPv4 and IPv6 HTTPS traffic during a quay.io blob pull. The retry occurred and reused completed layers, but the captured progress output was not clean

HarshwardhanPatil07: added a shared PullProgress abstraction 

testing it manually it works, showed the retry message and preserved cached layers.

HarshwardhanPatil07: Thank you for recommending the e.g. to inject iptables rules

crates/lib/src/deploy.rs
HarshwardhanPatil07: Filed it here: https://github.com/bootc-dev/bootc/issues/2466

crates/lib/src/deploy.rs
```diff
@@ -85,6 +85,39 @@ const BASE_IMAGE_PREFIX: &str = "ostree/container/baseimage/bootc";
...
+impl PullProgress {
+    pub(crate) fn new(quiet: bool, json: ProgressWriter) -> Self {
+        let bars = indicatif::MultiProgress::new();
+        if quiet {
+            bars.set_draw_target(indicatif::ProgressDrawTarget::hidden());
+        }
+        Self { bars, json, quiet }
+    }
+
+    fn println(&self, message: impl AsRef<str>) {
```
cgwalters, line 105: Hmm...but probably this *should* be a JSON progress event?

A different way to say this is: in an ideal world, our own indicatif (console) rendering is just a generic client of our progress JSON infrastructure, meaning that "3rd party" clients (esp GUIs) are equally first class.

cgwalters (dismissed):

HarshwardhanPatil07:
PTAL @Johan-Liebert1 @cgwalters

crates/ostree-ext/src/container/unencapsulate.rs
```diff
@@ -169,14 +170,45 @@ pub(crate) async fn join_fetch<T: std::fmt::Debug>(
...
         (Err(worker), Ok(())) => Err(worker),
     }
 }
 
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_join_fetch_preserves_typed_driver_error() {
```
Johan-Liebert1, line 186: Not quite sure what we're trying to test here

Johan-Liebert1 (approved):

crates/lib/src/deploy.rs
```diff
@@ -769,13 +812,69 @@ pub(crate) async fn pull_from_prepared(
...
+            Ok(value) => return Ok(value),
+            Err(error) if retries < PULL_MAX_RETRIES && is_retryable_pull_error(&error) => {
+                let retry_delay = pull_retry_delay(retries);
+                let attempt = retries + 1;
+                let description = format!(
+                    "Container image pull failed; retrying in {} seconds ({attempt}/{PULL_MAX_RETRIES}): {error}",
+                    retry_delay.as_secs()
+                );
+                progress
+                    .send(Event::ProgressSteps {
```
cgwalters, line 855: Hmm, I am not totally sure this matches the current concept of the progress steps which I think assumes a fixed number of steps?

Of course...the API is currently experimental anyways, but we may need to ensure it supports dynamism

crates/ostree-ext/src/container/unencapsulate.rs
cgwalters: I think it's that we can traverse the error chain?

cgwalters (approved):

PR #2445, opened 2026-09-08
lsm: Use libselinux for process availability

### LINKED ISSUE #2444: install: Handle osbuild making selinuxfs read-only
Regression from #2368 - basically osbuild makes /sys/fs/selinux read-only, and we started mixing direct calls to /sys/fs/selinux with using libselinux. We should consistently only use libselinux.

cgwalters:
Closes: https://github.com/bootc-dev/bootc/issues/2444

Basically us trying to use libselinux more to lookup the unlabeled type breaks in the osbuild scenario where it mounts /sys/fs/selinux readonly, which libselinux treats the same as not existing.

### DIFF
--- a/crates/lib/src/lsm.rs
+++ b/crates/lib/src/lsm.rs
@@ -35,8 +35,20 @@ fn unlabeled_type() -> Result<CString> {
     context_type(context.as_bytes())
 }
 
-#[context("Querying selinux availability")]
-pub(crate) fn selinux_enabled() -> Result<bool> {
+/// Whether SELinux is enabled for the current process.
+///
+/// libselinux caches this state process-wide, so mounting selinuxfs requires a
+/// re-exec before this result can change.
+pub(crate) fn selinux_enabled() -> bool {
+    selinux::kernel_support() != selinux::KernelSupport::Unsupported
+}
+
+/// Whether SELinux is enabled on the host, as observed from PID 1's mount namespace.
+///
+/// This bypasses libselinux's process-cached state and is only appropriate while
+/// bootstrapping the selinuxfs mount before re-exec.
+#[context("Querying host SELinux availability")]
+pub(crate) fn host_selinux_enabled() -> Result<bool> {
     Path::new("/proc/1/root/sys/fs/selinux/enforce")
         .try_exists()
         .map_err(Into::into)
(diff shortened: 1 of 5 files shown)


crates/lib/src/lsm.rs
```diff
@@ -35,8 +35,20 @@ fn unlabeled_type() -> Result<CString> {
...
 }
 
-#[context("Querying selinux availability")]
-pub(crate) fn selinux_enabled() -> Result<bool> {
+/// Whether SELinux is enabled for the current process.
+///
+/// libselinux caches this state process-wide, so mounting selinuxfs requires a
+/// re-exec before this result can change.
+pub(crate) fn selinux_enabled() -> bool {
+    selinux::kernel_support() != selinux::KernelSupport::Unsupported
```
Johan-Liebert1, line 43: I think we want to do similar operation here as we do in `host_selinux_enabled`. I was just testing around on how the `selinux` module checks for support and it just stats 

```c
statfs(
    "/sys/fs/selinux", 
    {f_type=SELINUX_MAGIC, f_bsize=4096, f_blocks=0, f_bfree=0, f_bavail=0, f_files=0, f_ffree=0, f_fsid={val=[0x16, 0]},
    f_namelen=255, f_frsize=4096, f_flags=ST_VALID| ST_NOSUID|ST_NOEXEC|ST_RELATIME}
) = 0
```

This wouldn't really check if selinux is "enabled" or not right as this directory would still exist even if selinux is disabled?

cgwalters: No, if SELinux is e.g. compiled into the Linux kernel but the system is booted with `selinux=0`, then `/sys/fs/selinux` which is an instance of selinuxfs doesn't exist at all. You can see this in e.g. `bcvk ephemeral run-ssh quay.io/centos-bootc/centos-bootc:stream10` 

```
[root@localhost ~]# cat /proc/cmdline 
systemd.journald.storage=volatile selinux=0
[root@localhost ~]# stat /sys/fs/selinux
stat: cannot statx '/sys/fs/selinux': No such file or directory
```

Johan-Liebert1: Okay, yeah, that's correct. I had updated `/etc/selinux/config` to disable but didn't read the note

```
# NOTE: In earlier Fedora kernel builds, SELINUX=disabled would also                                                                                                                           
# fully disable SELinux during boot. If you need a system with SELinux                                                                                                                         
# fully disabled instead of SELinux running with no policy loaded, you                                                                                                                         
# need to pass selinux=0 to the kernel command line. You can use grubby                                                                                                                        
# to persistently set the bootloader to boot with selinux=0: 
```

Johan-Liebert1 (approved):

PR #2446, opened 2026-09-09
export: Add --warn-unlabeled flag for SELinux label handling

gursewak1997:
Align SKIP_PATHS with ostree-ext's FORCE_CLEAN_PATHS (add `run`,  `var/cache`) and replace `--disable-selinux` with a `--selinux=enabled|warn-on-missing|disabled` flag so users can preserve labels for files the policy covers while skipping unmatched ones.


crates/lib/src/cli.rs
```diff
@@ -512,6 +512,12 @@ pub(crate) enum ContainerOpts {
         #[clap(long)]
         disable_selinux: bool,
 
+        /// Warn instead of failing when a file has no SELinux label in the policy.
+        /// By default, a missing label is a hard error. This flag allows the export
+        /// to continue, emitting unlabeled entries with a summary at the end.
+        #[clap(long, conflicts_with = "disable_selinux")]
+        warn_unlabeled: bool,
```
cgwalters: How about `--selinux=disabled|warn-on-missing|enabled` ?

crates/lib/src/container_export.rs
```diff
@@ -73,6 +77,16 @@ async fn export_tar(
     Ok(())
 }
 
+/// How to handle SELinux labeling during export.
+enum SepolicyState {
```
cgwalters, line 75: ...as that basically corresponds to the enum here

cgwalters (commented):
> Files with no matching policy rule cause a hard error, forcing users to pass --disable-selinux which loses all labels. Customers then need external scripts to reapply labels after extraction.

Wait but did this customer not know that one can add custom file paths?

Like this for example:

```
RUN semanage fcontext -a --equal /etc /extra-etc
```

This should ensure that any files in the container dropped into `/extra-etc` have the same label as `/etc`.

Or in short I think the fix almost always should be to add labels for the files, not to ignore unlabeled...

crates/lib/src/cli.rs
gursewak1997: Updated.

gursewak1997:
Also, working on the root cause. I am running a workflow right now.

crates/lib/src/container_export.rs
```diff
@@ -127,17 +154,18 @@ fn tar_header_dir_root() -> tar::Header {
...
-/// - `tmp` and `var/tmp` are meant to hold only ephemeral, runtime-created content (the
-///   same paths `ostree-ext::commit` always cleans before committing). They can end up
-///   containing arbitrary files dropped by package post-install scripts (e.g. `rhc`)
+/// - `run`, `tmp`, `var/tmp`, `var/cache` are ephemeral, runtime-created content that
+///   `ostree-ext::commit::FORCE_CLEAN_PATHS` already strips during regular ostree commits.
+///   They can contain arbitrary files dropped by package post-install scripts (e.g. `rhc`)
 ///   that the SELinux policy has no file-context entry for, which would otherwise turn
 ///   into a hard failure when computing labels for the tar entries.
-const SKIP_PATHS: &[&str] = &["sysroot/ostree", "tmp", "var/tmp"];
+const SKIP_PATHS: &[&str] = &["sysroot/ostree", "run", "tmp", "var/cache", "var/tmp"];
```
cgwalters: Hmm. I'm fine adding this but it feels like we're getting into heuristic territory.

Also - anyone who wants to have `var/cache` not be exported can just fix their container build today to have `RUN rm -rf /var/cache/*` which I think is just the better fix.

Arguably of course the same should be true of `/tmp` and `/var/tmp` and we shouldn't have included those here either.

gursewak1997: Fair point; the `--selinux=warn-on-missing` flag in commit 2 is the real fix for arbitrary unlabeled files. I can drop this commit if you'd rather keep SKIP_PATHS minimal. The main reason I added it was to match ostree-ext's FORCE_CLEAN_PATHS so the two code paths stay consistent, but I don't feel strongly.

cgwalters: Matching FORCE_CLEAN_PATHS could make sense, I am sure we are not at all consistent on this.

But it seems like a separate fix from this PR and also I'd just emphasize that I think the right thing in general is to not have the content in the container image at all.

gursewak1997: Just keeping the `--warn-unlabeled` flag commit for now.

cgwalters (dismissed):
Looks like this needs a rebase now

cgwalters (approved):

PR #2447, opened 2026-09-09
container: Preserve directory modes during export

cgwalters:
Ensure we pass the correct permissions when generating directory metadata.


gursewak1997 (approved):
Looks good

PR #2448, opened 2026-09-10
UKI Addons Support

Johan-Liebert1:
uki-addon: Update addons on update/switch

Introduce a function to gather all currently installed addons, scoped
and global. On upgrade/switch, gather all installed addons and if an
addon with the same name is found in the upgrade image, update that
particular addon automatically

---

uki/addon: Support global addons

We were partially supporting global addons, but they were lumped in with
scoped/local addons. Add a new cli option to composefs installs called
`--global-uki-addon` which would determine which global addon to
install.

---

cfs/upgrade/switch: Handle UKI Addons

Here is what we do now with UKI Addons

- Accept `--uki-addon` and `--global-uki-addon` cli options for bootc
  switch/upgrade commands

- If we find an installed addon with the same name as the one in the new
  image, we update it

---

global-uki-addons: Add prefix to name

Prefix global addon filenames with the bootc identifier in the ESP so
we can distinguish bootc-managed global addons from third-party ones

---

uki-addon: Add CLI for managing UKI Addons

Add `bootc uki-addon` subcommand with three operations:

- `bootc uki-addon list`: List installed UKI addons
  Supports `--json` for JSON output

- `bootc uki-addon add <name> <global|scoped>`: Install an addon from
  the booted image onto the ESP

- `bootc uki-addon remove <name> [deployment_id]`: Remove an addon

Add Display and Serialize to UkiAddonType/UkiAddonsList


supakeen:
Can we install UKI addons as part of a deployment with `bootc install`? These are (sometimes) used to have addons for mount options so the UKI isn't tightly coupled to a specific disk layout.

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -460,10 +465,18 @@ pub(crate) fn get_uki_addon_dir_name(depl_verity: &str) -> String {
     format!("{UKI_NAME_PREFIX}{depl_verity}{EFI_ADDON_DIR_EXT}")
 }
 
-#[allow(dead_code)]
-/// Returns the name of a UKI Addon given verity digest
-pub(crate) fn get_uki_addon_file_name(depl_verity: &str) -> String {
-    format!("{UKI_NAME_PREFIX}{depl_verity}{EFI_ADDON_FILE_EXT}")
+/// Returns the name of a scoped/local UKI Addon directory given name
+/// with or without the `.addon.efi` prefix
+pub(crate) fn get_scoped_uki_addon_name(name: &str) -> String {
```
cgwalters, line 470: I think it's really "canonicalizing"?

But the use case seems weird; when would we ever see a name *without* the suffix?

crates/lib/src/bootc_composefs/boot.rs
```diff
@@ -146,23 +147,20 @@ pub(crate) const BOOTC_UKI_DIR: &str = "EFI/Linux/bootc";
...
-/// Before recommending this feature for real use we should track which files here are
-/// bootc-owned, reconcile that set on every upgrade (installing newly-selected addons,
-/// removing ones we own that are no longer selected/present), and decide/document how
-/// this interacts with deployment rollback (a global addon update isn't reverted by
-/// rolling back to an older deployment).
+/// ESP will load whatever's here). Addons are now (re)installed on upgrade/switch
+/// (installed addons are auto-updated when the new image has a matching filename),
+/// but we still blindly overwrite same-named (global) files with no ownership tracking.
+/// Before recommending this feature for wider use we should track which global addon
+/// files are bootc-owned, and decide/document how this interacts with deployment
```
cgwalters, line 154: I think we can say a global addon is bootc-owned if *any* deployment ever installs it?

IOW when dropping a deployment/image we look at the set of all global addons and if any went from referenced ➡️  unreferenced then we drop it.

crates/lib/src/bootc_composefs/uki_addon.rs
```diff
@@ -0,0 +1,129 @@
...
+fn gather_addons_from_dir(
+    dir: &Dir,
+    addons: &mut Vec<UkiAddonsList>,
+    addon_type: UkiAddonType,
+) -> Result<()> {
+    for ent in dir.entries_utf8()? {
+        let ent = ent?;
+        let filename = ent.file_name()?;
+
+        if let Some(addon_name) = filename.strip_suffix(EFI_ADDON_FILE_EXT) {
```
cgwalters, line 54: Minor nit I prefer `let else .. {continue}`

crates/lib/src/bootc_composefs/uki_addon.rs
```diff
@@ -0,0 +1,129 @@
...
+            match addon_name.strip_prefix(UKI_NAME_PREFIX) {
```
cgwalters, line 55: Also let else here would reduce indentation

crates/lib/src/cli.rs
```diff
@@ -896,6 +906,36 @@ impl InternalsOpts {
...
+pub(crate) enum UkiAddonScope {
+    Global,
+    Scoped,
+}
+
+#[derive(Debug, clap::Subcommand, PartialEq, Eq)]
+pub(crate) enum UkiAddonCliOpts {
+    /// List all installed UKI Addons
+    List {
+        /// Output in JSON format
```
cgwalters, line 919: cc https://github.com/bootc-dev/bootc/issues/522

docs/src/man/bootc-uki-addon-add.8.md
```diff
@@ -0,0 +1,47 @@
+# NAME
+
+bootc-uki-addon-add - Add a UKI Addon to the current deployment
```
cgwalters, line 3: How about marking this as experimetnal first?

docs/src/experimental-composefs.md
```diff
@@ -187,6 +187,95 @@ For production environments with dedicated signing infrastructure:
...
+UKI addons are signed PE binaries that [systemd-stub](https://www.freedesktop.org/software/systemd/man/latest/systemd-stub.html) loads alongside the main UKI at boot. Each addon carries extra kernel command-line parameters (or other PE sections) that get merged into the boot. This lets you ship optional configuration — debug flags, hardware quirks, site-specific parameters — separately from the base UKI, without rebuilding or re-signing it.
+
+bootc supports two types of UKI addons:
+
+- **Scoped addons** are tied to a specific UKI (and therefore a specific deployment). They live in a `<uki>.efi.extra.d/` directory next to the UKI on the ESP, and are cleaned up by garbage collection when the deployment is removed.
+- **Global addons** apply to *every* UKI on the ESP. They live in `loader/addons/` and persist across deployments — they are not removed by GC.
+
+### Building Images with Addons
+
+Addons are built with `ukify` in the same Containerfile stage that produces the sealed UKI. Each addon is a separate `ukify build` invocation:
```
cgwalters, line 201: > Addons are built with `ukify` in the same Containerfile stage that produces the sealed UKI

There's no reason that has to be true?

crates/lib/src/install.rs
```diff
@@ -391,6 +391,23 @@ pub(crate) struct InstallConfigOpts {
...
 }
 
+#[derive(Debug, Default, Clone, clap::Parser, Serialize, Deserialize, PartialEq, Eq)]
+pub(crate) struct UkiAddonOpts {
+    /// Name of the local/scoped UKI addons to install without the ".efi.addon" suffix.
+    ///
+    /// This option can be provided multiple times if multiple addons are to be installed.
+    #[clap(long = "uki-addon")]
+    #[serde(default)]
+    pub(crate) scoped: Option<Vec<String>>,
```
cgwalters, line 401: BTW ideally we validate these eagerly at install time

cgwalters:
> Can we install UKI addons as part of a deployment with bootc install? 

That's in the docs here right?

cgwalters:
Or well actually, I think what you may be asking about is *external* UKI addons (i.e. not included in the container image), in which case it would totally make sense to do.

supakeen:
> Or well actually, I think what you may be asking about is *external* UKI addons (i.e. not included in the container image), in which case it would totally make sense to do.

Yea I wasn't super clear but that was what I was thinking about 🙂

crates/lib/src/bootc_composefs/boot.rs
Johan-Liebert1: > But the use case seems weird; when would we ever see a name without the suffix?

we shouldn't, this is just me being careful, as while testing the CLI opts I forgot to strip the suffix

crates/lib/src/bootc_composefs/boot.rs
Johan-Liebert1: Yes, I believe we do that, but I didn't document it or add test for that. Will update

Johan-Liebert1: We should be able to do this. I'll add it to the GC code

crates/lib/src/install.rs
Johan-Liebert1: Yeah. I'll add these to the UKI installation flow we have currently

docs/src/experimental-composefs.md
Johan-Liebert1: Currently it is. We don't have a mechanism of *ONLY* installing addons from oci artrifacts of another oci image for that matter

crates/lib/src/install.rs
Johan-Liebert1: On second thoughts, would you be fine with this as a follow up? This PR's getting huge as is, and this requires some refactoring in the `prepare_install` function

cgwalters: Oh yeah for sure can be a followup

docs/src/experimental-composefs.md
cgwalters: I cannot think of any reason why addons couldn't just be

```
COPY --from=quay.io/exampleos/uki-addons:latest /usr/lib/modules/$kver/foo.uki`
```

ref https://uapi-group.org/specifications/specs/unified_kernel_image/#locations-for-distribution-built-ukis-installed-by-package-managers

Right? What reason requires them to be *built* in the same phase?

docs/src/experimental-composefs.md
```diff
@@ -187,6 +187,95 @@ For production environments with dedicated signing infrastructure:
...
+# Inside the sealed-uki build stage, after building the main UKI:
+
+# Scoped addon: lives in <kver>.efi.extra.d/, loaded only by this UKI
+ukify build --cmdline 'debug loglevel=7' \
+    --output /out/${kver}.efi.extra.d/debug.addon.efi
+
+# Global addon: lives in loader/addons/, loaded by every UKI ergo every deployment
+mkdir -p /out/loader/addons
+ukify build --cmdline 'custom_param=value' \
+    --output /out/loader/addons/site-config.addon.efi
```
cgwalters, line 213: Uh is `/loader` right here?

cgwalters (commented):
How about `bootc install --included-uki-addon foo --external-uki-addon /path/to/external.efi` i.e. we always disambiguate between internal and external?

docs/src/experimental-composefs.md
Johan-Liebert1: Ah, of course that can be done. I thought we meant getting Addons from oci artifacts separately as we'd discussed in the past. I'll update the docs

docs/src/experimental-composefs.md
Johan-Liebert1: yes, the global addons go in `/boot/loader/addons` where `/boot` is the ESP. I made [this](https://github.com/bootc-dev/bootc/pull/2448/changes#diff-267511643d24177ba0428ef99f35bd67af384cbee1fff8e244931aa0add64e3aR44) changes in finalize-uki

Johan-Liebert1:
> How about `bootc install --included-uki-addon foo --external-uki-addon /path/to/external.efi` i.e. we always disambiguate between internal and external?

I don't understand what we mean by external here? I'm guessing it doesn't mean "global" addons

cgwalters:
external :arrow_right:  https://github.com/bootc-dev/bootc/pull/2448#issuecomment-5639781722

docs/src/experimental-composefs.md
cgwalters: One thing I find hard to understand here is where the somewhat bespoke build system we've grown stops and where the actual APIs we're offering starts.

It feels weird to me to use `/boot/loader` here, shouldn't the addons live literally right next to the UKI?

Johan-Liebert1: > It feels weird to me to use /boot/loader here, shouldn't the addons live literally right next to the UKI?

that's true for scoped UKI addons, but global addons do live in `/boot/loader`

From https://www.freedesktop.org/software/systemd/man/latest/systemd-stub.html

```
Addon files are sorted, loaded, and measured into TPM PCR 12 (if a TPM is present) and appended to the kernel command line. UKI command line options are listed first, then options from addons in /loader/addons/*.addon.efi, and finally UKI-specific addons.
[...]
Additionally, files /loader/addons/*.addon.efi are loaded and verified as PE binaries, and ".cmdline", ".dtb", ".initrd", and ".ucode" sections are parsed from them.
```

PR #2449, opened 2026-09-10
Add support for --from-delta for bootc switch and update

alexlarsson:
This adds native support for oci-delta files to bootc, for both the native composefs backend and the ostree backend.

The composefs backend just uses the existing support in composefs-rs, whereas the ostree backend has a bit more custom code. This code uses the generic delta code in composefs-rs to avoid having to reimplement delta stuff as we're applying deltas directly. This depends on https://github.com/composefs/composefs-rs/pull/394 and means we have a WIP commit that points to this MR. The intent is to get some release with that out before merging this PR.

Note: I have not really tested the native composefs codepath, because I didn't manage to get such a setup going. I will try to do this.

### DIFF
--- a/crates/lib/src/bootc_composefs/repo.rs
+++ b/crates/lib/src/bootc_composefs/repo.rs
@@ -316,10 +316,17 @@ async fn pull_composefs_unified(
 /// When `use_unified` is false (the default), the image is pulled directly
 /// into the composefs repo via skopeo.
 ///
+/// With a `delta`, the content is read from the delta file instead and no
+/// network access happens at all; `spec_imgref` then only names what the
+/// resulting deployment tracks. composefs-ctl recognises the delta artifact
+/// from the layout's manifest and reconstructs the changed layers against the
+/// source image in the repository, failing if that image is absent.
+///
 /// Checks for boot entries in the image and returns them.
 #[context("Pulling composefs repository")]
 pub(crate) async fn pull_composefs_repo(
     spec_imgref: &crate::spec::ImageReference,
+    delta: Option<&crate::delta::Delta>,
     allow_missing_fsverity: bool,
     use_unified: bool,
     quiet: bool,
@@ -327,7 +334,17 @@ pub(crate) async fn pull_composefs_repo(
 ) -> Result<PullRepoResult> {
     const COMPOSEFS_PULL_JOURNAL_ID: &str = "4c3b2a1f0e9d8c7b6a5f4e3d2c1b0a9f8";
 
-    let imgref = spec_imgref.to_image_proxy_ref()?;
+    let imgref = match delta {
+        Some(delta) => {
+            delta.validate_image_reference(spec_imgref)?;
+            crate::delta::reject_unified_storage(delta, use_unified)?;
+            if !quiet {
+                println!("Applying delta {}", delta.describe());
+            }
+            delta.pull_ref()?
+        }
+        None => spec_imgref.to_image_proxy_ref()?,
+    };
 
     tracing::info!(
         message_id = COMPOSEFS_PULL_JOURNAL_ID,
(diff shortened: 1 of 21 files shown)


alexlarsson:
New version that is not depending on the new composefs-rs, instead using the separated out delta parsing from the rust code in containers/oci-delta.

crates/lib/src/delta.rs
```diff
@@ -0,0 +1,657 @@
...
+/// Reject `--from-delta` on a storage backend that cannot apply one yet.
+pub(crate) fn reject_unsupported(path: Option<&Utf8Path>) -> Result<()> {
+    match path {
+        Some(path) => bail!("--from-delta ({path}) is not supported by the ostree backend yet"),
+        None => Ok(()),
+    }
+}
+
+/// Reject `--from-delta` for an image that also has to be in containers-storage.
+pub(crate) fn reject_unified_storage(delta: &Delta, use_unified: bool) -> Result<()> {
```
cgwalters, line 208: So...we are going to need a plan to make this work though...

I think the most viable thing will be extending https://github.com/podman-container-tools/container-libs/pull/651 to support pushes?

I guess we can merge this as is and just add "fix deltas" to the unified storage tracker TODO (can you do that as part of this PR?)

crates/lib/src/delta.rs
```diff
@@ -0,0 +1,657 @@
...
+    ensure!(
+        !use_unified,
+        "Cannot apply delta {}: deltas with unified storage not supported.",
+        delta.path,
+    );
+    Ok(())
+}
+
+/// Check the delta's internal consistency, to fail early
+fn validate(p: &ParsedDelta) -> Result<()> {
```
cgwalters, line 218: Old but good https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/

crates/lib/src/delta.rs
```diff
@@ -0,0 +1,657 @@
...
+        p.target_manifest.config().digest() == p.target_config_descriptor.digest(),
+        "Delta target manifest references config {}, but the embedded config is {}",
+        p.target_manifest.config().digest(),
+        p.target_config_descriptor.digest(),
+    );
+
+    let config = ImageConfiguration::from_reader(&p.target_config_raw[..])
+        .context("Parsing embedded target config")?;
+    let layers = p.target_manifest.layers();
+    ensure!(!layers.is_empty(), "Delta target image has no layers");
```
cgwalters, line 240: Like on this topic there is https://docs.rs/nonempty/0.12.0/nonempty/ for example (and arguably we should use that in oci-spec by default for images, or at least have a standard falliable accessor)

crates/lib/src/delta.rs
```diff
@@ -0,0 +1,657 @@
...
+            .as_ref()
+            .map(|t| t.to_string())
+            .unwrap_or_else(|| "none".into()),
+    );
+
+    parse_delta_manifest(&manifest, oci).await
+}
+
+/// Verify that `data` hashes to `expected`.
+fn verify_digest(what: &str, data: &[u8], expected: &Digest) -> Result<()> {
```
cgwalters, line 297: I think `what` should be an `impl Read` in the general case for this, I'd be surprised if we don't have a variant of this somewhere

crates/lib/src/bootc_composefs/repo.rs
```diff
@@ -316,18 +316,35 @@ async fn pull_composefs_unified(
...
 ) -> Result<PullRepoResult> {
     const COMPOSEFS_PULL_JOURNAL_ID: &str = "4c3b2a1f0e9d8c7b6a5f4e3d2c1b0a9f8";
 
-    let imgref = spec_imgref.to_image_proxy_ref()?;
+    let imgref = match delta {
+        Some(delta) => {
+            delta.validate_image_reference(spec_imgref)?;
+            crate::delta::reject_unified_storage(delta, use_unified)?;
+            if !quiet {
+                println!("Applying delta {}", delta.describe());
```
cgwalters, line 342: This needs to use the progress API.

We may need a blanket ban on `(e)println!` in `crates/lib`

cgwalters (commented):
Just an initial skim overall looks sane

crates/lib/src/delta.rs
cgwalters: (Also this should be part of `impl ParsedDelta` right)

PR #2450, opened 2026-09-10
docs: add guidance about resolv.conf

jmarrero:
Users keep copying the /etc/resolv.conf they see during a container
build into the image, or hitting a dangling resolver symlink after
install, because the file a container runtime bind-mounts during
`RUN` is not the booted host's configuration.  Document the one thing
bootc does about it -- removing a zero-length /etc/resolv.conf and
/etc/hostname left by the runtime at import -- and point at the
network stack (NetworkManager, systemd-resolved) for everything else,
including static DNS.  If a file really has to ship in the image, keep
/etc/resolv.conf a tmpfiles.d symlink to content under /usr, since a
regular file in /etc sticks across upgrades.


docs/src/building/dns.md
```diff
@@ -0,0 +1,192 @@
...
+
+Common policies include:
+
+- NetworkManager supplies per-link DNS configuration to `systemd-resolved`,
+  and `/etc/resolv.conf` points to the local resolver stub.
+- NetworkManager writes a regular `/etc/resolv.conf` directly.
+- Another network manager or resolver owns the file under a
+  distribution-specific policy.
+
+NetworkManager and `systemd-resolved` are not alternatives with the same role.
```
cgwalters, line 57: I think this is fine as is but if it got much more complex we should probably redirect to more in-depth guidance for a particular OS and/or upstream documentation (esp for NetworkManager)

docs/src/building/dns.md
```diff
@@ -0,0 +1,192 @@
...
+
+A base image using `systemd-resolved` should ensure that:
+
+- `systemd-resolved` is installed and enabled.
+- NetworkManager or the selected network manager is configured to supply DNS
+  information to it.
+- `/etc/resolv.conf` is a symlink to a resolver file generated below
+  `/run/systemd/resolve`, normally
+  `../run/systemd/resolve/stub-resolv.conf`.
+- The symlink can be created on boot if it was not materialized during the
```
cgwalters, line 78: I think this is what we expect to happen.

What we should be calling out here is that because of the `/etc/resolv.conf` being injected by container runtimes problem bootc actually always masks it (though just with the ostree backend today) see https://github.com/bootc-dev/bootc/pull/1167

docs/src/building/dns.md
```diff
@@ -0,0 +1,192 @@
...
+Fedora CoreOS also has an integration
+[test for its resolver policy](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/tests/kola/networking/resolv/systemd-resolved).
+
+This is an example of a complete base-image policy, not a requirement that all
+bootc images use `systemd-resolved`.
+
+## Base images using a regular `/etc/resolv.conf`
+
+A base image may instead have NetworkManager or another service write a
+regular `/etc/resolv.conf`. In that case it should ensure that:
```
cgwalters, line 120: Write at *runtime* not build time, probably worth calling out

docs/src/building/dns.md
```diff
@@ -0,0 +1,192 @@
...
+Selecting a resolver backend can change observable DNS behavior. Options from
+traditional `resolv.conf`, including lookup retry and search behavior, do not
+all have direct equivalents in `systemd-resolved`. Conversely,
+`systemd-resolved` provides features such as caching and per-link split DNS
+that are not provided by a regular file alone.
+
+Base image authors should select and document a resolver policy appropriate
+for their distribution. Workloads that depend on particular behavior such as
+`ndots`, timeouts, attempts, caching, or split DNS should be tested against
+that resolver. `bootc` does not translate options between resolver
```
cgwalters, line 148: bootc has nothing to do with `/etc/resolv.conf` other than the special case that we remove it if present in the container image.

I would just include links to the other relevant docs, we don't need to explain resolved vs resolv.conf semantics here.

docs/src/building/dns.md
```diff
@@ -0,0 +1,192 @@
...
+and NetworkManager's documentation for
+[`NetworkManager.conf`](https://networkmanager.pages.freedesktop.org/NetworkManager/NetworkManager/NetworkManager.conf.html)
+and
+[connection settings](https://networkmanager.dev/docs/api/latest/ref-settings.html)
+for the behavior and configuration supported by each component.
+
+## Static resolver configuration
+
+A static `/etc/resolv.conf` can be appropriate for a deliberately static
+system, but it is not the default recommendation. It prevents normal dynamic
```
cgwalters, line 162: NetworkManager definitely supports configuring the DNS through it *statically* e.g. for a static IP address, and I think should be the preferred mechanism for systems using NM.

docs/src/building/dns.md
```diff
@@ -0,0 +1,192 @@
...
+updates from DHCP, VPNs, and per-link configuration unless the network stack
+is also configured not to manage the file.
+
+If a static file is required:
+
+1. Explicitly configure the network stack not to own `/etc/resolv.conf`.
+2. Remove any inherited resolver symlink in a build step using
+   `RUN --network=none`.
+3. Add the intended regular file from build context; do not copy the build
```
cgwalters, line 171: I don't think this will work because of the above.

If someone wants to ship a static `/etc/resolv.conf` in their image, at the current time we probably need to recommend a systemd tmpfiles.d unit that symlinks it, to `/usr/lib/resolv.conf` or something else.

But again I think this is overall a bad idea, and it's better to configure DNS through the networking system (NM or resolved).

docs/src/building/dns.md
```diff
@@ -0,0 +1,192 @@
...
+   host's generated file.
+4. Treat its contents as machine-local configuration when deciding whether it
+   belongs in the reusable image or in provisioning.
+
+Remember that `/etc` is persistent and uses a three-way merge across bootc
```
cgwalters, line 176: Right this is why I think it's better to have it be a link to a file in `/run` (as systemd-resolved encourages) or a symlink to a file in `/usr` if it's truly static.

cgwalters (changes requested):

docs/src/building/dns.md
jmarrero: done

docs/src/building/dns.md
jmarrero: removed

docs/src/building/dns.md
jmarrero: updated in current iteration.

jmarrero:
I got Claude Code with Fable 5.1 to also verify each example by creating containers and verifying with BCVK.

docs/src/building/dns.md
```diff
@@ -0,0 +1,223 @@
...
+container build produces that file. In that case it should ensure that:
+
+- `systemd-resolved` is not enabled.
+- No tmpfiles rule recreates the `systemd-resolved` symlink.
+- `/etc/resolv.conf` is absent or is a regular file that the selected service
+  is configured to manage.
+
+For Fedora-derived images, removing the `systemd-resolved` package is generally
+clearer than only disabling its service: the package also owns the tmpfiles
+rule that creates the symlink. Do the removal in a `RUN --network=none` step:
```
cgwalters, line 139: Yeah though at some point it's cleaner to do a "from scratch" style build for this stuff.

cgwalters (approved):

PR #2453, opened 2026-09-11
install: Clarify systemd-boot backend error

### LINKED ISSUE #2093: install: improve error message when systemd-boot is selected without composefs backend
### Summary

When `--bootloader=systemd` is specified without `--composefs-backend`, the error message is:

```
error: Installing to disk: bootupd is required for ostree-based installs
```

This doesn't tell the user what to do. The fix is to use `--composefs-backend`, but the error message doesn't mention composefs at all. It reads like bootupd is a hard requirement, when actually systemd-boot works fine with the composefs backend.

### Reproduction

```bash
podman run --rm --privileged --pid=host \
    -v /dev:/dev -v /var/lib/containers:/var/lib/containers \
    quay.io/fedora/fedora-bootc:42 \
    bootc install to-disk --wipe --bootloader systemd --filesystem xfs /dev/sdX
```

### Suggested improvement

Something like:

```
error: systemd-boot is not supported with the ostree backend. Use --composefs-backend to enable systemd-boot support.
```

Or if the user explicitly asked for systemd-boot, auto-enable composefs (with a note).

### Context

The current error is at `install.rs:1818`. The bail message is technically accurate but not actionable for users who don't know about the composefs/ostree backend distinction.

ASVLCII:
Selecting `--bootloader systemd` with the ostree backend currently reports only that bootupd is required. The error now points users to `--composefs-backend`, which is the supported path for systemd-boot. The separate GrubCC error remains unchanged.

Closes: #2093

Generated-by: AI

The contributor has not performed an independent human Rust review yet. Local source inspection, `cargo fmt --all -- --check`, and `git diff --check` passed. The targeted test build could not link on this Windows host because Visual C++ `link.exe` is unavailable, so Linux CI is the build verification.


cgwalters (approved):

cgwalters:
Yeah, with stuff like this we really need a workflow that adds a signoff without making a maintainer do a force push, could be a label or `/signoff` chatops style command.

cgwalters:
Tried to do that in https://github.com/bootc-dev/infra/pull/261 but didn't succeed yet...

PR #2454, opened 2026-09-11
A few prep patches for https://github.com/bootc-dev/bootc/pull/2290

cgwalters:
The big one is moving CI to 26.04, see https://github.com/bootc-dev/actions/pull/53 which tries to hack around this on 24.04 but it really feels cleaner to just jump straight to 26.04.


cgwalters:
OK, looks like something going wrong with libvirt firmware selection.

gursewak1997 (approved):

PR #2458, opened 2026-09-12
Apply live bound images

jmarrero:
Basic validation done here, draft while I review further and test manually.

### DIFF
--- a/crates/lib/src/applylive.rs
+++ b/crates/lib/src/applylive.rs
@@ -1,0 +1,855 @@
+//! # Live application of staged content
+//!
+//! This module implements `bootc apply-live`, which makes content from the
+//! staged deployment visible on the running system without a reboot.
+//! See <https://github.com/bootc-dev/bootc/issues/76>.
+//!
+//! The only scope supported today is logically bound images
+//! (`bootc apply-live bound-images`). The image bytes are already shared
+//! between deployments via the bootc-owned container storage; what is
+//! pinned to a deployment is only the *definition*: the symlink in
+//! `/usr/lib/bootc/bound-images.d` and the quadlet file it references.
+//!
+//! Applying these live therefore does not require mounting anything.
+//! Changed quadlet files are written to `/run/containers/systemd/`, which
+//! podman gives precedence over `/etc` and `/usr`, and the affected units
+//! are restarted. Because `/run` is transient, the override disappears on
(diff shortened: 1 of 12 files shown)


crates/lib/src/applylive.rs
```diff
@@ -0,0 +1,855 @@
...
+/// Podman's highest-precedence quadlet search directory, relative to `/`.
+/// Paths are relative to the root (not `/run`) so that SELinux labels are
+/// computed for the real absolute path.
+const QUADLET_RUN_DIR: &str = "run/containers/systemd";
+/// Quadlet search directories relative to a deployment root, in decreasing
+/// precedence (excluding `/run`, which is what we write to).
+const QUADLET_SEARCH_DIRS: &[&str] = &["etc/containers/systemd", "usr/share/containers/systemd"];
+/// Directories which may be added/removed/changed in an ostree commit diff
+/// while still being in scope for `apply-live bound-images`. Note that
+/// `/etc` is stored as `/usr/etc` in ostree commits. The contents of an
```
cgwalters, line 56: Yeah but not in composefs...

docs/src/host-v1.schema.json
```diff
@@ -485,6 +496,65 @@
         "architecture"
       ]
     },
+    "LiveBoundImage": {
```
cgwalters, line 499: I'm not totally sure about this. A client tool can already discover some of this state by looking in `/run` directly?

We actually don't have status for *non-live* bound images either, note...Should we? Maybe. But IMO this also kind of quickly gets into https://github.com/bootc-dev/bootc/issues/522 territory.

crates/lib/src/applylive.rs
```diff
@@ -0,0 +1,855 @@
...
+    let Some(f) = run.open_optional(BOUND_IMAGES_STATE)? else {
+        return Ok(None);
+    };
+    let r = serde_json::from_reader(std::io::BufReader::new(f))
+        .with_context(|| format!("Parsing /run/{BOUND_IMAGES_STATE}"))?;
+    Ok(Some(r))
+}
+
+/// Read the live bound image state from the host `/run`, if any.
+pub(crate) fn read_state_from_host() -> Result<Option<LiveBoundImages>> {
```
cgwalters, line 287: What I'm thinking here is we should try to drive Source Of Truth here closer to the systemd unit files (perhaps with added comments?) instead of inventing a custom JSON file.

PR #2467, opened 2026-09-16
feat(switch): add --target-imgref to decouple pull source from upgrade origin

### LINKED ISSUE #2464: Add `--target-imgref` option to bootc switch
Following a customer case where they had lost connectivity to their registry and manually scp'd and oci image followed by podman load then `bootc switch --transport containers-storage ` which ended up not working out for upgrades as bootc would start fetching upgrades from containers-storage.

Having a `--target-imgref` option for switch would be extremely helpful here

Jorge-Polanco-Roque:
## Problem

Following a customer case: after losing connectivity to their registry, they side-loaded an OCI image (scp + `podman load`) and applied it with `bootc switch --transport containers-storage `. That worked for the immediate switch, but subsequent upgrades then tried to fetch from `containers-storage` — because the source used to pull is also persisted as the origin for upgrades. There was no way to say "pull from here now, but track this other imgref for upgrades."

Closes: #2464

## Approach

Add a `--target-imgref` option (plus `--target-transport`, defaulting to `registry`) to `bootc switch`, mirroring what `bootc install` already exposes via `InstallTargetOpts`. This decouples two concepts that `switch` previously conflated:

- **source** — where the image is pulled from *now* (`--transport` + the positional target).
- **origin** — the imgref persisted for *future* upgrades (`--target-imgref` if given, otherwise the source).

The underlying mechanism already existed: `deploy::pull`/`pull_unified` accept a `target_imgref` that `install` uses and `switch` was passing as `None`. This change wires `switch` to build that target ref the same way `install` does, so no new pull machinery is introduced.

The experimental composefs backend currently dereferences a single ref through `do_upgrade`; rather than silently ignore the new option there, it returns a clear error that `--target-imgref` is not yet supported on that backend.

## Testing

- Added `test_parse_switch_target_imgref` (clap parsing + propagation through both helpers): confirms source stays on `containers-storage` while the target-imgref resolves to the `registry` transport.
- Built and tested `bootc-lib` on Fedora (matching CI deps): `cargo test -p bootc-lib` passes, `cargo fmt --check` clean, and `cargo clippy` introduces no new warnings in the touched files.

## Notes

Open questions for maintainers, happy to adjust: whether to keep `--target-transport` (install parity) or require the transport inside the imgref; the `conflicts_with = from_downloaded` guard; and the composefs-backend follow-up.

Disclosure: this change was prepared with AI assistance for code review and drafting; a human authored and reviewed it and signed off (DCO).


crates/lib/src/cli.rs
```diff
@@ -2722,6 +2770,60 @@ mod tests {
         ));
     }
 
+    #[test]
+    fn test_parse_switch_target_imgref() {
```
cgwalters, line 2778: This test isn't wrong exactly, but it's only testing syntax. Our real test suite uses tmt.

crates/lib/src/cli.rs
```diff
@@ -1495,6 +1511,27 @@ pub(crate) fn imgref_for_switch(opts: &SwitchOpts) -> Result<ImageReference> {
     return Ok(target);
 }
 
+/// Build the optional `--target-imgref` for `switch`; this is the reference that will
+/// be recorded as the origin for subsequent updates, decoupled from the source the
+/// image is fetched from now. Returns `None` when `--target-imgref` was not provided.
+pub(crate) fn target_imgref_for_switch(
```
cgwalters, line 1517: This could also be an `impl SwitchOpts`

crates/lib/src/bootc_composefs/switch.rs
```diff
@@ -40,6 +40,10 @@ pub(crate) async fn switch_composefs(
         return apply_upgrade_from_downloaded(storage, booted_cfs, &host, &do_upgrade_opts).await;
     }
 
+    if opts.target_imgref.is_some() {
+        anyhow::bail!("--target-imgref is not yet supported with the composefs backend");
```
Johan-Liebert1: Let's keep both backends in sync

crates/lib/src/cli.rs
```diff
@@ -2722,6 +2774,74 @@ mod tests {
         ));
     }
 
+    #[test]
```
Johan-Liebert1, line 2777: We'd rather have tmt integration tests. Please take a look at the `tmt` directory in repository root

crates/lib/src/cli.rs
```diff
@@ -1517,15 +1554,24 @@ async fn switch_ostree(
...
-    let target = imgref_for_switch(&opts)?;
+    // The source we fetch the image from right now.
+    let source = imgref_for_switch(&opts)?;
+    // Optional decoupled reference to persist as the origin for subsequent updates.
+    let target_imgref = target_imgref_for_switch(&opts)?;
+    // What we record in the spec/origin for future upgrades: the `--target-imgref`
+    // if given, otherwise the same reference we're fetching from.
+    let origin_ref = match target_imgref.as_ref() {
+        Some(t) => ImageReference::from(t.clone()),
+        None => source.clone(),
```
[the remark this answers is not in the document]
Johan-Liebert1: Yeah, this is a very valid issue and a little bit more complicated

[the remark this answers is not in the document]
Johan-Liebert1: For composefs backend we can add another field in the JSON we store in `/run/composefs/staged-deployment`

Johan-Liebert1 (changes requested):
Thanks for starting this

Jorge-Polanco-Roque:
Thanks for the review — pushed a follow-up commit.

1. switch.rs "keep both backends in sync": done. The composefs backend now supports `--target-imgref` instead of bailing. It pulls from the source and records the decoupled reference as the origin, same as ostree.

2. tmt integration test: added `test-switch-target-imgref.nu` (test-49, skip_if_ostree). It derives a source image, switches with `--target-imgref`, and asserts the staged origin is the target ref, not the containers-storage source we pulled from. Kept the parse unit test too.

3. On persisting the origin: I found the decoupled ref already lands in the persistent `.origin` (`ORIGIN_CONTAINER`) via `write_composefs_state`, so plain `--target-imgref` switches don't need an extra field in `/run/composefs/staged-deployment` — the origin survives to the next upgrade through the existing state write. Let me know if you were picturing the field for a case I'm missing.

One thing I deliberately left out: the identical-content recovery case (source is a byte-identical copy of the *booted* image, re-pointing only the origin) still hits the existing fs-verity same-digest guard in `do_upgrade`/`validate_update` and errors. That guard is duplicated in two places, and relaxing it safely feels like your call on the design — happy to follow up with the "rewrite origin in place" semantics (this may be where the staged-deployment field earns its keep). Wanted to check the direction before touching a safety guard.

tmt/tests/booted/test-switch-target-imgref.nu
```diff
@@ -0,0 +1,49 @@
...
+# Fetch from the local containers-storage copy, but decouple the persisted origin.
+bootc switch --transport containers-storage --target-imgref $origin localhost/bootc-source
+
+# The staged deployment's origin must be the --target-imgref value, not the
+# containers-storage source we actually pulled from.
+let st = bootc status --json | from json
+assert ($st.status.staged != null) "Expected a staged deployment after switch"
+let staged = $st.status.staged.image
+assert equal $staged.image.transport "registry"
+assert equal $staged.image.image $origin
```
Johan-Liebert1, line 47: After this we should reboot and test upgrade to make sure it picks the right image to upgrade from

tmt/tests/booted/test-switch-target-imgref.nu
```diff
@@ -0,0 +1,49 @@
...
+#
+# Verify that `bootc switch --target-imgref` fetches the image from the given
+# source (here a local containers-storage copy) but persists the *decoupled*
+# reference as the origin for subsequent upgrades (issue #2464), on the
+# composefs backend.
+#
+# Skipped on UKI: the test switches to a freshly derived image, whose composefs
+# digest won't match the one sealed into the booted UKI, so the switch is
+# rejected before the origin is ever recorded (same limitation as
+# test-composefs-gc). The decoupling logic is boot-type agnostic and is covered
```
Johan-Liebert1, line 17: I don't think this is true. We can and should be able to do this with UKI as well, all we're doing is changing the origin so I don't get why this wouldn't work. See `make_uki_containerfile` in `tap.nu`

tmt/tests/booted/test-switch-target-imgref.nu
```diff
@@ -0,0 +1,49 @@
+# number: 49
+# tmt:
+#   summary: switch --target-imgref records a decoupled origin (composefs)
+#   duration: 30m
+# extra:
+#   skip_if_ostree: true
```
Johan-Liebert1, line 6: why skip for ostree? This should work for both backends

Johan-Liebert1 (changes requested):

Jorge-Polanco-Roque:
Thanks a lot for the thorough review, @Johan-Liebert1 and @cgwalters — the feedback was spot on. You're right that decoupling the origin should work on UKI too, and that my test was masking that by switching to a derived image rather than just re-pointing the origin. Doing this properly (UKI + the identical-digest recovery case + the fs-verity guard, plus a reboot/upgrade verification) is more of a design change than a quick fix, so I'll close this for now rather than leave a half-baked PR open. Happy to revisit if there's appetite for the fuller approach. Thanks again for the time.

PR #2243, opened 2026-06-10
Filter out BLS entries not managed by bootc

Dudecake:
Fixes #2228

The cases that the tests in the last commit check cannot occur anymore because of the changes in `get_sorted_type1_boot_entries_helper`. If they're unneeded, I can drop the commit.

PR #2475, opened 2026-09-19
install: Only remove bootloader dirs from the ESP

ericcurtin:
## Problem

`bootc install to-existing-root` (alongside mode) empties the entire EFI System Partition in `clean_boot_directories()`. The ESP is frequently shared with content bootc does not manage and cannot recreate.

Concrete case: converting a Fedora Asahi Remix 43 install on an Apple Silicon Mac mini to `quay.io/fedora-asahi-remix-atomic-desktops/base-atomic:44`. The Asahi installer keeps on the ESP:

- `m1n1/boot.bin` - stage 2 m1n1 + U-Boot, chainloaded by stage 1 from the Apple boot partition. Without it the machine cannot boot Linux at all.
- `vendorfw/` - Apple device firmware (WiFi, Bluetooth, ...) loaded by the Asahi dracut module at boot.
- `ubootefi.var` - U-Boot's EFI variable store.
- `asahi/` - installer metadata.

After the install the ESP contained only `EFI/`. The system was only recoverable because a copy of the ESP had been made beforehand. The same class of problem applies to any platform that stores firmware or earlier boot stages on the ESP (U-Boot based SBCs, some Chromebooks, ...).

There is already a `// TODO: we should also support not wiping the ESP.` at that spot.

## Change

- Only remove the `EFI/` and `loader/` directories from the ESP. These are the trees written by bootupd (shim/grub) and systemd-boot, and they are reinstalled immediately afterwards. Everything else on the ESP is preserved.
- Update the `to-existing-root` docs, which currently state that `/boot/efi` "will be reinitialized".
- Add a unit test (`test_clean_esp_bootloader_dirs`) covering removed vs. preserved content and idempotency.

## Testing

- `cargo test -p bootc-lib --lib install::tests::test_clean_esp_bootloader_dirs install::tests::test_remove_all_noxdev` passes (Fedora 44, aarch64).
- `cargo fmt --check` clean.
- The real-world install above was completed by manually restoring the wiped ESP content from a backup before reboot; with this change that step would not have been necessary.

## AI disclosure

Generated-by: AI

Per `AGENTS.md`, the commit intentionally carries no `Signed-off-by`; the human submitter will review and add it. The problem was diagnosed and reproduced on real hardware by the submitter; the code change and unit test are LLM generated and were compiled and tested as described above.


ericcurtin:
This seems like a safer default than what was suggested in the TODO @cgwalters "TODO: we should also support not wiping the ESP."

Disclosure: very much not AI writing this comment :)

crates/lib/src/install.rs
```diff
@@ -2351,6 +2351,35 @@ fn remove_all_except_loader_dirs(bootdir: &Dir, is_ostree: bool) -> Result<()> {
...
+
+/// Remove the bootloader content from the ESP, preserving everything else.
+///
+/// The ESP is commonly shared with earlier boot stages or platform firmware
+/// which bootc does not manage and cannot recreate. For example, on Apple
+/// Silicon the Asahi installer stores the stage 2 bootloader (`m1n1/boot.bin`),
+/// the device firmware payload (`vendorfw/`) and the U-Boot EFI variable store
+/// (`ubootefi.var`) on the ESP; wiping them leaves the machine unbootable.
+#[context("Removing bootloader content from EFI system partition")]
+fn clean_esp_bootloader_dirs(efidir: &Dir) -> Result<()> {
```
Johan-Liebert1, line 2392: People might also have custom bootloader config files in `ESP/loader`. I think we should read the config files and only remove the ones we own and the boot binaries referenced by them.

ericcurtin: Maybe but I wouldn't mind a second opinion on that... For example... Would the pre-existing boot entries be completely broken after this? I'm open to it, but a second opinion could be interesting

Johan-Liebert1: It's more or less related to https://github.com/bootc-dev/bootc/pull/2243. I don't think it's a blocker per se, but if we're going the distance to do this, then maybe we should just handle all the cases.

> Would the pre-existing boot entries be completely broken after this?

we're deleting the pre-existing boot entries... so, yes.

ericcurtin: Sorry, maybe I didn't explain myself well enough... Even if we left the pre-bootc boot entries around... Would they even boot at that point? Would they even be useful after the conversion to a bootc system? (genuine question, haven't checked how destructive "bootc install to-existing-root" is to userspace, initramfs, kernel, etc. This is useful in Asahi because of the files required above needed to boot, not because of rollback to pre-bootc case)

Johan-Liebert1: > Would they even boot at that point? 

if we leave both the `.conf` file and the PE binary it points to, inside of the ESP, then there's no reason why that entry won't work. These entries probably won't be booting a kernel but maybe could be something firmware related

cgwalters:
> This seems like a safer default than what was suggested in the TODO @cgwalters "TODO: we should also support not wiping the ESP."

Yes it's clearly safer. However, it's worth explaining the rationale behind the current default, which is that it was designed to support in-place reprovisioning from a default cloud image. In those cases, we can be generally sure we want to remove everything else.

One way we could reasonably weaken this default is via equivalent of `ConditionVirtualization=` perhaps.

But I don't like that much.

The thing is, if we only remove boot loader entires we are going to be leaking e.g. shim/grub or other things that we *do* expect to get reinitialized from the container state.

So one approach is to make this the OS problem - in fact kind of what's going on here with system-reinstall-bootc is we have custom package-manager plugins to support cleaning up the old OS state (which in some cases we do want to carry along!).

Hmm, it might be that that would actually just work today on def-managed systems, if we `rpm -e shim` that's going to wipe just those binaries from the ESP. Arguably we should do that *before* we run bootupd or install our boot loader.

ericcurtin:
> > This seems like a safer default than what was suggested in the TODO @cgwalters "TODO: we should also support not wiping the ESP."
> 
> Yes it's clearly safer. However, it's worth explaining the rationale behind the current default, which is that it was designed to support in-place reprovisioning from a default cloud image. In those cases, we can be generally sure we want to remove everything else.
> 

I guess in the x86_64/aarch64 UEFI case, if we consider nothing else.

> One way we could reasonably weaken this default is via equivalent of `ConditionVirtualization=` perhaps.
> 
> But I don't like that much.
> 
> The thing is, if we only remove boot loader entires we are going to be leaking e.g. shim/grub or other things that we _do_ expect to get reinitialized from the container state.
> 

Could we just overwrite these things with new ones? Since either way whether if we destroy everything or overwrite binaries, it's destructive to rollback to the pre-bootc state either way.

> So one approach is to make this the OS problem - in fact kind of what's going on here with system-reinstall-bootc is we have custom package-manager plugins to support cleaning up the old OS state (which in some cases we do want to carry along!).
> 
> Hmm, it might be that that would actually just work today on def-managed systems, if we `rpm -e shim` that's going to wipe just those binaries from the ESP. Arguably we should do that _before_ we run bootupd or install our boot loader.

There's a narrative in Fedora Asahi community (well one single person) that bootc is only interested in by the book UEFI platforms. I don't believe that to be true because there has been success in Android Bootloader-style platforms and suggestions of BIOS compatibility. The opinionated people (well again a single person) in Asahi is looking hard for reasons to not move forward with an Atomic variant.

But if we don't have a clear path forward here for the `bootc install to-existing-root` part, I guess that's true.

cgwalters:
> There's a narrative in Fedora Asahi community (well one single person) that bootc is only interested in by the book UEFI platforms. I don't believe that to be true because there has been success in Android Bootloader-style platforms and suggestions of BIOS compatibility.

I'm sure I know who that person is and they're obviously wrong; the bootloader space is very complicated but we definitely support classic x86 BIOS, and there's support for s390x zipl and as you know aboot in ostree, which bootc can enable too. (That said the Android Boot situation is quite complex)

> Could we just overwrite these things with new ones? Since either way whether if we destroy everything or overwrite binaries, it's destructive to rollback to the pre-bootc state either way.

Not necessarily, we would like to support that.

> But if we don't have a clear path forward here for the bootc install to-existing-root part, I guess that's true.

I think there's a pretty clear path per above: aim to keep the "reprovision cloud image" working, which could possibly be a change in default to `to-existing-root` installs and have system-reinstall-bootc pass that.

crates/lib/src/install.rs
ericcurtin: Agreed it'd be better, but keeping this PR minimal; added a TODO referencing #2243 to be more selective within `EFI/` and `loader/` as a follow-up.

ericcurtin:
Cloud images only have `EFI/` on the ESP, so reprovisioning behaves the same with this change. Happy to add an opt-in full wipe for system-reinstall-bootc if you'd still like one @cgwalters

crates/lib/src/install.rs
Johan-Liebert1: Sounds good to me

PR #2476, opened 2026-09-19
install: Read --root-ssh-authorized-keys before changing mounts

ericcurtin:
## Problem

`prepare_install()` mounts a fresh tmpfs over `/tmp` (`setup_tmp_mount`) and mirrors the host's `/var/tmp` into the container, and only *afterwards* reads the file given via `--root-ssh-authorized-keys`. A file bind mounted into the install container under `/tmp` is therefore hidden by then:

```
podman run ... -v ./authorized_keys:/tmp/authorized_keys:ro <image> \
    bootc install to-existing-root --root-ssh-authorized-keys /tmp/authorized_keys
...
error: Installing to filesystem: Reading /tmp/authorized_keys: No such file or directory (os error 2)
```

The file is visible in the container (`podman run ... cat /tmp/authorized_keys` works), so this is quite confusing. The docs happen to use `/target/root/.ssh/authorized_keys` in the example, which avoids the problem by accident.

## Change

- Read the file before any mount changes. The existing comment already says "eagerly read the file now to ensure we error out early"; it was just placed too late.
- `prepare_install()` may re-execute the process later on (mount namespace unshare, SELinux `install_t` re-exec into a copy of the binary in `/tmp`), and the re-executed process runs `prepare_install()` again with the tmpfs already mounted. To survive that, carry the content across via the environment: a small `bootc_utils::reexec::{set_reexec_env, apply_reexec_env}` helper records variables to pass to re-executions, and both re-exec sites (`reexec_with_guardenv` and the SELinux re-exec in `lsm.rs`) apply them. This avoids mutating our own environment, which is not thread safe.

## Testing

On Fedora 44 aarch64 (host SELinux enforcing, so the `install_t` re-exec path is exercised), with `~/.ssh/authorized_keys` bind mounted at `/tmp/authorized_keys`:

- Stock bootc 1.16.10 image: `bootc install to-disk --via-loopback --filesystem ext4 --generic-image --root-ssh-authorized-keys /tmp/authorized_keys /var/tmp/disk.img` -> `error: Installing to disk: Reading /tmp/authorized_keys: No such file or directory (os error 2)`.
- Same image with `/usr/bin/bootc` replaced by this branch: install completes; the resulting deployment's `/etc/tmpfiles.d/bootc-root-ssh.conf` contains the (base64) keys from the file.
- `cargo test -p bootc-lib --lib` unit tests pass; `cargo fmt --all --check` clean.

## AI disclosure

Generated-by: AI

Per `AGENTS.md`, the commit intentionally carries no `Signed-off-by`; the human submitter will review and add it. The submitter hit this on real hardware; the change was written by an LLM and tested end to end as described above.


crates/lib/src/install.rs
```diff
@@ -1667,6 +1671,29 @@ async fn prepare_install(
...
+        .root_ssh_authorized_keys
+        .as_ref()
+        .map(|p| -> Result<String> {
+            if let Some(v) = std::env::var_os(ROOT_SSH_AUTHORIZED_KEYS_ENV) {
+                return v.into_string().map_err(|_| {
+                    anyhow::anyhow!("Invalid UTF-8 in {ROOT_SSH_AUTHORIZED_KEYS_ENV}")
+                });
+            }
+            let contents = std::fs::read_to_string(p).with_context(|| format!("Reading {p}"))?;
+            bootc_utils::reexec::set_reexec_env(ROOT_SSH_AUTHORIZED_KEYS_ENV, &contents);
```
Johan-Liebert1: We should make sure it's UTF-8 compatible here itself

crates/utils/src/reexec.rs
```diff
@@ -1,9 +1,38 @@
...
 use std::os::unix::process::CommandExt;
 use std::path::PathBuf;
 use std::process::Command;
+use std::sync::Mutex;
 
 use anyhow::Result;
 
+/// Additional environment variables to pass along when we re-execute ourself;
+/// see [`set_reexec_env`].
+static REEXEC_ENV: Mutex<Vec<(OsString, OsString)>> = Mutex::new(Vec::new());
```
Johan-Liebert1, line 10: We only ever call this from one thread so I'm sure if the mutex is required. Also, not a fan of this being a global var. We should only ever re-exec when we are installing, so putting this in `prepare_install` would make sense and passing in the vector of env vars to `reexec_with_guardenv` as an `extra_args` param?

Johan-Liebert1 (changes requested):
Logic looks good to me. Could you please sign off your commit as we require a `Signed-off-by: <email>` line at the end of commit messages

crates/utils/src/reexec.rs
cgwalters: We need a mutex or equiv around any `static`, no problem with that from my PoV.

> Also, not a fan of this being a global var. 

Yes, but doing it differently would require threading this state from the install code into the lsm code...doable but ugly in a different way.

BTW I would generalize this slightly and e.g.:

- Define a struct we can serialize to JSON of stuff we need to save between re-exec
- In the install path, gather that state before we re-exec
- Serialize it to a memfd
- Set an env var `_BOOTC_INSTALL_REEXEC_STATE`
- Deserialize it early in the install path

crates/lib/src/install.rs
ericcurtin: Switched to `std::env::var()`, so the value is a `String` (UTF-8 enforced by the type) and a non-UTF-8 value is a proper error. The value we write is also a `String` from `read_to_string`, so it is UTF-8 by construction.

crates/utils/src/reexec.rs
ericcurtin: Kept the global per Colin (threading it from install into lsm was the alternative), but folded the env application together with the argv/argv0 setup into a single `prepare_reexec()` used by both re-exec sites, so the duplication there is gone too, plus a unit test.

On the memfd/JSON generalization: `set_reexec_env` is already generic over key/value, and the only payload today is a small text file, so I left the env approach for now. Happy to move to a memfd-backed struct if we grow more state to carry across re-exec (or if key files near the 128KiB per-env-string limit turn out to be a real concern).

crates/lib/src/deploy.rs
```diff
@@ -77,54 +76,6 @@ use crate::utils::async_task_with_spinner;
 // TODO use https://github.com/ostreedev/ostree-rs-ext/pull/493/commits/afc1837ff383681b947de30c0cefc70080a4f87a
 const BASE_IMAGE_PREFIX: &str = "ostree/container/baseimage/bootc";
 
-// Match podman's default registry retry policy. A failed attempt has to rebuild
```
cgwalters, line 80: It looks like you reverted other changes

cgwalters (changes requested):

crates/lib/src/deploy.rs
ericcurtin: Sorry about that, a bad rebase on my side (the squash was staged against a stale main). Fixed in 8c0cbb85: the diff against main is back to just the 5 intended files (Justfile, install.rs, lsm.rs, tests-integration/install.rs, reexec.rs).

ericcurtin:
The three `plan-44-shadow-fixup` failures in the last run were a shutdown deadlock in `bootc-finalize-staged.service` (`lsblk` stuck in `autofs_expire_wait` on the `/boot` automount, killed after `TimeoutStopSec=5m`), unrelated to this change. Fix proposed separately in #2488; re-pushed the same commit to re-run CI here.

